Issue #23, opened 2014-12-06
Export / import machines


bfirsh:
Something like this should work for transferring machines:

```
 $ machine export test | ssh anotherhost machine import
```

(Maybe we could use `machine inspect`?)

cjellick:
@bfirsh @waitingkuo This seems like a really useful feature that we'd use at Rancher (we'd like to use machine server-side).
 
Any reason the implementation (#29) seems to have stalled. Anything I could do to attempt to hep?

imton:
+1 would be awesome

Starefossen:
+1

urish:
+1

kevin-coyle:
+1

kevinsimper:
We have create a tool that we use ourselves, to make it easy to export docker-machines. Feedback is welcome!

https://github.com/blackbeardapp/docker-machine-export

cusspvz:
+1

uptownhr:
+!

cusspvz:
Currently we are using a git repo for sharing docker settings, problem is that some of us have virtualbox machines and thats the down-side

docteurklein:
@kevinsimper how do you import them back?

saada:
Does this issue include the idea of sharing swarm clusters between developers?

docteurklein:
@saada I think it only covers **credentials** sharing, not cluster config.
It would be awesome tho to have this ability in machine.
Currently I only know of https://github.com/efrecon/machinery and https://github.com/nathanleclaire/moby

micheletedeschi:
+1

pinecorn:
I am looking after the capability to share machine between users in the same hosts.
And is there a workaround for this ? I am copying the whole `~/.docker/machine/` folder for now

fuse:
+1

ffissore:
+1

shanewow:
+1

leandromoreira:
+1

montytorr:
+1

toh2526:
+1

lukaville:
+1

dimorinny:
+1

arpagon:
+1

gomex:
+1

mhubig:
+1

imjching:
+1

leandromoreira:
I think a better way to show your intention is to click on `Subscribe` (at right side **Notifications** )

schickling:
[...]

PR #29, opened 2014-12-06
add export and import

waitingkuo:
It's the implementation for #23 

To export a machine, you can type

    machine export MACHINE_NAME

to get a tar file.

Then you can use 

    machine import TAR_FILE

to import it in another computer.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -8,12 +8,12 @@ import (
 	"strings"
 	"sync"
 	"text/tabwriter"
-  "path"
-  "path/filepath"
-  "archive/tar"
-  "bytes"
-  "io"
-  "io/ioutil"
+	"path"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


ehazlett:
@waitingkuo would you mind syncing up with @sthulb on this?  I can see where this would be useful.  #260 is very close to this and I think we should all come together to decide how this should work.  There are a couple of things like the format, what should be exported, what to do with machines that cannot be exported (local vs. remote), etc.  Thanks!

ehazlett:
Closing as this has stagnated.  Please re-open to start the discussion again.  You can also put comments in #260 as well.  Thanks!

waitingkuo:
Hi @ehazleet Sorry for the late reply. Just came back from a trip. I'll take a look #260 and join the discussion soon.

PR #38, opened 2014-12-07
Proposal: pass env vars and custom options to the Docker Daemon 

dduportal:
Hi guys !

I'm in this use case : virtual box dev only, for running docker + fig  stack (and docker compose in near future).

We are facing two issues :
 - Ours b2d virtualbox's machines will run behind the corporate http_proxy to access the internet
 - We have a private internal registry which in running with no certificate : so, we'll have to add an ```--insecure-registry```switch to the DOCKER_OPTS

While one can just access the vbox machine, configure the /var/lib/boot2docker/profile file, it seems simpler to have easy configuration possibilities.

Given that, here is a proposal (early stage, need to discuss a bit before implementing) for passing docker options to the daemon, and passing env vars.

# Usage of the all thing :

I've written a little bit of doc, in order to approach the proposal from the user point of view. See the latest rendered version here : https://github.com/dduportal/machine/tree/proposal-conf-daemon-and-env-vars#configuring-the-docker-options-and-environment

## Direct passing to the command line using switches :

The idea is to define two switches that the *machine* command will use the fetch the custom options and inject them inside the newly created machine.
 - ```--docker_opts``` will a be a single KVP switch, its content will be append to the DOCKER_OPTS variable and applied to the Daemon.
 - ```--env``` will be a collection of kvp, in the same philosophy of the ```-e```switch of docker run command. (*Maybe* we can use a ```-e```switch and docker-run-syntax too in order to stay homogeneous ?)  

## Using a config file
I'm thinking about also using a config file (~/.machine_config for default and ```--config-file``` switch for custom file when dealing with a bunch of machines).

I haven't written any doc yet around the config file because i don't what and how to choose the correct syntax and i need help :
- INI-like file, with a global section for all machines, and specific sections (by driver, by machines, maybe by users or label)
- Same in YAML syntax ?
- Simple KVP text file ?
- Other idea ?

My preference is going to a YAML format, which seems homogeneous with the new docker compose core functionality.

# Implementation :

Here also, we have some things to think about, given that we are going to have a lot of target drivers AND OSes.

## Docker OPTS

We need to have a configurable and overload-able method in order to append DOCKER_OPTS, where we should :
- Associate file path of where to configure depending on the target OS (or at least the managing services system used) : *boot2docker* => var/lib/boot2docker/profile,  *ubuntu* => /etc/init/docker.conf, etc.
- The docker restart command : *boot2docker* => ```sudo /etc/init.d/docker restart```, *ubuntu* => ```sudo service docker restart```, etc.
- Conflict policy : i think we should do like in the init script of boot2docker : simply appending, which will overload the maybe-already-existing arguments

## Env vars

Same comment, because it depends what the user have in its mind. In my use cas, i just want to pass http_proxy and no_proxy variables to the context of the docker daemon and reload it.

But what if the user is starting up a machine based on a generic OS image and want machine to install docker inside ? Maybe it is outside the scope ? ut in this case, maybe the user want http_proxy to be accessible for its install processes and shell (bash, package manager, curl, etc.).
Any advise ?

My personal tough is : write that idea, make it outside the first scope, and saw if user try that or not.


ehazlett:
I like the general direction here.  We definitely need a way to specify the Docker opts for a Machine.  As for the `--env` are you imagining these for Docker only or system wide?

As for the config file, there is a PR open to add this functionality to Docker itself (https://github.com/docker/docker/pull/10151).

dduportal:
Thanks @ehazlett  for the return :)

For the env vars part, i was thinking to provide it to the Docker daemon only, using the principle (i like) of the "transparent hypervisor" : if you want to do something on that docker host, you should do it inside a container.

Whatever, maybe pushing also the env var to system wide could be useful for use case i'm not seeing yet.

As i am not sure to handle all the complexity related, i propose to implement Docker daemon in a first time and see what it got, what do you think ?

dduportal:
@ehazlett : concerning https://github.com/docker/docker/pull/10151, should be re-use the ".dockercfg" (whatever its realname would be) that @nathanleclaire  as stated, or should be use a ".machinecfg" (machine-dedicated one) to not pollute each others in case of directives ?

sthulb:
A `.machingcfg` could be nice, we could use it for auth tokens too.

nathanleclaire:
If boot2docker supported cloudinit would that take care of what we need to configure here?  Just curious as we shouldn't add more configuration options / interface surface area if we don't need to.

We definitely should _not_ write to `.dockercfg`.  I hadn't known it but that file is totally tangled up in all of the registry / auth logic and there are some cases where weird stuff happens like e.g. the whole file gets marshaled up and sent to the daemon.  I think a `~/.docker/machinecfg` file (probably in git config file format - that's where I'm leaning towards with Docker's config at least) would be fine though.

dduportal:
Hey, thank you @nathanleclaire and @sthulb  for your answers.

@nathanleclaire : you're right, cloud-init is one option. But i wonder how to handle other OSes ?
I wanted to abstract the underlying OSes regarding the wanted settings to the Docker daemon.

At the implementation level, it will imply to add a little API (i'll start gently with a switch case) which as to determine which OS we are running, and so define the file where to write docker opts and env vars, the keys to use (DOCKER_OPTS, OTHER_ARGS, etc.) and the daemon restart command (initd ? systemctl ? 
The user can provide its own if wanted :)

ehazlett:
@dduportal I would start with something simple like just passing DOCKER_OPTS.  I don't think we want to mess with system env vars (for now at least) and I want to see what comes out of the docker config discussions.  I think for now we could allow for something like `--docker-opts` during create that would allow for customizing at least the engine.

dduportal:
@ehazlett : i've done some preliminary implementation on a branch named wip-env on my repo : https://github.com/dduportal/machine/tree/wip-env if you want to see. I'll merge in the PR when i'll have a basic machine + vbox + b2d working use case (for example : using devicemapper instead of aufs in a basic b2d vanilla with machine)

dduportal:
OK, i added a simple PoC for virtual box driver + b2d as Host OS.

ping @ehazlett, @bfirsh  , @nathanleclaire , @sthulb :
First i see that my code is clearly redundant with some new changes recently pushed to master, lie drivers.(Start|Stop)Docker or host.ConfigureAuth. 
=> I think the next step here is to "merge" the actions around the docker-opts (like the one done in the ConfigureAuth). This is why I persisted it at the Host level, using a slice of string. The idea is to add your config directives, and at the end, we restart (or stop / start) the remote damon to apply all opts in one shot.
=> I haven't pushed until a map struct, given we can have multiple values for one key (exemple : -H/--host)
=> Since i'm beginner with Go, i'm not sure if there isn't an existing parsing lib. I looked at go-core flag like libs, but they parse from the binary codlin (and not from a string), and expect a finite-and-known collection of option : this is clearly a pain to maintain regading the lifecyle of docker opts, so i haven't investigated more.

Secondly, considering those elements :
* Docker daemon configuration file
* Docker daemon start/stop/restart commands
* The key of the variable to use to persist docker opts (EXTRA_ARGS / DOCKER_OPTS/...)
* Maybe another in the future
don't you think this is OS-type related, like :
* If boot2docker then ...
* if ubuntu-14.04 then ...
...
I begin with some crappy switch cases in host.go, but it seems that drivers implementation also might need theses elements Knowledge.
=> I was thinking about implementing something like that in utils (independent from host or driver) :
* A package named "dockerHostOSUtils"
* Which contains an interface, like for drivers, with those methods :
  * GetDockerHostOSType string
  * GetDockerHostOSConfigFile string
  * GetDockerHostOSStartCmd string
  * GetDockerHostOSStopCmd string
...

What do you think of that idea ? Is the naming and location seems ready ? Should i make first a standalone PR for that before connecting here ?

sthulb:
@dduportal Thanks! 

I like the idea of abstracting the docker commands away from the driver, I think it should go one step further and go into `docker.go`, something that could be moved into the `machines` package once #286 is merged, this would be in favour of @ehazlett's driver level ones.

We could abstract OS specific things into a `machine/os/type.go` (debian.go, tiny.go etc) file later and have `docker.go` or any other function wrap them later. (This is something I had planned to do after the 0.1.0 is released).

host.go
```diff
@@ -345,3 +359,157 @@ func (h *Host) SaveConfig() error {
...
+	}
+}
+
+func (h *Host) getDockerOptsVarName() (string, error) {
+
+	switch h.HostOSType {
+	case "boot2docker":
+		return "EXTRA_ARGS", nil
+	default:
+		return "", HostOSTypeNotKnown
```
sthulb: Errors should probably start with `Err`

commands.go
```diff
@@ -68,6 +68,11 @@ var Commands = []cli.Command{
 				),
 				Value: "none",
 			},
+			cli.StringFlag{
```
sthulb: Instead of parsing them out one by one, we can use the cli package's `StringSliceFlag` which allows you to pass the same option repeatedly and will return an array. Which means you could simplify `hosts.go`'s `parseDockerOpts()` The interface to the user would be: `$ docker-machine create -d <driver> --docker-opt="foo=bar" --docker-opt="baz=qux" ... <name>`

ehazlett: Thanks!

As you can see we are trying to move to some common operations.  There is
another PR that is doing similar things (
https://github.com/docker/machine/pull/286).  I think we are all thinking
along the same lines.  I would like to bring these together to come up with
a single "master plan" so to speak.  This will be very beneficial in the
near future, especially when we move to an API model.

As for the OS level, I'm a little torn on this.  Personally, I don't want
to try to support several distros in code.  An idea a while back was to use
cloudinit.  This would mean instead of stuff like we have here (OS
switching, deep inspection into the OS to figure out where things are, etc)
we could use cloudinit.  This way, if one wanted to use a different OS, all
they would have to do is update cloudinit (perhaps we could template it out
even).  In that way, we off-load some of the OS specific configuration to
that instead of re-creating it for machine.

+1 on refactoring the StartDocker, StopDocker, etc.  We have it for now to
support what is needed, however, I think we can plan this out for the near
future.

+1 for cli StringSlice.  +1 for prefixing errors with "Err".

host.go
dduportal: Changed, thanks :)

nathanleclaire:
@ehazlett I think we can close this with the recent addition of Engine Options.

dduportal:
@nathanleclaire i close it myself, since the Engine Options seems to solve my problem :)

Thanks !

nathanleclaire:
:+1: For anyone interested, https://github.com/docker/machine/pull/1040 is now merged and solves this issue.

ehazlett:
Thanks for the contribution and feedback @dduportal !

lasdecoeur:
Is the question related to env variables (http_proxy) resolved with this merge ?

PR #50, opened 2014-12-08
Use a script to bootstrap Docker.

smashwilson:
Rather than rely on each individual driver to bootstrap the hosts that it launches with a properly installed Docker service, it seems like it would be more manageable to centralize the install-and-configure logic in a single place.

I've started a bash script that uses https://get.docker.io and similar capability-probing logic to configure a host to act as a proper citizen with `machine`. I've been using it from a gist in the Rackspace driver as:

```go
if err := drivers.AddPublicKeyToAuthorizedHosts(d, "/.docker/authorized-keys.d"); err != nil {
	return err
}

sshCmd, err := d.GetSSHCommand(`curl -sSL https://gist.githubusercontent.com/smashwilson/1a286139720a28ac6ead/raw/41d93c57ea2e86815cdfbfec42aaa696034afcc8/setup-docker.sh | /bin/bash`)
if err != nil {
	return err
}
if err := sshCmd.Run(); err != nil {
	log.Infof("Error while bootstrapping docker: %v", err)
	log.Infof(`You'll need to log in with "machine ssh" and:`)
	log.Infof(" * Install Docker if necessary")
	log.Infof(" * Configure Docker to use identity auth and listen on all interfaces")
}
```

Do you think this is a useful approach?

### DIFF
--- a/script/setup-docker
+++ b/script/setup-docker
@@ -1,0 +1,113 @@
+#!/bin/bash
+#
+# This script is intended to be curled onto a host that was just provisioned with "machine create".
+# It should:
+#
+# * Install docker using the most suitable package manager.
+# * Configure docker to use identity auth and to listen on all interfaces.
+
+set -e
+
+has() {
+  command -v "$@" > /dev/null 2>&1
+}
+
+USER="$(id -un 2>/dev/null || true)"
+
+# Determine the command to run something as root.
+
+RUN='sh -c'
+if [ "$USER" != 'root' ]; then
+  if has sudo; then
+    RUN='sudo -E sh -c'
+  elif has su; then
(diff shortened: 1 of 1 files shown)


ehazlett:
Closing this.  We are most likely going to move to something like cloudinit to allow for more flexibility.  Feel free to give feedback there.  Thanks!

smashwilson:
:+1: Fine by me - my main agenda was to prevent setup from being re-invented from provider to provider. :bow:

ehazlett:
@smashwilson awesome.  please jump in when we start the cloudinit discussion (I'll try to remember to add you ;)).  i definitely want to hear your feedback!

PR #54, opened 2014-12-09
GCE Driver.

dlorenc:
This is a continuation of https://github.com/docker/machine/pull/7

I accidentally screwed up my remote branch, and then it got merged somehow.

I had a couple questions from that PR:

First, the GCE API uses oauth2 with refresh tokens. My current implementation caches tokens in driver.storePath, but this is obviously deleted when a machine is rm'ed, and only applies to a single machine. Would you be opposed to caching them one level up? So .docker/hosts/gce/ instead of .docker/hosts/gce//

Second, my test is more of an integration test and actually creates/rm's a machine. It's currently failing travis because it needs a real project ID to run against. Are you interested in having this kind of test? If so, how do you want me to block it from the travis run?

Signed-off-by: Dan Lorenc lorenc.d@gmail.com

### DIFF
--- a/drivers/google/integration_test.go
+++ b/drivers/google/integration_test.go
@@ -8,12 +8,12 @@ import (
 	"testing"
 
 	log "github.com/Sirupsen/logrus"
-
 	"github.com/docker/machine/state"
 )
 
 var (
-	project = flag.String("project", "", "Project")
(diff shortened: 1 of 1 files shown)


ehazlett:
@bfirsh i just tested again and works.  docs added in #201.

PR #104, opened 2014-12-15
[Fun] Introduce StreamSSHCommand function

nathanleclaire:
I thought it would be nice if we could get some output from the various drivers while we're doing provisioning.  E.g. if we call `curl get.docker.com | sh` over SSH it'd be quite nice to see the output of the script as its running to get a feel for how far along it is (or what went wrong if something does).  This function will stream the output of an SSH command as it is running.  It is used like:

```go
// suppose we are in digitalocean driver (d)
if err := drivers.streamSSHCommand(d, "echo foo"); err != nil {
    return err
}
if err := drivers.streamSSHCommand(d, "echo bar"); err != nil {
    return err
}
if err := drivers.streamSSHCommand(d, "echo quux"); err != nil {
    return err
}
```

Output:

```
digitalocean | foo
digitalocean | bar
digitalocean | quux
```

Wasn't sure about placement etc. so I put it here.  WDYT?

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -1,8 +1,10 @@
 package drivers
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
+	"io"
 	"os/exec"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


sthulb:
I'm not sure about the prefix. I like the ability to just copy output from my terminal.

ehazlett:
Actually as I think about it now, I don't think I like the prefix.  Since we only currently support running commands against a single machine, I'm not sure what purpose the prefix serves.  Other than that, LGTM

PR #128, opened 2014-12-23
use ubuntu 14.04 in azure; fix port errors and endpoint issues

ehazlett:
This fixes the azure provider.  It also moves to use plain Ubuntu to provision (in order to customize the Docker engine config and use the `2376` port).

Signed-off-by: Evan Hazlett ejhazlett@gmail.com


jeffmendoza:
:+1:

PR #129, opened 2014-12-25
cloudstack driver

svanharmelen:
This is a fully working and tested CloudStack driver for docker machine. All info needed to use this driver for your (private) CloudStack setup can be found in the updated README.

Of course there are a lot of use cases (different CloudStack setups) that could need more and/or other options to make the driver also work for those use cases. But since this is not predictable and this is currently tested with standard basic and advanced zones, most use cases should be covered without additional work.

Currently there also is an open WiP PR (#82) for a CloudStack driver, but @dduportal is in the loop and together we decided it was best to make this new PR in favour of the existing one. So @dduportal will contribute to this driver if needed and the PR #82 will be closed.


svanharmelen:
Ah, I missed the part where it should be in the commit. But coming to think of it, it's of course pretty obvious... Just updated the PR to comply with the requirements.

dduportal:
Tested in collaboration with @svanharmelen, in my company internal Cloudstack IaaS.
It work well !

Two settings tested :
* "Dev" zone : instances have one IP, on a flat private network where a DHCP server is used. Use case : dev run machine from windows like 
```
./machine_windows_amd64.exe create -d cloudstack --cloudstack-api-url=$CLOUDSTACK_API_URL --cloudstack-api-key=$CLOUDSTACK_APIKEY --cloudstack-secret-key=$CLOUDSTACK_SECRETKEY --cloudstack-machinename=b2d-dev-cs --cloudstack-template=boot2docker-1.4.1 --cloudstack-offering=little --cloudstack-zone=myzone--cloudstack-no-public-ip dev-cs
```
* "Prod" zone : you create your networks, then you spawn the instances telling it which network to use (manual IPs with public-ip param and auto-DHCP nets also)

=> Works great !

Testing scenario in each case :
* On Windows (7 x64) and Linux (Centos 7 x64)
* Create machine
* Set the DOCKER_HOST from URL (+ 2375 while i don't use the TLS & identity connexion)
* With the local docker client (yes Windows also :) ) : docker ps + docker run busybox echo ok + fig up with a basic 3 tiered docker-registry (nginx + registry + data volume)
* stop the machine
* start the machine
* move default machine from cloudstack to a vbox one and revert back
* list machine
* url machine
* rm the machine

=> LGTM !

csghuser:
This is great :+1:  However is there a plan to add a switch to select which security group the cloudstack instance will start with? Does it just startup with the default group at the moment?

Apologies if this is present and I have missed it.

svanharmelen:
@csghuser hmm... It should give a clear error if you go over the max userdata size. I'll see if I can update the go-cloudstack package so it does a post instead of a get, which updates the size from max 2K to max 32K. So then that will no longer be an issue, but still I wonder if you are hitting that limit since you don't get that error...

Do you have any more details? What is the command your using exactly? And from what OS are you running this? And to what kind of template?

Security group support is not in yet, as I didn't have a use case. Seems I do have one now, so will look into it...

csghuser:
@svanharmelen Many thanks for responding. The command I am using is:
` ./machine_linux_amd64 create -d cloudstack --cloudstack-template="coreos-beta" --cloudstack-offering="2vCPU, 4GB Memory" --cloudstack-no-public-ip --cloudstack-zone="zone01" dev`
I'm running this on a Ubuntu 14.04 machine, running within a virtualbox instance. I followed the instructions in the readme to create the coreos template. 

I've been through the cloudstack api logs, I can see the api requests come through when machine requests the UUID of the template, service offering and the zone. However it never seems to log the request to create the instance. I've ran tcpdump locally and I can see the request getting sent, and it does appear to be over 2K in size.

svanharmelen:
Hmm... Strange that the error on line 506 is triggered if that is the case. Could you (if your up for it) insert this line at line 505:

```
log.Debugf("What size is ud: %d", len(ud))
```

That should output the actual size of the user data after encoding it... You can also mail me directly for debugging this one if you like <sander@xanzy.io>.

Will have a look later today to update the go-cloudstack package to do a post while creating the machine so this limit is not so tight anymore.

csghuser:
Confirmed that the size of userdata was less than 2k:
`What size is ud: 1836`
I decided to try this outside of the virtualbox instance (actually a cloudstack instance), and it worked fine. Must be a strange networking issue that I need to look into further... :) Apologies for the distraction!

svanharmelen:
Cool, thanks for the feedback and cool that it works!

Will still have a look at changing the machine create call to a POST call to make sure there is always more then enough room for the userdata, but that code is not in this PR....

dduportal:
Ping @svanharmelen : we should rebase since there was a lot of PR merged recently.

Whats needs to be done today regarding the POST ? You want/need some help ?

bhaisaab:
w00t @svanharmelen !

svanharmelen:
@dduportal rebasing only is not enough as a lot is changed since my last commit. But I think I have all changes incorporated now, so will test some more and most likely create a new PR for this.

The POST is not really related to this PR, but as we now also would need to send certificates the userdata gets to big for the GET. This is something I need to change in the go-cloudstack package. Will have a look at that shortly and will try to get everything in by the end of the day.

Maybe we could/should also try to get in touch with @bfirsh and @ehazlett to see what else is needed to get this one merged. Would be really great if we could get this in before v0.1.0 is released...

dduportal:
Thanks for the quick response @svanharmelen !

I propose you that roadmap :
* Rebase from actual master (don't forget your git rerere :) ) => It seems to be done
* Correct validation/functionnal tests/DCOs/etc => It's on your own given that you are the owner of the PR :)
* End to end test at this current state : i'm ready to that.Ping @csghuser to help us test on cloudstack ?
* Ask validation for merging to @ehazlett / @bfirsh for 0.1.0
* In a second time, update the go-cloudstack lib for managing the  POST

?

svanharmelen:
That sounds about right... But the go-cloudstack changes should be done before I update the commit as we now need more space then is available when using a GET instead of a POST

svanharmelen:
@dduportal Almost done... go-cloudstack is updated and so is the driver.... Only thing I still see is some issues while trying to connect using the certificates. So I guess that needs a few little tweaks I will pick up tomorrow.

Did just push the current code so you can already play with it. It should do everything correctly, other then failing to connect using tls...

svanharmelen:
Argh... This is getting pretty annoying... So just when I expected a successful PR, again the interface changed so we are missing a function.

Will add that one tomorrow together with the last fixes to fully support the TLS connections. After that I will close this PR and open a new one as this one is getting really bad :wink:

thaJeztah:
Loving the Dutch "oeps" commits :smile:

svanharmelen:
Ok... All looks good now, so lets close this messy PR and open a brand new clean PR in order to make some progress in getting this merged again...

Will of course reference this PR in the new PR.

Issue #180, opened 2014-12-30
Discover current b2d instances and add to store


ehazlett:
Credit to shykes for this. This would allow for an easy transition from boot2docker to machine.

ehazlett:
This will be part of a "b2d migration path" doc.  I will rebase this to get up to date with the latest vbox driver.

teng-lin:
+1

nathanleclaire:
This has largely been covered by the `--virtualbox-use-boot2docker-vm` flag IMO.

PR #197, opened 2015-01-01
refs #180: add option to migrate from an existing b2d instance

ehazlett:
This adds a flag to import (copy) an existing boot2docker instance during creation of a virtualbox machine.  Perhaps later, we can add the additional "magic" to do a full sweep of existing b2d instances as mentioned in #180.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -27,12 +27,13 @@ import (
 )
 
 type Driver struct {
-	MachineName    string
-	SSHPort        int
-	Memory         int
-	DiskSize       int
-	Boot2DockerURL string
-	storePath      string
+	MachineName             string
+	SSHPort                 int
+	Memory                  int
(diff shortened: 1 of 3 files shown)


drivers/virtualbox/virtualbox.go
```diff
@@ -94,10 +100,11 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.Memory = flags.Int("virtualbox-memory")
 	d.DiskSize = flags.Int("virtualbox-disk-size")
 	d.Boot2DockerURL = flags.String("virtualbox-boot2docker-url")
+	d.boot2dockerInstanceName = flags.String("virtualbox-migrate-boot2docker-instance")
```
ehazlett: I'm all for a different name here :)

ehazlett:
This has stagnated.  Due to the upcoming refactoring we can revisit this later.

Issue #59, opened 2014-12-09
Set the hostname of created machines to match the machine name


robhaswell:
When creating and managing multiple machines, it is confusing to identify which machine you have logged into after using `machine ssh`. Setting the hostname to match the machine name would be useful in this scenario.

(woops I initially raised this against swarm by accident)

sthulb:
If I've read this issue correctly, the implementation should be:
  
  1. Set a hostname on creation
  1. add `machine move/rename` which will change a machine's name in the internal data store and change the hostname
  1. For move/rename: Error if the name is taken in the internal data store.

bfirsh:
I don't think renaming is necessary to start with - that's already being worked on here https://github.com/docker/machine/issues/59

At the moment we give hostnames long random names. I think we should set the name on the provider to be the name given by the user, which normally is also the hostname given to the machine. For example – when you create a droplet on digital ocean, you give it a name, which is also used by the hostname.

bfirsh:
Another thing to be aware of: the name given by the user might already exist on the provider. We should display useful error messages about this.

sthulb:
I absolutely agree with setting the name on the provider too. Things like EC2 won't care since the unique criteria will be the instance id.

As per: 

> I don't think renaming is necessary to start with - that's already being worked on here #59
That's this issue :)

ehazlett:
+1 for setting the name / hostname on the provider if supported.  I agree to leave renaming / moving out of this one -- I believe bfirsh meant #30 is being worked on for that.

sthulb:
I'll give it a go :)

bfirsh:
Oh whoops, yes, wrong issue. Sorry.

ehazlett:
Closing as #219 has been merged.

tlaguz:
How can I disable host renaming?

PR #219, opened 2015-01-05
Hostnames

sthulb:
This work sets the machine name in the Cloud API or Hypervisor. As well
as setting the hostname inside the VM.

I've added the machine name to the NewDriver func to allow for
identification by Cloud APIs and for the driver package.

Each driver will attempt to set the hostname for the VM, except for
Azure, which sets it based on the DNS name.

Refs: #59


drivers/virtualbox/virtualbox.go
```diff
@@ -363,6 +363,14 @@ func (d *Driver) Create() error {
 		return err
 	}
 
+	cmd, err = d.GetSSHCommand(fmt.Sprintf("sudo hostname %s", d.MachineName))
```
ehazlett: I don't believe this will persist reboots will it (need to update `/etc/hostname` at least).

bfirsh: Does that persist on boot2docker? Perhaps we don't bother setting the hostname on boot2docker... I think the big win from this pull request is the correct name showing up in your cloud provider's admin (e.g. list of instances on EC2), not the hostname when you log in.

sthulb: You're right, it won't persist reboots. However one needs to update `/usr/local/etc/hostname` for it to persist on a b2d image.

sthulb: @bfirsh I'm going for naming all of the things.

bfirsh:
Thanks for this @sthulb. The bit I'm really interested in is setting the correct name on cloud providers. E.g. on Digital Ocean, removing the concept of `DropletName` and just making the name of the droplet the name of the host.

sthulb:
EC2 support is kind of there, I can tag an instance with the machine ID, just the internal hostname remaining.

sthulb:
I moved how we set the hostname to the `driver.Create` func since different drivers need the hostname func to be called at different times.

drivers/amazonec2/amazonec2.go
```diff
@@ -225,9 +228,32 @@ func (d *Driver) Create() error {
 		return err
 	}
 
+	log.Infof("Settings tags for instance")
```
ehazlett: This should probably be Debug

ehazlett:
@sthulb i like this so far -- good work!

drivers/amazonec2/amazonec2.go
sthulb: If you prefer :)

bfirsh:
Great stuff. Since a driver is always attached to a machine, perhaps the name should be passed to `NewDriver`?

sthulb:
@bfirsh It's something that should be considered when we need it in other places, but for now, I think leaving it on `Create(name string)` seems ok.

sthulb:
@bfirsh I take back my comment... azure kind of forces your idea.

sthulb:
@bfirsh @ehazlett I think it's finished. someone should perhaps test the Azure & Google drivers

ehazlett:
awesome!  thanks -- I will check it out

ehazlett:
I'm getting "sudo: unable to resolve host test-ec2" with `amazonec2`.  The hostname isn't added to `/etc/hosts`.  You may need to silence this for the first sudo to add it but I think it would be good otherwise it will spit this out for every `sudo` command.

sthulb:
I'll alter the PR to add to the hosts file also.

ehazlett:
@sthulb thanks

ehazlett:
@sthulb i'm also getting the following with azure:

```
FATA[0004] A hosted service with the specified name already exists.
```

It's not happening in master.

sthulb:
I've fixed the sudo thing.

sthulb:
So Azure likes globally unique names... so the best we can do is: `foo-<timestamp>` for azure, which sucks. Unless someone knows differently?

jeffmendoza:
In Azure, your VM sits inside a "Cloud Service." The cloud service must have a globally unique name, as it's used for DNS: <name>.cloudapp.net. The VM can have any name, and that will be the hostname of the VM.

ehazlett:
@sthulb i think naming with a unique/timestamp is fine.  it's just the cloud service and not the actual VM hostname which is what we want to name anyway.

ehazlett:
@sthulb sorry to be nitpicky but I'm still getting a couple `sudo: unable to resolve host test-ec2`.  We may need to `> /dev/null 2>&1` or something to hide the initial set.  wdyt?

sthulb:
That's fine, ... I just need to track down the origin of it...

ehazlett:
@sthulb Hmm I keep getting this error with EC2:

```
Error creating host: Error decoding error response: Error decoding error response: http: read on closed response body
```

Have you ran into that?

sthulb:
Not at all.

ehazlett:
Hmm ok -- may be AWS.  I'll try again -- thx

sthulb:
Perhaps we can get @bfirsh try it too.

ehazlett:
@sthulb can you do a fresh build and confirm it's not doing it?  It's not doing it to me in master.

ehazlett:
@sthulb +1 /cc @bfirsh

drivers/digitalocean/digitalocean.go
```diff
@@ -147,9 +145,24 @@ func (d *Driver) Create() error {
 		return err
 	}
 
+	log.Debugf("Setting hostname: %s", d.MachineName)
+	cmd, err := d.GetSSHCommand(fmt.Sprintf(
+		"echo \"127.0.0.1 %s\" | sudo tee -a /etc/hosts && sudo hostname %s && echo \"%s\" | sudo tee /etc/hostname",
```
bfirsh: Doesn't Digital Ocean set the hostname to be the droplet name?

sthulb:
@ehazlett I think I've figured out why you get it... By any chance, do you already have a security group with the same group name?

ehazlett:
yep.  looks like it was left over from a previous run.  i will remove and
test.  thanks!

ehazlett:
LGTM

PR #232, opened 2015-01-07
Write test for getHostState

nathanleclaire:
This tests the getHostState method I refactored out from the existing `ls` command.

I also took the opportunity to move that method to the bottom of the file to follow the existing pattern of aux. functions in that file.

cc @ehazlett @bfirsh

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io/ioutil"
 	"os/exec"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


commands_test.go
```diff
@@ -0,0 +1,114 @@
...
+func (d *FakeDriver) Upgrade() error {
+	return nil
+}
+
+func (d *FakeDriver) GetSSHCommand(args ...string) (*exec.Cmd, error) {
+	return &exec.Cmd{}, nil
+}
+
+func TestGetHostState(t *testing.T) {
+	storePath := "~/.docker"
```
ehazlett: Should we change this to something else so we don't potentially smash current configs?  Not sure if it would or not but just to be safe perhaps?

ehazlett:
@nathanleclaire go fmt

commands_test.go
nathanleclaire: What do you mean smash current configs?  What else would we change it to?  I mostly just have it here as a dummy value so it can fill in the field.

nathanleclaire:
Fixed.  I think my editor's hook doesn't use `-s` which is annoying >_>

nathanleclaire:
@ehazlett I changed to use a temp dir

ehazlett:
LGTM

Issue #119, opened 2014-12-18
Machines are in a broken state when creation fails


nathanleclaire:
Consider this example:

```console
$ machine ls
NAME   ACTIVE   DRIVER   STATE   URL

$ machine create -d amazonec2 --amazonec2-access-key foo --amazonec2-secret-key bar --amazonec2-subnet-id quux spam
INFO[0000] Launching instance...
FATA[0000] Problem with AWS API call: Non-200 API response: code=401 message=AWS was not able to validate the provided access credentials

$ machine ls
ERRO[0000] error getting state for host spam: Problem with AWS API call: Non-200 API response: code=401 message=AWS was not able to validate the provided access credentials

ERRO[0000] error getting URL for host spam: Problem with AWS API call: Non-200 API response: code=401 message=AWS was not able to validate the provided access credentials

NAME   ACTIVE   DRIVER      STATE   URL
spam            amazonec2   Error

$ machine rm spam
ERRO[0000] Error removing machine spam: unabme to terminate instance: unknown instance
FATA[0000] There was an error removing a machine. To force remove it, pass the -f option. Warning: this might leave it running on the provider.

$ machine ls
ERRO[0000] error getting state for host spam: Problem with AWS API call: Non-200 API response: code=401 message=AWS was not able to validate the provided access credentials

ERRO[0000] error getting URL for host spam: Problem with AWS API call: Non-200 API response: code=401 message=AWS was not able to validate the provided access credentials

NAME   ACTIVE   DRIVER      STATE   URL
spam            amazonec2   Error

$ machine rm -f spam
ERRO[0000] Error removing host, force removing anyway: unabme to terminate instance: unknown instance

$ machine ls
NAME   ACTIVE   DRIVER   STATE   URL
```

User should not have to `machine rm -f` at all.  If there is _any_ error with `machine create`, machine should clean up the failed host create.

ehazlett:
What if we do the following:

If DEBUG is enabled, we do not automatically cleanup the reference nor attempt to remove the provisioned instance.

If not, we show the error and then remove the reference and attempt to remove the possibly provisioned instance.  We also show a warning that states all resources may not have been removed.

Thoughts?

bfirsh:
If a machine might have been created, we definitely want to keep around the local reference. Otherwise the user might not know they have created an instance and they'll be charged money for it.

Perhaps along with an error, we explain that your machine might be in a broken state, and give you instructions on how to remove it.

ehazlett:
[...]

PR #234, opened 2015-01-07
add cleanup for create failure

ehazlett:
Refs: #119 

This will automatically call Remove to cleanup a machine when Create fails if DEBUG is not specified and show an error message.  If DEBUG is specified, it behaves as normal:

Default

```
ehazlett@ejh-mbp ~/d/g/s/g/d/machine> machine create -d digitalocean test-do
ERRO[0000] Error creating host: POST https://api.digitalocean.com/v2/account/keys: 401 Unable to authenticate you.
WARN[0000] Removing provisioned machine.  Use --debug to prevent removal.
FATA[0001] You will want to check the provider to make sure the machine and associated resources were properly removed.
ehazlett@ejh-mbp ~/d/g/s/g/d/machine> machine ls
NAME   ACTIVE   DRIVER       STATE     URL
```

Debug

```
ehazlett@ejh-mbp ~/d/g/s/g/d/machine> ./machine -D create -d digitalocean test-do
DEBU[0000] executing: /usr/bin/ssh-keygen ssh-keygen -t rsa -N  -f /Users/ehazlett/.docker/machines/test-do/id_rsa

Your identification has been saved in /Users/ehazlett/.docker/machines/test-do/id_rsa.
Your public key has been saved in /Users/ehazlett/.docker/machines/test-do/id_rsa.pub.
The key fingerprint is:
dd:2a:e3:88:a2:e7:1d:9c:07:a4:57:ea:22:d1:92:a1 ehazlett@ejh-mbp.local
The key's randomart image is:
+--[ RSA 2048]----+
|                 |
|                 |
|.   . .          |
|.+ o o   . .     |
|E o +   S . .    |
| o + o     .     |
|. . = . o .      |
| .oo + o o       |
|.+..o . .        |
+-----------------+
ERRO[0000] Error creating host: POST https://api.digitalocean.com/v2/account/keys: 401 Unable to authenticate you.
FATA[0000] POST https://api.digitalocean.com/v2/account/keys: 401 Unable to authenticate you.
ehazlett@ejh-mbp ~/d/g/s/g/d/machine> machine ls
ERRO[0000] error getting state for host test-do: GET https://api.digitalocean.com/v2/droplets/0: 401 Unable to authenticate you.
ERRO[0000] error getting URL for host test-do: IP address is not set
NAME      ACTIVE   DRIVER         STATE     URL
test-do            digitalocean   Error
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -194,11 +194,12 @@ func cmdCreate(c *cli.Context) {
 		if c.GlobalBool("debug") {
 			log.Fatal(err)
 		} else {
-			log.Warnf("Removing provisioned machine.  Use --debug to prevent removal.")
+			log.Warnf("Removing created machine. You can run machine with the --debug flag to avoid this.")
 			// we know there was an error so do not check for the error on removal
 			// instead we will Fatal with a message to prevent spamming with error messages
-			store.Remove(name, true)
-			log.Fatalf("You will want to check the provider to make sure the machine and associated resources were properly removed.")
+			_ = store.Remove(name, true)
+			log.Warn("You will want to check the provider to make sure the machine and associated resources were properly removed.")
+			log.Fatal("Error creating machine")
 		}
 	}
 	if err := store.SetActive(host); err != nil {
--- a/commands.go
+++ b/commands.go
@@ -190,7 +190,16 @@ func cmdCreate(c *cli.Context) {
 
 	host, err := store.Create(name, driver, c)
 	if err != nil {
-		log.Fatal(err)
+		log.Errorf("Error creating host: %s", err)
+		if c.GlobalBool("debug") {
+			log.Fatal(err)
+		} else {
+			log.Warnf("Removing provisioned machine.  Use --debug to prevent removal.")
+			// we know there was an error so do not check for the error on removal
+			// instead we will Fatal with a message to prevent spamming with error messages
+			store.Remove(name, true)
+			log.Fatalf("You will want to check the provider to make sure the machine and associated resources were properly removed.")
+		}
 	}
 	if err := store.SetActive(host); err != nil {
 		log.Fatalf("error setting active host: %v", err)
--- a/host.go
+++ b/host.go
@@ -227,9 +227,7 @@ func (h *Host) Upgrade() error {
 
 func (h *Host) Remove(force bool) error {
 	if err := h.Driver.Remove(); err != nil {
-		if force {
-			log.Errorf("Error removing host, force removing anyway: %s", err)
-		} else {
+		if !force {
 			return err
 		}
 	}


nathanleclaire:
LGTM.  This is pretty much what I was envisioning and a quick local test indicates it works as I expect.  Just a few logging notes.

commands.go
```diff
@@ -110,7 +110,16 @@ var Commands = []cli.Command{
 
 			host, err := store.Create(name, driver, c)
 			if err != nil {
-				log.Fatal(err)
+				log.Errorf("Error creating host: %s", err)
+				if c.GlobalBool("debug") {
+					log.Fatal(err)
+				} else {
+					log.Warnf("Removing provisioned machine.  Use --debug to prevent removal.")
```
nathanleclaire: IMO: "Removing created machine.  You can run machine with the --debug flag to avoid this."

commands.go
```diff
@@ -110,7 +110,16 @@ var Commands = []cli.Command{
...
+					// we know there was an error so do not check for the error on removal
+					// instead we will Fatal with a message to prevent spamming with error messages
+					store.Remove(name, true)
```
nathanleclaire: Skipping an error check here?

commands.go
```diff
@@ -110,7 +110,16 @@ var Commands = []cli.Command{
...
+					log.Fatalf("You will want to check the provider to make sure the machine and associated resources were properly removed.")
```
nathanleclaire: I'd make this a `log.Warn` (you don't need to format anything) and make another `log.Fatal` on the next line simply saying "There as an error on create".

commands.go
ehazlett: The way things are it almost always will fail.  That is the reason I chose to just do it and then show the final error.  Otherwise you get multiple fails - which may be ok just a little confusing.

commands.go
ehazlett: Ah right -- good catch.

commands.go
ehazlett: Yep.  Sounds better.

commands.go
nathanleclaire: Interesting - I wonder why that is.  At any rate for this PR we can probably just use `_` (to visually demarcate that we're conciously discarding the error) and then later rework `store.Create` to be a little more flexible (since I don't think an error should be returned in this case, although there may well be a good reason I'm overlooking).

commands.go
```diff
@@ -190,7 +190,17 @@ func cmdCreate(c *cli.Context) {
 
 	host, err := store.Create(name, driver, c)
 	if err != nil {
-		log.Fatal(err)
+		log.Errorf("Error creating host: %s", err)
+		if c.GlobalBool("debug") {
+			log.Fatal(err)
+		} else {
+			log.Warnf("Removing created machine. You can run machine with the --debug flag to avoid this.")
```
nathanleclaire: Nit: Still have `Warnf` here with no format string

nathanleclaire:
One pedantic point; LGTM

ehazlett:
@nathanleclaire thx!

PR #252, opened 2015-01-09
Doc fix for vSphere driver.

frapposelli:
Quick fix for the README on using the vSphere driver with single ESXi hosts.


ehazlett:
LGTM

Issue #254, opened 2015-01-09
Use "standard" TLS instead of identity auth


ehazlett:
We have found that the identity auth branch is a bit invasive on the Docker Engine and will not be merged soon.  I am working on using the regular TLS mechanism for auth in machine.  This is the placeholder issue.  A PR will be coming soon.

sthulb:
+1 It would be nice to use a mainstream docker binary.

ehazlett:
@sthulb +1 -- and the regular b2d iso as well.

sthulb:
Are you working on the PR? If not, I can start it since I'm done with #219 and #241

ehazlett:
@sthulb coming momentarily

ehazlett:
Closing as #255 is merged.

PR #125, opened 2014-12-19
wip: standardize install and configuration of docker

ehazlett:
This makes the default OS Ubuntu 14.04 amd64 and uses a "standard" method of installing and configuring Docker.  This will handle installing Docker as well as setting the default options to be unified across providers.  It also makes for an easier upgrade path as drivers simply need to call `InstallDocker`.

Drivers implemented and testing:

- [x] Amazon EC2
- [x] DigitalOcean
- [x] Azure
- [ ] Google
- [ ] vCloud Air
- [ ] Softlayer
- [ ] Rackspace

PR #255, opened 2015-01-09
Use standard TLS for authentication

ehazlett:
This is a work in progress for using standard TLS for auth.

Currently working is digitalocean and amazonec2.  You no longer need a custom Docker binary.  You can use the standard 1.4.1.

Refs #254 

Drivers Updated and Working:

- [x] VirtualBox
- [x] Amazon EC2
- [x] DigitalOcean
- [x] Google
- [x] Azure
- [x] Fusion
- [x] vSphere
- [x] vCloud Air
- [x] Rackspace

There is also a new command to connect to the instances (until we get a global docker config): `config` - i.e. `machine config machine-test`.  There will be a help message that is printed upon machine creation as well.  Make sure to test with the standard Docker 1.4.1 client and not the identity builds we've had before.

Test Builds:

- Linux: https://docker-machine.s3.amazonaws.com/machine/linux/machine-tls-5f4df99
- Mac: https://docker-machine.s3.amazonaws.com/machine/darwin/machine-tls-5f4df99
- Win: https://docker-machine.s3.amazonaws.com/machine/windows/machine-tls-5f4df99.exe

Example Usage:

```bash
ehazlett machine> ./machine ls
NAME   ACTIVE   DRIVER   STATE   URL

ehazlett machine> ./machine create -d digitalocean test-do
INFO[0000] Creating Digital Ocean droplet...
INFO[0002] Waiting for SSH...
INFO[0082] "test-do" has been created and is now the active machine
INFO[0082] To connect: docker $(machine config test-do) ps

ehazlett machine> docker $(machine config test-do) version
Client version: 1.4.1
Client API version: 1.16
Go version (client): go1.3.3
Git commit (client): 5bc2ff8
OS/Arch (client): darwin/amd64
Server version: 1.4.1
Server API version: 1.16
Go version (server): go1.3.3
Git commit (server): 5bc2ff8

ehazlett machine> docker $(machine config test-do) ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

ehazlett machine> docker $(machine config test-do) run -ti busybox echo hello machine
Unable to find image 'busybox:latest' locally
busybox:latest: The image you are pulling has been verified
511136ea3c5a: Pull complete
df7546f9f060: Pull complete
ea13149945cb: Pull complete
4986bf8c1536: Pull complete
Status: Downloaded newer image for busybox:latest
hello machine
```

## Testing on Windows
This will eventually go into the official docs, but this is what I used to test on Windows:
- Install git for windows (http://msysgit.github.io/)
- Open a "Git Bash" prompt
- `machine-tls-5f4df99.exe create -d digitalocean --digitalocean-access-token MYTOKEN test-do`


nathanleclaire:
Rocking!  I will give this the once over when I have some time.

nathanleclaire:
Totally random idea but worth considering:  What about having some sort of `interface` for the type of auth?  It could define "verbs" which are common to various types of authorization (`ConfigureDaemon`, `AddKnownHost`, `GetPrivateKey`, `GetPublicKey`, etc. as some ideas).  I'm not sure what it would look like _exactly_ (or how we would handle that in the drivers - pass the `AuthHandler` in and use it there?) but it's definitely worth thinking about - and implementing this down the line would be much more painful than implementing it now.

Biggest advantage of this going forward would be testability (interfaces can easily be "mocked out" in unit tests) and swappability (don't like libtrust?  no problem, you can still use TLS).

ehazlett:
@nathanleclaire I like that idea.  I'll give it some more thought.  Thanks!

ehazlett:
@frapposelli any chance i could get you to gist me what is needed for vmtools to build a custom ISO?  Once we have that I think this can be reviewed / merged (only fusion/vsphere are left) :)

drivers/vmwarevcloudair/vcloudair.go
```diff
@@ -407,66 +411,17 @@ func (d *Driver) Create() error {
...
-
-		cmd, err = d.GetSSHCommand(dockerInstall)
-		if err != nil {
-			return err
-		}
-
-		if err = cmd.Run(); err != nil {
-			return err
-		}
+	dockerInstall := "curl -sSL https://get.docker.com/ | sudo sh"
```
tianon: Why is this piping to `sudo sh` and not just `sh`?  The script does intelligently `sudo` (or `su` as a fallback if `sudo` is unavailable, which isn't uncommon) itself only for the lines where `root` is necessary: https://github.com/docker/docker/blob/9f2c48614497dee2448b65c81aa6d44953484372/project/install.sh#L42-L53

ehazlett: Good question.  @frapposelli any idea?

frapposelli: That comes straight from https://docs.docker.com/installation/ubuntulinux/ but I'm all for changing that.

frapposelli:
@ehazlett we have two PRs open for that: boot2docker/boot2docker#638 and boot2docker/boot2docker#659 currently the ISO we use is built from [cloudnativeapps/boot2docker@`vmw-identity`](https://github.com/cloudnativeapps/boot2docker/tree/vmw-identity)

That's pretty much it.

drivers/vmwarevcloudair/vcloudair.go
ehazlett: @frapposelli thanks!  i think we can leave it for now since that's what the docs say :)

tianon: I don't see this line anywhere on that page. ;)

Those all have /ubuntu/, which is a different script that does require sudo
and is not recommended for actual use...  (Speaking as the maintainer of
both scripts.)

ehazlett: @tianon I think he's referring to the "Note" about 1/3 of the way down.  We actually do this for quite a few setups.  Should we be using a different method?  What would you recommend as the way to get the latest from the Docker repos?  Just setup the sources.list file and then apt-get install?

nathanleclaire: @ehazlett I think @tianon simply recommends we pipe to `sh` alone and not `sudo sh`.

tianon: If you want to continue reaping the benefits of that script, I absolutely recommend you drop the `sudo` here (as @nathanleclaire says).  You should be aware however that the future of that script is cloudy (note that the last time it was touched was [Nov 9, 2014](https://github.com/docker/docker/commits/master/project/install.sh) when the directory was renamed, and [a PR on Oct 22, 2014 that was merged on Nov 4, 2014](https://github.com/docker/docker/commits/master/hack/install.sh) before that.  It's not heavily maintained specifically because platform detection is very _very_ fragile.

If we know the platform we're trying to install into, I definitely recommend more direct installation.  For example, in this case, I'm definitely very much +1 to performing the steps manually:
```bash
echo deb https://get.docker.com/ubuntu docker main > /etc/apt/sources.list.d/docker.list
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
apt-get update
apt-get install -y lxc-docker
```

tianon: Also, look more closely at the Note you're referring to: :wink:
> ```console
> $ curl -sSL https://get.docker.com/ubuntu/ | sudo sh
> ```

frapposelli: Guys, not a problem, I just picked that from the page, but if there's a better way to do the install, I'm all for that.

And for the Note, copy and paste never lies :wink:

tianon: https://get.docker.com **/ubuntu/**

frapposelli: @tianon if you're referring to the `/ubuntu/` missing, I just omitted it because it's not necessarily ubuntu the target platform (it's the default, but can be passed as a parameter) I did some investigation on the script and figured out it would work on RHEL too, if I'm mistaken (which is likely the case) I'll fix that.

tianon: Right, that's exactly my point.  The scripts at /ubuntu/ and / are very,
very different, and the one at /ubuntu/ is meant to be a reference, not
something users actually run (and how that crept into the docs is beyond my
understanding -- it makes me want to turn the script into HTML explicitly
so people stop that).

frapposelli: @tianon right, I didn't know the two scripts were radically different, so what's the jury vote on that? maintain that and drop `sudo` in front of `sh` and that's it?

ehazlett: @tianon yep -- thanks!  @frapposelli I think it's fine for now.  We have a standardization PR (https://github.com/docker/machine/pull/125) that will use cloudinit and actually does exactly what @tianon is recommending.

thanks for the clarification all!

ehazlett:
@ggiamarchi is there anyway to support the external hard drive for Fusion?  I would like to try to use the standard B2D iso.  Do you think this is possible?

ehazlett:
There is an issue with provisioning machines using Windows.  It creates the paths on the remote machine as `\etc\docker\ca.pem` etc.  I think we could go ahead and merge this PR and open a new issue to go through and fix for Windows support.

Thoughts @bfirsh ?

sthulb:
There's already a windows path issue open.

sthulb:
I don't like how each driver is starting to dupe more and more functionality, things like starting/stopping docker should be abstracted out, perhaps into an init script based model and we guess which one to use at runtime for each machine.

sthulb:
The TLS generation in `hosts.go` could perhaps be moved into `tls.go` to separate it out a little.

frapposelli:
@ehazlett I'm trying to reproduce the same behaviour that virtualbox uses (magic string in the external hard drive) but right now we don't have a binary shipping with fusion that can do that. I'm building a quick and dirty golang package to manipulate VMDKs in that way but it's taking some time to develop.

utils/b2d.go
```diff
@@ -0,0 +1,62 @@
...
+		return "", fmt.Errorf("no releases found")
+	}
+
+	tag := t[0].TagName
+	url := fmt.Sprintf("https://github.com/boot2docker/boot2docker/releases/download/%s/boot2docker.iso", tag)
+	return url, nil
+}
+
+// Download boot2docker ISO image for the given tag and save it at dest.
+func DownloadISO(dir, file, url string) error {
```
sthulb: If this is being refactored, perhaps issue #217

cjellick:
While running off this PR, I stumbled upon some odd behavior:
If I create a machine named "foo" and then create another machine named "foo", I get the following:
```
ERROR Error creating host: Host "foo" already exists
WARN Removing created machine. You can run machine with the --debug flag to avoid this.
WARN You will want to check the provider to make sure the machine and associated resources were properly removed.
FATAL Error creating machine
```
And indeed it did delete foo when it obviously should have left it alone. Not sure if this a new bug introduced or something that will be resolved once you merge, but I wanted to point it out.

utils/b2d.go
ehazlett: Correct, that should take care of this.  I didn't want to blow this scope too much. :)

ehazlett:
@cjellick oh that's not good.  could you try it against master as well?  I don't believe there was anything in this that should do that behavior.  If so, I will update this.  Otherwise, I will fix it outside of this PR.

ehazlett:
@frapposelli ok thanks.  I'm not sure we will be able to get the bits needed into b2d for the 1.5 release so I was wondering if there was another way we could do it to enable the disk.  Thanks and let me know if there is anything I can do to help with the VMDK stuff.

cjellick:
Yes, should have mentioned: I tried against master and got the normal/expected behavior.

ehazlett:
@sthulb `GenerateCertificates` is pretty coupled to a host right now but I think it fits.  The actual cert generation is in `tls.go` and `GenerateCertificates` is centered around the actual host itself.  I do see where we can remove the param as it's not needed.

ehazlett:
@cjellick ok thanks I will debug :)

ehazlett:
@cjellick ok that has been reverted.  I will open (or re-open the old) issue for cleanup to properly address existing machines.

ehazlett:
Re-opened #119 to address the cleanup properly.

ehazlett:
One final integration test amongst available providers:

https://gist.github.com/ehazlett/cef8363e03466074dcf1

LGTM

sthulb:
Woooo! :tada:

nathanleclaire:
:clap: :clap: :clap:

Issue #62, opened 2014-12-09
Display better error message when "ssh-keygen" is missing


pandrew:
Digitalocean driver assumes ssh-keygen is installed, if not it leaves a node in an error state. I hade to `machine rm test` and install ssh-keygen and recreate the node. A check for public key and/or ssh-keygen before creating the node might solve this.
```
FATA[0000] Host "test" already exists
root@4a6e561799c5:/go# machine ls
ERRO[0000] error getting state for host test: GET https://api.digitalocean.com/v2/droplets/0: 404 The resource you were accessing could not be found.
ERRO[0000] error getting URL for host test: IP address is not set
NAME   ACTIVE   DRIVER         STATE   URL
test            digitalocean   Error
```

ehazlett:
Closing as #257 was merged.

PR #257, opened 2015-01-09
Adds check for SSH keygen command

sthulb:
Adds a check to find ssh-keygen in the user's $PATH, in the event it
can't find it, return an error with a friendly message

Fix for #62


nathanleclaire:
Why not [`exec.LookPath`](http://golang.org/pkg/os/exec/#LookPath)?

sthulb:
I didn't know it existed. I'll update.

sthulb:
Hat tip to @nathanleclaire :)

ehazlett:
LGTM

Issue #231, opened 2015-01-07
Alias "machine ps" to "machine ls"


bfirsh:
"machine ps" doesn't make sense because they're not processes, but lots of people are going to type it by mistake. We should probably alias "machine ps" to "machine ls" or at least say "You're looking for machine ls."

thaJeztah:
If there's any chance that a `machine ps` will exist in the future, I'd go for the latter, and not encourage people to get used to it. Just my 0.02c

ehazlett:
@thaJeztah thanks for that link.  Perhaps we follow suit and report "machine --help"?

thaJeztah:
I think it would be good to be consistent between docker projects yes

nathanleclaire:
Yes, I am +1 on just displaying "That's not a command, run `machine --help`."  Aliasing `machine ps` to `machine ls` would encourage bad habits and confuse users IMO.

sthulb:
I wouldn't alias it. Just do as others have mentioned.

bfirsh:
Alright. Doing nothing is my favourite solution.

PR #258, opened 2015-01-09
Add unknown command help

sthulb:
Spits out "machine: '<command>' is not a machine command see 'machine --help'" to be inline with Docker

Refs: #231


ehazlett:
Confirmed.  LGTM.  Thanks!

PR #260, opened 2015-01-10
Add "export"/"import" command.

sthulb:
So far, I've added a driver type (local/remote) for each driver, this is
required since local machines should be stopped before an export to
ensure data integrity

I've also added an "export" & "import" function to the store.

The implementation uses `archive/tar` to handle tars, there could be an issue with how I'm reading the VM files, since the disks could be quite big and cause performance issues.

Refs: #23 

---
I'm submitting another implementation since #29 seems to have stalled.


sthulb:
I think this is probably done now.
@ehazlett @bfirsh

sthulb:
Actually done now :) @bfirsh @ehazlett

ehazlett:
I'm not sure if this is setting up the keys/metadata properly.  I exported a virtualbox machine and imported.  However, when I try to run `ls` it is prompting me for the `docker@localhost` login info.  A quick inspect of the store config and they both have the same `MachineName` and `SSHPort`.

sthulb:
@ehazlett I fixed the naming, but I have to find a way to reconfigure the box for local drivers, since they have paths and custom ports, etc. 

I propose a `Configure()` func to the driver so I can regenerate it, but it causes issues, unless I also add a metadata method to return things like RAM, SSH ports and such. What do you think?

sthulb:
I think this is working, barring any regressions on the drivers.

ehazlett:
@sthulb what about just using the virtual disk for the local drivers?  This only works on local anyway right?  That would save the trouble of storing/restoring all config.  I did something similar in #197.

sthulb:
That's all I'm pretty much doing now. I regenerate the machine's metadata file.

ehazlett:
This will need a rebase.  Perhaps we should also wrap this in with the driver refactor as it touches the driver types.

sthulb:
Closing since this feature needs to be rethought of. 

I think there's a better way of doing this.

Issue #256, opened 2015-01-09
machine url panics 


cjellick:
### Description of problem:
If no machine is active, ```machine url``` panics

### Machine version:
Running latest of master of machine

### Environment details:
Running machine on Mac OSX Yosemite

### How reproducible:
Very

### Steps to Reproduce:
1. Get into a state where you have no active machine. Easiest is to just not create a machine. But I think deleting the active machine should also do the trick.
2. Run ```machine url``` with no machine/host specified

### Actual Results:
machine panics:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0xe90c2]

goroutine 1 [running]:
github.com/docker/machine/hosts.(*Host).GetURL(0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/cjellick/go/src/github.com/docker/machine/hosts/host.go:249 +0x92
main.cmdUrl(0xc20809c180)
	/Users/cjellick/go/src/github.com/docker/machine/commands.go:377 +0x41
github.com/codegangsta/cli.Command.Run(0x699510, 0x3, 0x0, 0x0, 0x7017b0, 0x18, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/cjellick/go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:101 +0xe42
github.com/codegangsta/cli.(*App).Run(0xc20809c000, 0xc20800a000, 0x2, 0x2, 0x0, 0x0)
	/Users/cjellick/go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:130 +0xb81
main.main()
	/Users/cjellick/go/src/github.com/docker/machine/main.go:36 +0x3ea

goroutine 5 [syscall]:
os/signal.loop()
	/usr/local/Cellar/go/1.4/libexec/src/os/signal/signal_unix.go:21 +0x1f
created by os/signal.init·1
	/usr/local/Cellar/go/1.4/libexec/src/os/signal/signal_unix.go:27 +0x35

goroutine 6 [runnable]:
text/template/parse.lexText(0xc208068100, 0x7c2580)
	/usr/local/Cellar/go/1.4/libexec/src/text/template/parse/lex.go:228 +0x37b
text/template/parse.(*lexer).run(0xc208068100)
	/usr/local/Cellar/go/1.4/libexec/src/text/template/parse/lex.go:198 +0x5d
[...]

PR #262, opened 2015-01-11
Fix for panics when no active host is set

sthulb:
By adding some better error checking on `store.GetActive`, we can stop
getting panics on some commands

As a part of this PR, we no longer care about `store.GetActive` failing. 

Refs: #256, #222

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -256,6 +256,15 @@ func cmdLs(c *cli.Context) {
 
 	for _, host := range hostList {
 		if !quiet {
+			tmpHost, err := store.GetActive()
+			if err != nil {
+				log.Errorf("There's a problem with the active host: %s", err)
+			}
+
+			if tmpHost == nil {
+				log.Errorf("There's a problem finding the active host")
+			}
+
 			go getHostState(host, *store, hostListItems)
 		} else {
 			fmt.Fprintf(w, "%s\n", host.Name)
@@ -399,6 +408,10 @@ func getHost(c *cli.Context) *Host {
 		if err != nil {
 			log.Fatalf("unable to get active host: %v", err)
 		}
+
+		if host == nil {
+			log.Fatal("unable to get active host, active file not found")
+		}
 		return host
 	}
 
@@ -426,7 +439,7 @@ func getHostState(host Host, store Store, hostListItems chan<- hostListItem) {
 
 	isActive, err := store.IsActive(&host)
 	if err != nil {
-		log.Errorf("error determining whether host %q is active: %s",
+		log.Debugf("error determining whether host %q is active: %s",
 			host.Name, err)
 	}
 
--- a/store.go
+++ b/store.go
@@ -64,6 +64,7 @@ func (s *Store) Remove(name string, force bool) error {
 	if err != nil {
 		return err
 	}
+
 	if active != nil && active.Name == name {
 		if err := s.RemoveActive(); err != nil {
 			return err
--- a/store_test.go
+++ b/store_test.go
@@ -189,6 +189,7 @@ func TestStoreGetSetActive(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
 	if host != nil {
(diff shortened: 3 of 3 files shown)


sthulb:
@bfirsh, @ehazlett: I'd consider this done. The only objectionable thing is how I treat the lack of an active machine.

commands.go
```diff
@@ -426,7 +426,7 @@ func getHostState(host Host, store Store, hostListItems chan<- hostListItem) {
 
 	isActive, err := store.IsActive(&host)
 	if err != nil {
-		log.Errorf("error determining whether host %q is active: %s",
+		log.Debugf("error determining whether host %q is active: %s",
```
bfirsh: Why swallow this error?

sthulb:
I found the error message annoying when you have a few machines... `store.GetActive` is called for every machine during `ls`, resulting in an error message for every machine. 

Not having an active machine doesn't impair `ls`

bfirsh:
Lots of this error swallowing looks scary. I think the interface should be `IsActive` always returns `false` if there's no active machine, `GetActive` should return `nil`.

sthulb:
Added back comment: 

@bfirsh I found the error message annoying when you have a few machines... store.GetActive is called for every machine during ls, resulting in an error message for every machine.

Not having an active machine doesn't impair ls

sthulb:
@bfirsh `IsActive` could be better for `getHostState()`

bfirsh:
Correct – if there is no active host, we should return false. If there is actually an error checking whether there is an active host (e.g. permissions are wrong on `~/.docker/machines/.active`) then we should throw the error so the user can fix the problem.

sthulb:
@bfirsh Sure, but we need to fix it so it's not shown for every machine.

sthulb:
@bfirsh I think my last commit is a fair solution, I still hide the error from IsActive, but I do a call to GetActive to display the error once.

store.go
```diff
@@ -59,10 +59,8 @@ func (s *Store) Create(name string, driverName string, flags drivers.DriverOptio
 }
 
 func (s *Store) Remove(name string, force bool) error {
-	active, err := s.GetActive()
-	if err != nil {
-		return err
-	}
+	active, _ := s.GetActive()
```
ehazlett: error check

store.go
```diff
@@ -113,12 +111,16 @@ func (s *Store) Load(name string) (*Host, error) {
...
+	fi, err := os.Stat(s.activePath())
+	if err != nil {
 		return nil, err
 	}
+
+	if fi.Size() == 0 {
+		return nil, fmt.Errorf("active file is empty")
+	}
+
+	hostName, _ := ioutil.ReadFile(s.activePath())
```
ehazlett: error check

store_test.go
```diff
@@ -185,10 +185,8 @@ func TestStoreGetSetActive(t *testing.T) {
 	store := NewStore("")
 
 	// No hosts set
-	host, err := store.GetActive()
-	if err != nil {
-		t.Fatal(err)
-	}
+	host, _ := store.GetActive()
```
ehazlett: error check

store_test.go
```diff
@@ -223,10 +221,8 @@ func TestStoreGetSetActive(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	host, err = store.GetActive()
-	if err != nil {
-		t.Fatal(err)
-	}
+	host, _ = store.GetActive()
```
ehazlett: error check

ehazlett:
@sthulb i know we were hiding some of these but I think we should check for them.  If it's for aesthetics, perhaps we can just format the error when bubbling it?

bfirsh:
Yep – we don't want to swallow errors. If we are receiving an error we expect, we should give the error a type and check for it.

bfirsh:
I'm not sure I'm following what's going on here – what is the error you're getting?

sthulb:
Now with error checking!

ehazlett:
LGTM

PR #264, opened 2015-01-12
WIP: The ability to suspend machines that run on VMWare Fusion or VirtualBox

Tebro:
Title says it all, give users the ability to save the state of a VM on these providers, and resume them later.

What is done: Working implementation for VMWare Fusion.

What is to be done: 
* Error handling
* New state when running "machine ls"
* Implementation for VirtualBox

If someone knows they can do this, or point me in the right direction, I would appreciate the help, first time I ever used GOlang.


Tebro:
Fusion does only tell us if the machine is running or not, no details about the state (Saved etc.). Any ideas?

PR #266, opened 2015-01-12
fixes #229: wait for docker before adding to known_hosts

### LINKED ISSUE #229: [DigitalOcean/Azure driver] Intermittent tls Handshake errors on create
A few times now I've gotten DigitalOcean errors on `create`:

```console
$ machine create -d digitalocean --digitalocean-access-token token foo
INFO[0001] Creating Digital Ocean droplet...
INFO[0003] Waiting for SSH...
FATA[0094] tls Handshake error: dial tcp 104.236.111.172:2376: connection refused
```

If I set that host to active and run docker commands against it it works as expected... it's just the `create` output that's blowing up.

Anyone else run into this?

ehazlett:
This adds a port check to make sure Docker is available before adding to `known_hosts`.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -332,6 +332,14 @@ func (d *Driver) Create() error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


bfirsh:
LGTM

PR #267, opened 2015-01-12
Stop machine from overwriting an existing machine

sthulb:
Stops machine from overwriting an existing machine in Create()

refs #249

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -180,6 +180,17 @@ func cmdCreate(c *cli.Context) {
 		log.Fatal("You must specify a machine name")
 	}
 
+	store := NewStore(c.GlobalString("storage-path"))
+
+	exists, err := store.Exists(name)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if exists {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Mostly LGTM, I'm not a fan of throwing away that error though.  Can we check it and echo it / `log.Fatal` on error instead?

sthulb:
We only care about the boolean, the error is rather superficial.

nathanleclaire:
If it's superficial then that function shouldn't be returning any error.  Discarding errors is a bad habit, let's please check it just in case.

bfirsh:
Agreed, we mustn't throw away errors. If the error is being thrown away then the error shouldn't be returned by the function.

sthulb:
No longer ignoring the error

ehazlett:
This LGTM.  @bfirsh / @nathanleclaire is this good for you also?

nathanleclaire:
LGTM Thanks @sthulb !

PR #276, opened 2015-01-14
Move VirtualBox configuration file to be inside the data store

sthulb:
It turns out, that we only stored the actual disk inside the data store,
this brings it back inline with vmware.


ehazlett:
LGTM

PR #284, opened 2015-01-14
change sign to ver 4 to support newer regions

### LINKED ISSUE #283: amazonec2 fails creating security group
Commit:  c2f1fea9dc8c3d8c2b182280de5124941ab904a1
Binary used: machine_darwin_amd64

```
$ machine create -d amazonec2 --amazonec2-access-key=$ACCESS_KEY --amazonec2-secret-key=$SECRET_KEY --amazonec2-vpc-id=$VPC_ID --amazonec2-region=eu-central-1 olibob --debug
INFO[0000] Launching instance...
DEBU[0000] executing: /usr/bin/ssh-keygen ssh-keygen -t rsa -N  -f /Users/orobert/.docker/machines/olibob/id_rsa

Your identification has been saved in /Users/orobert/.docker/machines/olibob/id_rsa.
Your public key has been saved in /Users/orobert/.docker/machines/olibob/id_rsa.pub.
The key fingerprint is:
75:fc:6f:6f:d8:c0:ff:45:dd:fc:7c:94:f6:7b:85:00 orobert@Cyberlab.local
The key's randomart image is:
+--[ RSA 2048]----+
|                 |
|          E.     |
|          ..o    |
|         . ... .+|
|        S    o.+*|
|              ==+|
|               =X|
|              ..X|
|               o=|
+-----------------+
DEBU[0000] creating key pair: olibob
DEBU[0000] creating security group in vpc-4d876924
FATA[0000] Please make sure you don't have a security group named: olibob
```

There is no security group with that name on the VPC I use. I can chose whatever name for the machine, it'll fail with this error.

3 security groups are defined with following names:
- DockerTLS
- Std - SSH
- default

sthulb:
This should fix #283

### DIFF
--- a/drivers/amazonec2/amz/ec2.go
+++ b/drivers/amazonec2/amz/ec2.go
@@ -149,7 +149,8 @@ func (e *EC2) awsApiCall(v url.Values) (http.Response, error) {
 		return http.Response{}, fmt.Errorf("error creating request from client")
 	}
 	req.Header.Add("Content-type", "application/json")
-	awsauth.Sign(req, awsauth.Credentials{
+
+	awsauth.Sign4(req, awsauth.Credentials{
 		AccessKeyID:     e.Auth.AccessKey,
 		SecretAccessKey: e.Auth.SecretKey,
 		SecurityToken:   e.Auth.SessionToken,


drivers/amazonec2/amz/ec2.go
```diff
@@ -149,11 +149,23 @@ func (e *EC2) awsApiCall(v url.Values) (http.Response, error) {
...
+			AccessKeyID:     e.Auth.AccessKey,
+			SecretAccessKey: e.Auth.SecretKey,
+		}
+	} else {
+		creds = awsauth.Credentials{
+			AccessKeyID:     e.Auth.AccessKey,
+			SecretAccessKey: e.Auth.SecretKey,
+			SecurityToken:   e.Auth.SessionToken,
+		}
+	}
```
md5: Isn't `creds.SecurityToken` going to end up as `""` when `e.Auth.SessionToken == ""` whether or not this `if` is there? Seems like you just need to change `awsauth.Sign` to `awsauth.Sign4`.

sthulb: @md5 It would end up as "", but it seemed to throw a tantrum about it...

sthulb:
@bfirsh, @ehazlett As good as it'll get :)

md5:
It looks like it would be nice if `go-aws-auth` were updated to use the right version of `Sign` based on region. It already has some logic to switch the signature version based on the service.

sthulb:
I think the driver could just do with some added robustness.

ehazlett:
I think in the near future we could possibly look at Stripe's Go AWS
package as well - better than rolling our own :)

sthulb:
I had a similar idea. I'll give it a go for me next PR :)

ehazlett:
LGTM

md5:
For future reference, I ended up opening this PR just now: https://github.com/smartystreets/go-aws-auth/pull/23

ehazlett:
Thanks!

PR #290, opened 2015-01-16
wip: Change AWS EC2 driver

sthulb:
The EC2 driver needs some robustness adding and a few other nice options adding.

I've decided to create an `awsec2` driver to keep my work separate. I have deliberately chosen to take some key args (`aws-access-key-id` and so on) since they're well known in other apps.

This implementation will use stripe/aws-go instead of the homebrewed solution that's currently being used in the `amazonec2` driver.


sthulb:
Closing, this needs to be replaced with something more inline with Machine's direction.

PR #340, opened 2015-01-18
Retrive latest boot2docker release tag name from VERSION file directly

saratovsource:
I think there are no limits for directly fetching file content via raw requests?

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -3,7 +3,6 @@ package virtualbox
 import (
 	"archive/tar"
 	"bytes"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


ehazlett:
@nathanleclaire you see any reason to not use this?

nathanleclaire:
Hm, I do like the idea of addressing this concern (API request limit) but I don't know what that file is used for currently (and consequently how important / likely to get updated correctly it is).  cc @SvenDowideit @tianon do you have any thoughts on this?  Would you implement it this way in b2d-cli?

ehazlett:
@SvenDowideit your thoughts on the above?

tianon:
Hey, sorry I missed this.  This will work now, and _should_ continue to work in the future, but it does definitely mean there'd be a period of anywhere from a few minutes to a few hours after the updated file is pushed to `master` before the ISO is pushed that people will have failing upgrades.

ehazlett:
@tianon No worries!  

I think the best route would be using the API since worst case scenario that would affect that requesting user while the file method would potentially affect a larger base.  Closing for now.  Thanks!

Issue #387, opened 2015-01-22
re-use machine security group


ehazlett:
When creating machines there can be issues if the security group does not get removed.  We should move to a single security group and re-use when creating machines.

sthulb:
I'm not sure that's the best idea. Eventually machine's hosts could be used for production purposes. I think a better idea would be to increment the group name if we find it's already in use.

We could end up with sec groups that look like swiss cheese because every other port is open.

ehazlett:
@sthulb  My idea was that we use this for now and then add the ability to specify a security group in a separate PR.  That way you get a default if none specified and full control if you want to place them -- the same behavior as subnet-groups.

ehazlett:
Discussion from IRC:

```
12:12 <sthulb> I'm not sold on the idea of sharing groups: )
12:12 <sthulb> :) *
12:13 <sthulb> Even with a flag to specify one
12:13 <ehazlett> see my comment
12:13 <sthulb> I did :)
12:13 <sthulb> In a world where we're trying to simplify things, sharing a group or telling a user to log in to the console and create one seems weird
12:14 <ehazlett> what about the default subnet?
12:14 <sthulb> I think perhaps incrementing group names would be better.
12:14 <ehazlett> everything is created in a single one
12:14 <ehazlett> i don't like the fact that you can have dozens of groups either
12:15 <ehazlett> (from a machine create)
12:15 <sthulb> You can specify the subnet and AWS doesn't have a default one, so it just slots them in to different AZs 
12:15 <ehazlett> idc how many you have in general
12:15 <ehazlett> right but we do in machine
12:15 <ehazlett> we find the first available and use it
12:15 <ehazlett> every time
12:15 <ehazlett> unless you specify
12:15 <sthulb> My driver just slots them in randomly
12:15 <sthulb> unless you specify 
12:16 <ehazlett> so here's the issue
12:16 <sthulb> (that bit isn't pushed yet)
12:16 <ehazlett> if you create a machine everything is fine
12:16 <ehazlett> unless you happen to name it the same as a group
12:16 <ehazlett> then it fails
12:16 <sthulb> yup
12:16 <ehazlett> or have an existing keypair, which fails
12:16 <ehazlett> (that's a separate issue)
12:17 <sthulb> We should have a loop of: Hey, does this resource exist? Yes? ok, what if we want to create <name>-1
12:17 <ehazlett> IMO we should make it as simple as possible to get started
[...]

PR #391, opened 2015-01-22
Create / Re-use existing EC2 security group

ehazlett:
This PR changes the way security groups are used in EC2.  Previously, a security group with the name of the machine would be created on each run.  This will use a default security group of `docker-machine` allowing for an optional name to be specified.  If the `docker-machine` or specified group does not exist, machine will create it.  Machine will then confirm (and update if necessary) the port authorizations needed (currently `22` for SSH and whatever the docker port is for the driver: default `2376`).

This PR also fixes a couple of race conditions in the EC2 driver found while testing: IP assigning and security group creation.

Refs: #387

### DIFF
--- a/drivers/amazonec2/amazonec2_test.go
+++ b/drivers/amazonec2/amazonec2_test.go
@@ -32,9 +32,9 @@ func TestConfigureSecurityGroupPermissionsSshOnly(t *testing.T) {
 
 	group.IpPermissions = []amz.IpPermission{
 		{
-			Protocol: "tcp",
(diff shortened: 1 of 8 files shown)


drivers/amazonec2/amazonec2_test.go
```diff
@@ -0,0 +1,94 @@
+package amazonec2
```
nathanleclaire: :+1: :+1: :+1:

ehazlett:
LGTM

PR #392, opened 2015-01-23
Remove chatty log errors which show up on machine ls

nathanleclaire:
Not sure what the general sentiment is around this, but throwing it out there.

Right now `docker-machine ls` is pretty chatty if there are machines in error state.

In general, I would argue against this style of logging and in favor of:

- A `machine error` command that would give you more information about
  _what_ kind of error state the machine is in, and/or:
- Finer granularity in the types of "Error" state we return (e.g. is it
  aborted?  Did something go wrong in create?  Is the network down?
  etc.)

@ehazlett @bfirsh @sthulb WDYT?

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -277,15 +277,6 @@ func cmdLs(c *cli.Context) {
 
 	for _, host := range hostList {
 		if !quiet {
-			tmpHost, err := store.GetActive()
-			if err != nil {
-				log.Errorf("There's a problem with the active host: %s", err)
-			}
-
-			if tmpHost == nil {
-				log.Errorf("There's a problem finding the active host")
-			}
-
 			go getHostState(host, *store, hostListItems)
 		} else {
 			fmt.Fprintf(w, "%s\n", host.Name)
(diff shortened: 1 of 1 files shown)


commands.go
```diff
@@ -277,15 +277,6 @@ func cmdLs(c *cli.Context) {
 
 	for _, host := range hostList {
 		if !quiet {
-			tmpHost, err := store.GetActive()
```
nathanleclaire: I'm not sure what the purpose of this code is in the happy case so let me know if there is something obvious I have missed.

sthulb: The purpose of this code was to trigger the errors, it's not really needed. It's code that was left in an old PR of mine and was merged.

ehazlett:
@nathanleclaire i like this.  could you explain a bit more about what you think for `machine error`?  Am I understanding correctly that by default it would a general fail would be reported and then you would run `machine error` to explain?

nathanleclaire:
@ehazlett Yes, that's exactly what I had in mind.  We could get drivers to dump API responses, for instance, to get more insight into what is happening behind the scenes when we deliberately ask for it.

ehazlett:
+1

I think this could be included as part of the logging refactor.  Thoughts?

sthulb:
Absolutely.

nathanleclaire:
> I think this could be included as part of the logging refactor. Thoughts?

+1.  Would be good to include `DumpError` method or similar on the driver interface to allow each driver to check for various things that might be going wrong on an individual basis.

nathanleclaire:
Unless I'm mistaken, this is covered by #591, so I'm closing

PR #393, opened 2015-01-23
Use binary file name instead of full path

radeksimko:
@phemmer has fair point:
https://github.com/docker/swarm/pull/282#issuecomment-71190899

If the binary is called with absolute path, there's probably no point in printing the whole path.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"os"
+	"path"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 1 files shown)


ehazlett:
LGTM

sthulb:
This need implementing in other places too.

ehazlett:
@sthulb I thought we were calling `app.Name` instead of `os.Args[0]`?  If not, we should be IMO :)

sthulb:
We should :) It's what I've been doing

ehazlett:
@sthulb ok so then shouldn't this be good?

sthulb:
This will need to be updated

PR #394, opened 2015-01-23
enable specifying arch along with os when building

ehazlett:
This enables specifying the `-arch` flag to gox.

By default, it will build binaries for windows, darwin and linux for 32 and 64 bit.  With this PR, you can now specify what arch to build for:

Example with just linux on 64 bit:

`./script/build -os="linux" -arch="amd64"`

### DIFF
--- a/script/build
+++ b/script/build
@@ -2,11 +2,17 @@
 set -e
 
 if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


ehazlett:
LGTM

PR #398, opened 2015-01-23
use ubuntu for base image

ehazlett:
We had originally started down this path with cloudinit but not all providers fully support it yet.  This consolidates the operating systems on the providers to give a consistent experience for machine.  In the future, we will improve this to be more performant and possibly move to "official" Docker images on the providers.  For now, this at least provides a consistent Docker experience in that Machines running on all providers now use the same version of Docker and the configuration with it (storage backends, TLS, etc).


ehazlett:
LGTM

Issue #383, opened 2015-01-22
Underscores in hostname cause error


ehazlett:
Thanks to @nathanleclaire for testing :)  If there is an underscore in the hostname it causes errors on creation:

```
DEBU[0057] Setting hostname: aws_lightning_fast
DEBU[0057] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -p 22 -i /Users/nathanleclaire/.docker/machines/aws_lightning_fast/id_rsa ubuntu@54.173.239.4 echo "127.0.0.1 aws_lightning_fast" | sudo tee -a /etc/hosts && sudo hostname aws_lightning_fast && echo "aws_lightning_fast" | sudo tee /etc/hostname
127.0.0.1 aws_lightning_fast
hostname: the specified hostname is invalid
ERRO[0062] Error creating machine: exit status 1
WARN[0062] You will want to check the provider to make sure the machine and associated resources were properly removed.
FATA[0062] Error creating machine
```

nathanleclaire:
:+1:

ehazlett:
I'm thinking instead of throwing errors perhaps we replace with dashes (`-`).  For example, in the above, it would become `aws-lightning-fast`.

Thoughts? /cc @sthulb @robhaswell

sthulb:
We should strip it out, or ensure it's apart of the validity check

ehazlett:
@sthulb are you saying remove them completely?  I think substituting with `-` would at least give a similar effect they are looking for.  Perhaps we should just error with `ErrInvalidHostname` or something?

robhaswell:
Think about the purpose of your "set hostname" feature. Is it purely for a way to identify machines to the user? Or is it for software to be able to identify the machines (i.e. DNS or orchestration).

For humans I was say translate the machine name into a human. For software I would say raise an exception.

I'm sorry, I know my hostname suggestion was designed for humans but I don't know if that was why this feature was merged.

ehazlett:
@robhaswell thanks for the input.  The main reason this was created was to be aware of the machine you are on when you ssh and to provide a consistent naming scheme across the providers.  For example, in the providers, the machines are now named the same as the Machine name (including metadata and the hostname).

Now that I think about it, we should probably raise an exception as automatically changing the name could lead to confusion to the user and programmatically.

Thoughts?

sthulb:
I'd return an invalid hostname error.

[...]

PR #399, opened 2015-01-23
Valid hostname check

sthulb:
Allowed hostname chars are: 0-9a-zA-Z . -

Refs: #383

### DIFF
--- a/host.go
+++ b/host.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


host.go
```diff
@@ -78,7 +80,7 @@ func LoadHost(name string, storePath string) (*Host, error) {
 
 func ValidateHostName(name string) (string, error) {
```
ehazlett: This seems like we could get some easy tests added for this?  Would you mind? :)

ehazlett: Otherwise looks good!  Thanks!

sthulb: Sure, not a problem.

sthulb:
I've added some tests :)

ehazlett:
LGTM

PR #401, opened 2015-01-24
add PreCreateCheck; check for existing ec2 keypair

### LINKED ISSUE #385: check for existing keypair
This stems from #381.  We should check for an existing keypair with the same name as this will almost always fail since Machine will not have access to the private key.

ehazlett:
This adds a method to the Driver: PreCreateCheck.  This can be used where you want to run some prerequisite checks before attempting to create the machine.  In the case of EC2, this is a check for an existingkeypair.  This can be used in the other drivers in the future as well.

Fixes #385

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -33,6 +33,10 @@ func (d *FakeDriver) GetState() (state.State, error) {
 	return d.MockState, nil
 }
 
+func (d *FakeDriver) PreCreateCheck() error {
+	return nil
+}
+
 func (d *FakeDriver) Create() error {
 	return nil
 }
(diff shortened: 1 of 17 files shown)


ehazlett:
@nathanleclaire @sthulb  thoughts?

sthulb:
Sounds like a good idea. Perhaps the check could be done in hosts instead though instead of in driver.Create

ehazlett:
@sthulb yeah i went around that idea too.  I will see if I can refactor a bit.

ehazlett:
LGTM

PR #402, opened 2015-01-24
Update to TLS Certificates

### LINKED ISSUE #397: Update TLS generation to create single CA and client certs
Currently, new CA, client and server certificates are generated for each machine.  This seems a bit much.  We should create a single CA and single client cert (for the current client) and then new server certs signed with the single CA for each machine.  This will also help integration with Swarm.

ehazlett:
This PR does the following:

- adds ServerAuth to the ExtKeyUsage (needed for some servers and swarm)
- no longer generates new CA and client certs for each machine - creates a single for the client (allows for easier sharing and a better CA workflow)
- uses a separate directory for the client cert and key (this allows for the Docker client to work using env vars)
- adds a `shellinit` command (refs: #251) for an easier local development workflow and b2d compatibility

This now enables the following (like b2d):

```
$>  docker-machine create -d virtualbox dev

$>  docker-machine shellinit dev
export DOCKER_TLS_VERIFY=yes
export DOCKER_CERT_PATH=/Users/ehazlett/.docker/machines/.client
export DOCKER_URL=tcp://192.168.99.104:2376

$>  $(docker-machine shellinit dev)
$>  docker ps
NAME        ACTIVE   DRIVER       STATE     URL
dev         *        virtualbox   Running   tcp://192.168.99.104:2376
```
Closes #397

### DIFF
--- a/utils/certs.go
+++ b/utils/certs.go
@@ -40,10 +40,10 @@ func newCertificate(org string) (*x509.Certificate, error) {
 
 }
 
-// GenerateCACert generates a new certificate authority from the specified org
+// GenerateCACertificate generates a new certificate authority from the specified org
 // and bit size and stores the resulting certificate and key file
 // in the arguments.
-func GenerateCACert(certFile, keyFile, org string, bits int) error {
+func GenerateCACertificate(certFile, keyFile, org string, bits int) error {
 	template, err := newCertificate(org)
 	if err != nil {
 		return err
--- a/utils/certs.go
+++ b/utils/certs.go
@@ -93,7 +93,7 @@ func GenerateCert(hosts []string, certFile, keyFile, caFile, caKeyFile, org stri
 	}
 	// client
 	if len(hosts) == 1 && hosts[0] == "" {
-		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}
+		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth}
 		template.KeyUsage = x509.KeyUsageDigitalSignature
 	} else { // server
 		for _, h := range hosts {
--- a/commands.go
+++ b/commands.go
@@ -30,6 +30,13 @@ import (
 	"github.com/docker/machine/utils"
 )
 
+type machineConfig struct {
+	caCertPath     string
+	clientCertPath string
+	clientKeyPath  string
+	machineUrl     string
+}
+
 type hostListItem struct {
 	Name       string
 	Active     bool
@@ -122,6 +129,11 @@ var Commands = []cli.Command{
 		Action: cmdRm,
 	},
 	{
+		Name:   "shellinit",
+		Usage:  "Display the shell commands to set up the Docker client",
+		Action: cmdShellinit,
+	},
+	{
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "command, c",
@@ -190,7 +202,7 @@ func cmdCreate(c *cli.Context) {
 		log.Fatal("You must specify a machine name")
 	}
 
-	store := NewStore(c.GlobalString("storage-path"), c.GlobalString("auth-ca"), c.GlobalString("auth-key"))
+	store := NewStore(c.GlobalString("storage-path"), c.GlobalString("tls-ca-cert"), c.GlobalString("tls-ca-key"))
 
 	host, err := store.Create(name, driver, c)
 	if err != nil {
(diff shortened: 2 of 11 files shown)


utils/certs.go
```diff
@@ -93,7 +93,7 @@ func GenerateCert(hosts []string, certFile, keyFile, caFile, caKeyFile, org stri
 	}
 	// client
 	if len(hosts) == 1 && hosts[0] == "" {
-		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}
+		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth}
```
cpuguy83: Setting this for ServerAuth here unfortunately doesn't do much since we aren't setting any hosts for it to be valid for.
If you try to use it for ServerAuth you end up with an error like so: `cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs` or `certificate is valid for , not localhost`

cpuguy83: Maybe the server keys can be created to do both ClientAuth and ServerAuth, and then to setup a swarm cert people can use the "none" driver, and use those server keys for swarmd?

cpuguy83:
What do you think about removing the server tls keys once they've been uploaded to the created machine?

utils/certs.go
ehazlett: Yes, I was actually seeing this in testing.  I will dig into the keys more.

As for the none driver, the way it's setup is to create the server keys per machine.  I don't think we want to re-use a separate set of server keys. 

So for the server certs to work with swarm, they just need the IP SAN of the swarm master right or am I misunderstanding?

ehazlett:
I think this would be ok.  I don't think we would need them after the machine is created.

utils/certs.go
cpuguy83: Correct, basically whatever address the user is using to connect to the swarm master.
This is why I was thinking about the "none" driver, which would let someone create a "machine" but for their swarm master... But I suppose that is a hack.

sthulb:
LGTM.. swarm is working well with this setup.

Host 1:
`docker run -P --rm sthulb/swarm join --discovery token://b55dd7af67426471e6118b9535fa41d0 --addr 172.16.50.156:2376`

Host 2: 
docker run -P --rm sthulb/swarm join --discovery token://b55dd7af67426471e6118b9535fa41d0 --addr 172.16.50.156:2376

Master:
```shell
swarm manage --tls --tlscacert ~/.docker/machines/ca.pem --tlscert ~/.docker/machines/.client/cert.pem --tlskey ~/.docker/machines/.client/key.pem --discovery token://b55dd7af67426471e6118b9535fa41d0 -H tcp://0.0.0.0:4243

./machine create -d none --url tcp://localhost:4243 master
```

sthulb:
I'd perhaps argue that the certs shouldn't have the issuer of "docker", but the username of the active user.

ehazlett:
@sthulb what about an issuer of "docker-machine"?  that seems to make sense to me.

sthulb:
That's a bit better. It just has to be clear that it's not actually a docker inc cert

ehazlett:
@sthulb now that I think about it, I like the username better.  updating.  thx!

sthulb:
You're welcome!

ehazlett:
LGTM

bfirsh:
Woot!

ryanwalls:
This https://github.com/docker/machine/pull/402#issuecomment-71541533 isn't working for me.  I startup my swarm manager with this:

```
swarm manage --tls --tlscacert ~/.docker/machines/ca.pem --tlscert ~/.docker/machines/.client/cert.pem --tlskey ~/.docker/machines/.client/key.pem  -H tcp://0.0.0.0:4444 token://<my token>
```

Then I run this:

```
docker-machine create -d none --url tcp://localhost:4444 master
```

Then I try to run something like this:

```
docker -H tcp://localhost:4444 ps
```

And I get this error on the swarm manager:
```
http: TLS handshake error from 127.0.0.1:58658: remote error: bad certificate
```

And on docker client side I get:
```
FATA[0000] An error occurred trying to connect: Get https://localhost:4444/v1.16/containers/json: x509: certificate is valid for , not localhost
```

Any ideas on what I'm doing wrong?

ehazlett:
There is an update coming soon (it's in my branch "swarm" if you want to
test) that adjusts the cert ext usage.  This should fix it.

ryanwalls:
@ehazlett FYI, the update works if I do the following:
```
docker $(docker-machine config master) run -d -p 8080:80 nginx
```

But fails if I run the shellinit first
e.g. 
```
$(docker-machine shellinit master)
docker run -d -p 8080:80 nginx
```
Gives me this error:
```
x509: certificate is valid for , not localhost
```

sthulb:
The shellinit command will need to wrapped in eval. Otherwise it does nothing

ehazlett:
@ryanwalls that should work (I work with it both ways).  I'm about to release the RC2 -- would you mind testing on that?

PR #406, opened 2015-01-26
Generic Driver

ehazlett:
This adds support for a "generic" driver.  This is very basic at the moment.  Basically you just need an Ubuntu server with SSH.  From there it provisions as normal.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1000,7 +1000,11 @@ Options:
 The DigitalOcean driver will use `ubuntu-14-04-x64` as the default image.
 
 #### Generic
-Create machines using an existing VM/Host with SSH
+Create machines using an existing VM/Host with SSH.
+
+This is useful if you are using a provider that Machine does not support
+directly or if you would like to import an existing host to allow Docker
+Machine to manage.
 
 Options:
 
@@ -1009,7 +1013,7 @@ Options:
  - `--generic-ssh-key`: Path to the SSH user private key
  - `--generic-ssh-port`: Port to use for SSH (default: `22`)
 
-> Note: you must use a base Operating System supported by Machine
+> Note: you must use a base Operating System supported by Machine.
 
 #### Google Compute Engine
 Create machines on [Google Compute Engine](https://cloud.google.com/compute/).  You will need a Google account and project name.  See https://cloud.google.com/compute/docs/projects for details on projects.
--- a/docs/index.md
+++ b/docs/index.md
@@ -999,6 +999,18 @@ Options:
 
 The DigitalOcean driver will use `ubuntu-14-04-x64` as the default image.
 
+#### Generic
+Create machines using an existing VM/Host with SSH
+
+Options:
+
+ - `--generic-ip-address`: IP Address of host
+ - `--generic-ssh-user`: SSH username used to connect (default: `root`)
+ - `--generic-ssh-key`: Path to the SSH user private key
+ - `--generic-ssh-port`: Port to use for SSH (default: `22`)
+
+> Note: you must use a base Operating System supported by Machine
+
 #### Google Compute Engine
 Create machines on [Google Compute Engine](https://cloud.google.com/compute/).  You will need a Google account and project name.  See https://cloud.google.com/compute/docs/projects for details on projects.
 
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -10,7 +10,6 @@ import (
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/drivers"
 	"github.com/docker/machine/provider"
-	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/state"
 	"github.com/docker/machine/utils"
 )
@@ -144,11 +143,7 @@ func (d *Driver) Create() error {
 		return err
 	}
 
-	log.Infof("Waiting for SSH...")
-
-	if err := ssh.WaitForTCP(fmt.Sprintf("%s:%d", d.IPAddress, d.SSHPort)); err != nil {
-		return err
-	}
+	log.Debugf("IP: %s", d.IPAddress)
 
 	return nil
 }
@@ -195,11 +190,7 @@ func (d *Driver) Remove() error {
 func (d *Driver) Restart() error {
 	log.Debug("Restarting...")
 
-	cmd, err := drivers.GetSSHCommandFromDriver(d, "sudo shutdown -r now")
-	if err != nil {
-		return err
-	}
-	if err := cmd.Run(); err != nil {
+	if _, err := drivers.RunSSHCommandFromDriver(d, "sudo shutdown -r now"); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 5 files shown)


thaJeztah:
Travis is failing

ehazlett:
I'll check thx

ehazlett:
Fixed tests.

hansbogert:
Perhaps the README.md in your branch should elaborate on the generic driver under the driver heading.

andrewwatson:
Should this driver fail if you try to use it as root?  Should it try one thing first that uses sudo to see if it works and then kindly inform the user that all hope is lost if it fails?

drivers/generic/generic.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return err
+	}
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+
+	log.Debug("Installing Docker")
+
+	cmd, err = d.GetSSHCommand("if [ ! -e /usr/bin/docker ]; then curl get.docker.io | sudo sh -; fi")
```
andrewwatson: this seems like a number of things could go wrong and it would be hard to know what it was... the curl could fail or you might not have sudo permissions...

drivers/generic/generic.go
```diff
@@ -0,0 +1,269 @@
...
+
+	return nil
+}
+
+func (d *Driver) GetURL() (string, error) {
+	ip, err := d.GetIP()
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("tcp://%s:2376", ip), nil
```
andrewwatson: should be noted in the README that TLS has to be set up properly, right?

drivers/generic/generic.go
```diff
@@ -0,0 +1,269 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (d *Driver) StartDocker() error {
+	log.Debug("Starting Docker...")
+
+	cmd, err := d.GetSSHCommand("sudo service docker start")
```
andrewwatson: this will only work on certain flavors of Linux.  Should the driver detect the flavor and bail out or just note it in the README and let people hit the wall at full speed? :)

ehazlett:
Updated to latest driver spec.

andrewwatson:
maybe it's just me but when i pull this locally and run "go build" i get a failure: 

```
drivers/openstack/client.go:423: provider.HTTPClient undefined (type *gophercloud.ProviderClient has no field or method HTTPClient)
```

ehazlett:
@andrewwatson you need to build with `godep` -- preferably using `script/build`

hairyhenderson:
So... what remains for this to get merged? It'd be awesome if this could be part of 0.3.0 :)

andrewwatson:
FYI I've been using this driver a lot the past week or so and it seems to
work really well. I've been driving machine from a local-exec provisioner
in terraform and loving it.

ehazlett:
@andrewwatson thx for the feedback!

andrewwatson:
I just submitted a PR to ehazlett/machine for a change to use DialTimeout() but other than that I think this is ready to ship!

ehazlett:
@nathanleclaire PTAL

nathanleclaire:
Will do!

ehazlett:
rebased to latest

nathanleclaire:
Maybe we could default the `--generic-ssh-key` option to `$HOME/.ssh/id_rsa`?

ehazlett:
@nathanleclaire +1

ehazlett:
@nathanleclaire added; PTAL

drivers/generic/generic.go
nathanleclaire: The `generic` driver takes care of it automatically due to the built-in provisioning.

libmachine/provision/ubuntu.go
```diff
@@ -75,7 +75,7 @@ func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.Pack
 
 func (provisioner *UbuntuProvisioner) dockerDaemonResponding() bool {
 	if _, err := provisioner.SSHCommand("sudo docker version"); err != nil {
-		log.Warn("Error getting SSH command to check if the daemon is up: %s", err)
+		log.Warnf("Error getting SSH command to check if the daemon is up: %s", err)
```
nathanleclaire: Nice catch :)

drivers/generic/generic.go
```diff
@@ -0,0 +1,221 @@
...
+	log.Debug("Restarting...")
+
+	if _, err := drivers.RunSSHCommandFromDriver(d, "sudo shutdown -r now"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d *Driver) Kill() error {
```
nathanleclaire: Sure we don't want to just have this run `sudo rm --no-preserve-root -rf /`?  JUST KIDDING!!!!!

drivers/generic/generic.go
```diff
@@ -0,0 +1,221 @@
...
+func (d *Driver) GetSSHPort() (int, error) {
+	if d.SSHPort == 0 {
+		d.SSHPort = 22
+	}
+
+	return d.SSHPort, nil
+}
+
+func (d *Driver) GetSSHUsername() string {
+	if d.SSHUser == "" {
```
nathanleclaire: In what situation would these be empty?  I can't think of why the user would set them that way.

drivers/generic/generic.go
```diff
@@ -0,0 +1,221 @@
...
+		cli.IntFlag{
+			Name:  "generic-ssh-port",
+			Usage: "SSH port",
+			Value: 22,
+		},
+	}
+}
+
+func NewDriver(machineName string, storePath string, caCert string, privateKey string) (drivers.Driver, error) {
+	return &Driver{MachineName: machineName, storePath: storePath, CaCertPath: caCert, PrivateKeyPath: privateKey}, nil
```
nathanleclaire: Can we put this struct across several lines please (the "pretty" way)?

docs/index.md
```diff
@@ -999,6 +999,18 @@ Options:
...
+Create machines using an existing VM/Host with SSH
+
+Options:
+
+ - `--generic-ip-address`: IP Address of host
+ - `--generic-ssh-user`: SSH username used to connect (default: `root`)
+ - `--generic-ssh-key`: Path to the SSH user private key
+ - `--generic-ssh-port`: Port to use for SSH (default: `22`)
+
+> Note: you must use a base Operating System supported by Machine
```
nathanleclaire: Missing a period here.

docs/index.md
```diff
@@ -999,6 +999,18 @@ Options:
 
 The DigitalOcean driver will use `ubuntu-14-04-x64` as the default image.
 
+#### Generic
+Create machines using an existing VM/Host with SSH
```
nathanleclaire: Missing a period here.  I'd also consider adding a little bit of explanation, for instance : "This is useful if you are using a provider Machine does not support directly or in the case that you want to use another tool to create the machines but import them into Docker Machine for usage".

nathanleclaire:
Have some comments, but all around looking good and really fun.

nathanleclaire:
I'm wondering if we might want to trim down the timeout on `GetState` to ~1 second.  Generally for me `ls` returns in under a second, and with this if you have even one generic host which isn't around any more, your `ls` runs will take a minimum of 5 seconds.  That's pretty long time to just hang at the CLI.

andrewwatson:
Sounds good to me.

ehazlett:
@nathanleclaire rebased and reduced GetState to 1s -- PTAL

ehazlett:
@nathanleclaire ok i think i hit all of your comments.  PTAL.  thanks!

drivers/generic/generic.go
ehazlett: Yes, I'm sure ;)

nathanleclaire:
LGTM

ehazlett:
@nathanleclaire "Generic is coming..."

PR #412, opened 2015-01-27
Ensure server cert generation uses env variables

### LINKED ISSUE #411: Error creating machine: error generating server cert with $MACHINE_TLS_CA_{KEY|CERT} set
`Host.ConfigureAuth` doesn't take into consideration `$MACHINE_TLS_CA_KEY` and `$MACHINE_TLS_CA_CERT` env variables when changing CA cert locations.

If the environment variables are set and you have valid client keys, the server is created and machine fails with the following error:

```
ERRO[0236] Error creating machine: error generating server cert: open /home/vagrant/.docker/machines/key.pem: no such file or directory
```

I think the culprit is in line 126 of host.go

The quick fix is to copy these keys into `/.docker/machines/{ca|cert}.pem`.

tonyhb:
If env variables exist for the CA and CA key paths we need to use these to
generate the server certificates instead of the defaults.

This should fix #411.

### DIFF
--- a/host.go
+++ b/host.go
@@ -123,8 +123,6 @@ func (h *Host) ConfigureAuth() error {
 		return err
 	}
 
-	caCertPath := filepath.Join(utils.GetMachineDir(), "ca.pem")
-	caKeyPath := filepath.Join(utils.GetMachineDir(), "key.pem")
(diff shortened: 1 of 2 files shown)


tonyhb:
Using another fix as @ehazlett pointed out in #411

PR #413, opened 2015-01-27
check properly for username on different platforms

### LINKED ISSUE #410: Machine fails to build from OS X
The current master is not building properly when using `script/build`.  The resulting binary appears to be fine but simply exits with a `0` status.  When using `godep go build` local from OS X the binary is fine and works as expected.  Linux appears to be working fine either way.

Thanks to @mbentley for reporting and debugging.

ehazlett:
This fixes the GetUsername util to properly check for usernames on different platforms.

Fixes #410

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,16 +11,12 @@ import (
 )
 
 func before(c *cli.Context) error {
-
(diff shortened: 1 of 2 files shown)


ehazlett:
LGTM

PR #414, opened 2015-01-27
use correct paths for custom TLS CA cert and key

### LINKED ISSUE #411: Error creating machine: error generating server cert with $MACHINE_TLS_CA_{KEY|CERT} set

ehazlett:
This fixes a bug where custom TLS CA cert and key is being ignored.

Closes #411

### DIFF
--- a/host.go
+++ b/host.go
@@ -123,8 +123,6 @@ func (h *Host) ConfigureAuth() error {
 		return err
 	}
 
-	caCertPath := filepath.Join(utils.GetMachineDir(), "ca.pem")
-	caKeyPath := filepath.Join(utils.GetMachineDir(), "key.pem")
 	serverCertPath := filepath.Join(h.storePath, "server.pem")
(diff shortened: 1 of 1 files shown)


ehazlett:
sthulb tested and is good (as reported in the issue).  LGTM

tonyhb:
Apologies — running through some releasing and testing today, meant to crack on with this last night!

ehazlett:
@tonyhb np -- please test if you can and report here to confirm :D

PR #418, opened 2015-01-28
check returned host of GetActive to avoid crash

wwwjfy:
when no active host, GetActive returns (nil, nil)


ehazlett:
Error handling was added as part of a larger refactor in https://github.com/docker/machine/commit/a60f0e672e3e22b62eb4a561b4f0cab4a6548236.  Thanks for the submission!

PR #423, opened 2015-01-28
add suspend command

wwwjfy:
Other drivers probably have this functionality, too, but I've only implemented for Vmware Fusion and VirtualBox


ehazlett:
I'm torn on this as a global command since not all drivers support it.

Thoughts @sthulb ?

sthulb:
I'd rather see it implemented as a flag on stop. So you end up with `machine stop --pause`

sthulb:
But I immediately regret it :p

Perhaps a better wording than pause would make more sense.

wwwjfy:
I don't see much a difference a flag on `stop` makes more clear for unsupported drivers. A message/info will be shown anyway. But if you think it's better, what about `matching stop --save`, to be same with the state `Saved`?

sthulb:
--save seems better.

Tebro:
I was hacking at this for VMWare fusion earlier (since boot2docker image was not persistant), my work can be found here: https://github.com/Tebro/machine/tree/feature/vmwarefusion-save, feel free to use it.

nathanleclaire:
I'm -1 on this because it can't be implemented in a uniform way.

nathanleclaire:
Going to close this since there's no way to implement for DigitalOcean, Amazon, etc.

PR #424, opened 2015-01-28
Download ISO from file schema - file:///...

rickard-von-essen:
Allow boot2docker-url arguments to take a local file as ISO. This simplifies testing of boot2docker.iso build.

### DIFF
--- a/utils/b2d.go
+++ b/utils/b2d.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 )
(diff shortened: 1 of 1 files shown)


ehazlett:
Thanks!

Could we add some checks around if a scheme exists (I mistakenly just tried to use a path and it failed) or perhaps if `http` isn't there then attempt to fallback to file automatically?

rickard-von-essen:
I'll update this to default to file if ```url.Scheme == nil```

I also did some testing with ```net/url.Parse``` and Windows paths on go playground and it didn't work very well with driver letters and backslashes. So someone with access to Windows should create a patch  with specific Windows path handling.

rickard-von-essen:
Done, now also posix paths work without the ```file://``` prefix.

ehazlett:
+1 for working without `file` scheme.  I think the last piece is the windows paths.  I will see if I can take a look.

nathanleclaire:
This should work now in master, so I'm closing this PR.

PR #427, opened 2015-01-28
remove name option from azure driver

ehazlett:
This removes the `name` option from the azure driver as it is automatically set to the machine name now.


ehazlett:
LGTM

PR #437, opened 2015-01-29
add timeout for b2d release check; warn instead of error if cannot check

### LINKED ISSUE #435: Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: lookup api.github.com: no such host
Hi All,
    I'm testing the docker machine tool, it looks really good, yesterday I could create a machine using VirtualBox driver, the image boot2docker.iso is stored in my filesystem.
Now I trying to create a new virtualbox machine, but I don't have internet acces from the current network, so my question is, Is there any parameter to say "ehhh don't ask for the latest image available, use my current image that is stored in my defaul location" ?

This is the error that I get when i try to create the machine:

 ./docker-machine create -d virtualbox dockerDEV2
ERRO[0000] Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: lookup api.github.com: no such host 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine

Basically the idea is to force docker machine to use my downloaded image, is it possible ?

Thanks

ehazlett:
This adds a timeout to the check for the latest b2d release.  It also changes to a "warn" in the event of an error (internet issues, etc).

Fixes #435

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -143,7 +143,7 @@ func (d *Driver) Create() error {
 		// until then always use "latest"
 		isoURL, err = utils.GetLatestBoot2DockerReleaseURL()
 		if err != nil {
-			return err
+			log.Warnf("Unable to check for the latest release: %s", err)
 		}
 
 		// todo: use real constant for .docker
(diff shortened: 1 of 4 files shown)


ehazlett:
LGTM

nathanleclaire:
Hey @ehazlett, did you ever take care of this?

> let's make sure to add some code to bail in the case that no local .iso exists and the network times out. Otherwise we try to careen into an ISO download that won't work.

ehazlett:
Yes an error would be thrown but I will add an informational message as well.

PR #438, opened 2015-01-29
Change message to use machine env.

Tebro:
So basically this is just a UI fix that informs users to use machine env, instead of docker $(machine config) to configure their docker client to connect to the machine.

This is my suggestion, the comments section is for discussion if there is a better way to say it.


ehazlett:
@Tebro this looks good -- could you squash your commits?  thx!

ehazlett:
LGTM

PR #439, opened 2015-01-29
[WIP] Support for exporting environment variables (e.g. HTTP_PROXY support)

tombee:
On our corporate network, we're required to specify the HTTP_PROXY and HTTPS_PROXY environment variables in order to fetch the images we are attempting to deploy as containers on the provisioned machines.

This applies to both our local environments using the VirtualBox driver and our staging environments that use vSphere.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -76,6 +76,11 @@ var Commands = []cli.Command{
 				),
 				Value: "none",
 			},
+			cli.StringSliceFlag{
+				Name:  "environment, e",
+				Usage: "Environment variables to export. e.g. -e HTTP_PROXY=http://example.org:8080 -e HTTPS_PROXY=http://example.org:8080",
+				Value: &cli.StringSlice{},
+			},
 		),
 		Name:   "create",
 		Usage:  "Create a machine",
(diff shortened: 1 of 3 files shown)


ehazlett:
Thanks!  I think this is a good start for a discussion around configuring the Docker daemon and supporting things like the environment.

These questions are not necessarily aimed at you, I'm just asking to start a discussion :)

In this case, this is specific to Virtualbox / Boot2Docker.  How would we add these to the other providers (without `/var/lib/boot2docker/profile`)?

I'm also cautious of what this might open up.  For example, in the case of setting system environment variables, this sort of crosses over into a config management area.  I do think we need to be able to configure the Docker daemon but I'm not sure about other system aspects.

One other possible solution is to use cloudinit.  We want to move in that direction anyway and that would allow almost any customization like this from the user.

tombee:
@ehazlett 
Cool, thanks.

From what I could see in host.go, "virtualbox", "vmwarefusion", "vmwarevsphere" drivers use the /var/lib/boot2docker/profile location in the generateDockerConfig method, for anything else it writes the config out to /etc/default/docker.

ehazlett:
actually it's a little different for the non-b2d hosts.  For non-b2d, we use `/etc/default/docker` to provide the options for the Docker daemon -- therefore the `HTTP_PROXY` etc would not work there.

Also, thinking ahead, another reason to support cloudinit is in the case of different distros.  For example, if you use a distro with systemd or non-ubuntu machine won't work because it most likely doesn't support `/etc/default/docker`.

tombee:
Cool -- well, sounds like cloudinit is the way to go :smile:.  Let me know if I should close the PR.

PR #440, opened 2015-01-29
use hostname for azure instead of generating

### LINKED ISSUE #433: [Azure driver] Allow custom DNS (cloud service) names
Recently the `--azure-name` parameter got removed (#219) from machine code and azure driver code now creates the Azure Cloud Service (DNS to the VM) in `machinename-YYYYMMDDHHmmss.cloudapp.net` format:

> Each driver will attempt to set the hostname for the VM, except for Azure, which sets it based on the DNS name.

This is done at: https://github.com/ehazlett/machine/commit/861b16dba9020ae6cb3c62e699683db9554300cc#diff-93a074128ed8940ea0dafbd000270978R170

I think this should not be done and user should be allowed to specify the DNS names (as it used to be). In my case, I want to spin up 10 different VMs like `mydocker-1`, `mydocker-2`... on Azure and I will be able to easily address them like `ssh -p 22 mydocker-1.cloudapp.net`. However this change gives me VM names like `mydocker-1-20150127152433`, `mydocker-2-20150127152750` which are not easily readable and manageable.

Another case is, if I want to deploy a VM like `ahmetsblog.cloudapp.net`, there is no way to do this with docker-machine. I think just limits the flexibility of the user.

I see that the date suffix is added to prevent the chances that the DNS name is already registered. I think the solution here should be giving the `--azure-name` parameter back to the user so that users can have a better control on what's going on. If the DNS name is already taken, user can simply be alerted "Cloud service name is already taken". (I think it used to be like that before #219) If it's really tedious and redundant, code can allow that parameter to be optional and can still use `fmt.Sprintf("%s-%s", driver.MachineName, date)`.

Also, this decision forces MachineName to comply with Azure Cloud Service Name rules (e.g. 3-25 chars and can contain only chars allowed in DNS). The date suffix already takes up 14 chars, so user has to limit their MachineName to *11 chars max*. That sounds wrong.

I'm not sure how the decision is made here but here's my 2c.

cc: @jeffmendoza @ehazlett @sthulb @chanezon
followup to #419.

ehazlett:
This removes the date append to the name of a machine in Azure.  Instead, the `PreCreateCheck` checks to see if the name is available and meets the Azure requirements.  If it doesn't an error will be reported and the machine will not be created.

Fixes #433

### DIFF
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -108,16 +108,7 @@ func GetCreateFlags() []cli.Flag {
 }
 
 func NewDriver(machineName string, storePath string, caCert string, privateKey string) (drivers.Driver, error) {
-	t := time.Now().Format("20060102150405")
-	name := fmt.Sprintf("%s-%s", machineName, t)
-
-	// trim name to 24 chars due to the azure dns name limit
-	if len(name) > 24 {
-		name = name[0:24]
-
-	}
-
-	driver := &Driver{MachineName: name, storePath: storePath, CaCertPath: caCert, PrivateKeyPath: privateKey}
+	driver := &Driver{MachineName: machineName, storePath: storePath, CaCertPath: caCert, PrivateKeyPath: privateKey}
 	return driver, nil
 }
 
@@ -173,6 +164,20 @@ func (driver *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 }
 
 func (driver *Driver) PreCreateCheck() error {
+	if err := driver.setUserSubscription(); err != nil {
+		return err
+	}
+
+	// check azure DNS to make sure name is available
+	available, res, err := vmClient.CheckHostedServiceNameAvailability(driver.MachineName)
+	if err != nil {
+		return err
+	}
+
+	if !available {
+		return fmt.Errorf(res)
+	}
+
 	return nil
 }
 
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -169,13 +169,13 @@ func (driver *Driver) PreCreateCheck() error {
 	}
 
 	// check azure DNS to make sure name is available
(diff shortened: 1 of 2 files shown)


ahmetalpbalkan:
@ehazlett lgtm overall. but you sure you want to restrict the machineName to the rules of Azure DNS name (like max 24 chars, has to be lowercase etc)? 

We should also probably drop a note saying

> the machine name will be used as DNS name for the Cloud Service (e.g. machinename.cloudapp.net)

in the [README](https://github.com/docker/machine#microsoft-azure).

ehazlett:
@ahmetalpbalkan well it's going to fail otherwise at the time of creation?

+1 for the readme -- I'll get that updated.

Thanks!

ahmetalpbalkan:
Well, if you allow user to keep Azure DNS Name `--azure-name` and Machine Name separate, you'll get full flexibility.

drivers/azure/azure.go
```diff
@@ -173,6 +164,20 @@ func (driver *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 }
 
 func (driver *Driver) PreCreateCheck() error {
+	if err := driver.setUserSubscription(); err != nil {
+		return err
+	}
+
+	// check azure DNS to make sure name is available
+	available, res, err := vmClient.CheckHostedServiceNameAvailability(driver.MachineName)
+	if err != nil {
```
ahmetalpbalkan: `res` would be better as `reason`, here it looks like `result`.

ehazlett:
that seems odd to me especially since we don't allow any other naming like that.  I think having it validate and throw the error at least allows for the customization and is consistent with the other drivers.

ahmetalpbalkan:
okay, sounds good. this may restrict the machine naming format people tend to follow in other cloud providers but I think it's better now since we got rid of extra 15 characters in the DNS name and it will always match the machine name.

sthulb:
LGTM

PR #441, opened 2015-01-29
Issue #435, new cmd parameter --ignoreISOUpdate

roldancer:
Using this global parameter  --ignoreISOUpdate  for virtualbox driver, it doesn't check for ISO updates, this parameter could be useful for private network without or limited internet access.

I have tested this parameters in my private network, it works ok.

./docker-machine_linux-amd64 --ignoreISOUpdate  create -d virtualbox dockerDEV3
INFO[0000] Creating VirtualBox VM...                    
INFO[0010] Starting VirtualBox VM...                    
INFO[0010] Waiting for VM to start...                   
INFO[0047] "dockerDEV3" has been created and is now the active machine 
INFO[0047] To connect: docker $(docker-machine_linux-amd64 config dockerDEV3) ps


drivers/virtualbox/virtualbox.go
```diff
@@ -131,27 +131,37 @@ func (d *Driver) Create() error {
...
-			return err
-		}
-	} else {
-		// todo: check latest release URL, download if it's new
-		// until then always use "latest"
-		isoURL, err = utils.GetLatestBoot2DockerReleaseURL()
-		if err != nil {
-			return err
+	// I have to ingnore ISO download if cmd has --ignoreISOUpdate
+	for _, f := range os.Args {
```
ehazlett: This should use the args in `commands.go` to add another arg for the virtualbox driver such as `--virtualbox-skip-update` or similar.

ehazlett:
This would also need added to the `vmwarefusion` driver and `vmwarevsphere` drivers.  We should probably refactor to move this to a common function.

roldancer:
Hi @ehazlett i could try to add the flag --virtualbox-skip-update to drivers, virtualbox, vmwarefusion, vmwarevsphere, what do you think ?

ehazlett:
Yeah like `--virtualbox-skip-update`, `--vmwarefusion-skip-update` etc.

sthulb:
If we're making things consistent, --vmwarefusion-memory-size & --virtualbox-memory could do with help :)

ehazlett:
@sthulb +1

ehazlett:
This needs `gofmt` and signed.  See https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work for details.  Thanks

PR #442, opened 2015-01-29
Fix network timeout preventing machine creation

nathanleclaire:
On an airplane, I attempted to create a machine using the VBox driver,
forgetting that we actually presume that network is available for a call
to Github to get the latest boot2docker release.

Users should not need to have access to the Internet in order to use
local drivers.  This fixes the issue and makes creation of a VM without
network connectivity possible :thumbsup:

cc @ehazlett @bfirsh @sthulb 

(To test this try disabling WiFi and running through `machine create -d virtualbox dev` before and after the patch)

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -28,6 +28,11 @@ const (
 	dockerConfigDir = "/var/lib/boot2docker"
 )
 
+var (
+	errNoIso              = fmt.Errorf("There was no boot2docker.iso found locally")
+	errReleaseCallTimeout = fmt.Errorf("Network call for boot2docker release timed out")
+)
+
 type Driver struct {
 	MachineName    string
 	SSHPort        int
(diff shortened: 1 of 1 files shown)


ehazlett:
Related to https://github.com/docker/machine/pull/437 and https://github.com/docker/machine/pull/441.

I'm up for closing mine but I'm not sure grep'g the error is the way to go.

sthulb:
I prefer this over #435 (--ignoreISOUpdate).

nathanleclaire:
@ehazlett Ah, I didn't see that.  Actually I like yours better in general (it's cleaner and covers other drivers) - but let's make sure to add some code to bail in the case that no local `.iso` exists and the network times out.  Otherwise we try to careen into an ISO download that won't work.

sthulb:
#437 seems nicer still.

ehazlett:
Ah thanks :)  Good idea -- I'll update for that.

ehazlett:
Closing.  Will track this fix in #437.

PR #443, opened 2015-01-29
Add shell function recommendation

nathanleclaire:
cc @bfirsh @ehazlett @sthulb 

I find this a lot easier and more pleasant to use than the other
currently supported methods (`env` and `config`) and may serve us well
as a stopgap until/if we can get true config files for Docker.  It's
definitely worth bringing up to users since the workflow is so nice (I
find that I forget the active machine often and want a way to access it
through the same interface I use to run Docker commands against it).

@bfirsh hopefully doesn't think I am trying to sneak in the "use" syntax I originally mentioned wanting :smile: .  I am not wildly attached to that part if you guys decide it is not best judgment (although I do find it _really_ nice to use as I forget my active machine quite often and want a quick way to see it) - the most important is to have something like `docker $(docker-machine config $(docker-machine active)) "$@"` in the function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ You can also get the commands to export environment variables to use with the Do
 
 Machine is still in its early stages. If you'd like to try out a preview build, [download it here](https://github.com/docker/machine/releases/latest).
 
+For information about getting started and an overview of all the Machine
+commands available to you, [please refer to the
+documentation](https://github.com/docker/machine/blob/master/docs/dockermachine.md).
+Note that the documentation in its current state is an incomplete "rough draft".
+
 ## Drivers
 
 ### VirtualBox
--- a/docs/dockermachine.md
+++ b/docs/dockermachine.md
@@ -262,6 +262,81 @@ NAME        ACTIVE   DRIVER    STATE     URL
 custombox   *        none      Running   tcp://50.134.234.20:2376
 ```
 
+## Shell Function Shortcut
+
+When you start using Machine, it is likely that you will want to create a few
+different hosts and run Docker commands against them using the subshell method
+suggested by Machine's console output, i.e. `docker $(docker-machine config dev)
+ps`.  However, that is a fair bit of typing and it can be difficult to remember
+which host is considered "active" by Machine.
+
+Therefore, we suggest that you may want to use a shell function such as
+the following to make switching machines and running commands against them
+easier and more fluid:
+
+```
+dkr () {
+    if [[ "$1" == "which" ]]; then
+        echo $(docker-machine active)
+        return
+    fi
+    if [[ "$1" == "use" ]]; then
+        docker-machine active "$2"
+        return
+    fi
+    docker $(docker-machine config $(docker-machine active)) "$@"
+}
+```
+
+To make this function available to you at all times, simply add it to your
+`.bashrc` or `.zshrc` file. This has been tested on `bash` and `zsh` and should
+work well with both (assuming there is no existing alias or function for `dkr`).
+
+With this function, you can run Docker commands as you normally would, except
+that you simply type `dkr` instead of `docker` and they will automatically run
+against the active machine.  For example:
+
+```
+$ dkr version
+Client version: 1.4.1
+Client API version: 1.16
+Go version (client): go1.3.3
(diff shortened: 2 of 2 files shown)


ehazlett:
I hope I don't offend anyone here, but why do we have "active" anyway (`/me ducks`) ?  I really don't see why it's necessary especially if you just export the settings (which also would mean no custom aliases, functions etc as above).

I'm not saying this isn't a good idea :)  I actually do something similar in my own `bashrc` now :)

bfirsh:
Active machines make sense when the Docker client configures itself automatically. We never got to that point though, unfortunately.

nathanleclaire:
I have a general thrust in that direction with https://github.com/docker/docker/pull/10151, but these things take time - I agree that active machine isn't really useful in the current implementation, but I really like the general principle as a workflow.  I'm pretty pleased with the shell function, though naturally I'm heavily biased.  Perhaps we could have a few different methods available for the alpha/beta releases, and see how things shake out in the community.

ehazlett:
As I said, I'm not opposed to it -- just trying to get on the same page.  Perhaps I'm just using machine wrong -- I tend to be explicit with commands by always specifying the machine name instead of using the "active" one.

nathanleclaire:
> As I said, I'm not opposed to it -- just trying to get on the same page. Perhaps I'm just using machine wrong -- I tend to be explicit with commands by always specifying the machine name instead of using the "active" one.

Right - to be clear from my end I think always being explicit about which machine you want to use with commands like `stop`, `start`, `ssh` and so on seems like the correct approach- I find "active" machine concept really useful for selecting a machine to run Docker commands against specifically.  The `config` method is pretty cumbersome to type (granted, I'm lazy) and `env` makes it is pretty easy to forget which one you've selected to run commands against (granted, I'm forgetful).

shykes:
One possibility: perhaps it's `machine` which will end up exposing its configuration in some way, and Docker will consume that, instead of the other way around. Then the "active" marker would be stored on the side of the Docker client, and machine wouldn't have to worry about it.

ehazlett:
That's an interesting idea.  I think the concept of "active" in docker machine causes confusion.  For example, if we remove the active machine then we get "error not active machine set" when things still work.  We also have to account for the varying invocations -  i.e. `machine url` and `machine url <name>` etc.  It just seems odd to me.  

However, if it makes sense to everyone else, I'm cool with it :)

Thoughts? /cc @sthulb @ggiamarchi @cpuguy83 @jeffmendoza @ ahmetalpbalkan @smashwilson

sthulb:
I'm not fond of the `active` command. I think enforcing the name param is nicer. When either Docker or Machine consumes a config file, we should perhaps have a `docker-machine config` command to setup docker or machine.

cpuguy83:
@ehazlett This is what I was thinking, at the very least the Docker client can see if there is a ~/.docker/active and use that, assuming no DOCKER_HOST or -H is present.

The docker client could have a concept of machines (or maybe hosts would be better here) even, which docker-machine worries about creating and populating. But on the Docker side it's litterally for selecting a known host and integration with supporting tools, and nothing else.
I think it's perfectly reasonable to have this functionality in Docker.

jeffmendoza:
I'm on the side where the docker client looks for a host in the order of
1. CLI arg
2. Env var
3. Config file

And machine would configure the file. In that case, `machine active` makes sense. Machine is built on top of docker, not the other way around. The docker client shouldn't know or care of machine's existence.

That said, I can see how having multiple named hosts in the docker config file would benefit docker as-is, and could be integrated with by other utilities than machine.

ehazlett:
Ok.  So I get the eventual usage for "active".  However, I'm not crazy of recommending a custom shell function for usage.  I think it deters from the regular usage and will just cause confusion.  I helped troubleshoot an instance where a user had `ssh` wrapped in a shell function that inserted newlines causing everything to break.  I am not crazy of wrapping these.

Perhaps there is some other way we can ease usage without resorting to an outside function?

nathanleclaire:
Hey @ehazlett I think this is a little too controversial to merge so please close?

ehazlett:
Cool thx

Issue #445, opened 2015-01-30
script error during DOCKER_OPTS injection


noodlehaus:
hi. i'm trying to use machine with openstack, and when it gets to the point where the `DOCKER_OPTS` are being injected into the host, the following error occurs.

```
DEBU[0154] Command: %s[sudo sh -c 'echo "export DOCKER_OPTS=\'--tlsverify --tlscacert=/etc/docker/ca.pem --tlskey=/etc/docker/server-key.pem --tlscert=/etc/docker/server.pem --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:2376\'" | sudo tee -a /etc/default/docker']  MachineId=8ce46b4a-8931-432a-9431-a5e6b42eacfe
DEBU[0154] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -p 22 -i /Users/noodlehaus/.docker/machines/34ce975260d8831c577ce7db92a257a6/id_rsa stack@10.64.252.30 sudo sh -c 'echo "export DOCKER_OPTS=\'--tlsverify --tlscacert=/etc/docker/ca.pem --tlskey=/etc/docker/server-key.pem --tlscert=/etc/docker/server.pem --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:2376\'" | sudo tee -a /etc/default/docker' 
bash: -c: line 0: unexpected EOF while looking for matching `"'
bash: -c: line 1: syntax error: unexpected end of file
ERRO[0156] Error creating machine: exit status 1        
WARN[0156] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0156] Error creating machine
```

This is on OS X.

sthulb:
This is probably linked with #434

ehazlett:
@sthulb I don't think so.  #434 is for the `ssh` command while this is during provisioning.

ehazlett:
@ggiamarchi any idea here?

LaynePeng:
This bug is caused by \' cannot work. Please change two  \' to  " \" .

LaynePeng:
It is caused by \' . Change it to \" will work.

LaynePeng:
I did the fix myself:

index a28fc68..61537ae 100644
--- a/host.go
+++ b/host.go
[...]

Issue #434, opened 2015-01-29
SSH command should be arguments, not an option


bfirsh:
I should be able to do something like `machine ssh name df -h`. If we want to make the name optional, perhaps we could make a `-n` (name) or `-m` (machine) option for the name.

Also see confusion in #247.

ehazlett:
I think if we are going to do this then we should force the name to be present.  I personally don't like how we have the inconsistency in the commands where some work on the "active" machine as well as specified and then we have this where it gets more tangled.

I would say we keep it simple and just make the command `machine ssh <name> <command>` like you said first.

ehazlett:
@bfirsh the reason we use the `-c` is the cli package validates the flags.  For example, `machine ssh test-machine free` works but `machine ssh test-machine free -m` fails with `Incorrect usage`.  I'm digging into the cli package to see what we can do.

sthulb:
There isn't much if I recall. We can't tell if the first arg is a machine name or a command or perhaps both (machine named ls for example).

A potential solution is:
If arg[0] is a machine name, run arg[1:] on arg[0], if arg[0] isn't a machine, run all args on the active machine.

nathanleclaire:
I am +1 on args over `-c`- I think support for the flag parsing terminator (`--`) got merged into codegangsta/cli lately, so another solution available is that we might be able to bump the version and use that.  Then commands would look like `machine ssh dev -- df -h`.

ehazlett:
Oooh nice!

sthulb:
@nathanleclaire @ehazlett This doesn't really help... in @codegangsta/cli, all the -- option does is stop anything after -- being picked up and parsed as a flag :/

ehazlett:
Closing as #446 is merged.

PR #446, opened 2015-01-30
Use whatever is after -- as the SSH command.

sthulb:
Updates the codegansta/cli package.

Change the way commands are detected

Refs: #434, #445, #247


ehazlett:
LGTM Thanks!

PR #453, opened 2015-01-31
Add structure to generate code coverage representation

nathanleclaire:
This adds a script which will spit out code coverage information to the terminal as well as serve up a pretty slick browsable version.  See the README changes for usage.

The test coverage for machine is really scary skimpy right now, I hope that this will be a step in the right direction.  I think it should be mandatory for driver maintainers to write tests for their drivers.  Perhaps we can ping them to do so if/when this gets merged in.  More enablement will be needed, for instance we should be wrapping a lot more things that we are using from external libraries in `interface`s so that they can be manipulated for testing.  Some refactoring will definitely be needed, for instance the `cmd*` methods are so hardcoded to `codegangsta/cli` right now that unit testing them is something we should be thinking about sooner rather than later.

cc @ehazlett @bfirsh @sthulb 


Dockerfile
```diff
@@ -3,6 +3,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends openssh-client
 RUN go get github.com/mitchellh/gox
 RUN go get github.com/aktau/github-release
 RUN go get github.com/tools/godep
+RUN go get code.google.com/p/go.tools/cmd/cover
```
nathanleclaire: I really want to vendor these as well at some point, but first things first.

README.md
```diff
@@ -304,6 +304,29 @@ run:
 
 If you have any questions we're in #docker-machine on Freenode.
 
+## Code Coverage
```
bfirsh: Maybe we should have a separate `CONTRIBUTING.md` document, like Fig's.

bfirsh:
Cool!

README.md
nathanleclaire: +1

sthulb:
We should definitely start caring about testing.

Dockerfile
thaJeztah: If you want to, that may warrant a TODO or FIXME comment?

ehazlett: +1 for more coverage unit wise. I wouldn't say we don't care about testing.
It's not apparent but I have integration tests that run against the cli.
It's not perfect but at least 80% of the drivers are regularly tested and
I'm working on getting the remaining in (I need bare metal for those). I'm
wanting to get a way to expose those to the public but right now I can't do
to keys.

This PR is a great step and I think we need it.

Thanks

ehazlett:
As this stands this will cause every test to fail here on out.  While perhaps that's a good thing to show how scary, dangerous, lacking etc. these are, it will make it difficult to get anything in.

Perhaps we move that test to the integration ones (I'm also up for refactoring the whole lot in favor of a better test framework for those too).

nathanleclaire:
> As this stands this will cause every test to fail here on out

I'm not sure what you mean.  It shouldn't- the changes to Travis should only make it so that if a new directory gets added which doesn't have any `*_test.go` files, the build will error out.  This is the only restriction.

>  it will make it difficult to get anything in.

It shouldn't - in the case where we need to add a new directory as an "emergency" without any real tests, contributor should be able to simply put in empty test files as I have here.

> Perhaps we move that test to the integration ones (I'm also up for refactoring the whole lot in favor of a better test framework for those too).

Are you talking about the Google one?  I noticed that that folder has an integration test included, it should probably be moved to the actual integration tests since by convention its location implies it is a unit test.

Travis build is passing now.  Woop woop!  It turns out that when you run the unit tests multiple times in the same environment, they fail (since there are machines that get created which don't get cleaned up).  That definitely needs to get fixed.  I have the workaround of `rm -rf $HOME/.docker` here, which is fine for Travis, but definitely not OK for someone running them natively on their computer.

.travis.yml
```diff
@@ -2,7 +2,14 @@ language: go
...
+install:
+- export GOPATH=${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH
+- export PATH=${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH
+- go get -t -v ./...
+- go get code.google.com/p/go.tools/cmd/cover
 script:
 - script/validate-dco
 - script/validate-gofmt
 - go test -v -short ./...
+- rm -rf $HOME/.docker
```
nathanleclaire: This is the "run the tests multiple times in same environment" workaround I referenced.

Dockerfile
nathanleclaire: Good callout @thaJeztah.  I will make a note to fix and maybe follow up with a PR to do so.

ehazlett:
@nathanleclaire yes I was talking about the Google one.

If we could get people to run the tests in containers as originally intended it wouldn't be a problem.  I guess we will have to add cleanup for those that don't.

Thanks

nathanleclaire:
I put in https://github.com/docker/machine/pull/471 so people will know how to run unit tests in container.

ehazlett:
LGTM

Issue #432, opened 2015-01-29
Machine creates CA and certificate when just getting usage


bfirsh:
```
$ docker-machine                                                                                                                                                               
INFO[0000] Creating CA: /Users/ben/.docker/machines/ca.pem
INFO[0000] Creating client certificate: /Users/ben/.docker/machines/.client/cert.pem
NAME:
   docker-machine - Create and manage machines running Docker.

USAGE:
   docker-machine [global options] command [command options] [arguments...]

...
```

This seemed a bit odd. I was thinking "what's it doing? I didn't ask it to do anything!"

ehazlett:
Instead of tying this to a specific command, I set it as a "before" to the CLI app.  We could move it out but then would have to make sure to call it from each command in case the user specifies different keys.  Perhaps we just take off the info logging?

nathanleclaire:
@ehazlett We won't need the CA and certificate until a machine actually gets created, right?  So why not tie it to `Create`?

ehazlett:
Yeah we shouldn't.  I'll update.  Thanks.

ehazlett:
Closing - #460 merged

PR #460, opened 2015-02-02
move cert creation to create command

sthulb:

Refs: #432


commands.go
```diff
@@ -59,6 +59,69 @@ func (h hostListItemByName) Less(i, j int) bool {
 	return strings.ToLower(h[i].Name) < strings.ToLower(h[j].Name)
 }
 
+func beforeFunc(c *cli.Context) error {
```
ehazlett: Should we name this something like `createBefore` or something to show that it's tied to the `Create`?  `beforeFunc` seems a little generic.

ehazlett:
LGTM

Issue #452, opened 2015-01-31
docker-machine config chokes when run against non-running machine


nathanleclaire:
When the machine is running everything is cool:

```console
$ docker-machine config dev
--tls --tlscacert=/Users/nathanleclaire/.docker/machines/.client/ca.pem --tlscert=/Users/nathanleclaire/.docker/machines/.client/cert.pem --tlskey=/Users/nathanleclaire/.docker/machines/.client/key.pem -H tcp://192.168.99.132:2376%
```

But with non-running machines, everything goes haywire:

```console
$ docker-machine stop dev

$ docker-machine config dev
FATA[0000] Error getting machine url: host is not running
```

Yikes!  That shouldn't ever get thrown into a subshell.

sthulb:
What would we expect the behaviour to be? Nothing? Something to stderr?

ehazlett:
Hmm that's a good question.  I guess nothing?  Technically this behavior is the same for any command that throws an error when you run as a subshell.

Personally I think that's the right error.  You are asking for a config of something that is not running.

@nathanleclaire thoughts?

nathanleclaire:
> What would we expect the behaviour to be? Nothing? Something to stderr?

It should be the same output as when the machine is running.  Now, this might be a little tricky, since in our current implementation a machine's IP address disappears when a machine is stopped due to the way that the store save mechanics work.  I'd argue that this shouldn't be the case; instead each machine should get a statically assigned IP address at create time which isn't freed up until the machine is destroyed.  That would make stuff like this possible.

> Personally I think that's the right error. You are asking for a config of something that is not running.

It's the wrong error if we want users to know what's actually happening with the way the current workflow is set up.  Otherwise, `docker` just barfs because the user is passing invalid input to the CLI and leaves them with no idea what might have gone wrong (it will spit out help text and `Error: Command not found: time="2015-02-02T11:00:16-08:00"`).

A machine doesn't need to be running for clients to attempt to connect to it.  A hanging `docker version` is something which is universally accepted as an idiom for "the daemon is not responding".  That's what should happen in this case.  "Oh, my `docker version` is hanging, so the daemon must not be running".

Curious what @bfirsh thinks as well.

ehazlett:
If we want to leave the user without knowing there is an error and just rely on the Docker CLI to tell us, fine.  @bfirsh ?

sthulb:
[...]

PR #462, opened 2015-02-02
remove error message for config failures

sthulb:
My initial thoughts are we should do nothing, except return a non-zero exit code.

Refs: #452

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -214,6 +214,7 @@ func cmdCreate(c *cli.Context) {
 func cmdConfig(c *cli.Context) {
 	cfg, err := getMachineConfig(c)
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s", err)
 		os.Exit(1)
 	}
 
--- a/commands.go
+++ b/commands.go
@@ -214,8 +214,9 @@ func cmdCreate(c *cli.Context) {
 func cmdConfig(c *cli.Context) {
 	cfg, err := getMachineConfig(c)
 	if err != nil {
-		log.Fatal(err)
+		os.Exit(1)
 	}
+
 	fmt.Printf("--tls --tlscacert=%s --tlscert=%s --tlskey=%s -H %s",
 		cfg.caCertPath, cfg.clientCertPath, cfg.clientKeyPath, cfg.machineUrl)
 }


ehazlett:
So if we run this (no subshell) and there are errors, we expect the user to just know what they are?

sthulb:
Pretty much. We can't output anything to stdout (which is all of the logging).

ehazlett:
Why can't we output anything?  `log.Fatal` will output and return `1` for the exit code.

sthulb:
Because all of the log.* messages go to stdout.

ehazlett:
I agree it would be better to output to `stderr` but I think swallowing the errors entirely is a bad idea.  Ideally this would probably be fixed in logrus maybe.

sthulb:
We can ensure that logrus uses Stderr for logging, but then there's a bigger issue that some things that should go to Stdout are going to Stderr.

This is why I asked in IRC if there was a reason we use logging for everything :)

We're not quite swallowing the error, we would be returning a non-zero exit code to signal there's an error.

After 0.1.0 is released we could perhaps revise how we do logging.

ehazlett:
Asking why logging goes through `log` and disgarding altogether are two different things :)

I understand that a non-zero exit code signals an error.  From the user perspective, unless they are checking error codes they will not know there is an error.  Furthermore, even if they do check the code, they won't know what the error is.

I am all for a proposal to refactor logging.  I'm not in favor of hiding errors from the user.

sthulb:
If a user is executing it a subshell, they won't see the error anyway. At most, my PR should be spitting something out manually with fmt.Fprintf to stderr

sthulb:
Perhaps:
```go
  214 func cmdConfig(c *cli.Context) {
  215   cfg, err := getMachineConfig(c)
  216   if err != nil {
~ 217     fmt.Fprintf(os.Stderr, "%s", err)
  218     os.Exit(1)
  219   }
  220
  221   fmt.Printf("--tls --tlscacert=%s --tlscert=%s --tlskey=%s -H %s",
  222     cfg.caCertPath, cfg.clientCertPath, cfg.clientKeyPath, cfg.machineUrl)
  223 }
```

ehazlett:
I think this works.  Since `logrus` appears to only support sending the entire Out (instead of splitting) this seems like feasible solution to me.

@nathanleclaire thoughts since you raised the original issue?

nathanleclaire:
Well, `logrus` I think is really meant for long-running server-ey applications.  So it is a bit of an odd fit for our use case here, at least until we move to make machine a server with an API instead of a CLI app.

In general I'd favor moving away from `logrus` completely on the client side (replacing it with good old fashioned `fmt.Println` and the like for fine grained control of what goes where).  This will work as an _OK_ fix for the interim, but I'd expect to see more things like this crop up this down the line if the underlying issue doesn't get cleaned up (e.g. what happens when someone does `machine env` on a non-running machine, or a machine in an error state).

sthulb:
@nathanleclaire Agreed. We need to look at error reporting and logging for future releases.

PR #464, opened 2015-02-02
Swarm Provisioning

ehazlett:
This adds Swarm provisioning to Machine.  Usage is in readme.


sthulb:
How do we add exiting machines or remove machines from a cluster?

ehazlett:
https://github.com/docker/machine/pull/464/files#diff-04c6e90faac2675aa89e2176d2eec7d8R65

Removing is simply `rm` the machine.

cpuguy83:
```bash
➜  machine git:(e822716) ✗ ./machine create --help                                                                                                  ) ✗git:(e822716
Incorrect Usage.

FATA[0000] machine create: '' is not a machine create command. See 'machine create --help'.
```

ehazlett:
@cpuguy83 thx -- for some reason the `app.Name` is getting that from `os.Args[0]`.  Investigating.  It's actually doing this in master too.

cpuguy83:
Is there a machine command to get the propper config to talk to the swarm master instead of the docker daemon running on that machine?

I know I can connect to that machine on port 3376 and it works, but might be nice to be able to grab this config without manually doing stuff.

host.go
```diff
@@ -24,6 +26,11 @@ var (
 	ErrInvalidHostname   = errors.New("Invalid hostname specified")
 )
 
+const (
+	swarmDockerImage              = "ehazlett/swarm:latest"
```
cpuguy83: I believe there is an official swarm image now.

cpuguy83:
The swarm daemon doesn't seem to be doing client verification.

ehazlett:
@cpuguy83 can you elaborate?  Is this something in how it's configured?

rossbachp:
Please change at README.md

L79 +export DOCKER_HOST=tcp://192.168.99.100:3376
to
 +export DOCKER_HOST=tcp://192.168.99.100:2376

Nice PR!

commands.go
```diff
@@ -275,6 +314,17 @@ func cmdCreate(c *cli.Context) {
 	log.Infof("Configure docker client with: $(%s env %s)", c.App.Name, name)
```
sthulb: This is fixed now.

host.go
rossbachp: yes: dockerswarm/swarm:latest, but why it isn't a config option?

ehazlett: There isn't an official yet.  It's coming.  I could make it an option as well -- would be good for testing.

ehazlett:
@cpuguy83 yes if you do `docker-machine env --swarm swarm-master` it will output the swarm connection info instead of the daemon.

cpuguy83:
Oh nice, I missed that one.

cpuguy83:
@ehazlett I was looking through the code and it seems like it should be doing it, but it's definitely not. Maybe I'm looking in the wrong spot.

ehazlett:
@cpuguy83 you mean the client verification?

cpuguy83:
@ehazlett Yes, the client verification (GH needs threaded comments!)

Tebro:
I just tried this build, and I must say it is awesome, this is how easy it should be!

One thins I would like to see added though is something like this:
    docker-machine create -d someDriver --swarm --num-nodes 3 devSwarm

That command should generate a token, create a master named "devSwarm" and 3 nodes that could be named something like devSwarm-node-00, devSwarm-node-01 etc.

This would really speed up the creation of swarms with machine.

This could probably also be expanded with some kind of:
    docker-machine scale-swarm --num-nodes 1337 devSwarm

ehazlett:
Tebro: thanks!  there are plans for a scale function but will probably happen later after the swarm creation has been tested.

thanks for the feedback!

aanand:
Just finished trying this out. Fabulous. Two things:

1. I found myself needing to specify a different scheduling strategy when creating the master, so I implemented a `--swarm-strategy` flag for `create`. However, I can see these options multiplying and Machine having to constantly keep pace with Swarm. Would it make sense to be able to pass in arbitrary args with a `--swarm-args` option?

2. I found myself needing to specify a different Swarm image, as I was testing a PR for Swarm simultaneously. This is also something I can see other people needing to do, so perhaps there should be a `--swarm-image` option for `create` too.

ehazlett:
@aanand yeah i agree.  we also need things at the daemon level for things like labels.  I'm not sure yet on how we should tackle.  We were thinking cloudinit for the daemon level config but that would require quite a bit of refactoring.  We are looking to refactor the core (drivers, etc) -- this will be a topic.

I think your route (`--swarm-args`) is probably the best for swarm arguments and `--swarm-image` makes sense to me as well - I've wanted this too.

vieux:
this would be super useful to build our integration tests in swarm!

bfirsh:
I discussed this with @ehazlett yesterday. We thought it might be a good idea to sneak this in as "experimental and certain to change" in the beta release. It makes it infinitely easier to create a Swarm and doesn't impact any other parts of Machine. @sthulb @aanand – thoughts?

@ehazlett I think we could remove the `create-swarm-token` command though now that is just `docker run swarm create`. It'd mean we wouldn't have to vendor Swarm then, right?

bfirsh:
It does concern me a bit that this requires changes across all of the drivers though... I'm guessing that'll be cleaned up by refactoring.

We need to do some more design thought around what the Swarm + Machine integration looks like. I want to make sure we aren't being locked into any particular way of doing it with this method.

For example, is it safe to completely remove this feature, even though there are Swarm things in the config files, etc?

bfirsh:
Why does `env` have a `--swarm` option? Surely we could simplifying it by making that the default?

bfirsh:
Could the master also be added as a node?

sthulb:
img_0025

So I made some notes about the commands for swarm, I'd like to see them as subcommands.

ehazlett:
@bfirsh as far as the token, if we remove it (which is fine) it doesn't solve the race condition where we can't create a swarm because we don't have a token and we can't `docker run` to create because we dont' have an engine.  If the user uses a different discovery mechanism it's fine but otherwise how do they get a token to start?

@bfirsh yes the refactoring should remove this from the drivers and they should just do the provider logic.  there is a bit of state we need as far as which is the swarm master but i think we could solve that without being in the driver.  Yes, we could pull the feature easily -- the config files would simply have fields that would be ignored.

@bfirsh i chose to have the `--swarm` option because we don't know what the UX is yet.  For example, you may want to connect to the engine directly instead of to swarm (if swarm is acting up, etc).  I think this will be solved once we figure out the long term ux and functionality.

@bfirsh yes the master could be a node - i don't know why i didn't do that already ;) -- +1!

@sthulb +1 for the notes -- i didn't go too far because we don't know what the ux between swarm and machine are yet and i didn't want to introduce different toplevel commands for swarm specifically.  i think it will be along those lines.

bfirsh:
@ehazlett I think you could bootstrap it like:

```
$ docker-machine create -d virtualbox dev
$ $(docker-machine env)
$ docker run swarm create
$ docker-machine create -d digitalocean --swarm ...
```

bfirsh:
@sthulb Thanks for this! Like the direction here, Swarm should certainly be some kind of top-level object. Evan and I have been talking about similar things. It's not 100% clear whether it should be a part of Machine though – perhaps there is a Swarm client that integrates Machine, or some other higher-level tool.

ehazlett:
@bfirsh ok so for now I am removing the token generation and we will just have multiple steps.  I think that's a good in between.  it demonstrates we can create the swarm but don't have the dependency in machine for the token service.  I still don't like how you have to create a machine just to get a token but we can figure that out later.

ehazlett:
LGTM

aanand:
The only thing I have to add is :confetti_ball::confetti_ball::confetti_ball:

ehazlett:
thx :)

PR #467, opened 2015-02-02
New Driver: CenturyLink Cloud

dpetersen:
This adds the code and documentation for a CenturyLink Cloud driver. It's failing consistently on two integration tests, but it's a pair of tests that are also failing for me using the DigitalOcean driver (they're both around what the CLI displays once a server is stopped). Are there some known failures in master right now?

I have the latest version of `godep` installed, and doing a `godep save` to add my single new dependency is causing an enormous number of unrelated dependencies to be removed from the manifest (mostly docker libraries!). Is there a special syntax to use for adding dependencies? I manually tweaked the manifest in my vendoring commit.

I couldn't help but notice that none of the drivers are implementing anything in `Authorize`/`DeauthorizePort`. I know my API is capable of adding and removing ports from a public IP address, but I just did a dummy implementation of that since it doesn't appear to be used anywhere in the code.

Last but not least, while I was in there I noticed an old integration test script that doesn't seem to be referenced anymore. I tacked that onto this PR, but I can knock that commit off if I'm missing something.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -7,6 +7,10 @@
 			"Comment": "weekly-56",
 			"Rev": "afe77d958c701557ec5dc56f6936fcc194d15520"
 		},
+                {
+                        "ImportPath": "github.com/CenturyLinkLabs/clcgo",
+                        "Rev": "8b702d20bfe3710be07422907796e97bd1a6acb9"
+                },
 		{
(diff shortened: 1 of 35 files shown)


dpetersen:
Hey @ehazlett, did you see the old email chain that I resurrected on 7/2 around this driver PR?

ehazlett:

@ehazlett & @nathanleclaire

PR #474, opened 2015-02-03
fix docker deps to standard rev

### LINKED ISSUE #472: machine vendors a revision of docker which doesn't exist
cc @bfirsh @ehazlett @sthulb 

Not sure how this happened, but see for yourself:

```console
$ cd /tmp && git clone https://github.com/docker/docker && cd docker && git checkout 831d09f8b535a1e2939bdacb02032d238e8dd249
Cloning into 'docker'...
remote: Counting objects: 70905, done.
remote: Compressing objects: 100% (41/41), done.
remote: Total 70905 (delta 11), reused 44 (delta 9)
Receiving objects: 100% (70905/70905), 39.38 MiB | 4.72 MiB/s, done.
Resolving deltas: 100% (46278/46278), done.
Checking connectivity... done.
fatal: reference is not a tree: 831d09f8b535a1e2939bdacb02032d238e8dd249
```

After wrangling with Godeps a bit this afternoon I'm starting to think we should rip it out and replace it with something else (probably custom - we could either steal `vendor.sh` as a base or make our own thing).

It's not very actively maintained (last commit is from two weeks ago), [seems to have issues supporting Go 1.4](https://github.com/tools/godep/issues/168) (which we should move to anyway ASAP), and so on.

At any rate I will look into vendoring `docker/docker` properly.

ehazlett:
This properly vendors the Docker dependency from a custom rev to a SHA from the standard repo.

Fixes #472


ehazlett:
Tests pass.  Build succeeds.

nathanleclaire:
Will you please write some docs on godep workflow (or a script to automate), because I spent entirely too much time wrangling with this sort of thing to no avail.  For instance, don't even know where to begin with adding `github.com/mitchellh/gox` to `Godeps`, since I think godep expects you to actually import the dependency.

Godeps/Godeps.json
```diff
@@ -147,6 +147,10 @@
 			"Rev": "1fbbd62cfec66bd39d91e97749579579d4d3037e"
 		},
 		{
+			"ImportPath": "golang.org/x/net/context",
```
nathanleclaire: What's up with this?

nathanleclaire:
Looks mostly pretty good but why the addition of the golang.org/x/net/context dependency?

ehazlett:
It was probably added in Docker.

nathanleclaire:
I can't find it reference anywhere in this PR or in the Docker source code at the specified revision using `grep`.  Correct me if I'm wrong.  I'm sure it's not _harmful_ per se, but why add it if it is not needed?  It will make the repo fatter and will make other people scared to take it out later on ("What if we break something?").

ehazlett:
This was a dependency added from my Swarm branch.  Swarm relies on it indirectly through etcd.  Good catch.  I've updated Godeps from master using the Docker v1.4.1 branch.

ehazlett:
Actually, I'm not sure how that was added.  If it was from my Swarm branch (which I'm almost positive it wasn't because I `git reset`) there would have been a lot more.  Not sure.

sthulb:
Great. This was on my list of things.

ehazlett:
LGTM

nathanleclaire:
:clap: :clap: :clap:

PR #475, opened 2015-02-03
docs for working with godep

ehazlett:
Docs on working with Godep.  Refs #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,3 +345,38 @@ You can set the path to the machine binary under test using the `MACHINE_BINARY`
 environment variable.
 
 To run, use the helper script `./script/run-integration-tests`.
+
+## Working with Godep
+This is basically copied from the [Godep Docs](https://github.com/tools/godep)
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Heh.  I meant unique to our case, as I have read the godep docs quite thoroughly (not much sense to reproduce them here).  For instance, do you run the `godep save` inside of a container or on your host?  How did you update Docker to a specific revision in #474 ?  In my experience `godep` has not played very well with containers due to it making assumptions about what or where `GOPATH` is.

ehazlett:
Ugh ok well I guess I will rip them out then.

I literally follow that for 99% of development.  Unless you dev in
containers you don't need to do anything besides that.

I can add those troubleshooting / fix steps for what I did there but I'm
afraid most people will mess things up when manually editing Godeps.json.
I would rather help someone on IRC instead of encouraging them to hack with
Godeps manually.

PR #481, opened 2015-02-03
fixes cli arg bugs

ehazlett:
This fixes cli arg bugs where the `os.Args` were being munged from the `before` func.

Refs #479 
Refs #464


sthulb:
LGTM. We should avoid `Before` like the plague until https://github.com/codegangsta/cli/blob/master/command.go#L126 is fixed.

ehazlett:
LGTM

PR #490, opened 2015-02-05
docs for running on windows

ehazlett:
Some tips for running machine on Windows.


ehazlett:
@jeffmendoza does this make sense?  is there a better way?

jeffmendoza:
Looks good, might also need openssl: http://slproweb.com/products/Win32OpenSSL.html

PR #495, opened 2015-02-06
only show fusion on darwin

ehazlett:
This will only show the VMware Fusion driver on `darwin` platforms.  Also added a comment to the `hyper-v` driver as well.


ehazlett:
LGTM

PR #506, opened 2015-02-08
use custom vmware b2d iso for fusion and vsphere

ehazlett:
This enables external disks and the ability to stop/restart the vm using VMware Fusion and vSphere.

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -31,6 +31,7 @@ const (
 	B2D_USER        = "docker"
 	B2D_PASS        = "tcuser"
 	dockerConfigDir = "/var/lib/boot2docker"
+	isoFilename     = "boot2docker-vmw.iso"
 )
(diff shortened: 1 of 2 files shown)


ehazlett:
@frapposelli this works however we do not have shared folders for fusion.  Is this something misconfigured or just not in the driver?

frapposelli:
AFAIK the shared folder driver is not there. Will discuss this matter this
week as I believe it should be there.

md5:
@frapposelli so it sounds like that ISO just has the changes discussed in https://github.com/boot2docker/boot2docker/pull/638, right?

frapposelli:
Has two patches applied:

boot2docker/boot2docker#638
boot2docker/boot2docker#659

ehazlett:
@frapposelli is there anyway we can get shared folders with this?

ehazlett:
Merging to get the correct ISO so we can at least restart.  Will approach the shared folders in another issue.

PR #535, opened 2015-02-12
Adds multiple cpu support for vmware fusion

jmmills:
This adds --vmwarefusion-cpu-count support.


ehazlett:
Thanks!  We might want to default to the host number of CPUs like the virtualbox driver does.

sthulb:
I thought the VMware driver did this.

ehazlett:
@sthulb By default only 1 vcpu is allocated to the VM.

jmmills:
It appears I fudged the help string for the optarg, what's what's the patch policy for this? Just a second pull request with an update to the string?

ehazlett:
Nah just use this and rebase.  I still think we should match the vbox driver and detect the host number of cpus instead of specifying as these will only apply for local hypervisors.

jmmills:
Okay, rebased the commit in my fork to fix the help line.
I'm not sure what the implications of detecting host cpu count would be specifically for vmwarefusion, it's local only - but I'm not an expert enough in vmware to know if it would crush the host machine.

ehazlett:
The converse is that it would always perform a fraction of what your machine is.  I think it depends on your usage.  I simply wanted to be consistent with the local drivers.

PR #546, opened 2015-02-16
WIP: OS abstraction

sthulb:
- Creates interface for an OS runtime
 - Creates a way of registering a runtime


drivers/amazonec2/amz/ec2.go
```diff
@@ -160,6 +160,9 @@ func (e *EC2) awsApiCall(v url.Values) (*http.Response, error) {
 		fmt.Printf("client encountered error while doing the request: %s", err.Error())
 		return resp, fmt.Errorf("client encountered error while doing the request: %s", err)
 	}
+
+	fmt.Print(resp)
```
nathanleclaire: I think this might be debugging code?

nathanleclaire:
I like the general idea and want to move everything away from being so Debian specific eventuallly - I think it will require some design and review though, so maybe you could put together a proposal and open an issue so we can discuss the architecture a bit?

drivers/amazonec2/amz/ec2.go
sthulb: It definitely is – it is WIP after all :)

sthulb:
@nathanleclaire Definitely.

machine/os/ubuntu.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func init() {
+	RegisterRuntime("ubuntu", &RegisteredRuntime{
+		Detect: UbuntuDetection,
+	})
+}
+
+func UbuntuDetection(d drivers.Driver) (*Runtime, error) {
+	out, err := sshCommand(d, "lsb_release -i | cut -f2 -d:")
```
nathanleclaire: I want to check with @tianon @SvenDowideit what the most robust approach would be here, since I'm sure they'll have valuable input.

machine/os/ubuntu.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"strings"
+
+	"github.com/docker/machine/drivers"
+)
+
+func init() {
+	RegisterRuntime("ubuntu", &RegisteredRuntime{
+		Detect: UbuntuDetection,
```
nathanleclaire: As noted in IRC, I'd prefer the actual function that does the detection (`CompatibleWithHost(h)`?) to be part of a generic "Provisioner" interface.

This should totally accept a factory function like the host drivers do, though (`New: NewUbuntuProvisioner` style).

machine/os/detect.go
```diff
@@ -0,0 +1,5 @@
+package os
```
nathanleclaire: As mentioned in IRC: Let's rename the package to `provision` since `os` is a common Go stdlib package.

machine/os/ubuntu.go
tianon: Relying on `lsb_release` to exist as a command is going to cause trouble down the road.  The current standard is `/etc/os-release`:
```console
$ for image in ubuntu:{precise,trusty,utopic,vivid} debian:{wheezy,jessie}; do (set -x && docker run --rm $image cat /etc/os-release); done
+ docker run --rm ubuntu:precise cat /etc/os-release
NAME="Ubuntu"
VERSION="12.04.5 LTS, Precise Pangolin"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu precise (12.04.5 LTS)"
VERSION_ID="12.04"
+ docker run --rm ubuntu:trusty cat /etc/os-release
NAME="Ubuntu"
VERSION="14.04.2 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.2 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
+ docker run --rm ubuntu:utopic cat /etc/os-release
NAME="Ubuntu"
VERSION="14.10 (Utopic Unicorn)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.10"
VERSION_ID="14.10"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
+ docker run --rm ubuntu:vivid cat /etc/os-release
NAME="Ubuntu"
VERSION="15.04 (Vivid Vervet)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu Vivid Vervet (development branch)"
VERSION_ID="15.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
+ docker run --rm debian:wheezy cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 7 (wheezy)"
NAME="Debian GNU/Linux"
VERSION_ID="7"
VERSION="7 (wheezy)"
ID=debian
ANSI_COLOR="1;31"
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support/"
BUG_REPORT_URL="http://bugs.debian.org/"
+ docker run --rm debian:jessie cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support/"
BUG_REPORT_URL="https://bugs.debian.org/"
```

nathanleclaire: Oh cool.  @sthulb it looks like we want to match on `ID` field in `/etc/os-release`.  I tested for the file in Debian, boot2docker, and Fedora and it was present in all of them.  (I tested the Debian and Fedora ones in containers)

nathanleclaire: @sthulb  The success or failure of running commands like `'cat /etc/os-release | egrep ^ID=boot2docker'` is probably what we're after, then.  I have to admit it's pretty tempting to just run `cat /etc/os-release | egrep ^ID` once and then do a pattern match on the output once instead of running a command over SSH for each possible driver though :)

tianon: Why not just cat the whole file and parse it in Go so you can use other fields later if you need to?

nathanleclaire:
Ooh, I'm really liking the direction this is headed in @sthulb.  Much appreciated.

When you get a chance, could you elaborate a little on how the `Service` and `Package` methods will be used?  I think I understand, just want to make sure.

host/provision/boot2docker.go
```diff
@@ -0,0 +1,39 @@
+package provision
+
+import (
+	"github.com/docker/machine/drivers"
+)
+
+func init() {
+	RegisterProvisioner("boot2docker", &ProvisionerDetection{
+		New: func(driver drivers.Driver) Provisioner {
```
nathanleclaire: Could you please pull this closure out into a named function?

host.go
```diff
@@ -265,6 +267,11 @@ func (h *Host) Create(name string) error {
 		return err
 	}
 
+	_, err = mos.DetectRuntime(h.Driver)
```
nathanleclaire: I think this should come after `SaveConfig` (actually after waiting for SSH to be up, but that code isn't merged in yet) as it is technically the first step in attempting provisioning.

nathanleclaire:
I like @tianon idea to parse the whole `/etc/os-release` file.  How does this sound?

1. Make a `OsRelease` struct that is embedded in the `Provisioner` structs which we demarshal the information from that file into.  For instance, `PRETTY_NAME` would become a `PrettyName string` field.
2. Then, the compatibility methods would be just to check that the field we expect to line up (`Id`) is correct.  We get nice separation of concerns in the drivers instead of one master function, and only have to make one SSH call.

sthulb:
@nathanleclaire I'm reading that file now, but I'm not sure we need to store it just yet...

host/provision/boot2docker.go
sthulb: not a problem :)

sthulb:
@nathanleclaire @ehazlett We should talk about how we're handling package paths, I introduced hosts/provider in this PR, but it's perhaps not a good fit with: #767

ehazlett:
I like this direction too :+1:

I also like @nathanleclaire comment about parsing `/etc/os-release`.

ehazlett:
What can I do to help this?

nathanleclaire:
It looks like this is getting pretty close to a merge-able state and I think that mostly what remains to be done is to:

- List the package dependencies
- Replace the existing provisioning in `host.go` with this method
- Talk about how cloudinit fits in I suppose?
- Play around with it and see how it works

I'll take a close look at this today

sthulb:
Closed as this became #827

PR #559, opened 2015-02-19
#445-Fix Patch.

LaynePeng:
Fix the bug of creating host using OpenStack driver. For the detail: script error during DOCKER_OPTS injection #445

Signed-off-by: LaynePeng <appamail@hotmail.com>


ehazlett:
LGTM

PR #560, opened 2015-02-19
support for fish shell in env command

databus23:
This is basically ported from boot2docker-cli

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -398,8 +398,14 @@ func cmdEnv(c *cli.Context) {
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


ehazlett:
LGTM

ehazlett:
@databus23 i'm having trouble getting this to work in general (i.e. `(docker-machine env dev)` is giving me an `fish: Illegal command name`.  Tried with fish v2.1.2.  any ideas?

databus23:
How are you sourcing the output of `docker-machine env dev`?

I do `docker-machine env dev | source`.

databus23:
Ah I overlooked the `()` in your post.

That does not work, command substitution in fish works different then in bash. You have to do ` ... | source` to set the environment in fish.

ehazlett:
Whew thanks!  I'll update the docs.

Issue #555, opened 2015-02-18
move ssh operations to crypto/ssh


ehazlett:
Remove the shell out to `ssh` and use `crypto/ssh` instead.

This should do the following:
- refactor the key generation from the `ssh` command into `crypto/ssh`
- refactor the functionality from GetSSHCommand from `ssh` into `crypto/ssh`

andrewwatson:
+1

bfirsh:
Work in progress here: https://github.com/docker/machine/pull/730

ahmetalpbalkan:
+1 this would improve windows command line prompt (cmd.exe) & PowerShell experience a great deal as Windows by default doesn't have ssh.exe in PATH.

ahmetalpbalkan:
A big +1 for this! This would improve windows command line prompt (cmd.exe) & PowerShell experience a great deal for users as Windows by default doesn't have ssh.exe in PATH.

hairyhenderson:
I think this was resolved by #909 - right? /cc @ehazlett

ehazlett:
@hairyhenderson yes should be.  @sthulb #909 replaced all of the ssh shell-out correct?

sthulb:
Ja!

ehazlett:
sweet.  closing.  thx!

PR #561, opened 2015-02-19
WIP: No more SSHell out :P

sthulb:
Refs: #555

### DIFF
--- a/docs/DRIVER_SPEC.md
+++ b/docs/DRIVER_SPEC.md
@@ -1,0 +1,85 @@
+> DRAFT
+
+# Machine Driver Specification v1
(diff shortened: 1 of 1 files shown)


ehazlett:
Thanks!  I wasn't clear in the issue -- I think we should remove the shell out completely (including the SSH commands that are ran remotely).  I've updated the issue.  Also, it looks like the tests aren't happy either :)

sthulb:
@ehazlett I plan to remove all shelling out for SSH commands in this PR. This just seemed like a good start.

ehazlett:
@sthulb ah cool.  thx!

PR #567, opened 2015-02-20
Add machine scp command

nathanleclaire:
See the docs for usage.

It's not perfect but it's definitely a start.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -193,6 +193,11 @@ var Commands = []cli.Command{
 		Action: cmdEnv,
 	},
 	{
+		Name:   "scp",
+		Usage:  "Move files from one machine to another.",
+		Action: cmdScp,
+	},
+	{
 		Name:   "ssh",
 		Usage:  "Log into or run a command on a machine with SSH",
 		Action: cmdSsh,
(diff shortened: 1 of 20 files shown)


nathanleclaire:
cc @ehazlett @bfirsh @sthulb

sthulb:
A good start. I think the scp code should be in the SSH package though.

ehazlett:
I think this is a good start.  I want to be careful about letting Machine do "too much".  That being said, I can see this being useful.  

+1 @sthulb - I'm not a fan of shelling out everywhere.  This is what I used in my poc "sync" app - http://godoc.org/github.com/pkg/sftp.  Worked pretty well -- not sure if it fits here or not.

nathanleclaire:
I don't think it will work with the current organization to put the scp logic inside of `ssh`, even though I'd like to.  A separate `scp` folder might work for now.

I initially put the scp logic in the `ssh` package but I immediately ran into issues with circular imports because I want to be able to use the driver logic to generate the scp commands (need to call out to get IP address, key path location, etc.) and the main package where store, etc. are kept that I need in order to do that imports the drivers, which in turn import `ssh`.  So, to import that package from `ssh` will not work.

I agree very much that the scope of what Machine can and/or should do should be limited (especially for now, until we get a very distinct picture of where it is best at delivering value), but I think having this as an alias for `scp` and `rsync` is reasonable (mostly because otherwise the user would have to cobble it together on their own with the "plumbing").  

For instance, deploying a static site would become as simple as mounting it in an nginx container and running `docker-machine scp --delta ./public blogprod:/path/to/blog`.  Super fast and convenient.

ehazlett:
@nathanleclaire well said.  i think some form of copy will be awesome.  i would rather it be as portable as possible but besides technical implementation details i agree with the concept :)  I think this will work nicely with the refactor (no more circular imports, store is a package, etc)

hairyhenderson:
@nathanleclaire - I presume this is obsoleted by #1140? :)

ehazlett:
@hairyhenderson thanks.  closing as #1140 is merged.

PR #574, opened 2015-02-21
Add docker-machine env unset

nathanleclaire:
User was confused about how to unset env variables previously set with `$(docker-machine env blah)`, so I figured there might as well be an automated way to unset them as well.

Also added docs for `env` since there were none.

Will this need to be something different for fish @ehazlett ?

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -232,6 +232,10 @@ var Commands = []cli.Command{
 				Name:  "swarm",
 				Usage: "Display the Swarm config instead of the Docker daemon",
 			},
+			cli.BoolFlag{
+				Name:  "unset, u",
+				Usage: "Unset variables instead of setting them",
+			},
 		},
 	},
 	{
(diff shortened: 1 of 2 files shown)


ehazlett:
Cool.  Yes you will need to use `set -e` I believe.  We can probably do something similar to https://github.com/docker/machine/blob/master/commands.go#L509 for fish.

nathanleclaire:
Made a `fish` version and rebased.

FYI checking `SHELL` is not a very versatile solution as, for instance, if I start a `fish` instance from my usual `zsh` instance, `$SHELL` is still `/bin/zsh`.

ehazlett:
i was simply recommending as the b2d way did

nathanleclaire:
Yep, unfortunately it's a non-trivial problem to solve- I was noting the fact so that we're not confused down the line if weird behavior shows up with users.

ehazlett:
LGTM

PR #588, opened 2015-02-23
Added a remote-build script that streams tgz'ed binaries to stdout

saturnism:
Signed-off-by: Ray Tsang <saturnism@gmail.com>

### DIFF
--- a/script/remote-build
+++ b/script/remote-build
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# A build script that allows building from a remote docker server that can't mount local volumes.
+# $ script/remote-build | tar xvf -
+set -e
+
+if [ -z "$1" ]; then
+    OS_PLATFORM_ARG=(-os="darwin linux windows")
+else
+    OS_PLATFORM_ARG=($1)
+fi
+
+if [ -z "$2" ]; then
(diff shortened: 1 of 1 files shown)


ehazlett:
I like this idea!  Do you think we could merge this into the existing build script as an option or something?

saturnism:
ehazlett - sure! I can take a look at that.

Before I make modifications though, here is the thought on what it may look like... Any preferences?

Regular build:
```
    $ script/build 
```

Remote build:
```
    $ script/build -r
```
Or, generically, output binaries in tgz format?
```
    $ script/build -o
    $ script/build -
````

sthulb:
Without looking over the changes, what's the behaviour? Output all binaries in one tgz or each in tgz format?

saturnism:
it will output all the binaries in one tgz

saturnism:
quick ping

nathanleclaire:
Definitely want to support this - we are in a bit of a shuffle to get some stuff in right now, but soon I want to integrate this style of build into a `make remote` option or otherwise.

ehazlett:
Thanks for the follow up. I think a -r option would work for now. Thanks!

nathanleclaire:
We've added this in a slightly different way now, thanks!

PR #590, opened 2015-02-23
do not show debug for docker install by default

### LINKED ISSUE #587: Digital Ocean driver prints out a bunch of debugging output
```
$ docker-machine create -d digitalocean --digitalocean-access-token=$DIGITALOCEAN_ACCESS_TOKEN --swarm --swarm-master --swarm-discovery=token://... swarm-master
INFO[0000] Creating Digital Ocean droplet...
INFO[0004] Waiting for SSH...
INFO[0073] Configuring Machine...
+ [ https://get.docker.com/ = https://get.docker.com/ ]
+ sh -c apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
gpg: requesting key A88D21E9 from hkp server keyserver.ubuntu.com
gpg: key A88D21E9: public key "Docker Release Tool (releasedocker) <docker@dotcloud.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1  (RSA: 1)
+ sh -c echo deb https://get.docker.com/ubuntu docker main > /etc/apt/sources.list.d/docker.list
+ sh -c sleep 3; apt-get update; apt-get install -y -q lxc-docker
+ sh -c docker version
INFO[0135] Configuring Swarm...
INFO[0142] "swarm-master" has been created and is now the active machine.
INFO[0142] To point your Docker client at it, run this in your shell: $(docker-machine env swarm-master)
```

ehazlett:
This hides the debug output from the Docker install for drivers.  It also refactors the install process out of the driver.

fixes #587

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -5,7 +5,6 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path"
-	"strconv"
 	"strings"
 	"time"
 
@@ -22,36 +21,35 @@ const (
 )
 
 type Driver struct {
-	AuthUrl             string
-	Username            string
-	Password            string
-	TenantName          string
-	TenantId            string
-	Region              string
-	EndpointType        string
-	MachineName         string
-	MachineId           string
-	FlavorName          string
-	FlavorId            string
-	ImageName           string
-	ImageId             string
-	KeyPairName         string
-	NetworkName         string
-	NetworkId           string
-	SecurityGroups      []string
-	FloatingIpPool      string
-	FloatingIpPoolId    string
-	SSHUser             string
-	SSHPort             int
-	Ip                  string
-	EnableDockerInstall bool
-	CaCertPath          string
-	PrivateKeyPath      string
-	storePath           string
-	SwarmMaster         bool
-	SwarmHost           string
-	SwarmDiscovery      string
-	client              Client
+	AuthUrl          string
+	Username         string
+	Password         string
+	TenantName       string
+	TenantId         string
+	Region           string
+	EndpointType     string
+	MachineName      string
+	MachineId        string
+	FlavorName       string
+	FlavorId         string
+	ImageName        string
+	ImageId          string
+	KeyPairName      string
+	NetworkName      string
+	NetworkId        string
+	SecurityGroups   []string
+	FloatingIpPool   string
+	FloatingIpPoolId string
+	SSHUser          string
+	SSHPort          int
+	Ip               string
+	CaCertPath       string
+	PrivateKeyPath   string
+	storePath        string
+	SwarmMaster      bool
+	SwarmHost        string
+	SwarmDiscovery   string
+	client           Client
 }
 
 type CreateFlags struct {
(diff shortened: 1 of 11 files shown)


bfirsh:
Nice.

sthulb:
Nice. There's probably a lot more that could be removed from the driver into this.

ehazlett:
thx -- i agree -- for now what do you think of just starting with this to get the debug hidden?  thoughts?

host.go
```diff
@@ -360,10 +361,17 @@ func (h *Host) Create(name string) error {
 		return err
 	}
 
+	// create the instance
 	if err := h.Driver.Create(); err != nil {
 		return err
 	}
 
+	// install docker
+	if err := h.Provision(); err != nil {
```
ggiamarchi: It would be great to have a flag to disable this step. On the OpenStack driver we had a flag `--openstack-docker-install` useful when we use an image with docker already installed and well configured.

`--no-provision` maybe ?

drivers/softlayer/driver.go
```diff
@@ -142,12 +140,6 @@ func GetCreateFlags() []cli.Flag {
 			Usage:  "OS image for machine",
 			Value:  "UBUNTU_LATEST",
 		},
-		cli.StringFlag{
-			EnvVar: "SOFTLAYER_INSTALL_SCRIPT",
-			Name:   "softlayer-install-script",
-			Usage:  "Install script to call after the machine is initialized (should install Docker)",
-			Value:  DockerInstallUrl,
-		},
```
ggiamarchi: The README should be updated to remove this option.

drivers/openstack/openstack.go
```diff
@@ -175,13 +173,6 @@ func GetCreateFlags() []cli.Flag {
 			Usage: "OpenStack SSH port",
 			Value: 22,
 		},
-		// Using a StringFlag rather than a BoolFlag because
-		// the BoolFlag default value is always false
-		cli.StringFlag{
-			Name:  "openstack-docker-install",
```
ggiamarchi: The README should be updated to remove this option.

ehazlett: Thanks.  I agree we need a way to disable.  Perhaps a global flag now that
it's centralized?  Sorry for updating the openstack one -- I wanted to at
least be consistent across the providers as far as what was provisioned.

sthulb:
The host name stuff could be here. Variances for b2d I guess though. :(. 

For the no provision flag, perhaps we could look for docker on the path, if found skip.

ehazlett:
it does some of that now (checks for /usr/bin/docker which most of the
installers put it in that place).

drivers/openstack/openstack.go
ggiamarchi: Yes, a global flag would be great. I originally did it in the OpenStack driver because it was not possible to make it global but with your refactoring now it is :+1:

host.go
nathanleclaire: I am +1 on this - originally the EC2 driver had something similar but I think it was removed.

I'd be a fan of checking to see if Docker's already installed as well (maybe we could provide the drivers with a function that checks for the presence of the Docker socket).

ehazlett: The provision step has this somewhat (it checks for `/usr/bin/docker`).  The reason these things were removed (there was nothing in ec2 fwiw) was to try and "standardize" on the ux across drivers.  Once we get things refactored (store, host, etc) this will be a lot easier to implement.

nathanleclaire: OK, great, I wasn't aware that had been implemented.

ehazlett:
@bfirsh so is this good to merge?

nathanleclaire:
This LGTM

PR #591, opened 2015-02-23
do not error with no active host for ls

### LINKED ISSUE #582: Error displayed when no active machine is set
```
$ docker-machine ls                                                                                                                                                           
ERRO[0000] There's a problem finding the active host
NAME   ACTIVE   DRIVER       STATE     URL                         SWARM
dev             virtualbox   Running   tcp://192.168.99.100:2376
```

This error should be displayed when you are trying to do something against an active machine, but not when listing hosts. It isn't an error.

ehazlett:
This will not report an error when there is no active host for the `ls` command.

fixes #582

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -395,21 +395,12 @@ func cmdLs(c *cli.Context) {
 
 	for _, host := range hostList {
 		if !quiet {
-			tmpHost, err := store.GetActive()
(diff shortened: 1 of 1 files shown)


sthulb:
LGTM

PR #599, opened 2015-02-24
add swarm label of provider by default

ehazlett:
This will add a default label of `provider=<provider>` when using Swarm.

This enables things like `docker run -e constraint:provider==digitalocean ...`

### DIFF
--- a/host.go
+++ b/host.go
@@ -327,10 +327,12 @@ func (h *Host) generateDockerConfig(dockerPort int, caCertPath string, serverKey
 
 	swarmLabels = append(swarmLabels, fmt.Sprintf("--label=provider=%s", h.Driver.DriverName()))
 
-	defaultDaemonOpts := fmt.Sprintf(`--tlsverify \
---tlscacert=%s \
---tlskey=%s \
---tlscert=%s %s`, caCertPath, serverKeyPath, serverCertPath, strings.Join(swarmLabels, " "))
+	defaultDaemonOpts := fmt.Sprintf(`--tlsverify --tlscacert=%s --tlskey=%s --tlscert=%s %s`,
+		caCertPath,
+		serverKeyPath,
+		serverCertPath,
+		strings.Join(swarmLabels, " "),
+	)
 
 	switch d.DriverName() {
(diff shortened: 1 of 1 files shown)


host.go
```diff
@@ -322,16 +322,19 @@ func (h *Host) generateDockerConfig(dockerPort int, caCertPath string, serverKey
...
-	// TODO @ehazlett: template?
+	if h.SwarmHost != "" {
+		swarmLabels = append(swarmLabels, fmt.Sprintf("--label=provider=%s", h.Driver.DriverName()))
+	}
+
 	defaultDaemonOpts := fmt.Sprintf(`--tlsverify \
 --tlscacert=%s \
 --tlskey=%s \
---tlscert=%s`, caCertPath, serverKeyPath, serverCertPath)
+--tlscert=%s %s`, caCertPath, serverKeyPath, serverCertPath, strings.Join(swarmLabels, " "))
```
sthulb: Not sure of the formatting for this line, I'd perhaps like to see:
```go

	defaultDaemonOpts := fmt.Sprintf(
		"--tlsverify --tlscacert=%s --tlskey=%s --tlscert=%s %s`, 
		caCertPath,
		serverKeyPath,
		serverCertPath, 
		strings.Join(swarmLabels, " ")
	)
```

host.go
```diff
@@ -322,16 +322,19 @@ func (h *Host) generateDockerConfig(dockerPort int, caCertPath string, serverKey
 		daemonOpts    string
 		daemonOptsCfg string
 		daemonCfg     string
+		swarmLabels   = []string{}
 	)
 
-	// TODO @ehazlett: template?
+	if h.SwarmHost != "" {
```
vieux: Why only for swarmhost ? I would do it on every host, so if later it become a swarm host, no problem.

ehazlett: @vieux yes sir -- that makes more sense :)  thx!

host.go
ehazlett: sure -- the reason i went multi-line is i was anticipating more engine options being added.  if we go with cloudinit it won't be there anyway.  i can update it - it should evaluate the same.

ehazlett:
@sthulb updated ptal

vieux:
LGTM

PR #601, opened 2015-02-24
Logging updates rackspace

ehazlett:
This moves the details (machineid, keypair name, etc) to debug messages to match the other drivers.

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -384,14 +384,15 @@ func (d *Driver) Stop() error {
 }
 
 func (d *Driver) Remove() error {
-	log.WithField("MachineId", d.MachineId).Info("Deleting OpenStack instance...")
+	log.WithField("MachineId", d.MachineId).Debug("deleting instance...")
+	log.Info("Deleting OpenStack instance...")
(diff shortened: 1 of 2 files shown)


drivers/openstack/openstack.go
```diff
@@ -384,14 +384,15 @@ func (d *Driver) Stop() error {
 }
 
 func (d *Driver) Remove() error {
-	log.WithField("MachineId", d.MachineId).Info("Deleting OpenStack instance...")
+	log.WithField("MachineId", d.MachineId).Debug("deleting instance...")
+	log.Info("Deleting OpenStack instance...")
```
ggiamarchi: Here you have both `Info` and `Debug` logs. It is what you want ?

ehazlett: I think so.  This way you can see the Machine ID if you enable debug.  What do you think?  Mainly I just wanted to get the same feel across drivers for the `Info`.

ggiamarchi: I agree, just wanted to be sure that had been done intentionally.

ggiamarchi:
LGTM

ehazlett:
@ggiamarchi thx!

Issue #539, opened 2015-02-14
How I can create a virtualbox machine offline?


rossbachp:
I tested with absolute file and file:/// scheme, but with no success.

Many thanks
Peter

```
$ docker-machine create -d virtualbox tomcat-2 --virtualbox-boot2docker-url ///Users/xxx/.docker/machines/tomcat/boot2docker.iso 
ERRO[0000] Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: stat /etc/resolv.conf: no such file or directory 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine                       
$ docker-machine rm tomcat-2
INFO[0000] machine does not exist, assuming it has been removed already 
$ docker-machine create -d virtualbox tomcat-2 --virtualbox-boot2docker-url /Users/xxx/.docker/machines/tomcat/boot2docker.iso 
ERRO[0000] Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: stat /etc/resolv.conf: no such file or directory 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine                       
$ docker-machine rm tomcat-2
INFO[0000] machine does not exist, assuming it has been removed already 
$ docker-machine create -d virtualbox tomcat-2 --virtualbox-boot2docker-url file:///Users/xxx/.docker/machines/tomcat/boot2docker.iso 
ERRO[0000] Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: stat /etc/resolv.conf: no such file or directory 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine                       
```

ehazlett:
We should skip the release check if a file is specified.  Thanks for reporting.

rossbachp:
Great:) Many thanks!

arun-gupta:
This is very relevant in a hackathon setup.

hairyhenderson:
Is this still an issue in HEAD? the code in virtualbox.go seems to imply that the release check only occurs if no b2d ISO URL was specified.

hairyhenderson:
Oh! This might just be a flag issue. I seem to recall running into this a while back.

Try putting a `=` between the flag and the argument:

```
[...]

Issue #537, opened 2015-02-13
support github enterprise urls for b2d


ehazlett:
We should support Github enterprise URLs for b2d.  See https://github.com/boot2docker/boot2docker-cli/pull/339

sthulb:
We should start putting these in a 0.2 milestone.

ehazlett:
Once we have it scoped we will.

razic:
@ehazlett does this have to do with the `--url` provisioner? if so, can you explain what the `--url` provisioner is?

ehazlett:
@razic the `--url` simply adds an existing host to the machine store.  however, i don't particular like it because it doesn't do anything with creds, certs, etc.  I would rather use the Generic Driver (#406) which imports the ssh key and could setup the auth.

databus23:
@ehazlett  @sthulb I started working on this and then discovered #603. Whats the status on this? Is this still being worked on. Is it worth looking into this or should I jus be patient. :)

sthulb:
@darabus23 go ahead. I closed my PR

nathanleclaire:
@databus23 Do you still have interest in working on this?

databus23:
@nathanleclaire I do. i got distracted by other tasks and this fell off the desk. That's being said I'm one week away from a two week vacation and won't be able to attend to this before the vacation. So this will linger for at least three more weeks.

nathanleclaire:
@databus23 No problem, just checking in.

dmp42:
Hey @databus23 - do you still have plans to work on this?

databus23:
@dmp42 Yes, looking into it now.

jeanlaurent:
Closing since this is fixed by #1856

Issue #447, opened 2015-01-30
One time storage for local ISOs


sthulb:
At the moment boot2docker iso files are stored centrally and for each machine using it, we should stop that and only use the central iso.

nathanleclaire:
The issue with this is that `docker-machine upgrade dev` should only work on that machine.  If the ISOs are shared, it will affect all of them, and also we would need to make sure they are all powered down, and so on.  So, I'm closing this.

PR #603, opened 2015-02-24
Better handling of ISO files

sthulb:
The aim of this PR is to:
 - [ ] allow GitHub enterprise installs (#537)
 - [ ] allow local ISO images to be used (#539)
 - [ ] version ISO files by type and version (#217)
 - [ ] to only use one copy of an ISO (#447)

## Boot2Docker
To allow for better flexibility of Boot2Docker downloads, this PR introduces three environment variables:
 - `MACHINE_B2D_GITHUB_WEB`: Web hostname (with protocol)
 - `MACHINE_B2D_GITHUB_API`: API hostname (with protocol)
 - `MACHINE_B2D_GITHUB_REPO` GitHub repo name (org/repo)

### DIFF
--- a/utils/image/download_strategy.go
+++ b/utils/image/download_strategy.go
@@ -1,0 +1,85 @@
+package image
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+)
+
+var (
+	ErrNoDownloadUrl = errors.New("No download URL found")
+)
+
+type Metadata map[string]string
+
+type DownloadStrategy interface {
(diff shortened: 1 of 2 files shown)


utils/image/download_strategy.go
```diff
@@ -0,0 +1,85 @@
+package image
```
sthulb: I already hate this implementation. I'll redo it.

ehazlett: lol

PR #604, opened 2015-02-24
Make multiple machine name arguments work for kill, restart, rm, start, stop, and upgrade

### LINKED ISSUE #563: Running machine commands with multiple machine names silently fails
```
$ docker-machine start machine1 machine2
```

Only starts machine1. If commands don't support >1 arguments, I would expect it to complain instead of silently not start machine2.

nathanleclaire:
Thanks to @ehazlett for the start of this one.

cc @bfirsh @sthulb PTAL

Fixes #563

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -9,9 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	"sync"
 	"text/tabwriter"
-	"time"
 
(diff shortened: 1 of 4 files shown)


ehazlett:
LGTM

nathanleclaire:
Woot!

PR #610, opened 2015-02-25
refactor directories to machines, certs, cache

ehazlett:
This refactors the directories in machine to be `machines` for all machines, `certs` for CA and client certs/keys, and `cache` for the image cache (b2d) to not pollute the root directory.  This will cause less technical debt for the future and eliminate the need for a migration path in the future.

I think it would be quite a bit of migration to regenerate certs for the old ones and make sure we don't overwrite any custom certs in the process, etc.  IMHO we should do this before the 0.1.0 release.

Refs #217 and #603

### DIFF
--- a/host_test.go
+++ b/host_test.go
@@ -4,13 +4,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
 
 	_ "github.com/docker/machine/drivers/none"
-	"github.com/docker/machine/utils"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


ehazlett:
Note: this will make your existing machines unusable.  This is one of the reasons we should do this before the beta release.

host_test.go
```diff
@@ -118,37 +117,6 @@ func TestValidateHostnameInvalid(t *testing.T) {
 	}
 }
 
-func TestGenerateClientCertificate(t *testing.T) {
```
bfirsh: Why did this need to be removed?

bfirsh:
LGTM

host_test.go
ehazlett: It was testing a function that wasn't being used so I removed them both.

bfirsh: Cool. :+1:

sthulb:
LGTM

nathanleclaire:
Already been merged but, LGTM.  Without any existing machines around there's no problem creating etc.    This kind of thing is definitely why we should have a big fat "WARNING: BETA SOFTWARE THAT MAY CAUSE BREAKING CHANGES" sticker in our README, though.

bfirsh:
@nathanleclaire agreed.

PR #611, opened 2015-02-25
Remove binaries in build after building image

bfirsh:
This will bail before deleting the binaries if it can't connect to a Docker host. Sometimes I build machine binaries when I haven't got a machine and then it removes the machine binaries and then I can't use machine to create a machine and then the universe implodes. :(


ehazlett:
+1

ehazlett:
LGTM

PR #615, opened 2015-02-25
switch to bats for integration testing

ehazlett:
This uses https://github.com/sstephenson/bats for integration testing using the compiled binary.  Huge thanks to @aluzzardi for recommending!  Details are in the readme.  This also removes the old integration tests (they were quirky anyway).

Note: To test a cloud provider, you will need the credentials loaded as environment variables.

You specify a `DRIVER` and bats will run the tests (commands) against that driver:

```
ehazlett machine> bats test/driver-virtualbox.bats
 ✓ virtualbox: machine should not exist
 ✓ virtualbox: create
 ✓ virtualbox: active
 ✓ virtualbox: ls
 ✓ virtualbox: url
 ✓ virtualbox: ip
 ✓ virtualbox: ssh
 ✓ virtualbox: stop
 ✓ virtualbox: machine should show stopped
 ✓ virtualbox: start
 ✓ virtualbox: machine should show running after start
 ✓ virtualbox: restart
 ✓ virtualbox: machine should show running after restart
 ✓ virtualbox: remove
 ✓ virtualbox: machine should not exist

15 tests, 0 failures
```


aluzzardi:
This looks awesome :)

Looking forward to use bats for https://github.com/docker/swarm/issues/376

test/cli.bats
```diff
@@ -0,0 +1,106 @@
...
+}
+
+@test "cli: show active help" {
+  run ./docker-machine_$PLATFORM-$ARCH active -h
+  [ "$status" -eq 0  ]
+  [[ ${lines[3]} =~ "command active"  ]]
+}
+
+@test "cli: show config help" {
+  run ./docker-machine_$PLATFORM-$ARCH config -h
```
aluzzardi: Minor nit:

I think you could define a function somewhere, say `machine` that does the run with $PLATFORM-$ARCH etc so that you could write instead:

```
machine config -h
```

So you won't have to repeat this every time.

sthulb:
This looks great! I could think of a few places to use it at work :)

nathanleclaire:
Sweet!  Will have to give it a whirl.

test/cli.bats
ehazlett: +1 thx!

ehazlett:
@nathanleclaire @sthulb PTAL

integration-tests/driver-vmwarevcloudair.bats
```diff
@@ -0,0 +1,111 @@
...
+  [[ ${lines[1]} == *"Running"*  ]]
+}
+
+@test "$DRIVER: remove" {
+  run machine rm -f $NAME
+  [ "$status" -eq 0  ]
+}
+
+@test "$DRIVER: machine should not exist" {
+  run machine active $NAME
```
sthulb: Is there a difference between this `machine` and where you're using `docker-machine_$PLATFORM_$ARCH`?

ehazlett: There is no difference.  `machine` is simply a function that returns `docker-machine_$PLATFORM_$ARCH`

sthulb:
LGTM. Can we get the core tests added to travis?

ehazlett:
@sthulb i'm going to merge for now.  i will check on adding the core tests to travis as i'm not sure what the variable exposure is or whether we want to place keys there.  thanks!

Issue #491, opened 2015-02-05
Problems using a custom boot2docker ISO hosted on an internal HTTPS site with self-signed cert


hairyhenderson:
So I'm getting this error:

```sh
$ VIRTUALBOX_BOOT2DOCKER_URL=https://my-internal-server.mydomain.com/boot2docker.iso
$ docker-machine create -d virtualbox --virtualbox-boot2docker-url=$VIRTUALBOX_BOOT2DOCKER_URL dev         
INFO[0000] Downloading boot2docker.iso from https://my-internal-server.mydomain.com/boot2docker.iso... 
ERRO[0000] Error creating machine: Get https://my-internal-server.mydomain.com/boot2docker.iso: x509: certificate signed by unknown authority 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine        
```

I need a custom `boot2docker.iso` because our internal HTTPS all uses an internal corporate Root CA for signing (so, basically self-signed). I had this working fine with boot2docker-cli (see boot2docker/boot2docker#347), but I want to try out docker-machine with the same ISO.

I'm on OS X, and I've installed the Root CA certs in such a way that most apps recognize it (browsers, etc...). Is there some special _other_ way to install my certs such that docker-machine will be OK with them?

Thanks!

hairyhenderson:
Some clarification - this actually doesn't work with boot2docker-cli either. I had built the ISO locally at first, so it never had to actually _download_ it. So I guess this is a more general issue. Anyone had any experience with this?

sthulb:
I'm not quite sure how this would work. I guess it all depends on how "net/http" reads CA certs from the system, I'd assume it'll use the keychain on OS X and the CA store on Linux...

hairyhenderson:
Yeah, I did some `dtruss`ing and found that it's opening the various keychains in the right order (local, followed by System, followed by System Roots). The certs are in the System keychain. Other (non-golang) commands work fine, such as `curl`. I `dtruss`ed curl and it looked like it was opening the same keychains... Obviously I have no idea what `net/http` is _doing_ with those keychains, but... ;)

databus23:
We are experiencing this problem as well. It seems to me this is caused by the cross-compilation on linux. If I `go run` a simple go file that creates a http client this same way as [getClient] (https://github.com/docker/machine/blob/master/utils/b2d.go#L23) it works but not when building the docker-machine binary with `script/build`

[...]

PR #625, opened 2015-02-26
respect proxy settings for b2d downloads

databus23:
This fixes docker-machine with boot2docker behind a (corporate) http_proxy

### DIFF
--- a/utils/b2d.go
+++ b/utils/b2d.go
@@ -22,8 +22,9 @@ func defaultTimeout(network, addr string) (net.Conn, error) {
 
 func getClient() *http.Client {
 	transport := http.Transport{
-		Proxy: http.ProxyFromEnvironment,
-		Dial:  defaultTimeout,
+		DisableKeepAlives: true,
+		Proxy:             http.ProxyFromEnvironment,
+		Dial:              defaultTimeout,
 	}
 
 	client := http.Client{
--- a/utils/b2d.go
+++ b/utils/b2d.go
@@ -22,7 +22,8 @@ func defaultTimeout(network, addr string) (net.Conn, error) {
 
 func getClient() *http.Client {
 	transport := http.Transport{
-		Dial: defaultTimeout,
+		Proxy: http.ProxyFromEnvironment,
+		Dial:  defaultTimeout,
 	}
 
 	client := http.Client{


shawnzhu:
+1.

@ehazlett I'm curious if it's worth to customize transport of http client just because we want http connect timeout (not https) to be 5 seconds. the default transport will provide proxy support. I also found that it customize transport in docker base because of customized `tlsConfig`.

ehazlett:
@shawnzhu i'm ok with removing the dialTimeout and use the default as long as it supported proxy (I haven't verified that).  I'm weary of a long timeout trying to reach the b2d endpoint, but I'm ok following upstream if docker does the same.

shawnzhu:
@ehazlett According to the [http client of docker registry client](https://github.com/docker/docker/blob/master/registry/registry.go#L46), the customized parameters are:

* `DisableKeepAlive`
* `Proxy`
* `TLSClientConfig`

Plus a timeout switch on ConnectionTimeout (5 seconds) or ReceiveTimeout (1min). So you're right about keeping the timeout.

So I prefer this PR plus option `DisableKeepAlive`

databus23:
@bfirsh Any specific reason this removed from the 0.2.0 milestone? Its a pretty low hanging fruit which improves the usability of machine in corporate networks with mandatory http_proxy considerably?

radiumx3:
@bfirsh @ehazlett  Is this going to be implemented anytime soon ? We were about to prototype a solution based on docker machine but this is the first struggle of our long journey.
Why it has been dropped from 0.2 without any communication about it ?!

ehazlett:
@radiumx3 @databus23 no just misclick in waffle ;)  I'm reviewing this today.

ehazlett:
@databus23 would you mind adding the suggestions from @shawnzhu ?  I think matching the upstream docker updates would be good.

@radeksimko @databus23 are you ok with testing?  i can get a proxy setup local but having some more testing would be great.  thanks!

databus23:
@ehazlett Do you mean adding the `DisableKeepAlive` option? Sure why not. I don't think we should add the whole `TlsClientConfig` from the registry though. I don't think the special certificate handling (/etc/docker/certs.d/...) makes sense for machine.

databus23:
@ehazlett Oh and I'm very much ok with testing. As it stands docker-machine does not work in my corporate network because of this. :)

sthulb:
+1 This should have been here for 0.1.0 :)

ehazlett:
@databus23 good point.  let's leave the TLS off -- we can always add it if neeede.  thanks!

databus23:
So I added the `DisableKeepAlive` option, rebased on master and tested that it works in my corporate environment with a mandatory http proxy.

sirlatrom:
Please *do* add it, if that will enable me to specify e.g. a CA to use when
communicating with specific hosts. Otherwise I will not be able to use
machine...

databus23:
@sirlatrom I don't think implementing a custom trust store for certificates is the way to go. Instead docker-machine should be using the default ca store of the operating system and if you use a self-signed certificate or custom CA it should be placed there. There is currently a bug with that at least on OSX which is tracked here https://github.com/docker/machine/issues/491

sthulb:
I think maybe we should add it. Explicitly for the OS X bug.

databus23:
@sthulb I would prefer a general solution to the problem as proposed in #491. This is a general problem when shipping cross-compiled (darwin) go binaries that do https communication. Adding this for the b2d download case only solves one specific case. This will come up again and again for different cases and requires to customize every `http.Transport`.
Also I don't think its a good practice  to maintain separate trust stores for https clients, they should use the system store IMO.

databus23:
In general I think the custom certificate discussion should be moved to #491. This issue is about fixing docker-machine behind a mandatory proxy (aka in a corporate environment). It is as easy as a one line change, doesn't break anything and I would very much like it to happen fast.
Even adding the `DisableKeepAlive` option was already out of scope IMO but I went with it in the hopes of getting closer to a merge. :)

sthulb:
You're right, I never meant it in terms of your issue, I was just voicing an opinion :)

But in general, it might be nice to add it back.

ehazlett:
I agree with @databus23.  The cert issue is different and I don't think we want to manually manage a cert store (I am not in favor of this).

ehazlett:
Tested with local Squid to GH for b2d -- works great.  Thanks!  LGTM

sirlatrom:
You're absolutely right about that. I'm looking forward to trying it out in
Windows behind a corporate proxy which inserts its own certificate in the
chain.

PR #626, opened 2015-02-26
add trailing newline when unsetting env in fish

databus23:
This fixes using docker-machine with boot2docker behind a (corporate) http_proxy


ehazlett:
LGTM

PR #628, opened 2015-02-26
add flags for ipv6, backups and private networking to the DigitalOcean driver

knarz:
This adds support for creating droplets with IPv6, backups and private networking.

I also started working on integrating checks to prevent creating droplets
in invalid regions, and for the future check if all the desired features are
available in the given region.

Signed-off-by: Sascha Hanse <shanse@gmail.com>


ehazlett:
thanks!  i'm not sure if we need the backups option or not.  I think perhaps that could stay in the DO UI.  other than that, I like the ipv6 and private networking.

/cc @nathanleclaire

nathanleclaire:
This LGTM

knarz:
Not sure what happened here. I'll just go ahead and submit a new pull request.

nathanleclaire:
Thanks @knarz

PR #633, opened 2015-02-26
Fix #629 by adding PrivateIPAddress to config.json

### LINKED ISSUE #629: Proposed: save public and private IP of ec2 instances in config.json
to make it possible to use "machine inspect" to get the private IP it would be helpful if it was saved in config.json.  Some systems like consul need to be configured to advertise the private IP and join gossip pools on the private IP space so this would make it easier to script the creation of a consul cluster created with machine.

andrewwatson:
Signed-off-by: Andrew Watson <andrew@ionicsecurity.com>

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -281,7 +281,7 @@ func (d *Driver) Create() error {
 	log.Debugf("created instance ID %s, IP address %s, Private IP address %s",
 		d.InstanceId,
 		d.IPAddress,
-		instance.NetworkInterfaceSet[0].PrivateIpAddress,
+		d.PrivateIPAddress,
 	)
 
 	log.Infof("Waiting for SSH on %s:%d", d.IPAddress, 22)
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -44,6 +44,7 @@ type Driver struct {
 	InstanceId        string
 	InstanceType      string
 	IPAddress         string
+	PrivateIPAddress  string
 	MachineName       string
 	SecurityGroupId   string
 	SecurityGroupName string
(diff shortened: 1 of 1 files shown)


drivers/amazonec2/amazonec2.go
```diff
@@ -327,8 +332,8 @@ func (d *Driver) GetIP() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	d.IPAddress = inst.IpAddress
-	return d.IPAddress, nil
+
```
andrewwatson: I didn't like that this Getter method was modifying the Driver struct.

ehazlett: +1

drivers/amazonec2/amazonec2.go
```diff
@@ -269,12 +270,19 @@ func (d *Driver) Create() error {
...
+	}
 
 	d.waitForInstance()
 
-	log.Debugf("created instance ID %s, IP address %s",
+	log.Debugf("created instance ID %s, IP address %s, Private IP address %s",
 		d.InstanceId,
-		d.IPAddress)
+		d.IPAddress,
+		instance.NetworkInterfaceSet[0].PrivateIpAddress,
```
nathanleclaire: Let's use `d.PrivateIPAddress` here instead - it is less verbose and will default to empty string if this field is not present in the response.

drivers/amazonec2/amazonec2.go
nathanleclaire: +1 if this is called from `docker-machine ip` it does not even save the config so it doesn't matter to set this field in the struct

drivers/amazonec2/amazonec2.go
andrewwatson: good call.

ehazlett:
LGTM

nathanleclaire:
LGTM

PR #646, opened 2015-02-27
Fix none driver socket issue

### LINKED ISSUE #643: Cannot use unix socket URL with none driver
I use ssh & socat to port forward a remote /var/run/docker.sock as a local unix socket.
I add this local unix socket using the none driver.

docker-machine ls output:
```
NAME    ACTIVE   DRIVER         STATE     URL                                             SWARM
dev              vmwarefusion   Running   tcp://192.168.2.130:2376
vultr   *        none                     unix:///Users/bsuh/.remote-sockets/vultr.sock
```

docker-machine env output:
```
docker-machine env vultr
set -x DOCKER_TLS_VERIFY yes
set -x DOCKER_CERT_PATH /Users/bsuh/.docker/machine/machines/vultr
set -x DOCKER_HOST unix:///Users/bsuh/.remote-sockets/vultr.sock
```

It tries to use TLS on a Unix socket URL.

nathanleclaire:
Fixes #643.

PLEASE DO NOT MERGE YET.  I want to add tests first.

cc @ehazlett 

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -39,6 +39,7 @@ type machineConfig struct {
 	clientCertPath string
 	clientKeyPath  string
 	machineUrl     string
+	protocol       string
 	swarmMaster    bool
 	swarmHost      string
 	swarmDiscovery string
@@ -336,6 +337,9 @@ func cmdConfig(c *cli.Context) {
 	}
 	dockerHost := cfg.machineUrl
 	if c.Bool("swarm") {
+		if cfg.protocol != "tcp" {
+			log.Fatal("Error: using swarm with a non-TCP protocol is not supported")
+		}
 		if !cfg.swarmMaster {
 			log.Fatalf("%s is not a swarm master", cfg.machineName)
 		}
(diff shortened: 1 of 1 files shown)


bsuh:
Should it unset environment variables and then set them? If your previous machine had TLS on and you change to a machine using unix sockets, it seems like TLS would remain on.

ehazlett:
I'm not a fan of this.  This will mean we will need to throw errors for `start`, `stop`, `ssh`, `upgrade`, and `kill`.  Seems counter-intuitive.  I can see the benefit of seeing the local instance if there is one, but why not just `$(docker-machine env -u)`?

bsuh:
The none driver already throws errors for those.

ehazlett:
@bsuh true, but I'm not sure where that is used outside of testing.  I guess it fits in this case but I'm not sure we should have a driver that negates the majority of the functionality.

bsuh:
To be honest, I'm just getting my feet wet with devops with side projects, so perhaps I'm not the target audience for machine. I'm not using Docker to build distributed systems, but I use it for isolation. I don't need to create a lot of Docker hosts automatically or bring them up and down, but I'd still like to use machine for keeping a list of hosts & connecting to those hosts that I create manually on unsupported hosting services. If the none driver already exists and supports TCP sockets, I don't see why it couldn't support Unix sockets.

PR #649, opened 2015-02-27
keep machine URL consistent in machine config and machine env

shawnzhu:
Now it output double-quoted string value for docker host only when performing command `machine config`:

    $ machine config
    --tls --tlscacert=/.client/ca.pem --tlscert=/.client/cert.pem --tlskey=/.client/key.pem -H="tcp://192.168.99.100:2376"

Which works for `docker` command only because of docker/docker#6147 but not for other project like [drone/drone](https://github.com/drone/drone) which uses `codegangsta/cli` that does not trim quotes for string flag all the time.

This PR will enable scenario like command `drone build $(docker-machine config)` on OS X, which doesn't work util it removes the double quotes of docker host from command `machine config`. (notice that it doesn't use double quotes in `machine env` too)

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -1,10 +1,13 @@
 package main
 
 import (
+	"bytes"
 	"flag"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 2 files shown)


ehazlett:
LGTM

commands_test.go
```diff
@@ -293,3 +296,49 @@ func TestRunActionForeachMachine(t *testing.T) {
...
+	c := cli.NewContext(nil, set, set)
+
+	cmdConfig(c)
+
+	if strings.Contains(testOutput.String(), "-H=unix:///var/run/docker.sock") {
+		t.Fatalf("Expect docker host")
+	}
+
+	os.Stdout = stdout
+	os.Setenv("MACHINE_STORAGE_PATH", "")
```
duglin: Should these type of "undo" things be in some `defer`s so that when the t.Fatal's are run we still reset the env?

ehazlett: absolutely +1 -- thx!

@shawnzhu I will do a follow up PR after this is merged to add all of the `defer` cleanup.

shawnzhu: I just updated this PR to use `defer`.

nathanleclaire:
Arg, I could have sworn there was a reason I implemented it this way, but unfortunately I can't remember why.  So I'm hoping it was just aesthetics - LGTM

ehazlett:
thx

PR #654, opened 2015-02-27
Provision OpenStack instances with Docker

gierschv:
Since 54f36078, Machine doesn't check anymore if Docker is installed
in the created instances, which breaks the compatibility between
v0.1.0-rc5 and v0.1.0 when using Machine with some cloud providers
like RunAbove.com.


ehazlett:
LGTM

ggiamarchi:
LGTM

PR #655, opened 2015-02-27
Add start of a change log

bfirsh:
We should probably also have a release checklist to remind us to update this!


ehazlett:
LGTM

PR #664, opened 2015-02-27
ec2: check subnet exists in vpc

### LINKED ISSUE #663: EC2: Seems like same VPC isn't used for subnet and security groups
Here are my steps to duplicate:

1. Create a new VPC on AWS.  CIDR=`10.0.0.0/24` Name=`fordocker` ID=`vpc-3a299c5f`
2. Create machine:
```
$ dmachine -D create -d amazonec2 --amazonec2-region us-west-2 --amazonec2-access-key $AWS_ACCESS_KEY_ID --amazonec2-secret-key $AWS_SECRET_ACCESS_KEY --amazonec2-vpc-id vpc-3a299c5f aws-test-1
INFO[0000] Launching instance...                        
DEBU[0000] executing: /usr/bin/ssh-keygen ssh-keygen -t rsa -N  -f /Volumes/FatMan/home/matt/.docker/machine/machines/aws-test-1/id_rsa
 
Your identification has been saved in /Volumes/FatMan/home/matt/.docker/machine/machines/aws-test-1/id_rsa.
Your public key has been saved in /Volumes/FatMan/home/matt/.docker/machine/machines/aws-test-1/id_rsa.pub.
The key fingerprint is:
47:01:6a:07:9b:b1:1c:64:b3:ed:1b:5a:14:99:37:0c matt@boneyard.sm
The key's randomart image is:
+--[ RSA 2048]----+
|     .B E*.      |
|     o %o.+.     |
|      O +...     |
|     . + .       |
|        S .      |
|       o +       |
|      . .        |
|                 |
|                 |
+-----------------+
DEBU[0000] creating key pair: aws-test-1                
DEBU[0000] configuring security group in vpc-3a299c5f   
DEBU[0000] creating security group (docker-machine) in vpc-3a299c5f 
DEBU[0000] waiting for group (sg-62d5f807) to become available 
DEBU[0001] configuring security group authorization for 0.0.0.0/0 
DEBU[0001] authorizing group  with permissions: [{tcp 22 22 0.0.0.0/0} {tcp 2376 2376 0.0.0.0/0}] 
DEBU[0001] launching instance in subnet subnet-f8ad088f 
ERRO[0001] Error creating machine: Error launching instance: Problem with AWS API call: Non-200 API response: code=400 message=Security group sg-62d5f807 and subnet subnet-f8ad088f belong to different networks.
 
WARN[0001] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0001] Error creating machine
```

The Security Group it made seems correct: Name=`docker-machine` ID=`sg-62d5f807` VPC=`vpc-3a299c5f (10.0.0.0/24)`

It chose to reuse an existing subnet that isn't even in the same VPC.  ID=`subnet-f8ad088f` VPC=`vpc-3f10b25a (10.0.0.0/16)` CIDR=`10.0.2.0/24`

I think that other times I ran it, it would create a subnet, but the subnet would be in the default VPC (not the one I just created).

ehazlett:
Fixes #663 

This also fixes the IP allocation eventual consistency.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -210,6 +210,42 @@ func (d *Driver) checkPrereqs() error {
 	if key != nil {
 		return fmt.Errorf("There is already a keypair with the name %s.  Please either remove that keypair or use a different machine name.", d.MachineName)
 	}
+
+	regionZone := d.Region + d.Zone
+	if d.SubnetId == "" {
+		filters := []amz.Filter{
+			{
+				Name:  "availabilityZone",
+				Value: regionZone,
+			},
+			{
+				Name:  "vpc-id",
+				Value: d.VpcId,
+			},
+		}
+
+		subnets, err := d.getClient().GetSubnets(filters)
+		if err != nil {
+			return err
+		}
+
+		if len(subnets) == 0 {
+			return fmt.Errorf("unable to find a subnet in the zone: %s", regionZone)
+		}
+
+		d.SubnetId = subnets[0].SubnetId
+
+		// try to find default
+		if len(subnets) > 1 {
+			for _, subnet := range subnets {
+				if subnet.DefaultForAz {
+					d.SubnetId = subnet.SubnetId
+					break
+				}
+			}
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


drivers/amazonec2/amazonec2.go
```diff
@@ -250,7 +250,7 @@ func (d *Driver) Create() error {
 		}
 
 		for _, s := range subnets {
-			if s.AvailabilityZone == regionZone {
+			if s.AvailabilityZone == regionZone && s.VpcId == d.VpcId {
```
md5: Both availability zone and VPC id can be passed as `Filter.N` params in the call to `DescribeSubnets`.

md5: It might also be reasonable to see if one of the subnets is the default for the AZ and prefer that one.

iffy:
This fixes my problem from #663 :+1:

drivers/amazonec2/amazonec2.go
ehazlett: +1 to both :)

ehazlett:
@md5 PTAL

ehazlett:
@md5 it should "fallback" safely.  It will only check for the default if there is more than one and will only set it if there is one marked.  Otherwise, it uses the first one it finds.

ehazlett:
@md5 i'm not sure what you mean by log the err -- if an error occurs while waiting on the IP, it's returned.

md5:
It's possible to have more than one subnet with none marked as
defaultForAz, I believe.

Regarding my other concern, I misspoke. I meant the case where ip keeps
being "" and err is also nil.

ehazlett:
Correct. In the case where one is marked as default it will use that one. Otherwise it will just use the one already set. 

We need a retry and timeout function in the amz lib itself. I'll look into this for better retry and timeout support. Thanks!

ehazlett:
Ah gotcha. Good point.

md5:
LGTM

I had missed the `d.SubnetId = subnets[0].SubnetId` in my first pass through the diff, so I see how it falls back now if there is no `defaultForAz`.

PR #665, opened 2015-02-27
Test SSH establishing a real SSH connection

### LINKED ISSUE #656: [openstack] SSH password
When creating a new docker machine using the openstack driver, I am asked for a ssh password:

    DEBU[0081] Waiting for the SSH server to be started...   IP=x.x.x.x MachineId=0d893949-aae6-4777-aa7e-378216f3c472
    DEBU[0098] generating server cert: /Users/user/.docker/machine/machines/test103/server.pem ca-key=/Users/user/.docker/machine/certs/ca.pem private-key=/Users/user/.docker/machine/certs/ca-key.pem org=test103
    DEBU[0100] Stopping Docker...
    DEBU[0100] Command: %s[sudo sh -c 'sudo service docker stop']  MachineId=0d893949-aae6-4777-aa7e-378216f3c472
    DEBU[0100] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -p 22 -i /Users/user/.docker/machine/machines/test103/id_rsa centos@x.x.x.x sudo sh -c 'sudo service docker stop'
    centos@x.x.x.x's password:

When I try to connect to that machine manually, I am not asked for a password though. This seems to be the same issue as described in #480 .
Is there a way to force docker-machine to wait for a specific period before connecting to the machine?

ggiamarchi:
It's done in the OpenStack driver here but i think it should be generalize because this issue can potentially affect any machine.

Fix #656

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -736,7 +736,37 @@ func (d *Driver) waitForSSHServer() error {
 		"MachineId": d.MachineId,
 		"IP":        ip,
 	}).Debug("Waiting for the SSH server to be started...")
-	return ssh.WaitForTCP(fmt.Sprintf("%s:%d", ip, d.SSHPort))
+
+	err = ssh.WaitForTCP(fmt.Sprintf("%s:%d", ip, d.SSHPort))
+	if err != nil {
+		return err
+	}
+
+	retryCount := 0
+	for {
+		log.WithFields(log.Fields{
+			"MachineId": d.MachineId,
+			"IP":        ip,
+			"SSHUser":   d.SSHUser,
+		}).Debug("Trying to establish a SSH connection...")
+
+		cmd, err := d.GetSSHCommand("exit 0")
+		if err != nil {
+			return err
+		}
+		if err := cmd.Run(); err != nil {
+			retryCount++
+			// Avoid infinite retries
+			// A big enough retryCount limit is arbitrary chosen
+			if retryCount == 100 {
+				return err
+			}
+			time.Sleep(time.Second)
+			continue
+		}
+		break
+	}
+	return nil
 }
 
 func (d *Driver) waitForInstanceToStart() error {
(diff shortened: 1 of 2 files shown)


drivers/openstack/openstack.go
```diff
@@ -736,7 +736,37 @@ func (d *Driver) waitForSSHServer() error {
 		"MachineId": d.MachineId,
 		"IP":        ip,
 	}).Debug("Waiting for the SSH server to be started...")
-	return ssh.WaitForTCP(fmt.Sprintf("%s:%d", ip, d.SSHPort))
+
+	err = ssh.WaitForTCP(fmt.Sprintf("%s:%d", ip, d.SSHPort))
```
ehazlett: Perhaps we update `ssh.WaitForTCP` to accepts a retries itself?

ctrlrsf: Or how about a new ssh.WaitForSSH function that would loop until "exit 0" could be successfully executed like @ggiamarchi implemented for the OpenStack driver? Other users of ssh.WaitForTCP could be changed to this as needed.

ehazlett: @ctrlrsf +1

ggiamarchi: I like the @ctrlrsf 's solution too. @ehazlett do you wish i replace calls of `WaitForTCP` by the new `WaitForSSH ` in all drivers in this PR ?

ehazlett: @ggiamarchi how about you just create the new retry wait.  then, as @ctrlrsf said, we can update the others as needed.

nathanleclaire: +1 to @ctrlrsf suggestion.  @ggiamarchi can you please update to create the generic function in `ssh` package and use it in the `openstack` driver in this PR please?  We will move the other drivers over to this new method when it is merged

nathanleclaire:
In the current master, there is now a generic (universal to all drivers) way that this is done.

@ehazlett perhaps we can close this issue?

Big thanks to @ggiamarchi for the original idea, by the way.

ehazlett:
Yep, this should be covered now.

@ggiamarchi thanks for the idea!

PR #668, opened 2015-02-28
show fish usage upon create

ehazlett:
This adds support for showing the proper usage for `env` upon creation when using the fish shell.

Example:

```
...
INFO[0038] "test-vbox" has been created and is now the active machine.
INFO[0038] To point your Docker client at it, run this in your shell: docker-machine_darwin-amd64 env test-vbox | source
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -325,8 +325,21 @@ func cmdCreate(c *cli.Context) {
 		log.Fatalf("error setting active host: %v", err)
 	}
 
+	info := ""
+	userShell := filepath.Base(os.Getenv("SHELL"))
+
+	switch userShell {
+	case "fish":
+		info = fmt.Sprintf("%s env %s | source", c.App.Name, name)
+	default:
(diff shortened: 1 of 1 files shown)


commands.go
```diff
@@ -325,8 +325,21 @@ func cmdCreate(c *cli.Context) {
...
+	case "fish":
+		info = fmt.Sprintf("%s env %s | source", c.App.Name, name)
+	default:
+		info = fmt.Sprintf("$(%s env %s)", c.App.Name, name)
+	}
+
 	log.Infof("%q has been created and is now the active machine.", name)
-	log.Infof("To point your Docker client at it, run this in your shell: $(%s env %s)", c.App.Name, name)
+
+	if info != "" {
```
sthulb: This if statement probably isn't needed, the default from the switch would cover it

sthulb:
LGTM -- I'll take your word that fish works that way :)

ehazlett:
@sthulb thx.

including https://github.com/docker/machine/pull/560#issuecomment-76477083 for reference

Issue #669, opened 2015-02-28
Use other iso than boot2docker to create machine based on virtualbox, but failed


TrumanLing:
Input to create machine:
```console
truman@truman:~/work/machine-bak$ sudo ./docker-machine_linux-amd64 -D create -d virtualbox --virtualbox-boot2docker-url=http://rnd-isourceb.huawei.com/docker/dockos/blob/master/DockOS.iso dev
```
output:
```console
DEBU[0000] STDERR:                                      
INFO[0000] Downloading boot2docker.iso from http://rnd-isourceb.huawei.com/docker/dockos/blob/master/DockOS.iso... 
ERRO[0039] Error creating machine: open /home/truman/.docker/machine/cache/boot2docker.iso/boot2docker.iso.tmp108446320: not a directory
```

shawnzhu:
The error message indicated that the downloading path is wrong. see [my comment](/docker/machine/issues/670#issuecomment-77972562)

shawnzhu:
@TrumanLing the option `--virtualbox-boot2docker-url` is fixed in #805

PR #670, opened 2015-02-28
Make improvement for supporting other iso with virtualbox

TrumanLing:
Fix issue #669 
Enforce support with other iso, deatail modification:
Dynamically get iso filename
Dynamically set dvddrive medium
Dynamically set docker configration directory

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -24,7 +24,7 @@ import (
 	"github.com/docker/machine/utils"
 )
 
-const (
+var (
 	dockerConfigDir = "/var/lib/boot2docker"
 	isoFilename     = "boot2docker.iso"
 )
@@ -131,8 +131,7 @@ func (d *Driver) Create() error {
 
 	b2dutils := utils.NewB2dUtils("", "")
 	imgPath := utils.GetMachineCacheDir()
-	isoFilename := "boot2docker.iso"
-	commonIsoPath := filepath.Join(imgPath, "boot2docker.iso")
+	commonIsoPath := filepath.Join(imgPath, isoFilename)
 	// just in case boot2docker.iso has been manually deleted
 	if _, err := os.Stat(imgPath); os.IsNotExist(err) {
 		if err := os.Mkdir(imgPath, 0700); err != nil {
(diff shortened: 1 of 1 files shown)


TrumanLing:
@bfirshv @ehazlett because of PR #693 , do you have plan to merge this PR

shawnzhu:
@TrumanLing @ehazlett do you think it's a good idea to fix option `--virtualbox-boo2docker-url` before adding more feature?

TrumanLing:
@shawnzhu done, thanks for your review. And I agree with you.

sthulb:
All instances of --<driver>-boot2docker-url should just become `--<driver>-image-url` for now.

drivers/virtualbox/virtualbox.go
```diff
@@ -553,9 +551,8 @@ func (d *Driver) diskPath() string {
 // Make a boot2docker VM disk image.
 func (d *Driver) generateDiskImage(size int) error {
 	log.Debugf("Creating %d MB hard disk image...", size)
-
-	magicString := "boot2docker, please format-me"
-
+	parts := strings.Split(isoFilename, ".")
+	magicString := fmt.Sprintf("%s, please format-me", parts[0])
```
sthulb: You can't assume that all images will support automatic disk formatting.

TrumanLing: @sthulb thanks for your review.I have removed code for dynamically set magic string.

sthulb:
@TrumanLing We're probably going to be stripping out all of the boot2docker stuff soon, in favour of cloudinit.

TrumanLing:
@sthulb I know your idea, but I think, same as what @shawnzhu said, it is better to fix it before addiong new feature.

bfirsh:
Thanks for this @TrumanLing. There aren't any immediate plans to remove boot2docker, so we should certainly still maintain what we have.

sthulb:
My comment was vague I guess, I meant in terms of supporting other distros, we might not have to worry about modifying boot2docker specific code to work with other images.

shawnzhu:
@TrumanLing @sthulb @bfirsh I believe we all want to fix the documented feature `virtualbox-boot2docker-url` because it doesn't work right now. a new PR as fix?

ehazlett:
@shawnzhu it actually does work but with b2d ISOs.  We are actually doing this with the vmwarefusion driver.  They use a custom b2d iso.

@TrumanLing thanks for the contribution but I'm not sure we want to change that much at the core of the virtualbox driver at this time.  Right now, the `--virtualbox-boot2docker-url` supports custom b2d ISOs from version 1.5+.

shawnzhu:
@ehazlett it is the option `--virtualbox-boot2docker-url` that doesn't work. see [Line147 of drivers/virtualbox/virtualbox.go](https://github.com/docker/machine/blob/master/drivers/virtualbox/virtualbox.go#L147), it should uses `imgPath` when saving iso file instead of `commonIsoPath`, or you will get error output like that in #669.

I would prefer to fix the option `--virtualbox-boot2docker-url` under this PR.

shawnzhu:
@ehazlett I created #826 to fix the option `--virtualbox-boot2docker-url` only.

ehazlett:
@shawnzhu thx -- making comments there.

nathanleclaire:
@TrumanLing Please take a look at making a new PR if you would still like to support this feature (this PR is somewhat outdated).

PR #686, opened 2015-03-02
ec2: do not try to auth swarm port if exists

### LINKED ISSUE #678: [amazonec2 driver] Attempt to create swarm master bails if SG rule already exists
We get an API error back because the SG rule for swarm already exists, but it's actually fine, it shouldn't affect the attempted creation at all.

screen shot 2015-03-01 at 3 19 04 pm

Edit:  Oops, wrong screenshot.

ehazlett:
Fixes #678

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -29,7 +29,11 @@ const (
 	ipRange                  = "0.0.0.0/0"
 	dockerConfigDir          = "/etc/docker"
 	machineSecurityGroupName = "docker-machine"
-	dockerPort               = 2376
+)
+
+var (
+	dockerPort = 2376
+	swarmPort  = 3376
 )
 
 type Driver struct {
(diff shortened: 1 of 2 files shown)


drivers/amazonec2/amazonec2.go
```diff
@@ -664,42 +658,52 @@ func (d *Driver) configureSecurityGroup(groupName string) error {
 	return nil
 }
 
-func configureSecurityGroupPermissions(group *amz.SecurityGroup) []amz.IpPermission {
+func (d *Driver) configureSecurityGroupPermissions(group *amz.SecurityGroup) []amz.IpPermission {
```
nathanleclaire: Ah, the test needs to be updated to reflect that this has been moved to a driver method.

ehazlett: ok tests fixed.  thanks!

nathanleclaire:
LGTM

ehazlett:
thanks!

PR #690, opened 2015-03-03
Introduce universal poll retries function

nathanleclaire:
I feel like we need something more uniform than the current situation - there are lots and lots of places where this pattern is needed, e.g. https://github.com/docker/machine/pull/665

Think it's important to have an answer so that contributors don't accidentally spin forever without a max attempts or spam a lot by forgetting a sleep.

We can move over instances of this pattern to this universal way as we see them.

Let me know what you think @ehazlett @sthulb 

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -11,7 +11,6 @@ import (
 	"path"
 	"strconv"
 	"strings"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
@@ -19,6 +18,7 @@ import (
 	"github.com/docker/machine/drivers/amazonec2/amz"
 	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/state"
+	"github.com/docker/machine/utils"
 )
 
 const (
@@ -279,6 +279,19 @@ func (d *Driver) PreCreateCheck() error {
 	return d.checkPrereqs()
 }
 
+func (d *Driver) instanceIpAvailable() bool {
+	ip, err := d.GetIP()
+	if err != nil {
+		log.Debug(err)
+	}
+	if ip != "" {
+		d.IPAddress = ip
+		log.Debugf("Got the IP Address, it's %q", d.IPAddress)
+		return true
+	}
+	return false
+}
+
 func (d *Driver) Create() error {
 	if err := d.checkPrereqs(); err != nil {
 		return err
@@ -309,18 +322,10 @@ func (d *Driver) Create() error {
 	}
 
 	d.InstanceId = instance.InstanceId
+
 	log.Debug("waiting for ip address to become available")
-	for {
-		ip, err := d.GetIP()
-		if err != nil {
-			return err
-		}
-		if ip != "" {
-			d.IPAddress = ip
-			log.Debugf("Got the IP Address, it's %q", d.IPAddress)
-			break
-		}
-		time.Sleep(5 * time.Second)
+	if err := utils.WaitFor(d.instanceIpAvailable); err != nil {
+		return err
 	}
 
 	if len(instance.NetworkInterfaceSet) > 0 {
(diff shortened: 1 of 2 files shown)


sthulb:
I like it, can you change the defaults to 10 retries over 10 seconds?

ehazlett:
+1 -- @sthulb i like it at 1 second.  This helps keep it fast as most operations will resolve within a few seconds and gives a good default 1 min timeout IMO.

sthulb:
I think 1 minute is too long.

ehazlett:
It's trying every 1 second for a max of 1 minute.  You think that's too long?

sthulb:
Yeah, I think the default should be lower, the things that we know require more can obviously just set it using the more verbose function.

ehazlett:
Ok so the interval of 1 second is fine, you just want a default max of 10 seconds?

sthulb:
yep!

ehazlett:
Cool.  I think we could try that and see what happens.  @nathanleclaire ?

nathanleclaire:
I'm OK with it.  I've just realized I should update the "wait until the instance is running" timeout to be a few minutes or so before we merge it in though, since it takes a while for newly created instances to boot up.

ehazlett:
@nathanleclaire lol yeah that probably would be a good thing :)

drivers/amazonec2/amazonec2.go
```diff
@@ -310,17 +310,18 @@ func (d *Driver) Create() error {
 
 	d.InstanceId = instance.InstanceId
 	log.Debug("waiting for ip address to become available")
-	for {
+	if err := utils.WaitFor(func() bool {
```
cjellick: I think pulling the anonymous function out of the if statement would make this code much more readable.

drivers/amazonec2/amazonec2.go
```diff
@@ -519,29 +521,6 @@ func (d *Driver) sshKeyPath() string {
 	return path.Join(d.storePath, "id_rsa")
 }
 
-func (d *Driver) updateDriver() error {
```
nathanleclaire: This didn't seem to be doing anything which wasn't already done by the main `Create` function or the other check for IP address, so I've taken it out.  Instance creation still works fine, so let me know if I'm overlooking something.

nathanleclaire:
OK I've pulled the closures out into real methods.  I'm waffling on the default timeout though; I have it set to 3 seconds interval / 3 minutes total here: @sthulb could you please elaborate on your concern?  I'd rather give a comfortable amount of space for resources to come up by default than accidentally bail when everything is fine.

drivers/amazonec2/amazonec2.go
sthulb: The IP address on EC2 instances normally changes on stop/start, I guess this function was for that use case.

sthulb:
@nathanleclaire I think 1 minute is too long. Perhaps 10 seconds is too short though, there's probably a middle ground somewhere :)

drivers/amazonec2/amazonec2.go
```diff
@@ -610,6 +593,17 @@ func (d *Driver) isSwarmMaster() bool {
 	return d.SwarmMaster
 }
 
+func (d *Driver) securityGroupAvailableFunc(id string) func() bool {
```
ehazlett: This is a bit nit picky but do we need to name this with "Func" ?

nathanleclaire: Hm, I don't really care either way but I called it that way to be more clear to the reader of the code that it actually returns a function as opposed to the others which return a boolean.  Happy to take it out if you think it's confusing or redundant.

ehazlett: nah that works

ehazlett:
I think this is good for now.  Until we get some more feedback I think the 3/3 is a good start to test with.  We can always update it if we find it's not enough or too much.

ehazlett:
LGTM

PR #692, opened 2015-03-03
Implement libvirt/KVM support

dhiltgen:
This adds a new driver for machine supporting KVM leveraging libvirt.  It presently supports networking models where libvirt runs a local dnsmasq for DHCP.

Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: go
 sudo: false
 go:
 - 1.3
+addons:
+  apt_packages:
+  - libvirt-dev
 install:
 - export GOPATH=${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH
 - export PATH=${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
 FROM golang:1.3-cross
-RUN apt-get update && apt-get install -y --no-install-recommends openssh-client
+RUN apt-get update && apt-get install -y --no-install-recommends openssh-client libvirt-dev
 
 # TODO: Vendor these `go get` commands using Godep.
 RUN go get github.com/mitchellh/gox
 RUN go get github.com/aktau/github-release
 RUN go get github.com/tools/godep
 RUN go get code.google.com/p/go.tools/cmd/cover
+RUN go get github.com/alexzorin/libvirt-go
 
 ENV GOPATH /go/src/github.com/docker/machine/Godeps/_workspace:/go
 ENV MACHINE_BINARY /go/src/github.com/docker/machine/docker-machine
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -20,6 +20,7 @@ import (
 	_ "github.com/docker/machine/drivers/generic"
 	_ "github.com/docker/machine/drivers/google"
 	_ "github.com/docker/machine/drivers/hyperv"
+	_ "github.com/docker/machine/drivers/kvm"
 	_ "github.com/docker/machine/drivers/none"
 	_ "github.com/docker/machine/drivers/openstack"
 	_ "github.com/docker/machine/drivers/rackspace"
--- a/drivers/kvm/kvm.go
+++ b/drivers/kvm/kvm.go
@@ -1,0 +1,1 @@
+package kvm
--- a/drivers/kvm/kvm_linux_amd64.go
+++ b/drivers/kvm/kvm_linux_amd64.go
@@ -1,0 +1,721 @@
+package kvm
+
+import (
+	"archive/tar"
+	"bytes"
+	"encoding/json"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"github.com/alexzorin/libvirt-go"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/codegangsta/cli"
+	"github.com/docker/machine/drivers"
+	"github.com/docker/machine/log"
+	"github.com/docker/machine/provider"
+	"github.com/docker/machine/ssh"
+	"github.com/docker/machine/state"
+	"github.com/docker/machine/utils"
+)
+
+const (
+	dockerConfigDir = "/var/lib/boot2docker"
+	isoFileName     = "boot2docker.iso"
+	dnsmasqLeases   = "/var/lib/libvirt/dnsmasq/%s.leases"
+	dnsmasqStatus   = "/var/lib/libvirt/dnsmasq/%s.status"
+
+	// TODO
+	// - Support multiple NICs
+	// - Support bridged networks via static IP configuration
+	domainXML = `<domain type='kvm'>
+  <name>%s</name> <memory unit='M'>%d</memory>
+  <vcpu>%d</vcpu>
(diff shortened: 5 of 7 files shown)


dhiltgen:
The build failed because libvirt-dev isn't installed on the build system.  My change includes an update to the Dockerfile to include it, so there must be another piece I need to update for the CI system.

drivers/libvirt/libvirt_linux_amd64.go
```diff
@@ -0,0 +1,695 @@
...
+		return err
+	}
+	switch nw.Forward.Mode {
+	case "nat":
+		fallthrough
+	case "route":
+		fallthrough
+	case "": // No forwarding
+		if nw.Ip.Address == "" {
+			log.Warnf("Network doesn't have valie IP range")
```
SvenDowideit: `valid` :)

drivers/libvirt/libvirt_linux_amd64.go
```diff
@@ -0,0 +1,695 @@
...
+	case "bridge":
+		log.Infof("Network %s bridged", name)
+		/*
+		 * TODO
+		 *  There isn't really a silver bullet for bridged networks.
+		 *  Since the DHCP server is most likely not on this box,
+		 *  we can't look up the IP address, and our ARP
+		 *  table probably wont have the MAC listed.  To support
+		 *  this we can ask the user to give us the IP.
+		 */
```
SvenDowideit: does this mean you're only creating one network  for the vm? B2D uses 2 to get around this problem - the first is a NATed one so that the cli can ask the vm for the current IP of the more general one (and at the same time, ask about the Docker daemon port and other configurables.

dhiltgen: Here's how I think it should probably work longer term:

Support multiple networks, and allow the user to specify zero or more on the command line.  If zero specified, use a default NAT style network.

If one of the chosen networks supports local dnsmasq, then use that to get the IP of the VM.

If no network supports local dnsmasq, then add one additional network using NAT.


That said, I was hoping the current behavior would be useful enough, so I was thinking the fancier logic could come in a subsequent change.

samalba:
What's the next step on this one? I need it.

ibuildthecloud:
Could this driver be called libvirt, not KVM.  I'm a huge KVM user but not so much libvirt.  I'd like to submit a pure KVM driver shortly, but I'd rather not have two :).

ehazlett:
@ibuildthecloud +1

I believe the intent is to have it KVM specific without exposing too much of libvirt as that opens a huge potential for maintenance difficulty (there are several providers available in libvirt and I could see it growing out of hand).

As an aside, I've been working on a side thing with KVM + Machine.  @dhiltgen perhaps we could sync and work together on a KVM specific, non-libvirt backed driver?  Thoughts? @ibuildthecloud @dhiltgen

dhiltgen:
I've always used libvirt to access KVM, so I don't have experience running it stand-alone.   That said, I'll be the first to admit libvirt is kind of a beast, I just saw it as a means to an end.  I just skimmed some of the docs floating around on "raw" KVM, and it looks pretty raw.  What were you guys thinking as far as networking topology/configuration?

ehazlett:
@dhiltgen in my other project I'm having awesome success with dnsmasq + vde2.  I would be happy to do a quick sync with thoughts.

@ibuildthecloud your thoughts on the networking, etc?

dave-tucker:
:+1: I need this driver!

Re: Networking, what's the benefit of using vde2? I've been a KVM + Libvirt + Open vSwitch user for a while now so not familiar with it. Using OVS, I can easily create L2 tunnels from host to host if I want a "distributed" switch. That said, you can achieve the same with a standard Linux bridge and it has the added benefit of not requiring additional packages on the host machine.

Question on libvirt vs pure KVM for the experts here... If machine creates VM's using pure KVM, will they still show up in `virt-manager`?

ehazlett:
@dave-tucker because i'm not a network guru and bridging is sketchy on my Debian Jessie install (periodic fails, interface drops / loss in the VM, etc).  I'm open to suggestions :)  I just want the most stable and simple setup possible to reduce the risk for issues.  That's my biggest fear with the KVM driver is the variety of network choices and potential failures.

vincentbernat:
One drawback of using VDE2 is that QEMU from Ubuntu is not compiled with its support because libvde is in universe while QEMU is in main and the quality of libvde is considered insufficient to move it to main.

I would suggest to keep libvirt since it handles network and storage and also it acts as some kind of orchestrator (start this one, stop this one, pause this one). All that should be developed again if KVM is used directly. Also, it enables the use of remote QEMU transparently (over SSH for example).

ehazlett:
@vincentbernat cool -- vde is out -- although it's funny to me that it's been the most stable in my testing ;)  My fear with libvirt is that (as mentioned above) others will "exploit" that to take machine to other providers that it is not intended for thus causing a maintenance burden.  Also, I personally don't use the libvirt stuff so we would need more to test.

clnperez:
I've been working on a driver for a product called Kimchi, which uses
libvirt under the covers. It's got a lot of the intelligence baked in
for ppc64, as well. For example, it will pick the right arch and machine
type:

x86:

<domain type='kvm'>
<os>
    <type arch='x86_64' machine='pc-i440fx-2.1'>hvm</type>
    <boot dev='hd'/>
</os>


Power:

<domain type='kvm'>
  <os>
    <type arch='ppc64' machine='pseries-2.2'>hvm</type>
    <boot dev='hd'/>
  </os>

There are other little things, like IDE disks are not supported on
Power, different memory minimums, etc.

That said, I'll definitely help with this one so that we make sure it's
flexible enough for multiple architectures. It seems basic enough (being
just a base 'machine') that there might not be really any issues to deal
with.

As for libvirt, the networking features alone are enough for me. And
libvirt has some things baked in for different arch's, IIRC.

Regards,

- Christy

SvenDowideit:
not sure why it was't linked, but #650 is a direct qemu/kvm driver

ehazlett:
@SvenDowideit thanks -- does that work?  I didn't see any hostonly network in that.

JohnStarich:
Has there been more progress on this?
I would really love a KVM driver for my personal development environment in Docker 1.7.
I've been using the driver @dhiltgen made and I've pulled in from docker/machine master without issue. Unfortunately, I'm still having some issues with networking on the machines where they will be provisioned correctly, set up and everything, yet somehow the certs aren't transferred over properly?

nathanleclaire:
Please see: https://github.com/docker/machine/issues/1626

dhiltgen:
I've ported my kvm driver over to take advantage of the new plugin architecture @nathanleclaire is working on.  The source tree now lives at: https://github.com/dhiltgen/docker-machine-kvm and downloadable binaries are available at https://github.com/dhiltgen/docker-machine-kvm/releases

Since the plugin framework hasn't merged yet, you'll have to build your own copy of Nathan's work, but after that, download my plugin, put it in your path, then you should be able to say something like:

```bash
docker-machine create --driver kvm --kvm-memory 2048 --kvm-disk-size 10000 --kvm-cpu-count 2 mymachine
docker-machine ls
docker-machine env mymachine
```

cnkcb:
@dhiltgen please update the docker-machine-kvm README with installation steps for the less saavy users.

warmchang:
@dhiltgen , nice job!

PR #700, opened 2015-03-03
EC2 fixes

nathanleclaire:
Fix a couple of issues I identified with the EC2 driver:

- If the instance is removed remotely via a method other than `docker-machine`, there will be a panic.  This will return the instance as a "Error" state instead.
- We were setting `d.IPAddress` to the wrong value, so this corrects that and fixes an issue where the IP would sometimes not be identified correctly


nathanleclaire:
cc @ehazlett @sthulb

sthulb:
LGTM

ehazlett:
LGTM thanks!

Issue #531, opened 2015-02-11
x509: certificate is valid for 192.168.99.103, not 192.168.99.100


stephenlawrence:
After having my machine running on the .103 IP I rebooted my Mac. Upon reboot my docker-machine switched to the .100 address. When I tried to perform any docker commands on my machine I get things like this:

`docker exec -it mycontainer bash`:
FATA[0000] An error occurred trying to connect: Post https://192.168.99.100:2376/v1.16/containers/mycontainer/exec: x509: certificate is valid for 192.168.99.103, not 192.168.99.100

ehazlett:
Can you give some more detail on the issue?

- What version of Machine?
- What driver (I'm assuming vbox based upon the IP)?
- Command line used to create machine

Thanks

stephenlawrence:
docker-machine version 0.1.0
virtualbox
docker-machine  create  -d virtualbox --virtualbox-disk-size '40000' --virtualbox-memory '4096' devbox

I also added an --insecure-registry to the daemon so I could talk to our private registry (if that matters).

ehazlett:
Would you mind testing from master (you can use https://docker-machine-builds.evanhazlett.com/latest/ if you don't want to build locally).  I have tested this with VMs changing IPs and everything works as expected:

```
ehazlett machine> docker-machine ls
NAME        ACTIVE   DRIVER       STATE     URL
test-vbox   *        virtualbox   Running   tcp://192.168.99.100:2376
ehazlett machine> docker-machine stop 
ehazlett machine> docker-machine create -d virtualbox test-vbox-2
...
ehazlett machine> docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL
test-vbox              virtualbox   Stopped   
test-vbox-2   *        virtualbox   Running   tcp://192.168.99.100:2376
ehazlett machine> docker-machine start test-vbox
...                  
ehazlett machine> docker $(docker-machine config test-vbox) ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
ehazlett machine> docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL
test-vbox              virtualbox   Running   tcp://192.168.99.101:2376
test-vbox-2   *        virtualbox   Running   tcp://192.168.99.100:2376

```

You can see that `test-vbox` switched IPs.

stephenlawrence:
Ok, here is what I did:

Stop the current machine
Start using the new machine bin
[...]

Issue #688, opened 2015-03-02
Proposal: Identity management


sthulb:
We should allow for the generate and regeneration of x509 keys for existing and new servers.

---

### Users
We should allow owners of machines to generate new identities to allow third parties to control Docker. The certificates would be based of the same CA root. 

The third party user would be informed of a way of collecting the new client certificates securely.

### Regeneration
We should allow users to regenerate x509 certificates on their machines.

sthulb:
More to come when I have time.

nathanleclaire:
Oh interesting - I was just thinking about this problem recently - how to grant access to remote Docker on a client that is different than the one where the machine was created.

sthulb:
I had the idea a month ago, I talked about it in the IRC channel, I've just got around to proposing it :)

sthulb:
This would have more benefit with a client/server model for machine though.

frapposelli:
Great idea, much better than keeping certs on dropbox :smile: 

Client/Server could be very interesting, maybe a passphrase to give access to a common shared machine?

**Alice**: `docker-machine create`, sets passphrase, share passphrase w/ Bob
**Bob**: `docker-machine attach`, use Alice passphrase to gain access

sthulb:
@frapposelli I'd expect new certs to be generated for every user. A token/passphrase would only be needed for collection. We could revoke "Alice's" certs at any time.

Disclosure: Docker doesn't support cert revocations yet.

frapposelli:
@sthulb of course, passphrase is a way to grant access to a cert generation on the shared machine, ideally the passphrase should be OTP.

sthulb:
Yep!

ehazlett:
+1  I think this will be a must for machine server.

docteurklein:
What is the current solution (even hacky) to share 1 machine configuration?
I saw https://github.com/docker/machine/pull/1476 and it seems great, once landed.

But what are the minimum set of files to move and what parts of config.json should I copy and change in order to have a working access to a docker created with machine ?

Thanks :)

docteurklein:
Aside that, it would be awesome to enlight me on the differences and roles of each folder in ~/.docker/machine.

What is certs/*, compared to machines/*/* ?

nathanleclaire:
> What is the current solution (even hacky) to share 1 machine configuration?

[...]

PR #702, opened 2015-03-04
rebuild-ca and regenerate-certs commands

ehazlett:
This adds `tls-rebuild-ca` and `tls-regenerate-certs` for rebuilding the local CA cert and keys as well as generating new certificates and propagating them to the specified machines.  This should resolve issues like #531 as well as provide a mechanism to completely re-key all machines.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -67,6 +67,21 @@ func (h hostListItemByName) Swap(i, j int) {
 func (h hostListItemByName) Less(i, j int) bool {
 	return strings.ToLower(h[i].Name) < strings.ToLower(h[j].Name)
 }
+func confirmInput(msg string) bool {
+	fmt.Printf("%s (y/n): ", msg)
+	var resp string
+	_, err := fmt.Scanln(&resp)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if strings.Index(strings.ToLower(resp), "y") == 0 {
+		return true
+	}
+
+	return false
+}
 
 func setupCertificates(caCertPath, caKeyPath, clientCertPath, clientKeyPath string) error {
 	org := utils.GetUsername()
@@ -179,6 +194,22 @@ var Commands = []cli.Command{
 		},
 	},
 	{
+		Name:        "env",
+		Usage:       "Display the commands to set up the environment for the Docker client",
+		Description: "Argument is a machine name. Will use the active machine if none is provided.",
+		Action:      cmdEnv,
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "swarm",
+				Usage: "Display the Swarm config instead of the Docker daemon",
+			},
+			cli.BoolFlag{
+				Name:  "unset, u",
+				Usage: "Unset variables instead of setting them",
+			},
+		},
+	},
+	{
 		Name:        "inspect",
 		Usage:       "Inspect information about a machine",
 		Description: "Argument is a machine name. Will use the active machine if none is provided.",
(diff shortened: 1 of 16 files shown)


mrevilme:
+1

sthulb:
Perhaps this can be linked into my PR about certs/identity #688

ehazlett:
@sthulb +1.  Right now, there isn't much to it.  It simply calls the other functions to regenerate.  I think your PR is needed for machine server.  This would temporarily give rotation capabilities and the ability to reset the TLS config on the engine (assuming you have the SSH keys).

sthulb:
Sure, implement them how you want for now, whilst we're < 1.0 we can fiddle with command structure.

ehazlett:
@bfirsh thoughts?  i went back and forth on the naming.  i'm not tied to them.

jeffdm:
:+1:

bfirsh:
A command for this would help people get out of that crappy situation, I suppose. Though I'm a bit concerned about adding this because we're not fixing the root cause of a bug.

Questions:
 - Are there any reasons why you want to rotate all of your keys, beyond fixing that bug and general good security practice?
 - Why would you want to rebuild your CA?
 - I wonder if we need both commands... If you rebuild your CA, don't you always need to regenerate your certs? If you want to rebuild all your certs, is there any harm in also rebuilding the CA?

ehazlett:
Unfortunately, I cannot get this reproduced.  What appears to be happening is the IP SAN isn't valid for the specified cert (even though changing IPs I cannot reproduce).  I don't think there is any other way to fix this unless we set a wildcard for all access.  Perhaps this would be ok for the local drivers but I think the potential would still be there for the cloud providers too.  Take ec2 for example, you could stop and get another IP.  I don't know any other way than regenerating certificates.

General good security practice for start.  If your CA gets compromised, you would want to build a new one.  If you generate a new CA, you don't have to regenerate all certs because we store a copy of the ca and cert in the machine directory to be self-contained.  This allows for a "rolling" cert re-issue.

bfirsh:
I think setting a wildcard hostname and verifying the exact cert is a good idea. I'm pretty sure if you don't need a CA for Docker's TLS. You can pass the host's certificate to `--tlscacert` directly and it'll verify that the host is presenting specifically that certificate.

bfirsh:
In terms of naming, I like commands which start with verbs. So if we're having one command, I would vote `regenerate-certs`. If we're having two commands, it's a bit more complex...

ehazlett:
Ok so for server certs we generate a wildcard?  I think it's less secure but would alleviate the issue and as long as we verify the cert I would be ok with it.

thoughts @sthulb @ggiamarchi @andrewwatson

ehazlett:
i don't think we would want to force a regeneration of the CA as that would cause confusion I think.  I think two commands is better.

bfirsh:
Why would you want to regenerate the CA? If that's something that doesn't happen that often, could we recommend running `rm -r ~/.docker/machine/certs; docker-machine regenerate-certs`?

ehazlett:
I think this could be modified to only generate the CA certs if they don't exist and thus a single command would suffice.

md5:
FYI, I don't think you can "wildcard" with an IP SAN, so you'd have to create a fake hostname/domain name for the Machine-created Docker hosts.

md5:
> I think setting a wildcard hostname and verifying the exact cert is a good idea. I'm pretty sure if you don't need a CA for Docker's TLS. You can pass the host's certificate to --tlscacert directly and it'll verify that the host is presenting exactly that certificate.

@bfirsh Could you clarify what you mean here?

md5:
Will Go not check that the certificate used as a trust root includes the CA extension?

md5:
I'm pretty sure it checks for the CA extension of the asserted trust roots here: https://github.com/golang/go/blob/c2317db2f9bf9f097f0bc297004a8f581b944206/src/crypto/x509/x509.go#L593

sthulb:
I'm ok with wildcard certs if we can generate them.

ehazlett:
it looks like @md5 is correct.  I just tried to create some using wildcards (`*` and `0.0.0.0`) which both do not work.  It looks like we are going to have to support either detection of failed connections or the ability to re-cert as we probably cannot rely or generate hostnames for all instances.

ehazlett:
@bfirsh so we ran into this again in irc.  i think this would be good to get in.  I think removing the CA command and defaulting to relying on removing manually would suffice for now.  thoughts?

bfirsh:
@ehazlett :+1: that sounds great. I think a `regenerate-certs` command could be easier to understand.

Actually, come to think of it, isn't the CA generated automatically on first start if it doesn't exist?

ehazlett:
@bfirsh correct -- if the ca/client certs do not exist, they get auto-generated on first machine create.

ehazlett:
Closing this PR in favor of #770.  It is the exact same (which the name as `regenerate-certs`) only created against the new driver refactor for easier merge.

sporkmonger:
FWIW, I just ran into this after running b2d via Kitematic and then spinning up two other docker-related VMs subsequently. The Kitematic-spawned b2d VM seems to have gotten a new IP in the process somehow and then `certificate is valid for 192.168.99.101, not 192.168.99.100`. Running `docker-machine regenerate-certs default` resolved successfully.

PR #706, opened 2015-03-04
rackspace: enable env var for flavor id

ehazlett:
Enable setting the Rackspace flavor id via environment variable.

### DIFF
--- a/drivers/rackspace/rackspace.go
+++ b/drivers/rackspace/rackspace.go
@@ -76,12 +76,12 @@ func GetCreateFlags() []cli.Flag {
 		cli.StringFlag{
 			Name:  "rackspace-image-id",
 			Usage: "Rackspace image ID. Default: Ubuntu 14.04 LTS (Trusty Tahr) (PVHVM)",
-			Value: "",
 		},
 		cli.StringFlag{
-			Name:  "rackspace-flavor-id",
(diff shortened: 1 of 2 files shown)


ehazlett:
@ggiamarchi would you mind reviewing this?  thanks!

drivers/rackspace/rackspace.go
```diff
@@ -76,12 +76,12 @@ func GetCreateFlags() []cli.Flag {
...
-			Value: "",
 		},
 		cli.StringFlag{
-			Name:  "rackspace-flavor-id",
-			Usage: "Rackspace flavor ID. Default: General Purpose 1GB",
-			Value: "general1-1",
+			Name:   "rackspace-flavor-id",
+			Usage:  "Rackspace flavor ID. Default: General Purpose 1GB",
+			Value:  "general1-1",
+			EnvVar: "RACKSPACE_FLAVOR_ID",
```
ggiamarchi: The others env var names have the OpenStack prefix `OS_`. I think we should use the same rule here (`OS_FLAVOR_ID`) for naming consistency.

ggiamarchi: This env var should be added in the documentation

ehazlett:
@ggiamarchi thanks!  updated!

ggiamarchi:
LGTM

Issue #666, opened 2015-02-27
Machine is trying to provision docker on top of Hyper-V's boot2docker iso


jeffmendoza:
see:
54f36078f4c7516d717987f94c4787c0566798c0

https://github.com/docker/machine/blob/master/host.go#L406

Hyper-V is missing from this line.

nathanleclaire:
Argh, this is why I would really like a `IsLocal` method or equivalent on drivers, it would definitely be more robust.

sthulb:
@nathanleclaire in the long run, we don't need that, I need to finish #546. So we can detect the distro family.

PR #709, opened 2015-03-04
Fix check for whether docker is installed

nathanleclaire:
Should fix https://github.com/docker/machine/issues/666 and is a more robust solution than checking whether the driver uses b2d by default or not (what if user wants to use Ubuntu VM locally?)

cc @ehazlett @sthulb 


ehazlett:
LGTM

PR #714, opened 2015-03-04
add the three new DO flags to the docs

knarz:
I totally forgot to add the flags to the docs, related to #695 

Any input on the wording? Not sure if it is clear that those are just toggles.

Signed-off-by: Sascha Hanse <shanse@gmail.com>


ehazlett:
+1 and I forgot to check for them :). LGTM

PR #715, opened 2015-03-05
Add --n-instances to specify number of instances to create

nathanleclaire:
This is much more of a "let's discuss design, play with it and try it out" than a request for merge, but I'm really excited about this functionality and I think others will be too.

Known issues:

- Sometimes (pretty regularly actually) drivers fail at creation, or at provisioning, and we should be able to handle
- What do we do with active host in this situation?  I've just passed the buck completely in this impl
- Right now if the machine requires shared resources before creation and it checks to see if they exist (e.g. security group in Amazon driver), then creates them if they don't, it will race with all the other goroutines to try and do so.  This == :(  and makes me think a separate `Precreate` step, or some kind of "skip shared resource creation" option, may well be a good thing.

Bonus points: Helps reduce some of the constant passing around of `cli.Context` struct.

I'll follow up later with more commentary but... PTAL

@ehazlett @sthulb 
@bfirsh This is pretty similar to something that was originally on the roadmap, so you might be interested.
@aluzzardi @vieux I think you'll be interested in this too.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -161,6 +161,11 @@ var Commands = []cli.Command{
 				Usage: "addr to advertise for Swarm (default: detect and use the machine IP)",
 				Value: "",
 			},
+			cli.IntFlag{
+				Name:  "n-instances, n",
+				Usage: "Number of instances to create - each instance's number will be appended to the name",
+				Value: 1,
+			},
 		),
 		Name:   "create",
 		Usage:  "Create a machine",
@@ -299,7 +304,17 @@ func cmdActive(c *cli.Context) {
 	}
 }
 
+type SwarmOptions struct {
+	Discovery string
+	Master    bool
+	Host      string
+	Addr      string
+}
+
 func cmdCreate(c *cli.Context) {
+	var (
+		swarmConf *SwarmOptions
+	)
 	driver := c.String("driver")
 	name := c.Args().First()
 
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Oh yeah, screenshots:

screen shot 2015-03-04 at 5 37 41 pm

screen shot 2015-03-04 at 5 38 22 pm

You can see that the swarm of 21 machines only took a little under 6 minutes to create and be fully ready to run containers.  Nice.

sthulb:
@nathanleclaire You should perhaps change the output to show the machine name as the prefix.

nathanleclaire:
@sthulb Oh yes, that's a huge priority and I want it a lot - it can be incredibly confusing to try and track down individual host issues without it.  It's a little tricky to do so because the drivers call `logrus` directly, but I think that eventually I want to (transparently) replace calling logrus directly with our own `log` that does some custom voodoo (e.g. `WithFields` driver names like you mentioned, supporting various logging formats such as JSON, etc.).

thaJeztah:
Cool!

sthulb:
@nathanleclaire Great! It's looking good. 

As for the logging, we will probably need something similar when we go to client/server :)

ehazlett:
I think this is cool.  I want to hold off until we get a grasp around all of the nuances around instance creation as there are some consistency issues among them.  While I think this will be great in the future, I'm afraid that adding concurrency on create will cause more issues around creation in general.  Once we get those ironed out I think this will be great.

bfirsh:
Thanks for this, @nathanleclaire. I have a feeling the way we want to create multiple hosts, though, is with some kind of Swarm interface. E.g.: `docker-machine swarm create -d digitalocean --size 20 myswarm`

As @ehazlett says, until we've figured out that design, I think we should hold off on this.

nathanleclaire:
@bfirsh With this PR that works out of the box (barring issues with instance creation):

Assuming that a master already exists:

```console
$ docker-machine create -d digitalocean -n 20 --swarm --swarm-discovery token://mytoken workerbee
```

(we might drop the `--swarm` as it can be deduced from `--swarm-discovery`)

nathanleclaire:
I'm going to close this and eventually implement a better solution.

bfirsh:
@nathanleclaire It'd be really cool to have a top-level Swarm command, which would also cover this use case...

```
$ docker-machine swarm create myswarm
$ docker-machine swarm scale myswarm=20
```

Boom.

PR #718, opened 2015-03-05
Modify description of StartDocker and StopDocker

cheng--zhang:
Modify RestartDocker to StartDocker and StopDocker in description.

Signed-off-by: Cheng Zhang <cheng.zhang@citrix.com>


ehazlett:
+1 LGTM thanks!

PR #722, opened 2015-03-05
Revert "Use random strategy for Swarms"

vieux:
I find it very confusing that machine doesn't have the same default as swarm.


vieux:
ping @ehazlett @bfirsh

ehazlett:
This makes sense to me.  I think we should match whatever the Swarm default is.  Eventually we will make this configurable so it won't be an issue.

@bfirsh ?

bfirsh:
Hmm, I suppose. The reason I added this in the first place is I think it's a bit surprising that when you first try out Swarm, all of your processes end up on a single node.

I understand that random scheduling also has issues, in that you won't be able to schedule large jobs.

I guess it's better to be consistent... all the better for improving our Swarm support. :)

ehazlett:
LGTM

PR #723, opened 2015-03-05
remove --swarm as it can be deduced from --swarm-discovery

vieux:
Signed-off-by: Victor Vieux <vieux@docker.com>

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -139,10 +139,6 @@ var Commands = []cli.Command{
 				Value: "none",
 			},
 			cli.BoolFlag{
-				Name:  "swarm",
-				Usage: "Configure Machine with Swarm",
-			},
-			cli.BoolFlag{
 				Name:  "swarm-master",
 				Usage: "Configure Machine to be a Swarm master",
 			},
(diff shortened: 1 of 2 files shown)


vieux:
@ehazlett please review careful, I'm afraid I missed something

nathanleclaire:
Nice!  This has been bugging me for a while.

ehazlett:
@vieux https://github.com/vieux/machine/pull/1

Note: with the `--swarm` removal direct access to the engines on a Swarm master via `config` and `env` is no longer possible.  Only the Swarm port will be used with those commands.  This was one of the reasons for `--swarm`.  I think this is OK since we are interacting with a Swarm master and should be the default.

nathanleclaire:
@ehazlett I'm in favor of that (defaulting to talking to swarm), although I still want to enable direct access to those Docker daemons via something like https://github.com/docker/machine/pull/605 .  Do you agree?  I just don't want to leave those machines stranded if something wacky happens with Swarm.

ehazlett:
@nathanleclaire that was the original intent with the `--swarm` flag.  We could keep these changes and add the `--engine` flag for engine access.  I will update it to see what it looks like.

ehazlett:
After working through this there are other things that are tied to `--swarm` (config, env, etc).  I think we should re-think the Swarm integration UX before moving further on this.  I'm in favor of removing `--swarm` but we need to look at everything it affects.

ehazlett:
I'm going to close this for now.  We use the `--swarm` option in a few other commands.  Perhaps when we move to subcommands we can revisit.

Thanks!

PR #728, opened 2015-03-06
use default http client to download iso

shawnzhu:
fix for #625 
The transport of default http client provides:
* proxy support from environment variables (`http_proxy`, `no_proxy` for go 1.3, `https_proxy` for go 1.4)
* 30 seconds http connect timeout
* 30 seconds for keep-alive
* 10 seconds for TLS handshake timeout

### DIFF
--- a/utils/b2d.go
+++ b/utils/b2d.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


andrewwatson:
I'm not sure i understand what this is going to accomplish.  Isn't the default http client used by default already?  I could see making it possible to pass in a different client for platforms like app engine that require it.

shawnzhu:
@andrewwatson we didn't use default transport when downloading boot2docker ISO file but a customized one. we can see the `DefaultClient` works for all the features we need, so I would prefer default option instead of customization.

According to platform specific requirement, `utils/b2d` is not used at all, if we need specific http client, more util functions make sense like those different transport in docker.

shawnzhu:
close since #625 is merged

Issue #699, opened 2015-03-03
Add provisioner so that the Engine and Swarm can be configured on the host


ehazlett:
This will remove the provisioning functionality from the `store` and drivers into a clean package that can be used in conjunction with #693 to provision and configure the Docker engine.  The long term goal should be to get cloudinit working.  However, it should be flexible to allow for fallback.

Refs #554

Provision PR: #546

PR #735, opened 2015-03-06
Add flag --pull-swarm-image

LaynePeng:
Add the option "pull-swarm-image" to let user choose if pull the latest version of Swarm image. Because in some network case, it is hard to access official docker hub.

Signed-off-by: LaynePeng <appamail@hotmail.com>

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -161,6 +161,11 @@ var Commands = []cli.Command{
 				Usage: "addr to advertise for Swarm (default: detect and use the machine IP)",
 				Value: "",
 			},
+			cli.StringFlag{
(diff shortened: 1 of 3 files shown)


sthulb:
Can you rename the flag to swarm-image? I think the pull part isn't required.

ehazlett:
Unfortunately right now we aren't going to merge any options to swarm as we need to discuss the UX integration.  You will be able to configure Swarm in the future via some form of config (file, cloudinint, etc).  Right now we don't want to clutter up the UI with flags.

Thanks for the contribution and please follow #699 for provisioning related operations.

PR #736, opened 2015-03-06
uses --tlsverify instead of --tls in config output

shawnzhu:
Now it outputs `export DOCKER_TLS_VERIFY=yes` when performing command `docker-machine env` but it outputs `--tls` only when performing `docker-machine config`.

Since it requires TLS by default and both client side and server side certs are presented, it should use `--tlsverify` all the time.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -300,14 +300,13 @@ func TestRunActionForeachMachine(t *testing.T) {
 func TestCmdConfig(t *testing.T) {
 	stdout := os.Stdout
 	r, w, _ := os.Pipe()
-	os.Stdout = w
 
+	os.Stdout = w
 	os.Setenv("MACHINE_STORAGE_PATH", TestStoreDir)
 
 	defer func() {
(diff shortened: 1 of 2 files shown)


commands_test.go
```diff
@@ -335,15 +329,29 @@ func TestCmdConfig(t *testing.T) {
...
-	c := cli.NewContext(nil, set, set)
+	go func() {
+		defer close(outStr)
+		var testOutput bytes.Buffer
+		io.Copy(&testOutput, r)
+		outStr <- testOutput.String()
+	}()
 
-	cmdConfig(c)
+	os.Setenv("MACHINE_STORAGE_PATH", "")
```
ehazlett: This cleanup won't happen if something above fails.

shawnzhu: Thanks! I changed back to `defer` to do the "undo" work

ehazlett:
LGTM

ehazlett:
thanks!

PR #738, opened 2015-03-06
consistent naming for DOCKER_TLS_VERIFY env var

### LINKED ISSUE #734: Be consistent with docker in DOCKER_TLS_VERIFY usage
docker-machine env output shows 

    export DOCKER_TLS_VERIFY=yes

although that might be working, that's not the value generally used in docker docs in general (`1` is the one in the docker docs) and when I first saw it I was confused. In fact this usage appears only in machine-related repos/discussion according to google. 

We should probably speak the same language with docker/docker repo. Treat as low-priority UX feedback.

ehazlett:
This sets the env var `DOCKER_TLS_VERIFY` to "1" instead of "yes" to be consistent with Docker.

Fixes #734

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -523,10 +523,10 @@ func cmdEnv(c *cli.Context) {
 
 	switch userShell {
 	case "fish":
-		fmt.Printf("set -x DOCKER_TLS_VERIFY yes;\nset -x DOCKER_CERT_PATH %s;\nset -x DOCKER_HOST %s;\n",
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
LGTM

ehazlett:
thanks!

ahmetalpbalkan:
@ehazlett prolly need to fix the sample output in the docs as well?

ehazlett:
thx @ahmetalpbalkan

PR #739, opened 2015-03-06
--swarm-strategy option added. Defaults to swarm's default: binpacking

remysaissy:


ehazlett:
Hey, thanks for the contribution.  Unfortunately, at this time we aren't adding more swarm options as command flags.  We are still working through the UX.  This will probably be in a configuration set of some sort (cloudinit, etc).  As for the swarm default strategy, this was reverted and is now binpacking by default (https://github.com/docker/machine/pull/722).  Thanks!

remysaissy:
Ok, I'm eager to see this released!

ehazlett:
@remysaissy thanks!  so am i ;)  i want to be able to configure my swarms too :)

PR #741, opened 2015-03-06
Begin separation of creation and provisioning

nathanleclaire:
- Implement methods to get SSH key path, port, username, and hostname
- Move hostname provisioning out of drivers and into main code path
- implement generic wait for SSH function like in @ggiamarchi PR https://github.com/docker/machine/pull/665

cc @ehazlett @sthulb 

### DIFF
--- a/host.go
+++ b/host.go
@@ -387,13 +387,17 @@ func (h *Host) Create(name string) error {
 		return err
 	}
 
-	// install docker
-	if err := h.Provision(); err != nil {
-		return err
-	}
+	log.Debug("Done with driver create")
 
 	// save to store
 	if err := h.SaveConfig(); err != nil {
+		return err
+	}
+
+	log.Debug("Time for provisioning")
+
+	// install docker
+	if err := h.Provision(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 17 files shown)


nathanleclaire:
Oh yes, and also I moved `SaveConfig` to be _before_ provisioning, since provisioning is often the step where things go awry.

Next step is to remove IP polling and waiting for instance startup out of drivers' `Create` method.

host.go
```diff
@@ -499,3 +522,38 @@ func (h *Host) SaveConfig() error {
 	}
 	return nil
 }
+
+func (h *Host) SetInstanceHostnameUbuntu() error {
+	log.Debugf("Setting hostname for: %s", h.Name)
+	cmd, err := h.Driver.GetSSHCommand(fmt.Sprintf(
```
sthulb: We could merge the two host name functions and supply a path to the hostname file.

sthulb:
I agree we shouldn't do SSH commands from inside the driver. 

Can you update the driver doc to mention SSH connection info must be exported?

host.go
nathanleclaire: I considered this but they are slightly different in other ways (mostly the `/etc/hosts` line) - is it okay to use the command from `SetInstanceHostnameUbuntu` in both styles?

sthulb: The B2D host won't mind.

ehazlett:
I like a lot of these ideas just not in the patch bomb format.  I've implemented some of the SSH operations in #756.  I think we can take bits of this and open smaller PRs against the issues listed in the repo (i.e. #744, #699) - especially the SaveConfig stuff - that will help tremendously.

Thanks!

Issue #744, opened 2015-03-06
Remove SSH operations from drivers


ehazlett:
The drivers should not be responsible for SSH operations.  This proposes to have functions in the driver to get the SSH configuration and a separate package for the operation.

Refs #554

nathanleclaire:
So, after https://github.com/docker/machine/pull/756, the next step probably will be to move the logic for waiting for the IP address into the `GetIP` method in the drivers - this will allow us to have just one catch-all "wait for ssh" check in what is currently `host.go`, and consequently the drivers will get to save their config a lot sooner, which is a big step forward because it narrows the failure window a lot and would allow us to retry provisioning if it blows up or hangs.

ehazlett:
ugh waffle.

ehazlett:
@nathanleclaire is this complete?  i believe we have the single wait for ssh don't we?

nathanleclaire:
So, we have the single wait for SSH, but there are still some situations where the drivers do it themselves like https://github.com/docker/machine/blob/cf136eacd70d5e510331de3d4c2fe50de46e06d3/drivers/azure/azure.go#L266 .  So we probably should make a "hitlist" of which drivers do it on their own still and which need to be removed.  It isn't _super terrible_ but if the drivers wait for SSH on their own it lengthens the wait to a config save and consequently the window of failure.

Also want to make sure we file a separate issue for that comment above- after we pull getting IP address out of the drivers, they'll all be in good shape in terms of saving config in a timely fashion, although we still need to figure out how we want to handle interrupts and things of that nature.

nathanleclaire:
I'll put together a checkbox list of which drivers fulfill this requirement here

nathanleclaire:
- [ ] amazonec2/ - needs removal
- [ ] azure/ - needs removal
- [ ] digitalocean/ - needs removal
- [x] fakedriver/ - 
- [ ] google/ - needs removal
- [ ] hyperv/ - needs removal
- [x] none/
- [ ] openstack/ - needs removal
- [ ] rackspace/ - need removal (depends on openstack I think)
- [ ] softlayer/ - needs removal
- [x] virtualbox/
- [x] vmwarefusion/
- [ ] vmwarevcloudair/ - needs removal
- [x] vmwarevsphere/

nathanleclaire:
Alright the above list has been updated to the best of my ability to indicate which drivers still wait for SSH vs. which drivers don't.

Checked means they don't, so they are good to go, unchecked means they do not.

[...]

PR #756, opened 2015-03-09
Driver interface refactor

### LINKED ISSUE #743: Define Driver Standard Interface
This will define the standard driver interface to be used in all drivers.   Most of it is already defined but this will finish it so we can merge the new drivers.

Refs #554

ehazlett:
This adds a few helpers for SSH so we can remove the ssh operations from provisioning.  It also removes the Docker specific operations like starting / stopping the daemon and `GetSSHCommand` which will be refactored in #744.

Closes #743


drivers/drivers.go
```diff
@@ -14,62 +13,57 @@ import (
...
 
-	// SetConfigFromFlags configures the driver with the object that was returned
-	// by RegisterCreateFlags
-	SetConfigFromFlags(flags DriverOptions) error
+	// GetIP returns an IP or hostname that this host is available at
+	// e.g. 1.2.3.4 or docker-host-d60b70a14d3a.cloudapp.net
+	GetIP() (string, error)
+
+	// GetSSHHostname returns hostname for use with ssh
+	GetSSHHostname() (string, error)
```
sthulb: I'd convert this to `GetSSHAddr(ess)` it's rare we actually use hostnames, and addr is generic.

ehazlett: Good point.  Thanks!

drivers/drivers.go
```diff
@@ -3,73 +3,78 @@ package drivers
...
 	DriverName() string
 
-	// SetConfigFromFlags configures the driver with the object that was returned
-	// by RegisterCreateFlags
-	SetConfigFromFlags(flags DriverOptions) error
+	// GetIP returns an IP or hostname that this host is available at
+	// e.g. 1.2.3.4 or docker-host-d60b70a14d3a.cloudapp.net
+	GetIP() (string, error)
+
+	// GetSSHHostname returns hostname for use with ssh
```
nathanleclaire: The comment says "GetSSHHostname", but it is named `GetSSHAddress` here

nathanleclaire:
It would probably be faster to split my existing PR () out into a few different pieces, one of which would be this piece, the new SSH functions.  I'd happily split it into a few different PRs if you want.

ehazlett:
@nathanleclaire yeah that sounds great -- especially the ssh and waitFunc stuff.  I'm mainly focusing on the driver specific implementation to match this updated interface.  thanks!

drivers/drivers.go
ehazlett: Oops - good catch.  Thanks!

nathanleclaire:
OK - I will put in PRs for those (maybe pulling out the hostname provisioning as well)

ehazlett:
@nathanleclaire i have the hostname stuff coming -- can you comment on it first?  i want to see what you think.

ehazlett:
@nathanleclaire fyi using `SaveConfig` after provisioning is making working with ec2 much much better :)  :+1:

nathanleclaire:
Nice.

Mostly looks pretty good, a few things to consider:

- I'd favor `GetSSHHostname` over `GetSSHAddress` - part of the reason this method needs to be implemented is to get a reliable hostname to run SSH commands against, one of which is `localhost` (in b2d/vbox's case)
- `HypervisorType` strikes me as a little misleading.  Types of hypervisors are ESXI, Xen, KVM, etc.  But at the same time I'm struggling to come up with something better, so it might be okay.

ehazlett:
@nathanleclaire hmm ok -- we do return addresses for most of those.  is there a reason we return "localhost" and not "127.0.0.1" or something?

I agree with HypervisorType -- I was in the same boat -- not sure what to call it.  I had a bool just returning IsLocal but that seemed a bit restrictive.

ehazlett:
Maybe `ProviderType` with `Local`, `Remote`?

nathanleclaire:
> hmm ok -- we do return addresses for most of those. is there a reason we return "localhost" and not "127.0.0.1" or something?

Hmmmm either way will probably work but doing it by hostname was @SvenDowideit's suggestion, so perhaps he can comment on it.

`ProviderType` sounds okay - although thinking about it, would you be opposed to adding a basic `Provisioner` interface, doing detection of the underlying OS (only Ubuntu and b2d implementations supported for now) and then using a variation of that interface `Boot2DockerProvisioner` or `UbuntuProvisioner`?  This would replace the `switch` statements with just one method call, and IMO get us closer to our eventual goal of being able to pull provisioning out of the drivers completely instead of asking them to self-report how they should be provisioned.

ehazlett:
@nathanleclaire yeah the idea was to get a Provisioner (#699).  I'm not opposed to that.  You want to do a PR with that?

sthulb:
I can finish my OS detection PR if it'll make #699 easier.

ehazlett:
@sthulb yeah that sounds great. thanks!

ehazlett:
@nathanleclaire for now I will change `HypervisorType` to `ProviderType` until we address with the provisioner.  I think we will still probably need to have the Driver inform what type of provider they are, no?

nathanleclaire:
@ehazlett It probably can't hurt to have the methods available so that SGTM.  Maybe I can coordinate with @sthulb to shape up the `Provisioner` stuff we're discussing too.

ehazlett:
@nathanleclaire perfect.  sounds great.  thanks!

sthulb:
@nathanleclaire catchup tomorrow?

ehazlett:
PTAL @sthulb @nathanleclaire

nathanleclaire:
Looking pretty good... I'll pull it down locally and run through some testing with it!

drivers/azure/azure.go
```diff
@@ -51,7 +52,7 @@ func init() {
 	})
 }
 
-// GetCreateFlags registers the flags this driver adds to
+// GetCreateFlags registers the flags this d adds to
```
ahmetalpbalkan: this doesn't seem like fixed?

ehazlett: Yeah I need to merge in master.  Doing that now.  Thanks!

nathanleclaire:
I thought I commented but so far in my testing I've tried DigitalOcean, VirtualBox, and Amazon EC2 and they all work great.  This is looking really good from my end.

@ehazlett Is your plan for dealing with breakage of older machines that lack the expected SSH fields still the same (default to 22 etc.)?  And do you plan to implement it in this PR, or in a follow-up?

sthulb:
Arse. My comment from yesterday didn't go through apparently. 

I've tried VMware + EC2, looking good so far :+1:

ehazlett:
@sthulb thx!

ehazlett:
@nathanleclaire i want to address the ssh port/user issue in this one too.  working on it today.  thx!

ehazlett:
@nathanleclaire ok so i have the `SSHUser` and `SSHPort` issues fixed.  They are not ideal as I updated each driver, however, I think it's the safest.  Each driver can specify their own usernames / ports and I think it's up to them to make sure they are correct.  I started with a central function in host that would check if missing but it seemed dirty having the logic there.  I think this puts the responsibility where it should be and resolves the issue.

What this also brought back up in my head was the need to version the configs but that can come in a separate PR :)

nathanleclaire:
Yeah, let's not get too ahead of ourselves.  I really want to get this PR in and start banging on it in combination with some of the other stuff we have in the works.

LGTM

sthulb:
Small PRs = Good :)

LGTM

ehazlett:
Cool thx!

PR #758, opened 2015-03-09
Implement generic function to wait for SSH to become available

nathanleclaire:
This depends on : https://github.com/docker/machine/pull/756

It implements a `WaitForSSH` utility function to have "one wait for SSH to rule them all" before provisioning takes place :ring: 

Right now most drivers take care of it themselves and it is a possible point of failure

cc @ehazlett @sthulb


nathanleclaire:
had some rebase issues so opening a new PR

PR #760, opened 2015-03-09
softlayer: fix provisioning bug where curl is not available

ehazlett:
This fixes a provisioning bug where `curl` is not available.


nathanleclaire:
This LGTM

Issue #727, opened 2015-03-06
Idea: call the thing that Machine creates "hosts"


bfirsh:
"Use Machine to create a machine."

[This is kind of my fault, sorry.](https://github.com/docker/machine/pull/84) In documentation and so on, I find it very confusing trying to explain this sort of thing. The only distinction we make between Machine and a machine is whether it's a proper noun or not.

How about we call the thing that Machine creates a "host"? That's a word we've been using for ages with Docker, and we often use that word in Machine-land too.

"Use Machine to create a host."

bfirsh:
(I'm calling this an "idea" because I don't have any strong opinions... something less forceful than a proposal. ;)

hairyhenderson:
Would this be just a documentation change, or would also change directory name(s) and source code?

Personally I prefer the term "host" too

ehazlett:
I personally like "instance" but I'm not very strong on it either.

sthulb:
Host is fine.

nathanleclaire:
Heh.  I'm -1 on changing on terminology inside the code or in our general approach to explaining things at this point (more important things to do IMO), but an FAQ / glossary in the docs (as well as a discussion in a style guide if we evolve one) would probably go a long way.

i.e.:

### Terms

**Docker Machine** - The open source project and binary which is used to create and control new compute instances which run Docker.
**Machine** - The created instance which is running Docker.  This could be a local virtual machine on a laptop or a virtual machine running in the cloud.  This is sometimes referred to as a "host", an "instance", or a "daemon" (referencing the Docker daemon which is running on them).
**Provisioning** - The step in creating a machine which installs Docker, properly configures the machine so that the Docker client can connect to it, and bootstraps the specification outlined in the provided `cloud-init` file

etc.

ehazlett:
Cool.  Hosts it is.  @bfirsh do you want to open smaller issues with where this needs updated?  I imagine at least the docs and readme but perhaps more.

thaJeztah:
Perhaps late to the party (catching up); when used in combination with Swarm, a `host` created by Machine will be called a `node` in swarm terminologies; will that be confusing for users? Does it need some explanation what the difference between a "node" and a "host" is?

bfirsh:
I am purely talking about the terminology we use to communicate through docs and the UI. That said, it probably makes sense for us to use the same terminology in the code too. (Which conveniently happens to be "host". ;)

[...]

PR #764, opened 2015-03-10
Wait for state before return on start, stop, restart, kill

### LINKED ISSUE #757: docker-machine should wait for proper state to be fulfilled before returning from stop, start, etc. commands
When we do a command like `docker-machine stop`, not all drivers wait for the machine to be down before returning.  Some fire off an API request and immediately return.  We should wait for the machine to be in the desired state before returning from the command.

This is needed for:

- start
- stop
- restart
- kill

nathanleclaire:
Fixes #757 

cc @ehazlett @sthulb 

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -574,10 +574,10 @@ func cmdSsh(c *cli.Context) {
 // We run commands concurrently and communicate back an error if there was one.
 func machineCommand(actionName string, machine *Host, errorChan chan<- error) {
 	commands := map[string](func() error){
-		"start":   machine.Driver.Start,
-		"stop":    machine.Driver.Stop,
-		"restart": machine.Driver.Restart,
-		"kill":    machine.Driver.Kill,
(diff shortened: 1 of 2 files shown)


sthulb:
Should we start refactoring things out to match @bfirsh's issue #727? So instead of calling things machines, we call them hosts?

sthulb:
I like it.

ehazlett:
@sthulb i believe most of the internals are actually `host` but we should probably line them up.  I don't think @nathanleclaire has to do that for this -- those actually came from me I believe ;)

sthulb:
That's fair enough, I just thought we should start at some point :)

ehazlett:
@nathanleclaire this needs a rebase.  i would like to get this in -- helps tremendously especially with integration tests :)

ehazlett:
LGTM

Issue #553, opened 2015-02-18
Create an internal API so Machine can be used as a library


ehazlett:
This will enable machine to be used as a lib and lay the groundwork for machine server.

This will provide a more manageable way to interact with machine instead of working directly with the store and host individually.  For example, right now we use `store.Create` which calls `host.Driver.Create` etc.  This will split up those operations so that the store does not handle provisioning, simply storing machine configuration data.

- [x] Utilize new Driver architecture (#554)
- [x] Move store into package
- [x] Provide abstraction around common operations (see https://github.com/docker/machine/pull/550/files#diff-604f4192deea5adf9c159e5f88f64630)
- [x] #767 

Refs: #550

nathanleclaire:
Before we jump headlong into implementing a server, perhaps a good intermediate step would be to create (and dogfood) some good Go bindings / library?  @ibuildthecloud expressed an interest in that for their use case and I think it would be good to get people going / using it right away until the API fully takes shape.

ehazlett:
That is what this is.

ehazlett:
Sorry I was responding to your email which had a different comment.

Yes, this is the placeholder for an internal, non-http API to be used as a lib.

sthulb:
- [ ] Share flags
- [ ] Driver structure
- [ ] Metadata
- [ ] Error handling
- [ ] Logging

ehazlett:
@sthulb added your checklist to the description -- thx!

nathanleclaire:
Yep, I realized that after my first comment so I updated it to just say we should loop potential consumers of the API / bindings in on it.

ehazlett:
Updated scope to be more manageable for this first pass.  We can iterate after feedback.

ehazlett:
This is pending #699 as they both affect the backend provisioning.

PR #767, opened 2015-03-11
Create internal library to use instead of directly using store and host

### LINKED ISSUE #708: --storage-path isn't used anymore
In the changes to the machine internal directory, `--storage-path` lost its ability to do anything.  `GetBaseDir` in `utils/utils.go` still checks `MACHINE_STORAGE_PATH`, but not the flag.

ehazlett:
This depends on #756.

This provides an internal API via the `libmachine` package.  All operations have been refactored to use this internal API (cli, commands, etc).  Store has also been moved into this package.

This also fixes #708.

Refs #553


ehazlett:
PTAL @sthulb @nathanleclaire @cjellick

main.go
```diff
@@ -22,6 +22,10 @@ func main() {
 	app.Name = path.Base(os.Args[0])
 	app.Author = "Docker Machine Contributors"
 	app.Email = "https://github.com/docker/machine"
+	app.Before = func(c *cli.Context) error {
```
sthulb: Not specifically related to this PR, but if this actually works (this being app.Before) without screwing up, we should move the debug flag to here.

nathanleclaire:
so @ehazlett I think this is the next step in terms of merging stuff for the refactor?

nathanleclaire:
Some initial thoughts:

- Why does `Machine` struct contain an instance of `Store`?  I would expect a `Store` to contain machines, not the other way around.
- I really like making `Store` an interface - glad that's a part of this PR.
- Wouldn't it be preferable that `libmachine` know nothing about the flags / flag parsing at all?  Ultimately, I had envisioned clients using `libmachine` something like this:

```go
// place to keep all of your machines
store, err := libmachine.NewFileStore("./storeDir/machines", ca, privateKey)
if err != nil {
    panic(err)
}
host, err := store.NewHost(&libmachine.HostCreateOptions{
    Driver: "virtualbox",
    Memory: 1024,
    DiskSize: 20000,
})
// now we can do some stuff with the host...
```

Isn't having flag parsing logic inside of the `libmachine` package coupling those together too tightly?

main.go
ehazlett: I know -- I'm hesitant to use it as well but this has to be set before the command in order to fix the weird bug we have with the storage path setting.  It works here.

ehazlett:
@nathanleclaire yes, this is the next step.  it's due for milestone 0.2.0.

@nathanleclaire thanks and +1.  i did not want to make pluggable stores part of this PR.

I agree we shouldn't handle the flag parsing.  I will work on that.

I like your intended usage -- we were close.  The problem with the current way (which I will take a look at) is that `store` and `host` are very much coupled.  Perhaps that's ok but it seemed odd to me to have the `store` do operations like "Create".  

In your example usage, you are still using "store" to create the actual host.  My intention was to abstract away working directly with the `store` and `host`.

ehazlett:
@nathanleclaire i was trying to keep `libmachine` to just handle calling to `host` and `store` separately (in the future being able to pass any store to libmachine) and then having libmachine handle call the store for setting info and host for host operations.

ehazlett:
@nathanleclaire i think i misread your comments perhaps.  I am trying something.  I like the generic HostCreateOptions.

ehazlett:
@nathanleclaire I think it will come down to this:  I see benefits for each:

Your recommendation:

```
// place to keep all of your machines
store, err := libmachine.NewFileStore("./storeDir/machines", ca, privateKey)
if err != nil {
    panic(err)
}
host, err := store.NewHost(&libmachine.HostCreateOptions{
    Driver: "virtualbox",
    Memory: 1024,
    DiskSize: 20000,
    SwarmOptions: &libmachine.SwarmConfig{
        Discovery: "token://6e0262602a9307376e3721d77b0d2453",
        IsMaster: true,
    },
})
if err != nil {
    panic(err)
}
if err := host.Create(); err != nil {
    panic(err)
}
// now we can do some stuff with the host...
```

This PR:

```
// place to keep all of your machines
store, err := libmachine.NewFileStore("./storeDir/machines", ca, privateKey)
if err != nil {
    panic(err)
}

mcn, err := libmachine.New(store)

if err := mcn.Create(&libmachine.HostCreateOptions{
    Driver: "virtualbox",
    Memory: 1024,
    DiskSize: 20000,
    SwarmOptions: &libmachine.SwarmConfig{
        Discovery: "token://6e0262602a9307376e3721d77b0d2453",
        IsMaster: true,
    },
}); err != nil {
    panic(err)
}

// now we interact through libmachine.Machine
if err := mcn.Stop("foo"); err != nil {
    panic(err)
}

if err := mcn.Restart("bar"); err != nil {
    panic(err)
}

if err := mcn.Upgrade("baz"); err != nil {
    panic(err)
}
```

I think the second way works better for the future Machine Server since there will be less boilerplate:

For example, instead of:
```
mcn.Stop("foo")
```

It would be this with the first method:

```
host, err := store.Get("foo")
if err := host.Stop(); err != nil {
    panic(err)
}
```

Like I said, I think there are benefits with both.  There is also no reason that we couldn't expose the `store` in `Machine` in case you wanted to get direct access to the store (although I think that kind of breaks and you should just explicitly create/use a `store' in those cases).

Thoughts?

@nathanleclaire @sthulb

nathanleclaire:
Why separate `libmachine` and `mcn`?  All of the things which get called on `mcn` object, I would expect `libmachine` to provide, and then `libmachine` would have some kind of `SetStore` method or equivalent if you don't want to use the default fs store at `$HOME/.docker`.  When I see `mcn`, I expect it to be an actual instance of one machine.

I'll take a look at the changes more closely today, thanks for staying right on top of this.

ehazlett:
Ok I think we are getting things mixed up.  So you are saying we should have something like `libmachine.Start` and `libmachine.Remove` ?  So then it would implicitly use the default store?  I still don't know how we would get around the boilerplate.  The `mcn` instance simply wraps the multiple calls to `store` and `host`.  For example, if you want to check `Active` and then `Remove` it does it as one instead of multiple calls to each (https://github.com/docker/machine/pull/767/files#diff-aed4f4134b47de343fff75c46f6d9dacR114).

`libmachine.Machine` is not a "machine" -- it's the internal API helper.   I thought we settled that the object was going to be called "host" in which case nothing changes here.  Host is still available but instead of having to know about `store` and `host` individually from a lib user perspective, it's simply provided via the internal api.  Therefore, if one is creating machines, they don't also have to remember to set active on the store, etc.

sthulb:
I think I prefer the PR implementation. @ehazlett is right, having to get the host first and then performing an action isn't ideal.

Issue #771, opened 2015-03-12
subcommands start, stop, restart, kill, upgrade will crash while no active machine


TrumanLing:
```console
$ sudo ./docker-machine_linux-amd64 ls
NAME   ACTIVE   DRIVER       STATE     URL                         SWARM
dev1            virtualbox   Stopped
dev2            virtualbox   Stopped
dev3            virtualbox   Running   tcp://192.168.99.100:2376
```

Detailed error messages are as following:

```console
$ sudo ./docker-machine_linux-amd64 kill
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x4072f9]

goroutine 16 [running]:
runtime.panic(0xa4ad00, 0x10358f3)
        /usr/src/go/src/pkg/runtime/panic.c:279 +0xf5
main.runActionForeachMachine(0xae4210, 0x4, 0xc208089548, 0x1, 0x1)
        /go/src/github.com/docker/machine/commands.go:606 +0x599
main.runActionWithContext(0xae4210, 0x4, 0xc208056600, 0x0, 0x0)
        /go/src/github.com/docker/machine/commands.go:656 +0x2d4
main.cmdKill(0xc208056600)
        /go/src/github.com/docker/machine/commands.go:680 +0x3e
github.com/codegangsta/cli.Command.Run(0xae4210, 0x4, 0x0, 0x0, 0xaf85f0, 0xe, 0xbde1b0, 0x5b, 0x0, 0x0, ...)
        /go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:113 +0xeb2
github.com/codegangsta/cli.(*App).Run(0xc20802a410, 0xc20800e000, 0x2, 0x2, 0x0, 0x0)
        /go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:156 +0xaea
main.main()
        /go/src/github.com/docker/machine/main.go:67 +0x7cd

goroutine 17 [runnable]:
runtime.MHeap_Scavenger()
        /usr/src/go/src/pkg/runtime/mheap.c:507
runtime.goexit()
        /usr/src/go/src/pkg/runtime/proc.c:1445

goroutine 18 [runnable]:
bgsweep()
        /usr/src/go/src/pkg/runtime/mgc0.c:1976
runtime.goexit()
        /usr/src/go/src/pkg/runtime/proc.c:1445

[...]

PR #772, opened 2015-03-12
function GetActive should return real error when file doesn't exist

TrumanLing:
fix issue #771

### DIFF
--- a/host_test.go
+++ b/host_test.go
@@ -161,12 +161,16 @@ func TestMachinePort(t *testing.T) {
 	}
 	flags := getTestDriverFlags()
 
-	_, err = store.Create(hostTestName, hostTestDriverName, flags)
+	host, err := store.Create(hostTestName, hostTestDriverName, flags)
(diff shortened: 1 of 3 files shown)


TrumanLing:
@ehazlett and @bfirsh ,as for issue #771, I made this PR, please help review, thanks!

ehazlett:
@TrumanLing i'll review this today.  sorry for the delay!

ehazlett:
+1 for fixing the #771 issue.  however, I'm not sure on the error messages.  i think it's confusing to the user to just respond with "cannot find file".  perhaps instead we catch that specific error and use a custom error saying "no active host specified"?  if it's any other error, we could just return it?

TrumanLing:
@ehazlett done,please review again.

ehazlett:
LGTM

ehazlett:
This has caused a regression.  The `rm` command is no longer to remove machines when there is no active host.  Reverting commit.

PR #777, opened 2015-03-13
Use DOCKER_HOST for active host instead of file

nathanleclaire:
This method of tracking the "active" host for machine simply checks which machine is set in `DOCKER_HOST`.

If `DOCKER_HOST` is set to point to an address which is registered with `docker-machine`, the `*` will appear next to the respective machine.

This means that:

screen shot 2015-03-12 at 5 07 03 pm

_It Just Works_

Interestingly, this implementation prevents the kinds of issues where users attempt to run commands against the active host when there isn't one, since it will simply error out if a user attempts a command of this style instead of trying to go forward (and probably panicing).

## Concerns

- `docker-machine stop` (without arguments, thereby indicating the active host) works out of the box with this method, because it recognizes the active host as the one to stop, but `docker-machine start` does not work with some drivers such as VBox which expect the machine to be up in order to retrieve the IP address.  It forces specificity in this case.
- This causes an additional call to `GetURL` per-goroutine in `ls` (since each goroutine calls to check if it is the active host), which may have (probably minor) performance implications for drivers which actually perform remote operations to get the URL such as VBox.  Most, AFAICT, just grab the "cached" value if it is around (they shouldn't, really; it might change- but that's for another PR).
- I couldn't really figure out how to integration test this (perhaps my own shortcoming, but I am not sure if `bats` messes with environment variables in unexpected ways), and may have missed removing some of those tests, so PTAL.

props to @tiborvass for the idea

cc @ehazlett @bfirsh @sthulb

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -164,7 +164,7 @@ func setupCertificates(caCertPath, caKeyPath, clientCertPath, clientKeyPath stri
 var Commands = []cli.Command{
 	{
 		Name:   "active",
-		Usage:  "Get or set the active machine",
+		Usage:  "Print which machine is active",
 		Action: cmdActive,
 	},
 	{
@@ -326,8 +326,6 @@ var Commands = []cli.Command{
 }
 
 func cmdActive(c *cli.Context) {
-	name := c.Args().First()
-
 	certInfo := getCertPathInfo(c)
 	defaultStore, err := getDefaultStore(
 		c.GlobalString("storage-path"),
@@ -343,25 +341,18 @@ func cmdActive(c *cli.Context) {
 		log.Fatal(err)
 	}
 
-	if name == "" {
-		host, err := mcn.GetActive()
-		if err != nil {
-			log.Fatalf("error getting active host: %v", err)
-		}
-		if host != nil {
-			fmt.Println(host.Name)
-		}
-	} else if name != "" {
-		host, err := mcn.Get(name)
-		if err != nil {
-			log.Fatalf("error loading host: %v", err)
-		}
+	if len(c.Args()) > 0 {
+		log.Fatal("Too many arguments given.")
+	}
 
-		if err := mcn.SetActive(host); err != nil {
-			log.Fatalf("error setting active host: %v", err)
-		}
+	host, err := mcn.GetActive()
+	if err != nil {
+		log.Fatalf("error getting active host: %v", err)
+	}
+	if host != nil {
+		fmt.Println(host.Name)
 	} else {
-		cli.ShowCommandHelp(c, "active")
+		log.Fatal("No active host is set.")
 	}
 }
 
@@ -417,14 +408,10 @@ func cmdCreate(c *cli.Context) {
 		},
 	}
 
-	host, err := mcn.Create(name, driver, hostOptions, c)
+	_, err = mcn.Create(name, driver, hostOptions, c)
 	if err != nil {
 		log.Errorf("Error creating machine: %s", err)
-		log.Warn("You will want to check the provider to make sure the machine and associated resources were properly removed.")
-		log.Fatal("Error creating machine")
-	}
-	if err := mcn.SetActive(host); err != nil {
-		log.Fatalf("error setting active host: %v", err)
+		log.Fatal("You will want to check the provider to make sure the machine and associated resources were properly removed.")
 	}
 
 	info := ""
@@ -437,7 +424,7 @@ func cmdCreate(c *cli.Context) {
 		info = fmt.Sprintf(`eval "$(%s env %s)"`, c.App.Name, name)
 	}
 
-	log.Infof("%q has been created and is now the active machine.", name)
+	log.Infof("%q has been created.", name)
 
 	if info != "" {
 		log.Infof("To point your Docker client at it, run this in your shell: %s", info)
@@ -1069,11 +1056,8 @@ func getHostState(host libmachine.Host, store libmachine.Store, hostListItems ch
 		}
 	}
 
-	isActive, err := store.IsActive(&host)
-	if err != nil {
-		log.Debugf("error determining whether host %q is active: %s",
-			host.Name, err)
-	}
+	dockerHost := os.Getenv("DOCKER_HOST")
+	isActive := dockerHost == url
 
 	hostListItems <- hostListItem{
 		Name:         host.Name,
--- a/commands_test.go
+++ b/commands_test.go
@@ -350,9 +350,11 @@ func TestCmdConfig(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if err := store.SetActive(host); err != nil {
-		t.Fatalf("error setting active host: %v", err)
+	url, err := host.Driver.GetURL()
+	if err != nil {
+		t.Fatal(err)
 	}
+	os.Setenv("DOCKER_HOST", url)
 
 	outStr := make(chan string)
 
@@ -452,9 +454,11 @@ func TestCmdEnvBash(t *testing.T) {
 		t.Fatalf("error loading host: %v", err)
 	}
 
-	if err := mcn.SetActive(host); err != nil {
-		t.Fatalf("error setting active host: %v", err)
+	url, err := host.Driver.GetURL()
+	if err != nil {
+		t.Fatal(err)
 	}
+	os.Setenv("DOCKER_HOST", url)
 
 	outStr := make(chan string)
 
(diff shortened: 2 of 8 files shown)


nathanleclaire:
travis NOOOOOOOO

ehazlett:
:+1: 

> ...but docker-machine start does not work with some drivers such as VBox which expect the machine to be up in order to retrieve the IP address.

Why would `Start` expect the machine the be up?  That sounds like a bug.

Perhaps we force at least a machine name to be specified instead of implicitly selecting the "active"?  Especially if this breaks some of the command usage.  (fwiw I don't like leaving off a machine name for any command).

bfirsh:
Neat. "Active" is a new word invented by us. This makes it mean something else, so I wonder if there's an existing word we can use. "Configured"... "the one in the environment variables"... err...

nathanleclaire:
> Why would Start expect the machine to be up? That sounds like a bug.

It doesn't, but in order to "deduce" the active host with this method you have to know if the IP matches, which isn't possible to check with VBox if the machine is stopped.  So, the short form of "stop" does not work like before.

> Perhaps we force at least a machine name to be specified instead of implicitly selecting the "active"? Especially if this breaks some of the command usage. (fwiw I don't like leaving off a machine name for any command).

Yes, I would much rather migrate to this method of doing things in general, and just treat active as "this is where Docker commands will be run against", nothing more.

> Neat. "Active" is a new word invented by us. This makes it mean something else, so I wonder if there's an existing word we can use. "Configured"... "the one in the environment variables"... err...

Heh.  I don't really mind keeping it as active (if we invented it, we can change the meaning).

sthulb:
+1 I like this more than a file.

ehazlett:
@nathanleclaire so i guess my question is how does this negatively impact the vbox (or similar driver) behavior with start/stop?

tiborvass:
@nathanleclaire why don't you ask the driver to get the IPs of *running machines* and combine it with the last known state. That way if the machine is stopped and there's no IP, we still have the "last known IP".

nathanleclaire:
> so i guess my question is how does this negatively impact the vbox (or similar driver) behavior with start/stop?

Well, `docker-machine start` (without any args) won't work like it does (or at least is supposed to) today.  The specific version will be fine though.

> why don't you ask the driver to get the IPs of running machines and combine it with the last known state. That way if the machine is stopped and there's no IP, we still have the "last known IP".

What for?  I just don't want to break the no-name-specified `docker-machine start` method without a team commitment that we are going to go forward with always being explicit.

ehazlett:
I don't think we should have a mix of commands that can use active and commands that can't. 

Personally, I'm in favor of being explicit with the names.  However, since we allow the "active" machine to be used today, I'm not sure we should break that as users could be using it.

@bfirsh thoughts?

tiborvass:
@ehazlett TBH, machine is still in a state where you can break things and if it makes more sense to do so, it's better. I just don't understand how this PR would break `docker-machine start`. That command should start the active host, which is whatever is in DOCKER_HOST if it can match it. If it cannot match DOCKER_HOST than it should error out.

nathanleclaire:
> TBH, machine is still in a state where you can break things and if it makes more sense to do so, it's better.

Yeah, I'd really just rather always be explicit, and I think it's totally OK to change things like that around right now.

> I just don't understand how this PR would break docker-machine start. That command should start the active host, which is whatever is in DOCKER_HOST if it can match it.

The sticky part is that some drivers (such as VirtualBox) require the instance to be up in order to retrieve the IP address.  So, with this PR, the match with `DOCKER_HOST` will fail, because the reported IP address for the "active" host will be an empty string.  We are tracking the name of the active host in a file on disk currently, so the current implementation doesn't have these issues.

nathanleclaire:
At any rate, if/before this merged, I should make sure to update the documentation.

tiborvass:
@nathanleclaire 

> So, with this PR, the match with DOCKER_HOST will fail, because the reported IP address for the "active" host will be an empty string.

That's why I suggested that machine caches the IPs in case the driver cannot retrieve the IP.

ehazlett:
I like the idea of this.  @nathanleclaire where does it stand?  I think we could for now, require start to have explicit args like rm perhaps.

nathanleclaire:
@ehazlett My preference would be to eliminate the short (non-explicit) form completely and have this merge this in.  We could use the existing `.active` file as a "backup" like Tibor mentioned but I don't really want to have yet another area we track state, I'd rather remove it.

Would you prefer that way as well?  If we are blocked on moving forward making a decision on it, @samalba suggested we get @shykes input as a decider.

ehazlett:
As I said, I am in agreement and I welcome any and all feedback.  I don't like the way it works now nor tracking multiple states.  Let's make each command explicit.

nathanleclaire:
OK cool - just making sure we're still on the same page - guess we need to corner @shykes and get his input.

tiborvass:
@nathanleclaire to be clear I just wanted to track the "last-known-IP" of the machines. You can think of it as some kind of state, but it's not the same thing as "what's the active host"-kind-of state.

Maybe that IP state should be driver specific, and only implemented in the Vbox driver.

nathanleclaire:
Update: I've revised it to be congruent with the recent libmachine changes in HEAD, the integration tests are still a WIP though (have a working one for vbox that I mostly will just need to copy to the other ones).

tiborvass:
@nathanleclaire does my comment make sense? https://github.com/docker/machine/pull/777#issuecomment-88600396

Let me know if it's not clear :S (Wanna make sure it doesn't get overlooked)

tiborvass:
Ping @nathanleclaire

tiborvass:
Ping @nathanleclaire @ehazlett

nathanleclaire:
Hi @tiborvass like Evan mentioned in IRC we really want to get this in to the 0.3.0 release if we can so I will be working on it ASAP.  We plan to deprecate the short form usage, which takes care of most of the issues around the machines' IPs not being match-able to `DOCKER_HOST`.

tiborvass:
@nathanleclaire what is the "short form usage" ?

nathanleclaire:
@tiborvass `docker-machine start` and `docker-machine stop` commands which look at the "active" machine to figure out which one they should run against rather than having the user specify the machine explicitly.

tiborvass:
@nathanleclaire does this include `docker-machine env` as well ?

If you remove the short form, then what's the purpose of this PR ? What can DOCKER_HOST be used for?

nathanleclaire:
@tiborvass Yep, all of the commands including `env` wouldn't support the short form any more.  The user will always have to be explicit.

The purpose is that the "active" host will be automatically tracked and queryable based on the value in `DOCKER_HOST`.  That way, if I run `docker-machine ls`, the star _always_ appears next to the thing which is set in `DOCKER_HOST`, and `docker-machine active` will always spit out the correct thing based on the environment variable.  That's not the case now: it drifts and the user manually has to set `docker-machine active foo` after running `eval "$(docker-machine env foo)"` instead of having the change picked up on automatically.

In the implementation I have now (I'm re-working this on a new branch), `docker-machine active` is a bit naive in that it checks each IP address to see if it matches up with what is set in `DOCKER_HOST`, and returns the name of that machine if so.  Therefore, it takes a little while to run (~700ms for 3 machines on my box) - but, my long-term goal would be to come up with a way to make it embeddable on terminal prompts (so you can always see which machine you are pointed at).  

One way to possibly do this would be to set a daemon label on the created machine with its machine name, and make a quick call to `docker info` inside of machine to get the information, so that you can just hit whatever's in `DOCKER_HOST` when you run `docker-machine active` and it will tell you.  It would be faster (~200ms).  

I'm not against an "on-disk cache" entirely, but part of the problem is I would like users to be able to have "per-terminal-session" active hosts, not just "one true active" (does that make sense?), and so any mechanism powering it would have to account for that as well.

tiborvass:
@nathanleclaire thanks for your long answer.

I agree that explicit is better. The label part is interesting!

I'm not sure I understand the last part though. I never mentioned "on-disk caching" for caching which host is active. I mentioned caching of the IP addresses for vbox machines, because vbox does not provide the IP address for stopped machines. Sorry if I explained this unintelligibly: https://github.com/docker/machine/pull/777#issuecomment-88600396 and  https://github.com/docker/machine/pull/777#issuecomment-83077404

Initially I thought about caching it for all providers, but now I think it makes sense to make that provider-specific. Let me know if something still isn't clear :S

ehazlett:
@nathanleclaire can we close this as #1103 is merged?

nathanleclaire:
Yep

PR #780, opened 2015-03-13
Set the User-Agent for OpenStack and Rackspace drivers

smashwilson:
This will be helpful for identifying requests made with `docker-machine` from server logs, to determine how heavily it's used, what versions are most common, and other interesting statistics.

### DIFF
--- a/drivers/openstack/client.go
+++ b/drivers/openstack/client.go
@@ -2,9 +2,11 @@ package openstack
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net/http"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/machine/version"
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/openstack"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/keypairs"
@@ -415,6 +417,8 @@ func (c *GenericClient) Authenticate(d *Driver) error {
 	if err != nil {
 		return err
 	}
+
+	provider.UserAgent.Prepend(fmt.Sprintf("docker-machine/v%s", version.VERSION))
 
 	if d.Insecure {
 		// Configure custom TLS settings.
--- a/drivers/rackspace/client.go
+++ b/drivers/rackspace/client.go
@@ -5,6 +5,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/machine/drivers/openstack"
+	"github.com/docker/machine/version"
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/rackspace"
 )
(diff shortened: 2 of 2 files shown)


drivers/drivers.go
```diff
@@ -175,3 +175,10 @@ func GetSSHCommandFromDriver(d Driver, args ...string) (*exec.Cmd, error) {
 
 	return ssh.GetSSHCommand(host, port, user, keyPath, args...), nil
 }
+
+// SetVersioner is an interface to be implemented by specific Drivers that need access to the
+// current version of docker-machine. If a driver implements SetVersioner, it will be called with
+// the version on initialization.
+type SetVersioner interface {
+	SetVersion(string)
```
smashwilson: This is a bit of an awkward way to get the `VERSION` constant into the driver. I can't reference it directly from the `main` package because it creates an import cycle. If there's a cleaner way to accomplish this that I'm not seeing, then I'm all ears :smile:

hairyhenderson:
Cool! Is there any reason this shouldn't be done on all drivers?

smashwilson:
Well, I'm not certain that all driver SDKs support setting a custom user-agent. We only added the capability to Gophercloud recently ourselves.

Certainly, this would smooth the way for the maintainers of the other drivers to implement something similar, if they so choose.

ehazlett:
+1 for the idea.  I think your idea is good as far as being able to set a specific driver version.  Perhaps we could add a util function to `Driver` that would just return the current version?  That way we don't have to inject into each driver and the driver could pull it if needed?

Something like:

`drivers/drivers.go`:
```
func GetVersion() string {
    return version.VERSION
}
```

smashwilson:
> Perhaps we could add a util function to Driver that would just return the current version? That way we don't have to inject into each driver and the driver could pull it if needed?

Hmm, sure, I can try that. I'll ping here once I get a chance to update this.

ehazlett:
@smashwilson thx

smashwilson:
@ehazlett That still causes an import cycle:

```
Error reading packages: exit status 1
Stderr: can't load package: import cycle not allowed
package github.com/docker/machine
	imports github.com/docker/machine/drivers
	imports github.com/docker/machine
```

What I can do is implement a `drivers.SetVersion()` function and call it during initialization to inject the version constant. Or even just set a `drivers.Version` var.

smashwilson:
@ehazlett There we go. How's that?

ehazlett:
How about something like this?

https://github.com/docker/machine/compare/master...ehazlett:driver-version?expand=1

What do you think?

smashwilson:
@ehazlett Is that diff missing the extraction of the VERSION constant from `version.go` to a `github.com/docker/machine/version` package?

That would also work to break the import cycle, but [some of your build infrastructure](https://github.com/docker/machine/blob/master/docs/Dockerfile#L10) seems to rely on that file containing that constant, so it would need to be updated, as well.

I'll give it a shot and see what I can do.

smashwilson:
Oh, I see, that package has been created on master since I started. That makes this much easier.

ehazlett:
@smashwilson great point.  I've updated the docs dockerfile to work with this.  Thanks!

smashwilson:
@ehazlett :+1: Cool.

How is this? Now that `VERSION` is in its own package, this becomes even more trivial :grin:

ehazlett:
LGTM  thanks!

smashwilson:
:bow: thanks!

ehazlett:
thank you!

PR #784, opened 2015-03-15
Wait for any activeTransactions to finish during Create

hairyhenderson:
This adds an additional wait phase to make sure any active Transactions
are completed before Create tries to SSH in to the new host. This is
sometimes necessary because SSH can become available before SoftLayer
is fully done setting up the host and strange things can happen...

### DIFF
--- a/drivers/softlayer/driver.go
+++ b/drivers/softlayer/driver.go
@@ -307,78 +307,77 @@ func (d *Driver) PreCreateCheck() error {
 	return nil
 }
 
+func (d *Driver) waitForStart() {
+	log.Infof("Waiting for host to become available")
+	for {
+		s, err := d.GetState()
+		if err != nil {
+			log.Debugf("Failed to GetState - %+v", err)
+			continue
+		}
+
+		if s == state.Running {
+			break
+		} else {
+			log.Debugf("Still waiting - state is %s...", s)
+		}
+		time.Sleep(2 * time.Second)
+	}
+}
+
+func (d *Driver) getIp() (string, error) {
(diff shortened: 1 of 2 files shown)


drivers/softlayer/driver.go
```diff
@@ -305,11 +313,41 @@ func (d *Driver) Create() error {
...
 			if s == state.Running {
 				break
+			} else {
+				log.Debugf("Still waiting - state is %s...", s)
+			}
+			time.Sleep(2 * time.Second)
+		}
+	}
+
+	waitForSetupTransactions := func() {
```
nathanleclaire: Any particular reason this is done for using a closure here?  I'd really prefer to move this and the other closures into real methods / helper functions.

hairyhenderson: Nope. Was just trying to be consistent with the code above. I thought of refactoring and moving the three out, but wanted to minimize the changes.

nathanleclaire: If you wanted to do so, I'd be OK with it!

hairyhenderson: ok, done :)

nathanleclaire:
FYI eventually we will be moving some of that logic out of the driver (shouldn't have to get the IP, wait for SSH, or run `apt-get` commands inside the `Create` method) but until then - LGTM

nathanleclaire:
cc @ehazlett

ehazlett:
+1 to what @nathanleclaire said -- eventually these are planned to be moved out.  

LGTM

PR #795, opened 2015-03-17
Add "eval" to every invocation of "docker-machine env"

tianon:
See http://unix.stackexchange.com/a/181581 for the rationale behind this change (ie, better compatibility in the future, especially if `env` ever has to get more complex logic performed within the user's shell, since a workaround for one of the issues this solves is already in place).

See also docker/docker#11429, boot2docker/boot2docker#716, and boot2docker/boot2docker#786.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,13 @@ INFO[0000] Creating VirtualBox VM...
 INFO[0007] Starting VirtualBox VM...
 INFO[0007] Waiting for VM to start...
 INFO[0041] "dev" has been created and is now the active machine.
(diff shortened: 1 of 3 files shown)


ehazlett:
LGTM

ehazlett:
Thanks!

tianon:
:heart:

huslage:
This breaks the user experience. No one wants to run `eval` against anything. It's basically treating an edge case as if it were a major issue.

md5:
> This breaks the user experience. No one wants to run eval against anything. It's basically treating an edge case as if it were a major issue.

It's an edge case that is very confusing for newcomers. I think this is a good change.

Issue #653, opened 2015-02-27
Pass own AWS/EC2 endpoint


MitchK:
We have an EC2-compliant stack on premise and are starting to evaluate ```docker-machine```

The EC2 CLI seems to take an EC2_URL environment variable, so we can use our own on-premise API endpoint.
```
export AWS_ACCESS_KEY={Write your access key}
export AWS_SECRET_KEY={Write your secret key}
export EC2_URL={Write the endpoint}
```

Will this work with docker-machine and the AWS driver? Does anyone already have experience with this?

Cheers,
Michael

MitchK:
I have tried it this way:
```
$ docker-machine create --driver amazonec2 --amazonec2-access-key=$AWS_ACCESS_KEY --amazonec2-secret-key=$AWS_SECRET_KEY --amazonec2-subnet-id=$AWS_SUBNET_ID dev
client encountered error while doing the request: Get https://ec2.us-east-1.amazonaws.com/?Action=DescribeKeyPairs&Version=2014-06-15: dial tcp: lookup ec2.us-east-1.amazonaws.com: no such hostERRO[0000] Error creating machine: Problem with AWS API call: client encountered error while doing the request: Get https://ec2.us-east-1.amazonaws.com/?Action=DescribeKeyPairs&Version=2014-06-15: dial tcp: lookup ec2.us-east-1.amazonaws.com: no such host 
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0000] Error creating machine
```

Seems that ```docker-machine``` does not read/pass environment variables to the driver. 

Is it possible to add an option like ```--amazonec2-url```?

andrewwatson:
Yes, that should be possible. look in drivers/amazonec2/amazonec2.go

sthulb:
You'll have to make it ignore the region flag I guess

ehazlett:
Thanks for the feedback!  I think this would be a good addition.

jeanlaurent:
This in done in #2535

PR #800, opened 2015-03-18
Fixes for on-premise EC2 endpoints

MitchK:
This pull request relates to this issue: https://github.com/docker/machine/issues/653

**WIP**
 - [x] Allow to set a custom on-premise EC2 endpoint (currently added for testing internally)
 - [x] Add command line argument, replace the environment variable
 - [ ] Allow to trust self-signed certificates for on-premise EC2 endpoints
 - [ ] Add tests

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -66,9 +66,11 @@ type Driver struct {
 	SwarmDiscovery     string
 	storePath          string
 	keyPath            string
+	Endpoint           string
 }
 
 type CreateFlags struct {
+	Endpoint           *string
 	AccessKey          *string
 	SecretKey          *string
 	Region             *string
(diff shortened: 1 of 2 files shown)


MitchK:
**Proposal**
Add two command line arguments:

 * ```--amazonec2-url=...``` - Covers protocol, host with region and port
 * ```--amazonec2-certificate=...``` - Allows to import a self-signed ```.pem``` certificate

Any comments?

ehazlett:
I think we could add the args as I see this would be beneficial to ec2 compatible providers.  However, I think if we provide the `endpoint` arg we should allow the `EC2_URL` env var as well (as we do with other providers).

MitchK:
Okay, you can now use ```--amazonec2-endpoint=https://xyz``` and also use the environment variable ```AWS_ENDPOINT```(instead of ```EC2_URL```).

Next thing to fix will be the certificate validation for self-signed certificates.

ehazlett:
Any update on this?

MitchK:
Yep, I'll try to add the certificate validation. ```--amazonec2-certificate=``` plus the environment variable will do it. Still my first steps with Go :)

bgokden:
Are there any other development on this?

MitchK:
Currently, I don't have too much time to dig into this. Closing for now. I also think that there are not many companies with on-premise AWS-like endpoints.

PR #802, opened 2015-03-18
revert "function GetActive should return real error when file doesn't ex...

ehazlett:
#772 caused a regression where the `rm` command was no longer to remove machines when there was no active machine set.  Reverting.

This reverts commit 036b6b07b28845ce54c3c826363a4c0c7d04cadd.

### DIFF
--- a/host_test.go
+++ b/host_test.go
@@ -164,16 +164,12 @@ func TestMachinePort(t *testing.T) {
 	}
 	flags := getTestDriverFlags()
 
-	host, err := store.Create(hostTestName, hostTestDriverName, flags)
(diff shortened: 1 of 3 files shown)


TrumanLing:
@ehazlett I don't think function Remove in file store.go need to check active host, because command rm need a machine name as its parameter, and logically check the parameter at the starting of function cmdRm in file commands.go.

ehazlett:
Yeah there has been some discussion around making all commands take explicit machine names.  I agree with you -- I don't think there needs to be an "active" check in `rm` as we already check for the arg.  A separate PR would be best.

Thanks!

PR #807, opened 2015-03-18
do hostname validation in commands.go instead of Host.Create()

crunchywelch:
this prevents broken host entries from hanging around in the "db" because of a bad hostname which require an rm -f

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -343,6 +343,11 @@ func cmdCreate(c *cli.Context) {
 		log.Fatal("You must specify a machine name")
 	}
 
+	name, err := ValidateHostName(name)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	if err := setupCertificates(c.GlobalString("tls-ca-cert"), c.GlobalString("tls-ca-key"),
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -343,6 +343,11 @@ func cmdCreate(c *cli.Context) {
 		log.Fatal("You must specify a machine name")
 	}
 
+	name, err := ValidateHostName(name)
```
hairyhenderson: Seems to me it'd be good to merge this with the block immediately above (ln 341), since an empty name would also fail the ValidateHostName check. Also it'd be useful to `cli.ShowCommandHelp(c, "create")` before erroring out.

crunchywelch:
@hairyhenderson ah sure, works for me, will hopefully have a moment tomorrow to update this...

nathanleclaire:
cc @ehazlett I'd close this as it's a dupe of that one just merged

Thanks for the contribution @crunchywelch !  Hope to see you around more.

ehazlett:
@crunchywelch as @nathanleclaire mentioned, this has been fixed in another commit.  thanks for the contribution!

crunchywelch:
thanks @nathanleclaire @ehazlett sounds good!

Issue #812, opened 2015-03-19
Compile Error on Red Hat Server 7.0


TrumanLing:
Environmental info:
```console
[root@localhost ~]# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.0 (Maipo)
```
Compile message:
```console
[root@localhost machine]# script/build 
Sending build context to Docker daemon 6.501 MB
Sending build context to Docker daemon 
Step 0 : FROM golang:1.3-cross
 ---> c9b4f7f2a976
Step 1 : ENV http_proxy http://s00318865:*uhb9ijn@proxyus-rd.huawei.com:8080
 ---> Using cache
 ---> e077be9980cd
Step 2 : ENV https_proxy https://s00318865:*uhb9ijn@proxyus-rd.huawei.com:8080
 ---> Using cache
 ---> c82427052e70
Step 3 : ENV no_proxy rnd-isourceb.huawei.com,/var/run/docker.sock,localhost,docker01.thstack.com
 ---> Using cache
 ---> 8bc2a0c75c97
Step 4 : RUN apt-get update && apt-get install -y --no-install-recommends openssh-client
 ---> Using cache
 ---> 6dd73a624882
Step 5 : RUN go get github.com/mitchellh/gox
 ---> Using cache
 ---> 8341b65f64cb
Step 6 : RUN go get github.com/aktau/github-release
 ---> Using cache
 ---> aed11abed523
Step 7 : ENV GIT_SSL_NO_VERIFY 1
 ---> Using cache
 ---> 3455ae41e461
Step 8 : RUN go get github.com/tools/godep
 ---> Using cache
 ---> f97fb52d21f0
Step 9 : RUN go get code.google.com/p/go.tools/cmd/cover
 ---> Using cache
 ---> 6f1f38bf410b
Step 10 : ENV GOPATH /go/src/github.com/docker/machine/Godeps/_workspace:/go
 ---> Using cache
 ---> 1aab22a50dd3
Step 11 : ENV MACHINE_BINARY /go/src/github.com/docker/machine/docker-machine
 ---> Using cache
 ---> 296bbe828d8b
Step 12 : ENV USER root
 ---> Using cache
 ---> 62d298819dff
Step 13 : WORKDIR /go/src/github.com/docker/machine
 ---> Using cache
 ---> c1038eef89d0
Step 14 : ADD . /go/src/github.com/docker/machine
 ---> Using cache
 ---> 7208dcff09c8
Successfully built 7208dcff09c8
[...]

PR #813, opened 2015-03-19
Fix compiling error on Red Hat Server 7.0

TrumanLing:
Fix issue #812

### DIFF
--- a/script/build
+++ b/script/build
@@ -15,4 +15,4 @@ fi
 
 docker build -t docker-machine .
 rm -f docker-machine*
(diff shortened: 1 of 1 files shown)


ehazlett:
Hmm I'm not sure we need to add `--privileged` here.  Perhaps we need to add a CAP or something for RedHat specifically.  It seems to build find everywhere else and I am generally not in favor of just adding privileged.

ehazlett:
Closing as we do not want to enable privileged containers just for the build.  I would rather find the correct specific CAP and use that.  Please feel free to re-open with that.  Thanks!

PR #815, opened 2015-03-19
vbox: fix issue where could not restart a stopped instance

ehazlett:
This fixes not being able to issue `restart` to a stopped instance in VirtualBox.

### DIFF
--- a/host.go
+++ b/host.go
@@ -643,12 +643,14 @@ func (h *Host) Kill() error {
 }
 
 func (h *Host) Restart() error {
-	if err := h.Stop(); err != nil {
-		return err
-	}
+	if h.MachineInState(state.Running)() {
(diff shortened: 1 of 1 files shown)


ehazlett:
PTAL @sthulb @nathanleclaire

nathanleclaire:
LGTM, although I think I would expect an attempt to "restart" a stopped machine to error out?

ehazlett:
@nathanleclaire well before it would work on some drivers but others it wouldn't.  I guess if we want to error we should do it for all.

nathanleclaire:
Since you need it for testing, let's merge this in for now and if we decide that we don't like the behavior later we can change it.

ehazlett:
SGTM

PR #821, opened 2015-03-19
godep: fix upstream errors; update naturalsort

ehazlett:
This fixes the errors with the azure vendor rev as well as a custom SHA for Docker.  Docker now uses the 1.5.0 tag.  Also, `naturalsort` was updated to fix https://github.com/docker/machine/issues/539#issuecomment-83699610 This also removes unnecessary vendored libs too.


nathanleclaire:
LGTM

Issue #819, opened 2015-03-19
Virtualbox CPU Support


stephenlawrence:
Add an option to docker-machine Virtualbox integration to allow for the number of cpu's to be adjusted during create.

hairyhenderson:
Currently the behaviour is to use the same number of CPUs as the host has. Any reason you would want or need to adjust this?

greentg:
It's already an option for some of the other providers, so there should at least be some kind of consistency.  Thoughts?

dnvriend:
Reason: on my machine, MacBook Pro i7 end 2011 and Akka/play packaged with Docker, performance is best using 1 cpu per real core, and not the virtual cores (HyperThreading). I must manually change this setting. Not a real big issue, but it would be nice.

hairyhenderson:
@dnvriend - ah. Makes sense.

pbiggar:
Trying to do this during the open-source-a-thon

hairyhenderson:
I think this was a duplicate of #822

PR #822, opened 2015-03-19
add flag to support custom number of virtualbox processors

greentg:
This is an open feature request at #819 
Presently there are several other providers that expose this as a tunable, so there's an argument to be made that it should be included for consistency. 

The default behavior is to continue using the number of the CPUs on the machine of the flag isn't specified.

Flag is: --virtualbox-processors

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -73,7 +73,7 @@ func GetCreateFlags() []cli.Flag {
 	return []cli.Flag{
 		cli.IntFlag{
 			Name:  "virtualbox-processors",
-			Usage: "Number of virtual CPUs cores for host",
+			Usage: "Number of virtual CPUs cores for host. Defaults to number of available cores",
 			Value: GetCPUCores(),
 		},
 		cli.IntFlag{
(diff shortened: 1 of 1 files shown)


stephenlawrence:
I was putting together a pull request for this, you beat me to it. :)

drivers/virtualbox/virtualbox.go
```diff
@@ -58,11 +59,24 @@ func init() {
...
+	return cpus
+}
+
 // RegisterCreateFlags registers the flags this driver adds to
 // "docker hosts create"
 func GetCreateFlags() []cli.Flag {
 	return []cli.Flag{
 		cli.IntFlag{
+			Name:  "virtualbox-processors",
+			Usage: "Number of virtual CPUs cores for host",
```
hairyhenderson: Should probably mention that it defaults to the number of cores on the host. The default might seem to make it obvious, but it's easy to miss the correlation ;)

greentg: Thanks @hairyhenderson, added a description for default number of cores.

ehazlett:
Sorry we had another submission and I forgot about this one!  Closing as the functionality has been merged.  Thanks for the contribution!

PR #824, opened 2015-03-20
validate host config while load

ehazlett:
This depends on #767.

This adds a small shim that will "migrate" the 0.1.0 configs to be compatible with 0.2.0.  It's transparent to the user.


ehazlett:
This has been merged into #767 as it contains vital fixes for that.  Closing this PR.

PR #805, opened 2015-03-18
File support for b2d

ehazlett:
This fixes #539 as well as a couple manual URL bugs.

PR #826, opened 2015-03-20
fix option `--virtualbox-boot2docker-url`

shawnzhu:
fix for #669 by addressing [PR 670 (comment)](/docker/machine/pull/670#issuecomment-77972562)


shawnzhu:
the problem is already addressed in #805 and will be closed once #805 is merged

shawnzhu:
fixed in #805

PR #827, opened 2015-03-20
Add provisioning based on OS

nathanleclaire:
Closes https://github.com/docker/machine/issues/699

Continuation of work originally PRed in https://github.com/docker/machine/pull/546

This also fixes a swarm regression in master.

cc @ehazlett @sthulb


nathanleclaire:
Update: I also migrated the "config" naming scheme to "options" like we discussed.

libmachine/provision/utils.go
```diff
@@ -0,0 +1,274 @@
...
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/machine/libmachine/auth"
+	"github.com/docker/machine/libmachine/provision/pkgaction"
+	"github.com/docker/machine/libmachine/swarm"
+	"github.com/docker/machine/utils"
+)
+
+type DockerConfig struct {
```
nathanleclaire: use of `DockerConfig` and `EngineConfig` persists here

test/integration/helpers.bash
```diff
@@ -12,6 +12,19 @@ else
     ARCH="386"
 fi
 MACHINE_BIN_NAME=docker-machine_$PLATFORM-$ARCH
+BATS_LOG=${MACHINE_ROOT}/bats.log
+
+touch ${BATS_LOG}
```
nathanleclaire: I don't have `truncate` on my Mac, so I've changed this for now - we can modify it later to something more elegant if we want.

ehazlett: Ah ok -- no problem.

ehazlett:
Tested against several drivers, swarm etc.  LGTM

Thanks!

PR #830, opened 2015-03-21
update changelog

ehazlett:
Update for new `eval` usage.


sthulb:
@ehazlett we should make updating the changelog a part of PRs

ehazlett:
@sthulb +1 -- I'm also going to try to keep it updated as we go instead of all at once :)

PR #850, opened 2015-03-23
feature to fix #737 - openstack driver can use nova to allocate floating ips

### LINKED ISSUE #737: support openstack-floatingip-pool without neutron
## Summary
I expect docker-machine's openstack driver to support floating ip arguments even if I don't have a neutron service running.
This is likely a feature request... definitely a documentation bug ( documentation should specify that neutron is required for floating-ip at this point)... and maybe a bug too.

## Analysis
Given an openstack cluster with purely nova-based networking.

I expect that I can use docker-machine create to support floating ip pools.

Instead, the client fails with:
```
ERRO[0000] Error creating machine: No suitable endpoint could be found in the service catalog.
WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed.
FATA[0000] Error creating machine
```

Omitting all floating ip arguments and flags allows the create command to continue merrily.

## Next steps

Per discussion below, gophercloud already supports this. A PR will be made to add a service check, and if neutron is not found, the openstack client will just use the nova api to nab an ip address.

dudymas:
See #737 for details on issue.

This code change creates a new option for the openstack driver, in order to amend the problem.

While this doesn't get points for cleanliness or testing, it does work.

Let's discuss what concerns might come about from this feature, if necessary.

### DIFF
--- a/drivers/openstack/client.go
+++ b/drivers/openstack/client.go
@@ -11,6 +11,7 @@ import (
 	"github.com/docker/machine/libmachine/version"
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/openstack"
+	compute_ips "github.com/rackspace/gophercloud/openstack/compute/v2/extensions/floatingip"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/startstop"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/flavors"
@@ -40,7 +41,7 @@ type Client interface {
 	GetNetworkId(d *Driver) (string, error)
 	GetFlavorId(d *Driver) (string, error)
 	GetImageId(d *Driver) (string, error)
-	AssignFloatingIP(d *Driver, floatingIp *FloatingIp, portId string) error
+	AssignFloatingIP(d *Driver, floatingIp *FloatingIp) error
 	GetFloatingIPs(d *Driver) ([]FloatingIp, error)
 	GetFloatingIpPoolId(d *Driver) (string, error)
 	GetInstancePortId(d *Driver) (string, error)
@@ -98,6 +99,8 @@ type FloatingIp struct {
 	Ip        string
 	NetworkId string
 	PortId    string
+	Pool      string
+	MachineId string
 }
 
 func (c *GenericClient) GetInstanceState(d *Driver) (string, error) {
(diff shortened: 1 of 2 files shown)


drivers/openstack/openstack.go
```diff
@@ -165,6 +166,11 @@ func GetCreateFlags() []cli.Flag {
 			Value: "",
 		},
 		cli.StringFlag{
+			Name:  "openstack-ip-pool-type",
+			Usage: "OpenStack floating IP type, defaults to 'neutron', also accepts 'nova'.",
+			Value: "",
```
hairyhenderson: Can you set this to `neutron` here? Then it's _actually_ the default :)

drivers/openstack/openstack.go
```diff
@@ -555,7 +565,7 @@ func (d *Driver) resolveIds() error {
 		}).Debug("Found image id using its name")
 	}
 
-	if d.FloatingIpPool != "" {
+	if d.FloatingIpPool != "" && d.IpPoolType != "nova" {
```
hairyhenderson: `d.FloatingIpPool` is set explicitly to `nova` when it's empty and the `d.IpPoolType == "nova"` (ln 277), so this check is unnecessary

hairyhenderson: Scratch that - I wasn't paying attention to this. Carry on... Nothing to see here... :P

drivers/openstack/openstack.go
```diff
@@ -165,6 +166,11 @@ func GetCreateFlags() []cli.Flag {
 			Value: "",
 		},
 		cli.StringFlag{
+			Name:  "openstack-ip-pool-type",
```
ggiamarchi: It would be better to have a flag to switch between neutron and nova-network not only for IP pool. We could use a boolean flag `--openstack-nova-network` with the default value to `false` as long as neutron is the mainstream networking service.

drivers/openstack/client.go
```diff
@@ -270,7 +273,37 @@ func (c *GenericClient) GetServerDetail(d *Driver) (*servers.Server, error) {
...
+	return err
+}
+
+func (c *GenericClient) assignNovaFloatingIP(d *Driver, floatingIp *FloatingIp) error {
+	var err error
+	if floatingIp.Ip == "" {
+		f, err := compute_ips.Create(c.Compute, compute_ips.CreateOpts{
+			Pool: d.FloatingIpPool,
+		}).Extract()
+		if err == nil {
```
ggiamarchi: It's better to immediately return on error.

```go
if err != nil {
    return err
}
```

dmp42:
@ggiamarchi @hairyhenderson is this something we still need / want?

hairyhenderson:
@dmp42 - yeah, I think we still want this. It's not critical for _me_, but I'm sure it's an annoyance to lots of openstack users.

dmp42:
@dudymas still with us? :-)

@hairyhenderson if you would be willing to carry this through with @ggiamarchi feedback, that would be awesome!

hairyhenderson:
@dmp42 - hrm... I don't have access to an OpenStack instance anymore, so I'd need @ggiamarchi's help (or someone's!) testing it out. But I'm open to it if @dudymas isn't able.

ggiamarchi:
@dmp42 The content of this pull request makes docker-machine to work on old OpenStack installations (where networking is not handled by the neutron service but by the nova service).

I did comments on this PR few month ago but nothing have been done not so far. I can finish the work on it if nobody else can.

dudymas:
sorry for being AWOL

dudymas:
This code has been tested on openstack juno and kilo

dmp42:
Thanks @dudymas !

@ggiamarchi @hairyhenderson can you take a look?

Making sure the driver is as solid as possible is paramount now, with the release nearing :)

dudymas:
np. I'm really digging the drivers system, and how you pulled the ssh stuff out and moved things into a base driver type.

Please -do- test this on your own openstack deployments. Fact is: openstack solutions vary greatly and this code might be using gophercloud in a manner that only supports a subset of those.

hairyhenderson:
Thanks @dudymas!

The code looks fine to me, but I won't be able to test this out on an openstack deployment.

dudymas:
Well, if no one can test but me, I personally think it's good enough to ship.

The only deployment I could possibly imagine needing more logic would be multi-nic nova-network setups. I just don't have hypervisors with multiple public networks, or I'd easily confirm that works too. Even then, I'm not even sure if docker-machine knows how to pick a particular nic on a vm to configure for docker access.

dmp42:
Ok, thanks.
@ggiamarchi any last revised opinion on this before it goes?

ggiamarchi:
@dudymas @dmp42 Code looks good but i would like to run some tests to ensure nothing is broken. I have not had time to do it so far. I'll do it tomorrow.

dudymas:
Sure thing. If you think the bool warrants a unit test, just holler. I'm thinking there are ways to mock up gophercloud clients for that. Not sure if the change in logic (essentially just an 'if' statement) would be worth it.

nathanleclaire:
@ggiamarchi I'm going to go ahead and merge this, if you get to test it and things look problematic, we can revert.  We're going to be releasing faster (every two weeks) after 0.5.0, so we can fix + release soon if needed.

LGTM

Issue #820, opened 2015-03-19
Getting Error Using 'vmwarefusion-boot2docker-url' command option


aaronj314:
To reproduce run the following:

    docker-machine --debug create -d vmwarefusion --vmwarefusion-boot2docker-url "https://github.com/cloudnativeapps/boot2docker/releases/download/v1.5.0-vmw/boot2docker-1.5.0-vmw.iso" osxdock

I get the following:

    INFO[0000] Downloading boot2docker.iso from https://github.com/cloudnativeapps/boot2docker/releases/download/v1.5.0-vmw/boot2docker-1.5.0-vmw.iso...
    ERRO[0000] Error creating machine: open /Users/xxxx/.docker/machine/cache/boot2docker-1.5.0-GH747.iso/boot2docker-1.5.0-GH747.iso.tmp638961045: no such file or directory
    WARN[0000] You will want to check the provider to make sure the machine and associated resources were properly removed.
    FATA[0000] Error creating machine

nathanleclaire:
There have been many changes to this code since then, so please re-test and re-open if it is still an issue.  Thanks!

PR #857, opened 2015-03-23
fix download destination for boot2docker iso

greentg:
First argument passed to DownloadISO should be a directory, not a filename.  This is a fix for issue #820 

Signed-off-by: Tim Green <timsgreen@gmail.com>


ehazlett:
Thanks for the contribution!  This will be fixed in #805.  Feel free to put feedback there.

PR #859, opened 2015-03-23
Fix location of validating hostname functionality

nathanleclaire:
Similar in functionality to #807 (thanks for the tipoff @crunchywelch!), I'd prefer to validate the hostname before we attempt to make the host / serialize it to disk at all.

There seems to be no need for multiple returns as the regex match only returns bool, so I've updated that as well.

cc @ehazlett @sthulb 

### DIFF
--- a/libmachine/machine.go
+++ b/libmachine/machine.go
@@ -20,6 +20,10 @@ func New(store Store) (*Machine, error) {
 }
 
 func (m *Machine) Create(name string, driverName string, hostOptions *HostOptions, driverConfig drivers.DriverOptions) (*Host, error) {
+	validName := ValidateHostName(name)
+	if !validName {
+		return nil, ErrInvalidHostname
+	}
 	exists, err := m.store.Exists(name)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


ehazlett:
I don't think this is related to your PR but should we be showing the entire help test for an invalid hostname?

nathanleclaire:
> I don't think this is related to your PR but should we be showing the entire help test for an invalid hostname?

Do you mean the "check on resources you created" message?

sthulb:
I can't actually see anywhere that calls `ValidateHostName(name string)`. We seem to have a regression there.

libmachine/machine.go
```diff
@@ -20,6 +20,10 @@ func New(store Store) (*Machine, error) {
 }
 
 func (m *Machine) Create(name string, driverName string, hostOptions *HostOptions, driverConfig drivers.DriverOptions) (*Host, error) {
+	validName := ValidateHostName(name)
```
nathanleclaire: @sthulb It gets called right here.  Am I am missing something?

ehazlett:
@nathanleclaire this was an error is my usage -- disregard :)

@sthulb does that answer your question?

libmachine/machine.go
sthulb: Apparently "ag" missed it...

ehazlett: cool thx!

ehazlett:
LGTM

PR #866, opened 2015-03-24
update usage for eval

ehazlett:
@hairyhenderson mentioned issues with using quotes around `eval` for loading the environment.  This updates the usage info after the create command to remove the quotes.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -434,7 +434,7 @@ func cmdCreate(c *cli.Context) {
 	case "fish":
 		info = fmt.Sprintf("%s env %s | source", c.App.Name, name)
 	default:
-		info = fmt.Sprintf(`eval "$(%s env %s)"`, c.App.Name, name)
+		info = fmt.Sprintf(`eval $(%s env %s)`, c.App.Name, name)
 	}
 
 	log.Infof("%q has been created and is now the active machine.", name)


nathanleclaire:
Hang on-  @tianon had quotes on it in https://github.com/docker/machine/pull/795 .  Do we have a specific reproducible use case where we know this is an issue?

tianon:
Quotes are definitely important for proper whitespace preservation, but I can't think of any current common issues with either way.

nathanleclaire:
OK - if we aren't aware of any issues with the status quo then I'm all for removing the quotes today.  Should keep it in mind in the future though.

nathanleclaire:
I think there are some docs that will need to be updated with this too.

ehazlett:
@hairyhenderson can you provide some feedback into this?

#869 looks related too.

tianon:
@nathanleclaire well hold up; if there aren't concrete problems removing them fixes, why remove them?  They're definitely more "correct", AFAIK.

ehazlett:
Ok I'm going to close this one.  I can't get any issue re-produced.

@hairyhenderson if you run into the issue again, can you file an issue with details on how to reproduce?

thanks all -- sorry for the noise.

PR #872, opened 2015-03-25
Test migration functions

nathanleclaire:
I feel pretty confident in the functionality we have now so we should test it for posterity's sake when things get moved around again.

I've renamed the functions and file (`validate.go` => `migrate.go`) since I find it more descriptive, but if you totally hate it just let me know and I'll change it back.

cc @ehazlett 


ehazlett:
I'm not super crazy about the "FillNested" but I can live it it for now.

:+1: for the tests

LGTM

Issue #877, opened 2015-03-25
Print usage hint on `docker-machine env`


aanand:
In light of confusing experiences like #851 (which I just experienced myself), it might be nice to show a usage hint in the output of `docker-machine env` - commented out, of course:

```
$ docker-machine env

export DOCKER_TLS_VERIFY=1
export DOCKER_CERT_PATH="/Users/aanand/.docker/machine/machines/dev"
export DOCKER_HOST=tcp://192.168.99.101:2376

# Run this command to configure your shell: eval $(docker-machine env)
```

(It needs to come afterwards because of shell expansion - if the hash is at the beginning, the whole thing will be treated as a comment :smirk:)

mattreyuk:
#dibs

ehazlett:
@aanand +1

hairyhenderson:
This was fixed by #882

PR #882, opened 2015-03-26
Added env usage hint for bash and fish shells

tombee:
This is an attempt at the usage hint mentioned in #877 

Signed-off-by: Tom Barlow <tomwbarlow@gmail.com>

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -728,11 +728,13 @@ func cmdEnv(c *cli.Context) {
 
 	switch userShell {
 	case "fish":
-		fmt.Printf("set -x DOCKER_TLS_VERIFY 1;\nset -x DOCKER_CERT_PATH %q;\nset -x DOCKER_HOST %s;\n",
-			cfg.machineDir, dockerHost)
+		usageHint := fmt.Sprintf("# Run this command to configure your shell: eval (docker-machine env %s)", cfg.machineName)
+		fmt.Printf("set -x DOCKER_TLS_VERIFY 1;\nset -x DOCKER_CERT_PATH %q;\nset -x DOCKER_HOST %s;\n\n%s\n",
+			cfg.machineDir, dockerHost, usageHint)
 	default:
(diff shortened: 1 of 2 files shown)


tombee:
@ehazlett Let me know what you think of this :smile:

tombee:
@ehazlett Let me know what you think :smile:

aanand:
Nice.

commands.go
```diff
@@ -722,16 +722,41 @@ func cmdEnv(c *cli.Context) {
...
-			cfg.machineDir, dockerHost)
+		fmt.Printf("export DOCKER_TLS_VERIFY=1\nexport DOCKER_CERT_PATH=%q\nexport DOCKER_HOST=%s\n\n%s\n",
+			cfg.machineDir, dockerHost, usageHint)
 	}
 }
 
+func generateUsageHint(c *cli.Context, userShell string) string {
+	msg := "# Run this command to configure your shell: "
+	cmd := ""
+	machine := c.Args().First()
```
ehazlett: Since we only use the context for the machine name, we could probably have that as the first arg to avoid passing the entire context.

commands.go
```diff
@@ -722,16 +722,41 @@ func cmdEnv(c *cli.Context) {
...
+		}
+	default:
+		if machine != "" {
+			cmd = fmt.Sprintf("eval $(docker-machine env %s)", machine)
+		} else {
+			cmd = "eval $(docker-machine env)"
+		}
+	}
+
+	return fmt.Sprintf("%s %s\n", msg, cmd)
```
ehazlett: We could also just have the message as part of the `Sprintf` to avoid the extra var too.

tombee:
Great, thanks for the feedback @ehazlett -- I'll get those changes done now.

tombee:
@ehazlett How's this?

ehazlett:
LGTM

Issue #883, opened 2015-03-26
Strip out dependencies 


sthulb:
We should remove some of our dependencies, a good example is: `github.com/docker/docker/api`, it's only used by the `none` driver for validating a host.

sthulb:
Q: Are a lot of the things in our `Godeps.json` file there because they're dependencies of dependencies?

cc @ehazlett @nathanleclaire

ehazlett:
@sthulb yes i believe so

nathanleclaire:
Unfortunately, at the moment we don't have a great way to audit dependencies, although I do think a sweep to clean up would be good at some point.  Let's revisit this when we get some direction from Go upstream about their vendoring solution.

PR #884, opened 2015-03-26
Remove (nearly) useless deps

sthulb:
Refs: #883


ehazlett:
so did we just re-implement the docker validation here?  why not just use what is used against the engine?

ehazlett:
ping @sthulb

PR #887, opened 2015-03-26
Fix unix:// not working inside VM issue

nathanleclaire:
With the way that provisioning was implemented, there was an issue
getting connection to the UNIX socket to work with the new boot2docker
1.6rc2 ISO.  This issue is related to the fact that the docker
service was not truly stopped during the provisioning.  This PR fixes
the issue and provides some tests to help ensure that it does not come
up again.


nathanleclaire:
cc @ehazlett

ehazlett:
This looks mostly good however there appears to be some debug during creation:

```
Got an error it was dial tcp 192.168.99.109:2376: connection refused
```

nathanleclaire:
Good catch - I changed it to use `log.Debug` instead.

ehazlett:
LGTM

PR #892, opened 2015-03-27
hooked a PostCreateCheck for checking the configuration setting 

### LINKED ISSUE #843: Validate instance creation for Virtualbox using driver flags
**Problem**

Test the flags for VirtualBox

**Solution**

Validate instance creation using various flags for VirtualBox

There are several flags available for each driver.  Test each flag to ensure it does the intended behavior.

For example,

`docker-machine create -d virtualbox --virtualbox-memory "1024" test-vbox`

This should create a machine in virtualbox with 1024 MB of memory.

soleo:
Added a method called PostCreateCheck to enable to check if the configuration matches with the params provided or default value after the virtual machine is created. The same method could apply to each driver to do after creation checking. 

 Fix #843

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -304,6 +304,10 @@ func (d *Driver) PreCreateCheck() error {
 	return d.checkPrereqs()
 }
 
+func (d *Driver) PostCreateCheck() error {
+	return nil
+}
+
 func (d *Driver) instanceIpAvailable() bool {
 	ip, err := d.GetIP()
 	if err != nil {
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -229,6 +229,10 @@ func (d *Driver) PreCreateCheck() error {
 	return nil
 }
 
+func (d *Driver) PostCreateCheck() error {
+	return nil
+}
+
 func (d *Driver) Create() error {
 	if err := d.setUserSubscription(); err != nil {
 		return err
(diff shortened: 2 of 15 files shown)


ehazlett:
@soleo thanks!  we didn't implement `PostCreate` as the checking for creation is at start and once that is done it's handed to the provisioner.  I'm not sure we would need to check anything after creation since we just use SSH and that is checked as part of the provisioning process.

/cc @nathanleclaire @sthulb thoughts?

sthulb:
@ehazlett you could use it to verify the host created. However that could be implemented in `Create`.

ehazlett:
@sthulb yeah essentially that is what the SSH connect is doing isn't it, albeit not via a remote API but still checking what we expect to have.  I could see a potential case where the provider API says it is created but we still can't connect.  I'm not sure what this would gain us.

sthulb:
Me neither.

nathanleclaire:
Yes, I'm -1 on this change.  I can see the utility of smoke checks like this but I don't think this implementation is the correct way to approach it.  Eventually I would prefer to move over to declarative configuration of machines and do checks like this to calculate the "delta" to move to our desired state.

@soleo I definitely encourage opening an issue/proposal to discuss the problem you want to solve with this functionality and get the ball rolling on a conversation about it.

soleo:
I didn't think too much about it. The initial thought for this is just to have universal way  to validate the configuration after the vm is created. And, I saw several issues related with this. @nathanleclaire , probably it will be better used as a utility func, and it can also be used for health check of virtual machines, something like "brew doctor" in homebrew

ehazlett:
Closing this for now.  @soleo please watch #773 for the machine declaration functionality @nathanleclaire was mentioning -- that should achieve what you are looking for.  Thanks!

Issue #888, opened 2015-03-27
machine didn't start downloading boot2docker.iso first time to use


soleo:
```
bash-3.2$ ./docker-machine_darwin-386 -v
docker-machine_darwin-386 version 0.2.0 (45f08d9)
bash-3.2$ ./docker-machine_darwin-386 create --driver virtualbox dev
INFO[0000] Creating CA: /Users/xinjiang/.docker/machine/certs/ca.pem 
INFO[0001] Creating client certificate: /Users/xinjiang/.docker/machine/certs/cert.pem 
WARN[0004] Unable to check for the latest release: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: write udp 8.8.8.8:53: no route to host 
INFO[0004] Downloading boot2docker.iso to /Users/xinjiang/.docker/machine/cache/boot2docker.iso... 
ERRO[0004] Error creating machine: open : no such file or directory 
WARN[0004] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0004] Error creating machine                       
bash-3.2$ ls -l  /Users/xinjiang/.docker/machine/cache/boot2docker.iso
ls: /Users/xinjiang/.docker/machine/cache/boot2docker.iso: No such file or directory
```

It works after remove dev and create again

nathanleclaire:
It looks like we try to go forward anyway even though the release check / ISO download failed (might be related to the Github DDoS, but that's neither here nor there).  I have a fix in mind for this, thanks for the bug report.

nathanleclaire:
cc @ehazlett this should be fixed now

ehazlett:
@soleo please re-open if you experience this again.  thanks for reporting!

NoUseFreak:
I am having the same problem.

```bash
$ docker-machine -v
docker-machine version 0.2.0 (8b9eaf2)
```

```bash
$ docker-machine create -d virtualbox docker-dev
INFO[0000] Creating SSH key...                          
INFO[0000] No default boot2docker iso found locally, downloading the latest release... 
ERRO[0005] Error creating machine: Get https://api.github.com/repos/boot2docker/boot2docker/releases: dial tcp: i/o timeout 
WARN[0005] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0005] Error creating machine  
```

nathanleclaire:
@NoUseFreak Do subsequent `create` invocations run or are they failing as well?

[...]

PR #895, opened 2015-03-27
Fix release checking / ISO download paradox

nathanleclaire:
Fixes https://github.com/docker/machine/issues/888

Ensure that we don't proceed with creation if the release check failed
and we have no ISO locally.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -3,6 +3,7 @@ package virtualbox
 import (
 	"archive/tar"
 	"bytes"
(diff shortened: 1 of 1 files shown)


nathanleclaire:
cc @ehazlett

ehazlett:
We probably need this in hyperv, fusion, and vsphere too.

nathanleclaire:
With the recent `upgrade` changes this should be fixed.  We only make a release check if there is no `boot2docker.iso` around already or if we explicitly do one (which is what we do in the case of `upgrade`)

PR #898, opened 2015-03-27
Chagning VirtualBox version to the latest one

mtesselH:
Signed-off-by: Marianna <mtesselh@gmail.com> 
This upgrades VirtualBox to the most recent version 
cc: @nathanleclaire @ehazlett

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -47,7 +47,7 @@ Let's take a look at using `docker-machine` to creating, using, and managing a D
 host inside of [VirtualBox](https://www.virtualbox.org/).
 
 First, ensure that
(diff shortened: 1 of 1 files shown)


nathanleclaire:
To clarify the motivation for this change: at https://www.virtualbox.org/wiki/Downloads provides only a link to get the sub-version this PR updates to, not our previously referenced version.  It will be easier to get users to upgrade to the latest version of VirtualBox than downgrade to some specific sub-version if things go wrong, so I'd suggest that we always work with the latest.

ehazlett:
LGTM

Issue #776, opened 2015-03-13
Scope down the usage message when typing `docker-machine create -d driver` to only list flags relevant to that driver


hairyhenderson:
Right now, `docker-machine create -d openstack` spits outs 113 lines of usage message, 104 lines of which are flags. Only 18 of these lines are actually relevant to openstack.

It's a pretty minor papercut, but if there's a trivial way to scope it down, that'd be real slick.

sthulb:
@hairyhenderson I agree, we definitely need to change this. I'm unsure what the best UX would be for this.

hairyhenderson:
@sthulb maybe something like:

```
$ docker-machine create -d openstack
NAME:
   create - Create a machine

USAGE:
   command create [command options] [arguments...]

OPTIONS:
   --openstack-auth-url 												OpenStack authentication URL [$OS_AUTH_URL]
   --openstack-endpoint-type 												OpenStack endpoint type (adminURL, internalURL or publicURL) [$OS_ENDPOINT_TYPE]
   --openstack-flavor-id 												OpenStack flavor id to use for the instance
   --openstack-flavor-name 												OpenStack flavor name to use for the instance
   --openstack-floatingip-pool 												OpenStack floating IP pool to get an IP from to assign to the instance
   --openstack-image-id 												OpenStack image id to use for the instance
   --openstack-image-name 												OpenStack image name to use for the instance
   --openstack-insecure													Disable TLS credential checking.
   --openstack-net-id 													OpenStack network id the machine will be connected on
   --openstack-net-name 												OpenStack network name the machine will be connected on
   --openstack-password 												OpenStack password [$OS_PASSWORD]
   --openstack-region 													OpenStack region name [$OS_REGION_NAME]
   --openstack-sec-groups 												OpenStack comma separated security groups for the machine
   --openstack-ssh-port "22"												OpenStack SSH port
   --openstack-ssh-user "root"												OpenStack SSH user
   --openstack-tenant-id 												OpenStack tenant id [$OS_TENANT_ID]
   --openstack-tenant-name 												OpenStack tenant name [$OS_TENANT_NAME]
   --openstack-username 												OpenStack username [$OS_USERNAME]
   --swarm														Configure Machine with Swarm
   --swarm-master													Configure Machine to be a Swarm master
   --swarm-discovery 													Discovery service to use with Swarm
   --swarm-host "tcp://0.0.0.0:3376"											ip/socket to listen on for Swarm master
[...]

PR #901, opened 2015-03-29
Filter create command flags by driver

djfarrelly:
This implements the idea proposed in #776. When specifying a driver w/ the `-d` flag but not specifying a name, the command help only displays relevant flags to use with that specified driver. Example:

```
$ docker-machine create -d virtualbox
NAME:
   create - Create a machine

USAGE:
   command create [command options] [arguments...]

OPTIONS:
   --virtualbox-boot2docker-url 	The URL of the boot2docker image. Defaults to the latest available version [$VIRTUALBOX_BOOT2DOCKER_URL]
   --virtualbox-cpu-count "-1"		number of CPUs for the machine (-1 to use the number of CPUs available) [$VIRTUALBOX_CPU_COUNT]
   --virtualbox-disk-size "20000"	Size of disk for host in MB
   --virtualbox-memory "1024"		Size of memory for host in MB

Run 'docker-machine create --help' for all options.

FATA[0000] You must specify a machine name
```

**Note:** I'm very new to writing Go and the Docker machine project so please do let me know if I can make any adjustments to improve my change. Thanks!

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -161,6 +161,40 @@ func setupCertificates(caCertPath, caKeyPath, clientCertPath, clientKeyPath stri
 	return nil
 }
 
+var CreateFlags = []cli.Flag{
+	cli.StringFlag{
+		Name: "driver, d",
+		Usage: fmt.Sprintf(
+			"Driver to create machine with. Available drivers: %s",
+			strings.Join(drivers.GetDriverNames(), ", "),
+		),
+		Value: "none",
+	},
+	cli.BoolFlag{
+		Name:  "swarm",
+		Usage: "Configure Machine with Swarm",
+	},
+	cli.BoolFlag{
+		Name:  "swarm-master",
+		Usage: "Configure Machine to be a Swarm master",
+	},
+	cli.StringFlag{
+		Name:  "swarm-discovery",
+		Usage: "Discovery service to use with Swarm",
+		Value: "",
+	},
+	cli.StringFlag{
+		Name:  "swarm-host",
+		Usage: "ip/socket to listen on for Swarm master",
+		Value: "tcp://0.0.0.0:3376",
+	},
+	cli.StringFlag{
+		Name:  "swarm-addr",
+		Usage: "addr to advertise for Swarm (default: detect and use the machine IP)",
+		Value: "",
+	},
+}
+
 var Commands = []cli.Command{
 	{
 		Name:   "active",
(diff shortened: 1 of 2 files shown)


ehazlett:
Thanks for the contribution!  I think I would like to look into the codegangsta subcommands (actually sub-sub commands) instead of the regex.  Would you want to look into that?  I think it would be cleaner and perhaps less chance of breakage.  Thanks!

djfarrelly:
Hey @ehazlett, I appreciate the review! I do agree the regex is not ideal. That's a great idea, I think it might take moving the driver flag and driver-specific flags to a sub-command rather than append all of the driver flag and driver-specific flags to the "create" command. I'd be happy to look further into this! Will report back when I have a new approach.

ehazlett:
Thanks!

hairyhenderson:
Nice, @djfarrelly! Is it possible to also display "global" flags? For example, stuff like `--swarm-discovery` would be useful to see all the time, regardless of which driver's being used.

ehazlett:
@hairyhenderson i think this works with codegangsta

djfarrelly:
**tl;dr** - Using [subcommands](https://github.com/codegangsta/cli#subcommands) turned out to be a log bigger change than I originally thought. Instead I've created a non-regex way to filter the commands and updated this PR.

The new changes will also allow the Swarm flags to remain on the filtered list as proposed by @hairyhenderson :)

-

I spent a bit of time researching & working on Subcommands and ran into an issue:

In order to nest flags with [codegangsta's cli](https://github.com/codegangsta/cli#subcommands), I believe we'd need to make each driver it's own command to attach each driver's respective flags to. The main usage would change to `docker-machine create virtualbox myname`. This feels a lot different as *virtualbox* would be the command. As a side effect each driver could potentially have more simplified flags like `--cpu-count` instead of `--vmwarevcloudair-cpu-count`. Example output:

```
$ docker-machine create virtualbox --help
NAME:
   virtualbox - Create a machine using the virtualbox driver

USAGE:
   command virtualbox [command options] [arguments...]

OPTIONS:
   --virtualbox-memory "1024"		Size of memory for host in MB
   --virtualbox-cpu-count "-1"		number of CPUs for the machine (-1 to use the number of CPUs available) [$VIRTUALBOX_CPU_COUNT]
   --virtualbox-disk-size "20000"	Size of disk for host in MB
   --virtualbox-boot2docker-url 	The URL of the boot2docker image. Defaults to the latest available version [$VIRTUALBOX_BOOT2DOCKER_URL]
```
**Bottom line:** This feels like a much bigger change than a basic filter on flags. I'm not sure it's worth that fundamental change to cleanup the flag list, but I don't have the experience contributing to this and other Docker projects to make that call!

Would love any feedback or thoughts you had @ehazlett. Thanks so much for your time!

hairyhenderson:
@djfarrelly - The current commit looks good to me.

I'm on the fence about going to subcommands - you're right that it's a big change. A potential command line could look like:

```
$ docker-machine create --swarm-master --swarm-discovery token://abcd1234 \
  virtualbox --memory 2048 --cpu-count 4 swarm-master
```

On the one hand, it's nice to have the non-driver-specific `create` options isolated from the driver-specific options. And it involves less typing if the flags can drop the driver name.

On the other hand, it's a big change ;)

ehazlett:
@djfarrelly thanks for the write up and great feedback!  i agree that would be a much larger change and affect the UX quite a bit.  This might not be bad when we move to a plugin model where the subcommand would actually be desired.  For now, I will do some testing around this and filtering sounds good to me given we have 100s of options in total :)

/cc @nathanleclaire thoughts?

djfarrelly:
Awesome thoughts and feedback @hairyhenderson and @ehazlett! Agree it could be pretty cool w/ a plugin model :)

nathanleclaire:
I really like this.  There are a few nitpicky things that I want to change before merging and it might just be easier for me to make a PR on your repo (@djfarrelly), so stay tuned.

djfarrelly:
Sounds great @nathanleclaire, thanks for taking a look at this! I'm new to Go and contributing to Docker projects, looking forward to see what changes you're planning to make :)

djfarrelly:
Are there any rebasing or tweaks I should make for this to be ready?

ehazlett:
@nathanleclaire ping

nathanleclaire:
LGTM

nathanleclaire:
Let's please merge https://github.com/djfarrelly/machine/pull/3 before merging this PR though @djfarrelly thanks!

djfarrelly:
:+1: https://github.com/djfarrelly/machine/pull/3 is merged

ehazlett:
LGTM

djfarrelly:
Thanks guys! Glad I could contribute a bit :smile:

ehazlett:
@djfarrelly you are welcome :)  thanks for contributing!

PR #909, opened 2015-03-29
SSH Client

sthulb:
I'm opening another PR for this since I've split out the code from another PR and rebasing the other PR will take too long.

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -3,7 +3,6 @@ package drivers
 import (
 	"errors"
 	"fmt"
-	"os/exec"
 	"sort"
 
 	log "github.com/Sirupsen/logrus"
@@ -174,21 +173,32 @@ type DriverOptions interface {
 	Bool(key string) bool
 }
 
-func GetSSHCommandFromDriver(d Driver, args ...string) (*exec.Cmd, error) {
+func RunSSHCommandFromDriver(d Driver, args string) (ssh.Output, error) {
+	var output ssh.Output
+
 	host, err := d.GetSSHHostname()
 	if err != nil {
-		return nil, err
+		return output, err
 	}
 
 	port, err := d.GetSSHPort()
 	if err != nil {
-		return nil, err
+		return output, err
 	}
 
 	user := d.GetSSHUsername()
 	keyPath := d.GetSSHKeyPath()
 
-	return ssh.GetSSHCommand(host, port, user, keyPath, args...), nil
+	auth := &ssh.Auth{
+		Keys: []string{keyPath},
+	}
+
+	client, err := ssh.NewClient(user, host, port, auth)
+	if err != nil {
+		return output, err
+	}
+
+	return client.Run(args)
 }
 
 func MachineInState(d Driver, desiredState state.State) func() bool {
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -452,19 +452,17 @@ func (d *Driver) GetIP() (string, error) {
 	if s != state.Running {
 		return "", drivers.ErrHostIsNotRunning
 	}
-	cmd, err := drivers.GetSSHCommandFromDriver(d, "ip addr show dev eth1")
+	output, err := drivers.RunSSHCommandFromDriver(d, "ip addr show dev eth1")
 	if err != nil {
 		return "", err
 	}
 
-	// reset to nil as if using from Host Stdout is already set when using DEBUG
-	cmd.Stdout = nil
-
-	b, err := cmd.Output()
-	if err != nil {
+	var buf bytes.Buffer
+	if _, err := buf.ReadFrom(output.Stdout); err != nil {
 		return "", err
 	}
-	out := string(b)
+
+	out := buf.String()
 	log.Debugf("SSH returned: %s\nEND SSH\n", out)
 	// parse to find: inet 192.168.59.103/24 brd 192.168.59.255 scope global eth1
 	lines := strings.Split(out, "\n")
--- a/libmachine/host.go
+++ b/libmachine/host.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"regexp"
 
@@ -127,23 +126,49 @@ func (h *Host) Create(name string) error {
 	return nil
 }
 
-func (h *Host) GetSSHCommand(args ...string) (*exec.Cmd, error) {
+func (h *Host) RunSSHCommand(command string) (ssh.Output, error) {
+	var output ssh.Output
+
 	addr, err := h.Driver.GetSSHHostname()
 	if err != nil {
-		return nil, err
+		return output, err
 	}
-
-	user := h.Driver.GetSSHUsername()
 
 	port, err := h.Driver.GetSSHPort()
 	if err != nil {
-		return nil, err
+		return output, err
 	}
 
-	keyPath := h.Driver.GetSSHKeyPath()
+	auth := &ssh.Auth{
+		Keys: []string{h.Driver.GetSSHKeyPath()},
+	}
 
-	cmd := ssh.GetSSHCommand(addr, port, user, keyPath, args...)
-	return cmd, nil
+	client, err := ssh.NewClient(h.Driver.GetSSHUsername(), addr, port, auth)
+
+	return client.Run(command)
+}
+
+func (h *Host) CreateSSHShell() error {
+	addr, err := h.Driver.GetSSHHostname()
+	if err != nil {
+		return err
+	}
+
+	port, err := h.Driver.GetSSHPort()
+	if err != nil {
+		return err
+	}
+
+	auth := &ssh.Auth{
+		Keys: []string{h.Driver.GetSSHKeyPath()},
+	}
+
+	client, err := ssh.NewClient(h.Driver.GetSSHUsername(), addr, port, auth)
+	if err != nil {
+		return err
+	}
+
+	return client.Shell()
 }
 
 func (h *Host) Start() error {
@@ -326,13 +351,9 @@ func sshAvailableFunc(h *Host) func() bool {
 			log.Debugf("Error waiting for TCP waiting for SSH: %s", err)
 			return false
 		}
-		cmd, err := h.GetSSHCommand("exit 0")
-		if err != nil {
+
+		if _, err := h.RunSSHCommand("exit 0"); err != nil {
 			log.Debugf("Error getting ssh command 'exit 0' : %s", err)
-			return false
-		}
-		if err := cmd.Run(); err != nil {
-			log.Debugf("Error running ssh command 'exit 0' : %s", err)
 			return false
 		}
 		return true
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"os/exec"
 	"path"
 
 	log "github.com/Sirupsen/logrus"
@@ -12,6 +11,7 @@ import (
 	"github.com/docker/machine/libmachine/auth"
 	"github.com/docker/machine/libmachine/provision/pkgaction"
 	"github.com/docker/machine/libmachine/swarm"
+	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/state"
 	"github.com/docker/machine/utils"
 )
@@ -36,16 +36,13 @@ type Boot2DockerProvisioner struct {
 
 func (provisioner *Boot2DockerProvisioner) Service(name string, action pkgaction.ServiceAction) error {
 	var (
-		cmd *exec.Cmd
 		err error
 	)
-	cmd, err = provisioner.SSHCommand(fmt.Sprintf("sudo /etc/init.d/%s %s", name, action.String()))
-	if err != nil {
+
+	if _, err = provisioner.SSHCommand(fmt.Sprintf("sudo /etc/init.d/%s %s", name, action.String())); err != nil {
 		return err
 	}
-	if err := cmd.Run(); err != nil {
-		return err
-	}
+
 	return nil
 }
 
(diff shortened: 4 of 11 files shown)


ehazlett:
Looking forward to this.  Let me know when it's ready to test!

sthulb:
@ehazlett the SSH command works now. So that bit can be tested.

sthulb:
@ehazlett Fixed... ready to actually test now.

libmachine/host.go
```diff
@@ -127,23 +126,48 @@ func (h *Host) Create(name string) error {
 	return nil
 }
 
-func (h *Host) GetSSHCommand(args ...string) (*exec.Cmd, error) {
+func (h *Host) GetSSHCommand(args string) (*ssh.Output, error) {
```
ahmetalpbalkan: why do we keep this guy around?

libmachine/host.go
```diff
@@ -127,23 +126,48 @@ func (h *Host) Create(name string) error {
...
+	port, err := h.Driver.GetSSHPort()
+	if err != nil {
+		return err
+	}
+
+	auth := &ssh.Auth{
+		Keys: []string{h.Driver.GetSSHKeyPath()},
+	}
+
+	client, err := ssh.NewClient(h.Driver.GetSSHUsername(), addr, port, auth)
```
ahmetalpbalkan: err not checked

drivers/drivers.go
```diff
@@ -174,7 +173,7 @@ type DriverOptions interface {
 	Bool(key string) bool
 }
 
-func GetSSHCommandFromDriver(d Driver, args ...string) (*exec.Cmd, error) {
+func GetSSHCommandFromDriver(d Driver, args string) (*ssh.Output, error) {
```
ahmetalpbalkan: what you're saying is probably `RunSSHCommandOnDriver`, right?

libmachine/provision/boot2docker.go
```diff
@@ -188,8 +180,8 @@ func (provisioner *Boot2DockerProvisioner) Provision(swarmOptions swarm.SwarmOpt
 	return nil
 }
 
-func (provisioner *Boot2DockerProvisioner) SSHCommand(args ...string) (*exec.Cmd, error) {
-	return drivers.GetSSHCommandFromDriver(provisioner.Driver, args...)
+func (provisioner *Boot2DockerProvisioner) SSHCommand(args string) (*ssh.Output, error) {
+	return drivers.GetSSHCommandFromDriver(provisioner.Driver, args)
```
ahmetalpbalkan: This is way too complicated. `SSHCommand` calls `GetSSHCommandFromDriver` which probably actually runs some command but none of these have the word `Run` in it.

Also there are few other occurrences of this one-line methods. I don't think we really need those and they're not really consistently named across drivers.

Also why don't we can't just do `provisioner.Driver.RunSSHCommand(args)` instead of going back to the `drivers` package? Sounds like we can do better.

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+		authMethods = append(authMethods, ssh.Password(p))
+	}
+
+	return &ssh.ClientConfig{
+		User: user,
+		Auth: authMethods,
+	}, nil
+}
+
+func (client *Client) Run(command string) (*Output, error) {
```
ahmetalpbalkan: you don't need `*Output` here, what it holds are already pointers in it.

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/terminal"
```
ahmetalpbalkan: erm this library works quite poorly on windows and we did [some huge work](https://github.com/docker/docker/pulls?q=is%3Apr+ANSI+is%3Aclosed+created%3A%3E%3D2015-02-01+sort%3Acreated-asc) around terminal code in Docker 1.6 to port docker client to windows (and more will coming). 

Basically we need to emulate ANSI sequences on Windows (because cmd.exe/powershell doesn't do it). 

We should be importing `github.com/docker/docker/pkg/term` here instead for a better experience. Because even docker's Unix term package was initially ripped off/forked from this ssh/terminal but they diverged a lot over time as Docker changed it to accommodate the needs. This would also provide consistency across projects.

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+	var stdout, stderr bytes.Buffer
+
+	session.Stdout = &stdout
+	session.Stderr = &stderr
+
+	if err := session.Run(command); err != nil {
+		return nil, err
+	}
+
+	return &Output{
```
ahmetalpbalkan: protip: if you construct `Output{}` above (before `if err:=...`) you can eliminate the if block and just say:

return `output, session.Run(command)

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	session, err := conn.NewSession()
+	if err != nil {
+		return err
+	}
+
+	defer session.Close()
+
+	session.Stdout = os.Stdout
```
ahmetalpbalkan: use struct initializer where possible :)

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+type Auth struct {
+	Passwords []string
+	Keys      []string
+}
+
+type Output struct {
+	Stdout io.Reader
```
ahmetalpbalkan: can be `Stdout, Stderr io.Reader` really

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+
+	session.Wait()
+
+	return nil
+}
+
+type Auth struct {
+	Passwords []string
```
ahmetalpbalkan: Can be `Passwords, Keys []string`

ssh/client.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, err
+	}
+
+	client := &Client{
+		Config:   config,
+		Hostname: host,
+		Port:     port,
+	}
+
+	return client, nil
```
ahmetalpbalkan: you can merge this return statement with initialization above.

ahmetalpbalkan:
I'm super excited about this one, let's move this forward @sthulb :heart:

ssh/client.go
sthulb: It can be, I don't particularly like that syntax though :)

libmachine/host.go
sthulb: Thanks, added  a check.

drivers/virtualbox/virtualbox.go
```diff
@@ -452,19 +452,19 @@ func (d *Driver) GetIP() (string, error) {
...
-	cmd, err := drivers.GetSSHCommandFromDriver(d, "ip addr show dev eth1")
+	output, err := drivers.RunSSHCommandFromDriver(d, "ip addr show dev eth1")
 	if err != nil {
 		return "", err
 	}
 
-	// reset to nil as if using from Host Stdout is already set when using DEBUG
-	cmd.Stdout = nil
+	var buf bytes.Buffer
+	buf.ReadFrom(output.Stdout)
```
ahmetalpbalkan: this guy indeed returns err too, why we never check errs in `ReadFrom` calls?

ahmetalpbalkan: is you intention here to see if you can read something from `output.Stdout`, if so you can probably do:

```
nn, err := io.Read([]byte{1})
// check err
// if nn == 0 { compose some error here. }
```

drivers/virtualbox/virtualbox.go
```diff
@@ -452,19 +452,19 @@ func (d *Driver) GetIP() (string, error) {
...
 
-	b, err := cmd.Output()
-	if err != nil {
+	if len(buf.Bytes()) == 0 {
 		return "", err
```
ahmetalpbalkan: What did you assign to `err` before? :) In case of `== 0` it sounds like you're returning an empty err.

ssh/client.go
sthulb: The code would come out worse if I used `ssh.Session` as an initialiser. `NewSession()` does too much.

sthulb:
@ahmetalpbalkan I've addressed your latest concerns.

@ehazlett Back to testing I guess :)

ssh/client.go
ahmetalpbalkan: Ah my bad I didnt see NewSession above.  Actually if we always set stdin stdout on session maybe move these to NewSession as parameters?

sthulb: NewSession doesn't take any args unfortunately.

ahmetalpbalkan: Hmm. Ok I just brought that up assuming it's our code as well.

sthulb: Unfortunately not... it's from `golang.org/x/crypto/ssh`

ahmetalpbalkan:
@sthulb let me know if you need a windows box to test, I can give one :) my guess is unless we use docker/pkg/term (adding my comment which got lost in a commit below), ssh behavior on windows will probably get worse than what we got today.

> image

sthulb:
@ahmetalpbalkan I've already switched out for the one from docker as you suggested. 

I can use the windows vms from modern.ie or use one from AWS. :) thanks for the offer.

ahmetalpbalkan:
@sthulb ooh this is gonna be interesting.

we have some code to enable ANSI emulation on windows. check `StdStreams` method in `docker/docker` and use the streams it gives you instead of `os.Stdin/os.Stderr` (or you can modify `os.Stdin/os.Stderr` with those in main() or init() of main pkg too)

sthulb:
@ahmetalpbalkan the streams stuff sounds good. I can certainly look into this in another PR. 

I think machine could do with some other changes for this to work. And I'm definitely in favour of better windows support.

sthulb:
Ugh. Again.

ahmetalpbalkan:
you like red buttons  :D :dancer:

sthulb:
GitHub has pretty bad usability on a phone. :)

thaJeztah:
> GitHub has pretty bad usability on a phone. :)

Couldn't agree more. *Huge* buttons to "close" or "merge", but *no* button to edit your comment. To make things worse; (if you're on an iPhone), those buttons are directly beneath the "close" button to dismiss your keyboard.

(Recently pressed on "merge", and it was almost impossible to navigate/scroll away from the "confirm" button :fearful:).

ehazlett:
This is causing a panic on Windows (msysgit w/bash):

```
$ docker-machine_windows-amd64.exe ssh test-vbox
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x2 pc=0x567486]

goroutine 16 [running]:
runtime.panic(0xaeb240, 0xf4e502)
        /usr/src/go/src/pkg/runtime/panic.c:279 +0x11f
github.com/docker/machine/ssh.(*Client).Shell(0xc0820086a0, 0x0, 0x0)
        /go/src/github.com/docker/machine/ssh/client.go:127 +0x476
github.com/docker/machine/libmachine.(*Host).CreateSSHShell(0xc08201ed00, 0x0, 0x0)
        /go/src/github.com/docker/machine/libmachine/host.go:169 +0x2e5
github.com/docker/machine/commands.cmdSsh(0xc082054240)
        /go/src/github.com/docker/machine/commands/ssh.go:66 +0x790
github.com/codegangsta/cli.Command.Run(0xb9a5f0, 0x3, 0x0, 0x0, 0xc64650, 0x2f, 0xca6970, 0x57, 0x0, 0x0, ...)
        /go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:113 +0xeb9
github.com/codegangsta/cli.(*App).Run(0xc08201e270, 0xc082005da0, 0x3, 0x3, 0x0, 0x0)
        /go/src/github.com/docker/machine/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:156 +0xaf1
main.main()
        /go/src/github.com/docker/machine/main.go:73 +0x5aa
```

sthulb:
@ehazlett 
screen shot 2015-04-20 at 16 33 23

ehazlett:
LGTM

/cc @ahmetalpbalkan

drivers/virtualbox/virtualbox.go
```diff
@@ -452,19 +452,26 @@ func (d *Driver) GetIP() (string, error) {
...
 
-	// reset to nil as if using from Host Stdout is already set when using DEBUG
-	cmd.Stdout = nil
+	var buf bytes.Buffer
+	if _, err := buf.ReadFrom(output.Stdout); err != nil {
+		return "", err
+	}
 
-	b, err := cmd.Output()
+	n, err := buf.Read([]byte{1})
```
ahmetalpbalkan: this is not how to initialize a slice of 1 element :) You're creating a slice here with a SOH byte initialized with it.  Please use:

    make([]byte, 1)

ahmetalpbalkan: also I'm still not sure why are we reading from SSH instead of just connecting.

First I thought this is just to check if we can connect to the instance, but turns out this is actually to run the command. Now a buffer of size 1 doesn't make sense.

ssh/client.go
ahmetalpbalkan: do you agree on the comment above @sthulb?

ahmetalpbalkan:
@ehazlett I still have a few outstanding comments. I really didn't dig through the code btw, I just read a few quick passes on my bus rides several times. :)

ehazlett:
@ahmetalpbalkan cool thx!  @sthulb could you address those?  I think after that it should be good.

ahmetalpbalkan:
@sthulb awesome things in that screenshot! thanks for moving this forward. ANSI emulation in windows console should be done once we use the StdStreams from pkg/term (maybe a separate pr?).

sthulb:
@ahmetalpbalkan exactly. Another PR will be incoming with StdStream, once I figure out how to use it everywhere in machine.

ehazlett:
I agree - a second PR for that would be best.

ssh/client.go
sthulb: Yep.. doing so now.

sthulb:
All issues resolved hopefully. :) @ehazlett @ahmetalpbalkan

ehazlett:
LGTM :+1:

libmachine/host.go
ahmetalpbalkan: @sthulb it looks like this one is still missing the err check?

PR #916, opened 2015-03-30
Add the --openstack-insecure option  description of Openstack driver

Fangfenghua:


ggiamarchi:
The build fails because you didn't sign your commits. Please see https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

You should also squash your 3 commits.

PR #918, opened 2015-03-30
dont fail on trailing newline or empty line in ParseOsRelease()

crunchywelch:
if there is a trailing newline (or blank line really), ParseOsRelease() fails when it sees empty values, but we don't really care about that, so let it continue


nathanleclaire:
Nice!

Could I get you to write a unit test case for this and then squash your commits into one please?

crunchywelch:
@nathanleclaire sure!

crunchywelch:
@nathanleclaire gonna actually start over with a new fork, can close this one for now

PR #919, opened 2015-03-30
dont fail on trailing newline or empty line in ParseOsRelease()

crunchywelch:

if there is a trailing newline (or blank line really), ParseOsRelease() fails when it sees empty values, but we don't really care about that, so let it continue. also including a test for centos, which is where this condition was occurring.

### DIFF
--- a/libmachine/provision/os_release.go
+++ b/libmachine/provision/os_release.go
@@ -50,6 +50,10 @@ func (osr *OsRelease) setIfPossible(key, val string) error {
 }
 
 func parseLine(osrLine string) (string, string, error) {
+	if osrLine == "" {
+		return "", "", nil
+	}
+
 	vals := strings.Split(osrLine, "=")
 	if len(vals) != 2 {
 		return "", "", fmt.Errorf("Expected %s to split by '=' char into two strings, instead got %d strings", osrLine, len(vals))
--- a/libmachine/provision/os_release_test.go
+++ b/libmachine/provision/os_release_test.go
@@ -18,7 +18,8 @@ PRETTY_NAME="Ubuntu 14.04 LTS"
 VERSION_ID="14.04"
 HOME_URL="http://www.ubuntu.com/"
 SUPPORT_URL="http://help.ubuntu.com/"
-BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"`)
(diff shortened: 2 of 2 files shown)


nathanleclaire:
Could you please help me understand the logic of this patch a little bit better?  I'm having trouble wrapping my head around it.  If the scanner hits an empty line, the length of the split-ed string should be 0, right?  And we should get an error back?

crunchywelch:
yeah, so the parseLine() function that parses the individual lines returns an error on an empty line as it probably should. But on provisioning where ParseOsRelease() calls parseLine() it fails when it really doesn't need to, which causes the entire install to fail. You'll see in the centos test there is an empty line at the end, which I believe happens on Centos in the os-release file for some reason. Does that make sense?

nathanleclaire:
Hm, OK, I see why this works now.  I think it would be a little more clear to actually add this condition to `parseLine` function instead.  I don't think `parseLine` should return an error at all for empty lines, that's a valid condition which I forgot to account for.  Can you please update it to make it so `parseLine` doesn't return an error for this case instead?

crunchywelch:
Sure, I had considered doing it that way as well anyhow. I'll issue a new
pr in a bit.

On Mon, Mar 30, 2015 at 3:06 PM, Nathan LeClaire <notifications@github.com>
wrote:

> Hm, OK, I see why this works now. I think it would be a little more clear
> to actually add this condition to parseLine function instead. I don't
> think parseLine should return an error at all for empty lines, that's a
> valid condition which I forgot to account for. Can you please update it to
> make it so parseLine doesn't return an error for this case instead?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/machine/pull/919#issuecomment-87793433>.
>



-- 

Aaron Welch
SVP of Product
---
212.933.9785 x301
welch@packet.net

nathanleclaire:
Great, thanks for the persistence.

nathanleclaire:
Code looks good - just need to run gofmt on `libmachine/provision/os_release_test.go` and squash your commits into one signed one please.

crunchywelch:
sorry, need to rebase

nathanleclaire:
LGTM cheers

nathanleclaire:
cc @ehazlett

ehazlett:
LGTM

PR #920, opened 2015-03-30
Provisioner upgrade

nathanleclaire:
cc @ehazlett @sthulb 

This PR implements upgrade feature for the Ubuntu and boot2docker provisioners.  In Ubuntu we run `sudo apt-get upgrade lxc-docker` and for boot2docker we download the latest ISO, then replace the one the machine is currently using with that ISO.

Noteworthy changes:

- `WaitForState` has been pulled into `drivers` instead of `libmachine` in order to make it easier to use cross-package
- The majority of code related to handling moving ISOs around has been moved out of the `virtualbox` driver and into `utils` so it can be used in multiple places.  Come to think of it, maybe I should go update the other local drivers to use this common code as well?
- Adds new docs and integration tests

PTAL

### DIFF
--- a/libmachine/host.go
+++ b/libmachine/host.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/machine/libmachine/auth"
 	"github.com/docker/machine/libmachine/engine"
 	"github.com/docker/machine/libmachine/provision"
+	"github.com/docker/machine/libmachine/provision/pkgaction"
 	"github.com/docker/machine/libmachine/swarm"
 	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/state"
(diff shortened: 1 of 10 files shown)


ehazlett:
Thanks!  Taking a look.

libmachine/host.go
```diff
@@ -167,7 +155,7 @@ func (h *Host) Start() error {
 		return err
 	}
 
-	return utils.WaitFor(h.MachineInState(state.Running))
+	return utils.WaitFor(drivers.MachineInState(h.Driver, state.Running))
```
ehazlett: This is starting to get clunky.  I'm not saying it's bad - it just seems getting a bit much with the nesting.  Not sure what to do.

test/integration/driver-virtualbox.bats
```diff
@@ -316,6 +322,15 @@ findCPUCount() {
   [ "$status" -eq 0  ]
 }
 
+@test "$DRIVER: upgrade should work" {
+  buildMachineWithOldIsoCheckUpgrade
+  [ "$status" -eq 0 ]
```
nathanleclaire: I'm not so great with bats, so I'm open to suggestions of conditionals to more thoroughly test that the upgrade was successfully accomplished if you have ideas.

libmachine/host.go
nathanleclaire: Yeah I agree it's hairy.  Maybe we could have an additional short-hand method available in `utils`, like `utils.WaitForDriverState(h.Driver, state.Running)`.  I just don't want to re-implement "run an action, then wait for the state that action is trying to reach to be fulfilled" in different places.

test/integration/driver-virtualbox.bats
ehazlett: Perhaps using `docker version` with some grep magic to make sure it's the version?

libmachine/host.go
ehazlett: I agree -- I think it's ok for now.

ehazlett:
LGTM

PR #921, opened 2015-03-31
Adding --format/-f option to inspect subcommand

hairyhenderson:
This is inspired by docker's `inspect --format` flag.

For example:

```
$ machine inspect -f "{{json .HostOptions.EngineOptions}}"
{"Dns":null,"GraphDir":"","Ipv6":false,"Labels":null,"LogLevel":"","StorageDriver":"","SelinuxEnabled":false,"TlsCaCert":"","TlsCert":"","TlsKey":"","TlsVerify":false,"RegistryMirror":null}
```

Or:

```
$ machine inspect -f "{{prettyJSON .Driver}}"                   
{
    "MachineName": "dev-fusion",
    "IPAddress": "172.16.62.129",
    "Memory": 1024,
    "DiskSize": 20000,
    "CPUs": 8,
    "ISO": "/Users/hairyhenderson/.docker/machine/machines/dev-fusion/boot2docker-1.5.0-GH747.iso",
    "Boot2DockerURL": "",
    "CaCertPath": "/Users/hairyhenderson/.docker/machine/certs/ca.pem",
    "PrivateKeyPath": "/Users/hairyhenderson/.docker/machine/certs/ca-key.pem",
    "SwarmMaster": false,
    "SwarmHost": "tcp://0.0.0.0:3376",
    "SwarmDiscovery": "",
    "CPUS": 8,
    "SSHUser": "docker",
    "SSHPort": 22
}
```

~~:warning: some stuff that seems like it should work doesn't:~~
```
$ machine inspect -f "{{.Driver.IPAddress}}"   
FATA[0000] template: :1:9: executing "" at <.Driver.IPAddress>: can't evaluate field IPAddress in type drivers.Driver 
```

~~This is because `IPAddress` isn't part of the `drivers.Driver` interface, so technically it makes sense, but it's probably going to be surprising to users. I'm not totally sure how to handle this, but I have a vague idea. I'm open to suggestions :)~~

**Update:** This works now:
```
$ machine inspect -f "{{.Driver.IPAddress}}"   
172.16.62.129
```

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -142,6 +142,10 @@
 			"Rev": "1f0db8c0ee6362470abe06a94e3385927ed72a4b"
 		},
 		{
+			"ImportPath": "github.com/stretchr/testify/assert",
+			"Rev": "e4ec8152c15fc46bd5056ce65997a07c7d415325"
+		},
+		{
 			"ImportPath": "github.com/tent/http-link-go",
 			"Rev": "ac974c61c2f990f4115b119354b5e0b47550e888"
 		},
--- a/Godeps/_workspace/src/github.com/stretchr/testify/assert/assertions.go
+++ b/Godeps/_workspace/src/github.com/stretchr/testify/assert/assertions.go
@@ -1,0 +1,805 @@
+package assert
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"reflect"
+	"regexp"
+	"runtime"
+	"strings"
+	"time"
+)
+
+// TestingT is an interface wrapper around *testing.T
+type TestingT interface {
+	Errorf(format string, args ...interface{})
+}
+
+// Comparison a custom function that returns true on success and false on failure
+type Comparison func() (success bool)
+
+/*
+	Helper functions
+*/
+
+// ObjectsAreEqual determines if two objects are considered equal.
+//
+// This function does no assertion of any kind.
+func ObjectsAreEqual(expected, actual interface{}) bool {
+
+	if expected == nil || actual == nil {
+		return expected == actual
+	}
+
+	if reflect.DeepEqual(expected, actual) {
+		return true
+	}
+
+	// Last ditch effort
+	if fmt.Sprintf("%#v", expected) == fmt.Sprintf("%#v", actual) {
+		return true
+	}
+
+	return false
+
+}
+
+func ObjectsAreEqualValues(expected, actual interface{}) bool {
+	if ObjectsAreEqual(expected, actual) {
+		return true
+	}
+
+	actualType := reflect.TypeOf(actual)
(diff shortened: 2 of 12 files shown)


sthulb:
I guess for the IP address stuff you'd have to dynamically add the IP address from `GetIP` to the JSON.

I'm unsure about adding a testing lib though.

hairyhenderson:
@sthulb - hrm, yeah. I was thinking I could also `json.Marshal()`/`json.Unmarshal()` - that might work too.

re: adding the `testify` dependency, @ehazlett mentioned he was OK with that last night. Any specific reason this would be a bad idea? Seems to me making it easier to write tests would be a bonus, given the current state of coverage :)

sthulb:
My  issue  is that 2% of tests are  using it :) I guess another PR converting the 98% is needed

ehazlett:
This seems pretty good.  I cannot get the IP though.  Returning `<no value>` for me for a running machine.

ehazlett:
ping @hairyhenderson

hairyhenderson:
@ehazlett - Hmm. I must've been testing with vmware-fusion and not virtualbox. I can see it with fusion, but I also see `<no value>` with virtualbox. Though I can't see the IPAddress at all with the vanilla `machine inspect`, so maybe something's up with the virtualbox driver... I'll dig into it :)

hairyhenderson:
We've chatted a bit about this in IRC, but the short story is - virtualbox.Driver doesn't actually _have_ a IPAddress field, so it's totally expected behaviour...

ehazlett:
Ah ok this makes sense.  I think it will be odd though to have some drivers return IPs and some not.

hairyhenderson:
@ehazlett - true. but I think that's more something that should be harmonized among all the drivers... This PR is really just about templating the `inspect` output...

ehazlett:
@hairyhenderson valid point.  LGTM

ehazlett:
ping @nathanleclaire

hairyhenderson:
Ok, this is rebased after #923 - PTAL @ehazlett & @nathanleclaire :)

ehazlett:
LGTM

nathanleclaire:
Just a few things:

1. Can we please call the function `prettyjson` instead of `prettyJSON`?  I think it would be slightly more idiomatic for Go template functions.
2. After discussion, @ehazlett and I are both OK with the introduction of the additional testing library, but in the future please introduce such a dependency in a separate PR with a proposal (it doesn't have to be big and elaborate, just some motivations and potential concerns).

hairyhenderson:
@nathanleclaire - sure, I'll rename `prettyJSON` to `prettyjson`.

No problem - wasn't totally sure about how to go about introducing a dependency. A separate PR makes a lot of sense - will do that in future :)

Thanks!

nathanleclaire:
LGTM

PR #923, opened 2015-03-31
Refactoring commands.go into new commands package

hairyhenderson:
`commands.go` is pretty massive, so it might be nice to split it into a bunch of smaller files in a separate `commands` package.


nathanleclaire:
I like this idea in general, however IMO we should hold off until after the release to merge this though (just skittish about big changes).

I'd definitely be into the idea of calling this package `cli` and using it to gradually consolidate all of the flag/CLI-related logic that we have littered through the code right now.

hairyhenderson:
@nathanleclaire - oh, this definitely should be post-0.2.0.

I like the idea of `cli`, if for no other reason than it's 62% shorter ;) I'll rename packages and work on updating this...

ehazlett:
I like this.  We won't merge this for 0.2.0 but still is awesome to split it out -- nice work!

hairyhenderson:
@nathanleclaire - actually, `cli` conflicts with `codegangsta/cli` in a lot of places. Not unsurmountable, but unnecessary ugly IMO. What do you think of a two-phase approach where we refactor to a `commands` package first, then isolate all the CLI-related stuff into it later on and think about renaming to `cli` at that point?

ehazlett:
@hairyhenderson +1

ehazlett:
This doesn't appear to be passing some things.  Getting this error upon create:

```
Creating dynamic image with size 0 bytes (0MB)
```

hairyhenderson:
@ehazlett - hrm. That's... not good... I'll dig into it!

nathanleclaire:
> actually, cli conflicts with codegangsta/cli in a lot of places. Not unsurmountable, but unnecessary ugly IMO. What do you think of a two-phase approach where we refactor to a commands package first, then isolate all the CLI-related stuff into it later on and think about renaming to cli at that point?

Ah, right, always forget about that.  That sounds good to me.

hairyhenderson:
@ehazlett - finally getting around to checking this out - I can reproduce the error.

hairyhenderson:
@ehazlett - it's fixed now :)

ehazlett:
Tested all commands.  Works as expected.

LGTM

/cc @nathanleclaire can you review as well since this is a significant change?

nathanleclaire:
LGTM

ehazlett:
LGTM

nathanleclaire:
:+1: Thanks for the persistence on this @hairyhenderson

PR #934, opened 2015-03-31
driver doc updates

ehazlett:
This updates the driver docs with the latest changes and an overview on driver creation and integration.

### DIFF
--- a/docs/DRIVER_SPEC.md
+++ b/docs/DRIVER_SPEC.md
@@ -85,12 +85,13 @@ will need to make sure it is compliant with the Docker license terms (non-GPL).
 For more information, contact a project maintainer.
 
 # Implementation
-The following describes what is needed to create a Machine Driver.
+The following describes what is needed to create a Machine Driver.  The driver
+interface has methods that must be implemented for all drivers.  These include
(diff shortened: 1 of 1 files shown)


nathanleclaire:
If we update the Driver interface in the future (likely) we'll have to also update this document, so perhaps we should link to the relevant file instead (at least for the interface bits).

I think I'd also introduce the interface first, and then follow up with a small explanation that "And in order to provide this functionality most drivers end up as a struct that implements these methods with parameters similar to the following", since it's not strictly necessary to have those exact fields.  

Also, tangentially, since those fields are very common, we could probably provide a `struct` that drivers can embed which has them like you mentioned in that one issue, and recommend that as a way to keep things consistent across drivers.

ehazlett:
Yeah I was going to just have the link but then figured it would be good to have the docs in one place.  I'll update to just have the link as our docs are out of sync now in places and I'm sure this would get worse.

I like the idea of embedding too.  I will leave that out of this doc update though.

Thanks

ehazlett:
@nathanleclaire This has been updated - PTAL

nathanleclaire:
LGTM thanks for doing this.  I want to do some focusing on docs today as well.

ehazlett:
Thanks

PR #955, opened 2015-04-03
Keystone v3 domains

### LINKED ISSUE #679: Openstack Identity v3 support
Gophercloud supports the Openstack Identity v3 API, wich brings in a lot of new features. It has been the preferred API since Openstack Havana (2013), and the v2 API is somewhat deprecated.

http://gophercloud.io/docs/identity/v3/

Could you look into supporting the new auth mechanisms?

If you need access to an Openstack cloud with V3 API for testing I can arrange that.

We only support the v3 API since we are using domains and other related features.

For the Docker Machine use case, the v3 API is mostly used for acquiring a project scoped authtoken to talk to the nova, neutron etc. APIs.

ggiamarchi:
Fix #679 

~~__NB.__~~

> ~~This PR should not yet be merged. It needs https://github.com/rackspace/gophercloud/pull/408~~
> ~~At the moment, i have updated the gophercloud code in the godep workspace to use the code corresponding to my gophercloud PR~~

Most of the work is done in Gophercloud. Here, i have just added flags to handle keystone domains in the docker-machine CLI.


ehazlett:
LGTM

PR #958, opened 2015-04-04
New Driver: Vultr

janeczku:
Vultr driver updated to the new driver specs. Replaces PR #538 by @JamesClonk.     
    
* Conforms with new driver specs
* Added flags for ipv6, automatic backups and private networking
* Added pre-create validations for regionID, planID and API credentials
* Added bats integration tests
* Added unit tests
* Added driver documentation
* Added client request throttling to avoid hitting API rate-limit
    
bats-test-result
    
### DIFF
--- a/drivers/vultr/vultr.go
+++ b/drivers/vultr/vultr.go
@@ -1,0 +1,519 @@
+package vultr
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/ChimeraCoder/tokenbucket"
+	vultr "github.com/JamesClonk/vultr/lib"
+	"github.com/codegangsta/cli"
+	"github.com/docker/machine/drivers"
+	"github.com/docker/machine/log"
+	"github.com/docker/machine/ssh"
+	"github.com/docker/machine/state"
+)
+
+var vultrTokenBucket = tokenbucket.NewBucket(1*time.Second, 1)
+
+type Driver struct {
+	APIKey            string
+	MachineID         string
+	MachineName       string
+	IPAddress         string
+	PrivateIP         string
+	OSID              int
+	RegionID          int
+	PlanID            int
+	SSHKeyID          string
+	SSHUser           string
+	SSHPort           int
+	IPv6              bool
+	Backups           bool
+	PrivateNetworking bool
+	CaCertPath        string
+	PrivateKeyPath    string
+	DriverKeyPath     string
+	SwarmMaster       bool
+	SwarmHost         string
+	SwarmDiscovery    string
+	storePath         string
+	scriptID          int
+	bucket            *tokenbucket.Bucket
+}
+
+func init() {
+	drivers.Register("vultr", &drivers.RegisteredDriver{
+		New:            NewDriver,
+		GetCreateFlags: GetCreateFlags,
+	})
+}
+
+// GetCreateFlags registers the flags this driver adds to "docker hosts create"
+func GetCreateFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			EnvVar: "VULTR_API_KEY",
+			Name:   "vultr-api-key",
+			Usage:  "Vultr API key",
+		},
+		cli.IntFlag{
+			EnvVar: "VULTR_OS",
+			Name:   "vultr-os-id",
+			Usage:  "Vultr operating system ID. Default: 160 (ubuntu-14-04-x64)",
+			Value:  160,
+		},
+		cli.IntFlag{
+			EnvVar: "VULTR_REGION",
+			Name:   "vultr-region-id",
+			Usage:  "Vultr region ID. Default: 1 (New Jersey)",
+			Value:  1,
+		},
+		cli.IntFlag{
+			EnvVar: "VULTR_PLAN",
+			Name:   "vultr-plan-id",
+			Usage:  "Vultr plan ID. Default: 29 (768 MB RAM)",
+			Value:  29,
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_IPV6",
+			Name:   "vultr-ipv6",
+			Usage:  "Enable IPv6 for VPS",
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_PRIVATE_NETWORKING",
+			Name:   "vultr-private-networking",
+			Usage:  "Enable private networking for VPS",
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_BACKUPS",
+			Name:   "vultr-backups",
+			Usage:  "Enable automatic backups for VPS",
+		},
+	}
+}
+
+func NewDriver(machineName string, storePath string, caCert string, privateKey string) (drivers.Driver, error) {
+	d := &Driver{MachineName: machineName, storePath: storePath, CaCertPath: caCert, PrivateKeyPath: privateKey, bucket: nil}
+	// throttle API queries according to 1/s rate-limit
+	d.bucket = vultrTokenBucket
+	return d, nil
+}
+
+func (d *Driver) AuthorizePort(ports []*drivers.Port) error {
+	return nil
+}
+
+func (d *Driver) DeauthorizePort(ports []*drivers.Port) error {
+	return nil
+}
+
+func (d *Driver) GetMachineName() string {
+	return d.MachineName
+}
+
+func (d *Driver) GetSSHHostname() (string, error) {
+	return d.GetIP()
+}
+
+func (d *Driver) GetSSHKeyPath() string {
+	return filepath.Join(d.storePath, "id_rsa")
+}
+
+func (d *Driver) GetSSHPort() (int, error) {
+	if d.SSHPort == 0 {
+		d.SSHPort = 22
+	}
+
+	return d.SSHPort, nil
+}
+
+func (d *Driver) GetSSHUsername() string {
+	if d.SSHUser == "" {
+		d.SSHUser = "root"
+	}
+
+	return d.SSHUser
+}
+
+func (d *Driver) DriverName() string {
+	return "vultr"
+}
+
+func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
+	d.APIKey = flags.String("vultr-api-key")
+	d.OSID = flags.Int("vultr-os-id")
+	d.RegionID = flags.Int("vultr-region-id")
+	d.PlanID = flags.Int("vultr-plan-id")
+	d.IPv6 = flags.Bool("vultr-ipv6")
+	d.PrivateNetworking = flags.Bool("vultr-private-networking")
+	d.Backups = flags.Bool("vultr-backups")
+	d.SwarmMaster = flags.Bool("swarm-master")
+	d.SwarmHost = flags.String("swarm-host")
+	d.SwarmDiscovery = flags.String("swarm-discovery")
+	d.SSHUser = "root"
+	d.SSHPort = 22
+
+	if d.APIKey == "" {
+		return fmt.Errorf("vultr driver requires the --vultr-api-key option")
+	}
+	return nil
+}
+
+func (d *Driver) PreCreateCheck() error {
+
+	log.Info("Validating Vultr VPS parameters...")
+
+	if err := d.validateRegion(); err != nil {
+		return err
+	}
+
+	if err := d.validatePlan(); err != nil {
+		return err
+	}
+
+	if err := d.validateApiCredentials(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d *Driver) Create() error {
+	var userdata string
+	log.Debug("Generating SSH key...")
+	key, err := d.createSSHKey()
+	if err != nil {
+		return err
+	}
+	d.SSHKeyID = key.ID
+
+	log.Info("Creating Vultr VPS...")
+	// RancherOS iPXE boot prep
+	if d.OSID == 159 {
+		log.Info("Deploying RancherOS")
+		d.SSHUser = "rancher"
+		if err := d.createBootScript(); err != nil {
+			return err
+		}
+		log.Debugf("Uploaded iPXE boot script ID %d", d.scriptID)
+		userdata, err = d.getCloudConfig()
+		if err != nil {
+			return err
+		}
+		log.Debugf("Using the following cloud-config file:")
+		log.Debugf("%s", userdata)
+	}
+
+	client := d.getClient()
+	<-d.bucket.SpendToken(1)
+	machine, err := client.CreateServer(
+		d.MachineName,
+		d.RegionID,
+		d.PlanID,
+		d.OSID,
+		&vultr.ServerOptions{
+			SSHKey:            d.SSHKeyID,
+			IPV6:              d.IPv6,
+			PrivateNetworking: d.PrivateNetworking,
+			AutoBackups:       d.Backups,
+			Script:            d.scriptID,
+			UserData:          userdata,
+		})
+	if err != nil {
+		return err
+	}
+	d.MachineID = machine.ID
+
+	log.Info("Waiting for IP address to become available...")
+	for {
+		<-d.bucket.SpendToken(3)
+		machine, err = client.GetServer(d.MachineID)
+		if err != nil {
+			return err
+		}
+		d.IPAddress = machine.MainIP
+		d.PrivateIP = machine.InternalIP
+
+		if d.IPAddress != "" && d.IPAddress != "0" {
+			break
+		}
+		log.Debug("IP address not yet available")
+	}
+
+	if d.PrivateIP == "0" {
+		d.PrivateIP = ""
+	}
+
+	log.Infof("Created Vultr VPS ID: %s, Public IP: %s, Private IP: %s",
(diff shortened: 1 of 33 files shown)


hairyhenderson:
@janeczku - I'd like to try this out, can you rebase it please? :)

janeczku:
@hairyhenderson Can do - probably tomorrow.

janeczku:
Rebased the PR. Please test to break :) @hairyhenderson

hairyhenderson:
Thanks @janeczku!

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/docker/machine/ssh"
+	"github.com/docker/machine/state"
+)
+
+type Driver struct {
+	APIKey            string
+	MachineID         string
+	MachineName       string
+	PublicIP          string
+	PrivateIP         string
```
hairyhenderson: Can you add an `IPAddress` property too? It should be the same as `PublicIP`, or `PrivateIP` if `PrivateNetworking` is `true`. (see #1043)

hairyhenderson: I misunderstood `PrivateNetworking` :)... So, can you _either_ add `IPAddress` which is in sync with `PublicIP`, or can you rename `PublicIP` to `IPAddress`?

hairyhenderson:
Ok, so here's an issue, but I don't know if it's in this code or if has to do with some of the recent SSH changes:

```
$ machine create -d vultr --vultr-api-key XXX dm-vultr
Creating SSH key...
Creating Vultr virtual machine...
Error dialing TCP: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
Error dialing TCP: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
Error dialing TCP: ssh: handshake failed: ssh: unable to authenticate, attempted methods [publickey none], no supported methods remain
Error dialing TCP: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
Error dialing TCP: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
... (repeat a billion times)
To see how to connect Docker to this machine, run: machine env dm-vultr
$
```

It _appears_ that the resulting VM is working fine, so it might be nothing.

/cc @nathanleclaire @ehazlett @sthulb - I vaguely recall something recently that might be related?

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,362 @@
...
+		d.PublicIP = machine.MainIP
+		d.PrivateIP = machine.InternalIP
+
+		if d.PublicIP != "" && d.PublicIP != "0" {
+			break
+		}
+	}
+
+	if d.PrivateIP == "" || d.PrivateIP == "0" {
+		d.PrivateIP = "N/A"
```
hairyhenderson: Any reason this is `N/A` instead of empty string? The convention in other drivers is that it gets set to `""`

hairyhenderson:
@janeczku - looks like the start command is asynchronous. Can you change `Start()` (and probably `Stop()` too) so that it waits until the machine is reachable before returning? I _think_ most other drivers behave this way.

```
$ machine -D stop dm-vultr
command=stop machine=dm-vultr
$ machine -D start dm-vultr
command=start machine=dm-vultr
$ machine ls
NAME         ACTIVE   DRIVER         STATE     URL                        SWARM
dm-vultr     *        vultr          Running   tcp://45.63.7.71:2376      
$ machine -D ssh dm-vultr
dial tcp 45.63.7.71:22: connection refused
$ machine -D ssh dm-vultr
dial tcp 45.63.7.71:22: connection refused
...
$ machine -D ssh dm-vultr
Welcome to Ubuntu 14.04.2 LTS (GNU/Linux 3.13.0-52-generic x86_64)
...
```

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,362 @@
...
+		cli.IntFlag{
+			EnvVar: "VULTR_PLAN",
+			Name:   "vultr-plan-id",
+			Usage:  "Vultr plan ID (VPSPLANID). Default: 29 (768 MB RAM, 15 GB SSD, 1.00 TB BW)",
+			Value:  29,
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_IPV6",
+			Name:   "vultr-ipv6",
+			Usage:  "enable ipv6 for droplet",
```
hairyhenderson: Maybe:
```
Usage: "Enable IPv6 for VPS",
```

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,362 @@
...
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_PRIVATE_NETWORKING",
+			Name:   "vultr-private-networking",
+			Usage:  "enable private networking for this virtual machine",
```
hairyhenderson: Should probably say "VPS" instead of "virtual machine"

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,362 @@
...
+		},
+		cli.BoolFlag{
+			EnvVar: "VULTR_BACKUPS",
+			Name:   "vultr-backups",
+			Usage:  "enable automatic backups for this virtual machine",
```
hairyhenderson: Should probably say "VPS" instead of "virtual machine"

hairyhenderson:
@janeczku - overall, this is looking pretty good. Might be nice too to add a bit more info-level logging, especially when long operations (during `create` for example) are happening.

drivers/vultr/vultr.go
janeczku: I think its best to just rename `PublicIP`. `IPAdress` is expected to be public anyway.

hairyhenderson: :+1:

drivers/vultr/vultr.go
janeczku: yes indeed, it should be an empty string if the IP is not set.

ehazlett: @hairyhenderson i've seen this when you use an invalid username.

drivers/vultr/vultr.go
janeczku: Done.

drivers/vultr/vultr.go
janeczku: OK

janeczku:
@hairyhenderson thanks for your valuable observations!

> looks like the start command is asynchronous. Can you change Start() (and probably Stop() too) so that it waits until the machine is reachable before returning? I think most other drivers behave this way.

They should :) But at least the DO driver respectively its API package has asynchronous start/stop commands as well. I will implement this in the Vultr driver and since they share most code we can then easily add this change to the DO driver as well.

>Ok, so here's an issue, but I don't know if it's in this code or if has to do with some of the recent SSH changes: Error dialing TCP: ssh: handshake failed ....

I have investigated this. It's not really the driver's problem. This error occurs when the SSHKey has not yet been provisioned to the new instance (but the SSHd is already running). It takes up to 4 minutes until the Vultr instance is fully provisioned. One basically just needs to bump up the `utils.WaitFor()` timeout a minute:

```Go
func WaitFor(f func() bool) error {
	return WaitForSpecific(f, 90, 3*time.Second)
}
```

janeczku:
Merged reviewers suggestions, rebased again and squashed.

hairyhenderson:
@janeczku looks like you'll need to rebase _again_

janeczku:
Hey @hairyhenderson, would you take a final look at the patch and (hopefully) give your OK? I rebased again and added a tiny bit of code making `RancherOS` the default OS for new instances.

/cc @ibuildthecloud @imikushin
[Test builds](https://github.com/janeczku/machine/releases/tag/v0.3.0-vultr) / [Free trial for VULTR](https://www.vultr.com/freetrial/)

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,517 @@
...
+		return err
+	}
+	for _, v := range plans {
+		if v == d.PlanID {
+			return nil
+		}
+	}
+	return fmt.Errorf("Plan ID %d not available in this region. Available plans: %v", d.PlanID, plans)
+}
+
```
janeczku: @hairyhenderson FYI, the code below has been added since you last took a look (to allow booting RancherOS).

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,517 @@
...
+	if err != nil {
+		return err
+	}
+	d.scriptID, err = strconv.Atoi(script.ID)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
```
janeczku: @hairyhenderson FYI, these two functions have been added after you last took a look (to allow booting RancherOS).

hairyhenderson:
> Hey @hairyhenderson, would you take a final look at the patch and (hopefully) give your OK? I rebased again and added a tiny bit of code making RancherOS the default OS for new instances.

Yep! I'll have a look. Might need to wait until Monday or Tuesday, but I'll get to this :)

> Test builds / Free trial for VULTR

Thanks for the build! And, too bad the free credit is only for new users ;)

janeczku:
@hairyhenderson Thank you!

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,517 @@
...
+			return nil
+		}
+	}
+	return fmt.Errorf("Plan ID %d not available in this region. Available plans: %v", d.PlanID, plans)
+}
+
+// RancherOS - Create iPXE boot script
+func (d *Driver) createBootScript() error {
+	content := `#!ipxe
+set base-url https://releases.rancher.com/os/latest
```
imikushin: Would `https:` work w/ iPXE?

janeczku: @imikushin Not in every environment. HTTPS support has to be specifically enabled when compiling the iPXE firmware. So for Vultr HTTPS works, for other iPXE environments one might need to use plain HTTP.

janeczku:
ping @hairyhenderson  :smirk:

hairyhenderson:
@janeczku - hey, sorry, I've been super-busy with a new job and a new baby (born yesterday!) so it might be a little while before I can take a good look at this... In the meantime, looks like there's some merge conflicts again? I'll try to take a look in the next few days :)

janeczku:
> hey, sorry, I've been super-busy with a new job and a new baby (born yesterday!)
@hairyhenderson, thanks for the feedback and congrats!

drivers/vultr/vultr.go
```diff
@@ -0,0 +1,517 @@
...
+		cli.StringFlag{
+			EnvVar: "VULTR_API_KEY",
+			Name:   "vultr-api-key",
+			Usage:  "Vultr API key",
+		},
+		cli.IntFlag{
+			EnvVar: "VULTR_OS",
+			Name:   "vultr-os-id",
+			Usage:  "Vultr operating system ID. Default: 159 (RancherOS)",
+			Value:  159,
```
hairyhenderson: So... While RancherOS is definitely cool, and is considered "supported" from docker-machine's point of view, the default for remote drivers (like Vultr) is Ubuntu (see https://github.com/docker/machine/blob/master/docs/index.md#specify-a-base-operating-systems). So IMO you should change the default to Ubuntu (probably 14.04 - i.e. `160`), just to keep in line with what everyone expects and assumes.

imikushin: Not that I disagree, but...
RancherOS can be configured to look like Ubuntu :)
It's just a matter of what console container to use. Currently, RancherOS
does use ubuntu-console by default in cloud deployments.

Just my 2 cents.

hairyhenderson:
@janeczku - so far this looks pretty good! Just the one thing about the default for `--vultr-os-id`, then IMO this is ready to merge :)

/cc @ehazlett @nathanleclaire

drivers/vultr/vultr.go
hairyhenderson: Yup, sure. But even though it might _look_ like Ubuntu, it's not Ubuntu ;) I really think all of the remote drivers need to have the same default image.

janeczku: @hairyhenderson I have changed the default OS back to Ubuntu. Since the iPXE deployment of RancherOS is the better choice for Vultr (deploying Ubuntu takes up to 5 minutes) i have added instructions to the docs concerning the possibility to use RancherOS as base OS.

/cc @imikushin

janeczku:
@ehazlett @nathanleclaire you wanna chime in?  i'd be happy to finally see this merged :)

djmaze:
Thanks for your work @janeczku! I just tried out this fork with my Vultr account. It does not matter if I choose RancherOS or Ubuntu, `docker-machine create` always fails like this:

```
Validating Vultr VPS parameters...
Creating Vultr VPS...
Waiting for IP address to become available...
Error creating machine: json: cannot unmarshal string into Go value of type float64
```

Looking at my Vultr console, the servers have been created successfully. Also, I am able to login using the SSH keys created by docker-machine (in `~/.docker/machine/...`). 

Any idea what's going wrong here? How to debug this?

janeczku:
@djmaze thanks for your feedback. This is caused by recent changes to Vultr's API protocol. It is already fixed in the Vultr API client and i will push it to the machine driver later today.

janeczku:
@djmaze this is now fixed. You can try the dev builds available here: https://github.com/janeczku/machine/releases/tag/v0.4.0-dev
Please report back if you find any other issues!

djmaze:
@janeczku Great, it works now. Thanks for the update!

djmaze:
After creating multiple machines on Vulr, `docker-machine ls` gives the following:

```
error getting state for host docker3: Rate limit reached - please try your request again later.  Current rate limit: 1 request/sec
```

It seems rate limiting does not work as needed.

janeczku:
@djmaze Thanks for your feedback. The rate limit error is due to the `ls` command querying the state of the hosts asynchronously (dispatching a goroutine for every host). The driver has no way to throttle these concurrent queries. This has to be addressed upstream in the libmachine code: https://github.com/docker/machine/blob/master/libmachine/host.go#L381

djmaze:
Excuse my limited Go knowledge. The tokenbucket library uses a mutex. Shouldn't this work across goroutines already?

Ah I see, there is one driver instance per machine, so one bucket per host. That's unfortunate ;/

janeczku:
@djmaze You got that right. Vultr really needs to change their 1/sec rate limit policy to something more reasonable like x/minute or x/hour.

djmaze:
So forgive me if this is stupid, but I just made the token bucket global:

```diff
diff --git a/drivers/vultr/vultr.go b/drivers/vultr/vultr.go
index 11249b7..98a3b04 100644
--- a/drivers/vultr/vultr.go
+++ b/drivers/vultr/vultr.go
@@ -96,10 +96,12 @@ func GetCreateFlags() []cli.Flag {
        }
 }
 
+var vultrTokenBucket = tokenbucket.NewBucket(1*time.Second, 1)
+
 func NewDriver(machineName string, storePath string, caCert string, privateKey string) (drivers.Driver, error) {
        d := &Driver{MachineName: machineName, storePath: storePath, CaCertPath: caCert, PrivateKeyPath: privateKey, bucket: nil}
        // throttle API queries according to 1/s rate-limit
-       d.bucket = tokenbucket.NewBucket(1*time.Second, 1)
+       d.bucket = vultrTokenBucket
        return d, nil
 }
```

This makes the `ls` command work for me. The tockenbucket code itself should be safe from parallel use in goroutines because it uses mutex and channels, right? 

Of course we can do better and initialize the bucket on first use instead (using a mutex). Apart from that, is there anything wrong with this change?

janeczku:
Thats interesting! I would not have expected a single tokenbucket to work across multiple goroutines. It was my believe that when you have multiple receivers (`<-d.bucket.SpendToken(1)`) on the same channel,  they would all unblock once there is a single transmit on that channel and therefore breaking the throttling mechanism. It turns out thats not the case (with Golang channels there is only one active receiver at a time, the other ones are "waiting in line").
Again, great find :+1:

janeczku:
Confirmed working 1req/s API query throttling.

> Janek$ ./docker-machine_darwin-amd64 -D ls
GetState() timestamp: 2015-07-21 00:45:**38**.064903604 +0200 CEST
GetState() timestamp: 2015-07-21 00:45:**39**.06484465 +0200 CEST
GetState() timestamp: 2015-07-21 00:45:**40**.064840547 +0200 CEST
GetState() timestamp: 2015-07-21 00:45:**41**.06476442 +0200 CEST

djmaze:
Unfortunately, now I am not able to build machines based on RancherOS with the Vultr driver. This is the output when using `--vulr-os-id=159`:

```
Validating Vultr VPS parameters...
Creating Vultr VPS...
Deploying RancherOS
Error creating machine: json: cannot unmarshal number into Go value of type string
```
With the default Ubuntu OS the installation works.

djmaze:
Got it working again by making [SCRIPTID](https://github.com/JamesClonk/vultr/blob/master/lib/scripts.go#L7) an `int` and adjusting the rest of the code accordingly. So again, this has to be fixed upstream first.

janeczku:
@djmaze Pushed fix to the Vultr API client repo. It was not enough to change the SCRIPTID field to integer since the API continues to return this field as a string in other calls. See here: https://github.com/JamesClonk/vultr/pull/7/files

ehazlett:

@ehazlett & @nathanleclaire

PR #960, opened 2015-04-04
for ubuntu provision append hostname to 127.0.1.1 line in /etc/hosts

crunchywelch:
we should do it this way instead of appending a record for 127.0.0.1 to the file itself. Two reasons for this:

1) you should not have multiple lines in /etc/hosts with the same ip address
2) ubuntu/debian use 127.0.1.1 for this purpose as described here:

https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

and further discussed here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=316099

### DIFF
--- a/libmachine/provision/ubuntu.go
+++ b/libmachine/provision/ubuntu.go
@@ -128,8 +128,16 @@ func (provisioner *UbuntuProvisioner) Hostname() (string, error) {
 
 func (provisioner *UbuntuProvisioner) SetHostname(hostname string) error {
 	if _, err := provisioner.SSHCommand(fmt.Sprintf(
-		"sudo hostname %s && echo %q | sudo tee /etc/hostname && echo \"127.0.0.1 %s\" | sudo tee -a /etc/hosts",
+		"sudo hostname %s && echo %q | sudo tee /etc/hostname",
 		hostname,
+		hostname,
+	)); err != nil {
+		return err
+	}
+
+	// ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution
+	if _, err := provisioner.SSHCommand(fmt.Sprintf(
+		"if grep -xq 127.0.1.1.* /etc/hosts; then sudo sed -i 's/^127.0.1.1.*/127.0.1.1 %s/g' /etc/hosts; else echo '127.0.1.1 %s' | sudo tee -a /etc/hosts; fi",
 		hostname,
 		hostname,
 	)); err != nil {
--- a/test/integration/driver-digitalocean.bats
+++ b/test/integration/driver-digitalocean.bats
@@ -52,6 +52,12 @@ export MACHINE_STORAGE_PATH=/tmp/machine-bats-test-$DRIVER
   run machine ssh $NAME -- docker version
 }
 
+# currently this only checks debian/ubuntu style using 127.0.0.1
+@test "$DRIVER: hostname should be set properly" {
+  run machine ssh $NAME -- "grep -Fxq '127.0.1.1 $NAME' /etc/hosts"
+  [ "$status" -eq 0  ]
+}
+
 @test "$DRIVER: stop" {
   run machine stop $NAME
   [ "$status" -eq 0  ]


nathanleclaire:
I'm for it.

Could I get you to add an integration test which checks for this after create please?

crunchywelch:
@nathanleclaire I'll give it a shot :)

crunchywelch:
@nathanleclaire ok, how does that look? for some reason the busybox test was failing for me, but maybe that's because I'm on a plane and the wifi is spotty, the rest worked just fine...

libmachine/provision/ubuntu.go
```diff
@@ -124,7 +124,7 @@ func (provisioner *UbuntuProvisioner) Hostname() (string, error) {
 
 func (provisioner *UbuntuProvisioner) SetHostname(hostname string) error {
 	cmd, err := provisioner.SSHCommand(fmt.Sprintf(
-		"sudo hostname %s && echo %q | sudo tee /etc/hostname && echo \"127.0.0.1 %s\" | sudo tee -a /etc/hosts",
+		"sudo hostname %s && echo %q | sudo tee /etc/hostname && sudo sed -i 's/^127.0.1.1.*/127.0.1.1 %s/g' /etc/hosts",
```
nathanleclaire: @crunchywelch Why the additional `sed` command instead of just replacing 127.0.0.1 with 127.0.1.1?

crunchywelch: Because we want to keep the 127.0.0.1 entry, ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames:

https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

crunchywelch:
@nathanleclaire does the test look ok?

nathanleclaire:
@crunchywelch Yep, looks good - thanks for following up.  Just one last request: Could you please split the bit after the `&&` into a separate `SSHCommand` command?  I think it will make the code more readable.  This will be LGTM for me after that.

nathanleclaire:
Oh, and also a comment inline explaining the motivation (similar to what you have here on the inline code commentary) would be fantastic!

crunchywelch:
@nathanleclaire okies, here you go!

crunchywelch:
@nathanleclaire actually, this won't work, for the same reason I wanted to reconfigure the provision logic from issue #961 because you return an ssh cmd object, you don't actually execute the command within that method. I could run the first set directly, then return the second cmd, but that seems wrong. Would you like me to make a bigger PR moving the logic around to make this work if you want multiple commands?  For now, I'm reverting to a single command...

nathanleclaire:
> (fwiw, I think the single line way is more readable just due to error checks, etc. happy to change it back if you agree)

I'd rather keep it "one concern per command" - one sets the hostname, the other replaces `127.0.0.1` with `127.0.1.1`.

Just FYI, it's slightly more idiomatic to do this if a function call only returns one value, like so:

```
if err := cmd.Run(); err != nil {
    return err
}
```

You can change to that form if you want, but I don't really think of it as a hard requirement. 

LGTM

crunchywelch:
too right, updated!

libmachine/provision/ubuntu.go
```diff
@@ -124,9 +124,22 @@ func (provisioner *UbuntuProvisioner) Hostname() (string, error) {
...
+		"sudo hostname %s && echo %q | sudo tee /etc/hostname",
 		hostname,
 		hostname,
+	))
+
+	if err != nil {
+		return err
+	}
+
+	if err = cmd.Run(); err != nil {
```
nathanleclaire: This should be a `:=` - otherwise you are setting the error from above (therefore it's not self-contained and anyone who changes the above code might have to change this too)

libmachine/provision/ubuntu.go
```diff
@@ -134,7 +147,11 @@ func (provisioner *UbuntuProvisioner) SetHostname(hostname string) error {
 		return err
 	}
 
-	return cmd.Run()
+	if err = cmd.Run(); err != nil {
```
nathanleclaire: same here

nathanleclaire:
OK one last comment - LGTM - cc @ehazlett for review

crunchywelch:
cool, updated...

nathanleclaire:
Great, thanks for being responsive.

ehazlett:
@crunchywelch would you mind rebasing?  thanks!

crunchywelch:
ooook

nathanleclaire:
cc @ehazlett

ehazlett:
I'm getting a duplicate entry for the hostname -- is that expected?

```
127.0.1.1 test-do test-do
```

crunchywelch:
Only if you somehow ran this twice on the same host, is that possible?

ehazlett:
It was a fresh host in DigitalOcean.  I'll try again.

ehazlett:
This also doesn't appear to be working on EC2:

```
127.0.0.1 localhost
```

That's the only thing in `/etc/hosts` (besides the ipv6 entries).

ehazlett:
Same with Rackspace

crunchywelch:
hrm, ok, let me play with it and figure out what's up.

ehazlett:
@crunchywelch thanks!

crunchywelch:
@ehazlett  this works as expected for me on DO:

```
root@ubuntutest:~# cat /etc/hosts
# Your system has configured 'manage_etc_hosts' as True.
# As a result, if you wish for changes to this file to persist
# then you will need to either
# a.) make changes to the master file in /etc/cloud/templates/hosts.tmpl
# b.) change or remove the value of 'manage_etc_hosts' in
#     /etc/cloud/cloud.cfg or cloud-config from user-data
127.0.1.1 ubuntutest
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
```

Can you tell me the parameters you are using for the create command to make sure I'm not missing something?

Looking into rackspace and aws now...

crunchywelch:
@ehazlett ok, so that double hostname entry is the default behaviour on DO, this was the result of me creating a droplet from the control panel, not using machine:

```
root@hostnametester:~# cat /etc/hosts
# Your system has configured 'manage_etc_hosts' as True.
# As a result, if you wish for changes to this file to persist
# then you will need to either
# a.) make changes to the master file in /etc/cloud/templates/hosts.tmpl
# b.) change or remove the value of 'manage_etc_hosts' in
#     /etc/cloud/cloud.cfg or cloud-config from user-data
127.0.1.1 hostnametester hostnametester
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
root@hostnametester:~#
```

So I suspect the provision failed, or did not run this code at very least. I've been getting this on digital ocean, which I'm troubleshooting now, but results in a working droplet, but that isn't fully configured:

```
INFO[0001] Creating Digital Ocean droplet...            
DEBU[0003] Created droplet ID 5022362, IP address 104.236.200.167 
DEBU[0003] Getting to WaitForSSH function...            
ERRO[0137] Error creating machine: Process exited with: 1. Reason was:  () 
WARN[0137] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0137] Error creating machine
```

Will continue to look into rackspace and ec2, I think that just requires a check for the line and if it not there appending it, or replacing if it exists...

crunchywelch:
@ehazlett ok, got this working on rax, DO and aws, let me know if you notice any funny business.

As an aside, might be nice to add some debug statements in the provision method so you can see where things fail if they do.

crunchywelch:
@ehazlett polite bump, would love to get this one in so I can merge on my other branches and get these other PRs issued if possible, is looking ok?

ehazlett:
@crunchywelch looking now -- thanks for the bump!

ehazlett:
@crunchywelch hey if i run this on DO i still get the duplicate on the first entry -- not a huge deal.  On ec2 I get the following:

```
ubuntu@mcn-test-ec2:~$ cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
127.0.0.1 mcn-test-ec2
```

The hostname is appended at the end.  I'm not sure what is going on.

crunchywelch:
@ehazlett that's really odd, there was a rev with a typo at one point that appended the localhost IP that you see there, but it should be gone now. Are you certain you compiled against this newest rev? 

Here are my results just now:

AWS - https://gist.github.com/crunchywelch/1cb0616d33be039d9bd1
DO - https://gist.github.com/crunchywelch/f34f4170151a4f13a568
Rax - https://gist.github.com/crunchywelch/7e06a949b4aac52a93e5

ehazlett:
@crunchywelch thanks -- it was a fresh fetch -- i will try again -- thx!

ehazlett:
Fresh fetch and this looks good.  Thanks!

ehazlett:
LGTM

PR #966, opened 2015-04-06
Add a few improvements to stability

nathanleclaire:
This improves on a few issues which are not frequent but do sometimes
come up:

1. Check that the daemon is up before attempting to do anything
   docker-ey in provisioning Ubuntu
2. Set a max retries for SSH so that if a request fails it attempts
   again before giving up

Also have annotated a few of the SSH options slightly better.

### DIFF
--- a/libmachine/provision/ubuntu.go
+++ b/libmachine/provision/ubuntu.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"os/exec"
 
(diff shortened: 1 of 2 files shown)


nathanleclaire:
cc @ehazlett

ehazlett:
Hit a few provisioning issues but appear to be vbox related.  After creating almost 50 vbox + cloud VMs this seems to be working well.

LGTM

PR #968, opened 2015-04-07
Run BATS test inside docker-machine container 

janeczku:
After building, just do    
>  script/run-bats-test cli.bats
   
or    
   
>  script/run-bats-test driver-digitalocean.bats DIGITALOCEAN_ACCESS_TOKEN=ABC123
   


ehazlett:
Thanks for the contribution!  While I love running things in containers ;) I'm not sure I like not being able to run all of the tests in the same place (since we can't run these for local providers).

Thoughts? @nathanleclaire @sthulb

nathanleclaire:
While I appreciate the effort, unfortunately for the "native virtualization" providers we're stuck running the tests on the host system.  So, in order to keep things as simple as possible, I'd prefer to stick to running the integration tests only one way.  Thanks for the contribution though @janeczku !

janeczku:
@ehazlett @nathanleclaire  Guess i hadn't thought that through :) You are right of course!

ehazlett:
@janeczku thanks for the contribution -- i would love to be able to run all of these in a container -  maybe some day :)

PR #973, opened 2015-04-07
make ssh username configurable for ec2

wlan0:
@ehazlett 

Im working on provisioning rancherOS instance using machine. For aws, the SSH user needs to be 'rancher' for rancherOS instead of the default 'ubuntu'. 

I made the SSH user name a configurable flag to support this.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -146,6 +146,12 @@ func GetCreateFlags() []cli.Flag {
 			Name:  "amazonec2-iam-instance-profile",
 			Usage: "AWS IAM Instance Profile",
 		},
+		cli.StringFlag{
+			Name:   "amazonec2-ssh-user",
+			Usage:  "set the name of the ssh user",
+			Value:  "ubuntu",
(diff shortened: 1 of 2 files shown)


drivers/amazonec2/amazonec2.go
```diff
@@ -199,9 +205,13 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.SwarmMaster = flags.Bool("swarm-master")
 	d.SwarmHost = flags.String("swarm-host")
 	d.SwarmDiscovery = flags.String("swarm-discovery")
-	d.SSHUser = "ubuntu"
+	d.SSHUser = flags.String("amazonec2-ssh-user")
 	d.SSHPort = 22
 
+	if d.SSHUser == "" {
```
ehazlett: This isn't needed as the cli.StringFlag will always be set to "ubuntu" as the default.

wlan0: Nice catch. Fixed it.

ehazlett:
LGTM

PR #978, opened 2015-04-08
driver contribution guidelines

ehazlett:
Add driver contribution guidelines to readme.


nathanleclaire:
Nice start.  LGTM

I have some things I want to add/consider but we can do them in a followup PR (noting them here so I remember):

- Caveat that the project moves pretty quickly so merging new code in requires persistence
- Notice of intention to eventually move over to a plugin model (i.e. just because it is in the "core" now does not mean that it will be "core" forever)
- Preferably, pretty specific (objective) criteria about "what gets in" and "when we yank".  This will have to be discussed thoroughly, of course, so it's more of a long-term goal.  An example for the second, for instance, could be, "If we haven't heard from the maintainer or their organization in 30 days after notifying them that we need help, we will yank the driver".

bfirsh:
Looks great! Perhaps this should be under the contributing section?

More generally we need to make sure that we route people to the correct place. It's super important that "if you want to install and use machine, go to the docs" is super prominent, because this is what 95% of users will be looking for. Contributing docs can be less prominent.

For Compose, we very clearly fork off these two routes at the bottom, tucking away all of the contributing docs into a separate page so that users don't miss the big "here's the documentation!" link: https://github.com/docker/compose

ehazlett:
@bfirsh i like the idea of separating this out to a contributing page.

ehazlett:
Updated re: @bfirsh reqs.  PTAL

ehazlett:
Merging. We will continue to update in future PRs.

Issue #904, opened 2015-03-29
Boot2Docker Migration Support


ehazlett:
This will enable migration from boot2docker to machine.  This includes migrating the disk as well as document the cli changes.

ehazlett:
closing as migration is in and docs are created.

PR #979, opened 2015-04-08
VirtualBox: enable ability to import b2d instances

ehazlett:
This adds support for importing a b2d instance (in this case, just the disk).

Discussion:  should we do a full clone of the VM to bring in any changes the user may make to the VM itself?  I didn't do that in the current state as this might add things that Machine does not support (i.e. network configuration, etc).

Refs #904

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -186,10 +186,32 @@ func (d *Driver) Create() error {
 
 	// import b2d VM if requested
 	if d.Boot2DockerImportVM != "" {
-		log.Debugf("Importing disk image...")
-		if err := d.importB2DVM(d.Boot2DockerImportVM); err != nil {
+		name := d.Boot2DockerImportVM
+
+		// make sure vm is stopped
+		_ = vbm("controlvm", name, "poweroff")
+
+		diskInfo, err := getVMDiskInfo(name)
+		if err != nil {
 			return err
 		}
+
+		if _, err := os.Stat(diskInfo.Path); err != nil {
+			return err
+		}
+
+		if err := vbm("clonehd", diskInfo.Path, d.diskPath()); err != nil {
+			return err
+		}
+
+		log.Debugf("Importing VM settings...")
+		vmInfo, err := getVMInfo(name)
+		if err != nil {
+			return err
+		}
+
+		d.CPU = vmInfo.CPUs
+		d.Memory = vmInfo.Memory
 
 		log.Debugf("Importing SSH key...")
 		keyPath := filepath.Join(utils.GetHomeDir(), ".ssh", "id_boot2docker")
@@ -214,6 +236,9 @@ func (d *Driver) Create() error {
 		"--register"); err != nil {
 		return err
 	}
+
+	log.Infof("CPUS: %d", d.CPU)
+	log.Infof("Mem: %d", d.Memory)
 
 	cpus := d.CPU
 	if cpus < 1 {
@@ -612,26 +637,6 @@ func zeroFill(w io.Writer, n int64) error {
 		}
 		n -= int64(k)
 	}
-	return nil
-}
-
-func (d Driver) importB2DVM(name string) error {
-	// make sure vm is stopped
-	_ = vbm("controlvm", name, "poweroff")
-
-	diskInfo, err := getVMDiskInfo(name)
-	if err != nil {
-		return err
-	}
-
-	if _, err := os.Stat(diskInfo.Path); err != nil {
-		return err
-	}
-
-	if err := vbm("clonehd", diskInfo.Path, d.diskPath()); err != nil {
-		return err
-	}
-
 	return nil
 }
 
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -197,13 +197,13 @@ func (d *Driver) Create() error {
 			return err
 		}
 	} else {
-		log.Debugf("Creating disk image...")
-		if err := d.generateDiskImage(d.DiskSize); err != nil {
+		log.Infof("Creating SSH key...")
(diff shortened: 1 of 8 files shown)


nathanleclaire:
Cool.  That was fast!

If we're going to do the migration this way, we will need:

1. Documentation on how to do it (I don't think most users will know that they need to specify by name `boot2docker-vm`)
2. Game-plan for deprecating this feature eventually (2 releases?  3?)

> Discussion: should we do a full clone of the VM to bring in any changes the user may make to the VM itself? I didn't do that in the current state as this might add things that Machine does not support (i.e. network configuration, etc).

I'm not too worried about importing settings that the user has made manually, since that seems to me that it could be a large amount of work for a not so common use case, _but_ I do think we need to parse the b2d config file and convert it into machine terms.  If my b2d configuration was set to allow 4096MB RAM and the new machine doesn't have the same amount, that's going to be an annoying surprise.

ehazlett:
@nathanleclaire thanks!  yeah i need to do a separate guideline -- i think that should be a separate issue / PR -- thoughts?

+1 for planning for the deprecation

ehazlett:
added the ability to import cpu and memory settings from the b2d vm.  PTAL @nathanleclaire

utils/utils.go
```diff
@@ -79,6 +79,15 @@ func CopyFile(src, dst string) error {
 		return err
 	}
 
+	fi, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+
+	if err := os.Chmod(dst, fi.Mode()); err != nil {
```
nathanleclaire: What is this solving?  I'm not doubting that it's needed, just curious.

drivers/virtualbox/vm.go
```diff
@@ -0,0 +1,55 @@
...
+type VirtualBoxVM struct {
+	CPUs   int
+	Memory int
+}
+
+func parseVMInfo(r io.Reader) (*VirtualBoxVM, error) {
+	s := bufio.NewScanner(r)
+	vm := &VirtualBoxVM{}
+	for s.Scan() {
+		line := s.Text()
```
nathanleclaire: Could you move this little block into its own function please?

drivers/virtualbox/vm.go
```diff
@@ -0,0 +1,55 @@
...
+		if line == "" {
+			continue
+		}
+		res := reEqualLine.FindStringSubmatch(line)
+		if res == nil {
+			continue
+		}
+		switch key, val := res[1], res[2]; key {
```
nathanleclaire: Hm, I think I'd prefer separate lines for the assignment and the switch here for clarity (`switch key`) -- do you agree?

nathanleclaire:
Code is looking good, modulo a few nitpicks- I'll give this functionality a whirl soon.

drivers/virtualbox/vm.go
ehazlett: Ok works for me.  We should change it in the original too -- that's how it is to find the network details too.

utils/utils.go
ehazlett: If you copy a key it will not have the correct permissions.  When you then try to ssh, it will fail because we shell out to ssh which the file must be `0600`.  Here I copy the original file permissions to match the existing key perms.

drivers/virtualbox/vm.go
ehazlett: Not sure what you mean here.  The scanning?

nathanleclaire: The bit after `line := s.Text()`.  Should work fine to just pass `line` into it right?

ehazlett: Yes, but then we would have to pass the ref to `VirtualBoxVm{}` to set the fields or make it generic to return the val and then parse.  I think it's concise the way it is and it's tested as well.

ehazlett:
@nathanleclaire updated ptal

drivers/virtualbox/virtualbox.go
```diff
@@ -81,6 +82,11 @@ func GetCreateFlags() []cli.Flag {
 			Usage:  "The URL of the boot2docker image. Defaults to the latest available version",
 			Value:  "",
 		},
+		cli.StringFlag{
+			Name:  "virtualbox-import-boot2docker-vm",
+			Usage: "The name of a Boot2Docker VM to import",
+			Value: "",
```
nathanleclaire: I keep thinking that maybe we should default this value to `boot2docker-vm`, since that's by far the most common name and I have a feeling users will run this flag without args expecting it to "just work".

nathanleclaire:
So should we clean up the old `boot2docker-vm` automatically or prompt users to `boot2docker destroy`?

nathanleclaire:
I added an issue to write docs for the 0.3.0 milestone.

LGTM

drivers/virtualbox/virtualbox.go
tianon: +1, I'd certainly expect that myself -- I don't know that's it's very common for users to actually name their boot2docker VMs (even though it's supported to do so)

ehazlett:
@nathanleclaire Yeah I think that would be in the docs.  I don't want to remove it automatically -- that seems too destructive (especially if the import fails for some reason).  I think a note in the migration doc would suffice.  Thoughts?

ehazlett:
@nathanleclaire added docs -- this should close #1047.  PTAL

nathanleclaire:
> Yeah I think that would be in the docs. I don't want to remove it automatically -- that seems too destructive (especially if the import fails for some reason). I think a note in the migration doc would suffice. Thoughts?

Recommending `boot2docker destroy` (and maybe an uninstall of boot2docker as well) seems reasonable to me.

nathanleclaire:
:+1: :+1: :+1:

PR #991, opened 2015-04-10
fakedriver: remove useless methods

vincentbernat:
Since the refactor done in #756, some methods have been removed from the
driver interface. Remove those methods from `fakedriver`:

 - `StartDocker()`
 - `StopDocker()`
 - `GetDockerConfigDir()`
 - `GetSSHCommand()`

### DIFF
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -22,10 +22,6 @@ import (
 	"github.com/docker/machine/state"
 )
 
-const (
-	dockerConfigDir = "/etc/docker"
-)
-
 type Driver struct {
 	MachineName             string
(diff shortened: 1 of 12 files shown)


vincentbernat:
I have added a second commit for Digital Ocean, but I am unsure if I am correct: all drivers seem to define `dockerConfigDir` but I don't see where it is used.

vincentbernat:
Generalized the last commit to all drivers. The use of this constant has been removed since 3b2c8f9845de897acb38518ad97fa10267d9ea0d

ehazlett:
Tested and LGTM.  Thanks!

Issue #529, opened 2015-02-11
virtualbox driver on windows doesn't mount to the local file system yet 


cvuijst:
As discussed on IRC, after running a "docker-machine create" on Windows, the local file system is not mounted in the created docker environment. It works out of the box gracefully on OSX.
This leads possibly to a non implemented feature on Windows: See: https://github.com/docker/machine/blob/master/drivers/virtualbox/virtualbox.go#L289
Currently as a workaround, you can do this manually on the windows CLI having VBoxManage in your path through e.g.:
     $/> VBoxManage controlvm dummy_env acpipowerbutton
     $/> VBoxManage sharedfolder add dummy_env --name Users --hostpath c:/Users --automount
     $/> VBoxManage startvm dummy_env --type headless
Where "dummy_env" is the name of the created environment.
It would be nice to have this on Windows as well.

ehazlett:
@SvenDowideit @jeffmendoza what would be the best path to mount?  is there an env var for the user home dir on windows?  i would think so.

@SvenDowideit perhaps b2d already has this solved and i could reference it from somewhere?

cvuijst:
I took a look at the VMs created in the VBox management UI on OSX and Windows, it looks like for boot2docker:
* Windows: "c/Users" folder gets mounted to "c:\Users" 
* OSX: "Users" folder gets mounted to "/Users"
There is a platform difference there. I'm not sure that was the original intention though.
Currently, for docker-machine on OSX:
* "Users" gets mounted to "/Users"

jeffmendoza:
C:\Users would be fine

dsebastien:
There's actually a variable for the user's home folder: %USERPROFILE%.

atomi:
Is it safe to assume that volume mounting will never work on Windows using docker proxy to boot2docker and virtual box? It seems the best solution right now is to use an ubuntu vagrant box, mount your preferred folders and run all commands via a putty terminal. We may as well run docker machine on virtual box as well. Or wait for docker daemon to get ported to Windows which will be extremely surprising to see.

ehazlett:
@atomi that's the goal.  the shared folder support should work we just need to map it.

carletes:
Any plans to have this working also for Linux VirtualBox hosts (not just Windows and OS X) for 0.2.0?

sthulb:
@carletes Yes, there should be universal support for all of our supported OSs

atomi:
Yeah so this won't be on 0.2.0?

[...]

PR #993, opened 2015-04-10
Windows vbox share

ehazlett:
This adds `/c/Users` to the VM mounted as `/c/Users/`.  We still have a ways to go as bash in msysgit rewrites the paths so you still cannot use `docker run -v /c/Users/foo/dev:/app` for example.   However, this is a first step :)

Refs #529

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -330,10 +330,13 @@ func (d *Driver) Create() error {
 
 	var shareName, shareDir string // TODO configurable at some point
 	switch runtime.GOOS {
+	case "windows":
+		shareName = "c/Users"
+		shareDir = "c:\\Users"
 	case "darwin":
 		shareName = "Users"
 		shareDir = "/Users"
-		// TODO "linux" and "windows"
+		// TODO "linux"
 	}
 
 	if shareDir != "" {


nathanleclaire:
Could you elaborate a little bit more on what's the issue with the path transforms?

ehazlett:
When you attempt to run `docker run -v $(pwd):/app foo` on Windows, msysgit (bash) will translate these to something like this:
2015-04-13-154455_1070x98_scrot

ehazlett:
I think @ahmetalpbalkan can shed some light :)  please? :)

nathanleclaire:
Hm, yeah, I'm not sure if it's msysgit or the docker client doing something funny here (or how the two are supposed to interact).  Looks like `C:\\Users\\admin\\Downloads` is the correct path, but that's getting transformed to `c:\\Users\\admin\\Downloads;C:\\Program Files (x86)\\Git\\downloads`.

ahmetalpbalkan:
@ehazlett @nathanleclaire the error in the screenshot is happening because anything before `:` is parsed like a letter drive probably.

MSYS bash tries to convert

     /c/Users/admin/foo:/app

to 

    c:\Users\admin\foo:/app

and **then** passes to docker.exe.

Try using cmd.exe (it doesn't do such path conversion). In order to avoid path conversion in MSYS you can probably surround the argument with single quotes.

Actually this is a sign that using msys is confusing enough. Maybe we should make the docker-machine ready for cmd.exe/powershell and do not endorse msys as much as we can.

ehazlett:
@ahmetalpbalkan thx -- +1 to using cmd/powershell on windows -- i would rather as that would be "native" and not a third party thing to install.  We would, however, need to get ssh etc. as well.

hairyhenderson:
@ehazlett - want to add Linux support in this PR as well, or should that be in a separate PR?

StefanScherer:
@ehazlett I also prefer a native solution for Windows without fancy shells and extra tools. GO has direct support of SSH. I know and use https://www.packer.io on Windows to build Linux VMs and packer uses SSH for provisioning the VM's and no ssh.exe is needed on Windows hosts. Perhaps packer's ssh communicator ( https://github.com/mitchellh/packer/tree/master/communicator/ssh ) could be a good starting point.

hairyhenderson:
@StefanScherer - there's a PR for SSH being worked on - #909, so the dependency on `ssh.exe` is not long for this world ;)

ehazlett:
@StefanScherer as @hairyhenderson mentioned we want to move to crypto/ssh for this -- would eliminate the callout to ssh.exe and should get us in a good spot :)

StefanScherer:
Great things happen here!

nathanleclaire:
I will go steal Andy's laptop again and test this when I get a chance :)

ehazlett:
nice!

nathanleclaire:
So the functionality on this PR looks good from my end, but let's please add documentation.  Maybe we can make a new issue for that, since I think ultimately I'd like to include a "Folder Sharing" section that explains the mechanics of what we do, links to the ruby script which will allow you to use NFS on Mac if you are interested, and includes a digression about all the Windows issues that we are encountering here.

nathanleclaire:
Hard dep for release: https://github.com/docker/machine/issues/1116 Other than that, LGTM

ehazlett:
Merging.  Added #1116 to the milestone to track docs.  Thanks!

ehazlett:
:+1:

PR #995, opened 2015-04-11
Don't let drivers wait for SSH/Docker/State

vincentbernat:
Waiting for SSH to start or Docker to be available is a step done
directly in the framework (in libmachine). There is no need for the
driver to do that.

Those changes are just untested! Since people copy from drivers, I think this is important to keep the minimal code from each driver. Maybe we should ask each driver maintainer to make the appropriate changes, test them and do a PR. In this case, feel free to discard this one.

The SoftLayer driver still waits for SSH but it needs to `apt-get install curl`. Maybe it could be modified to put this step in some provisioning function instead.

There are separate commits when it is not obvious if the change is correct. For example, in the openstack driver, there is a wait for the machine to be in the correct state. I don't think this is useful since libmachine also does that, but maybe I am mistaken.

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -194,13 +194,6 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 		return err
 	}
 
-	instance, err = c.instance()
-	if err != nil {
-		return err
-	}
-	ip := instance.NetworkInterfaces[0].AccessConfigs[0].NatIP
-	c.waitForSSH(ip)
-
 	// Update the SSH Key
 	sshKey, err := ioutil.ReadFile(d.GetSSHKeyPath() + ".pub")
 	if err != nil {
@@ -278,12 +271,6 @@ func (c *ComputeUtil) waitForGlobalOp(name string) error {
 	return c.waitForOp(func() (*raw.Operation, error) {
 		return c.service.GlobalOperations.Get(c.project, name).Do()
 	})
-}
-
-// waitForSSH waits for SSH to become ready on the instance.
-func (c *ComputeUtil) waitForSSH(ip string) error {
-	log.Infof("Waiting for SSH...")
-	return ssh.WaitForTCP(fmt.Sprintf("%s:22", ip))
 }
 
 // ip retrieves and returns the external IP address of the instance.
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -398,12 +398,6 @@ func (d *Driver) Create() error {
 		d.PrivateIPAddress,
 	)
 
-	log.Infof("Waiting for SSH on %s:%d", d.IPAddress, 22)
-
-	if err := ssh.WaitForTCP(fmt.Sprintf("%s:%d", d.IPAddress, 22)); err != nil {
-		return err
-	}
-
 	log.Debug("Settings tags for instance")
 	tags := map[string]string{
 		"Name": d.MachineName,
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -3,12 +3,10 @@ package azure
 import (
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"time"
 
 	azure "github.com/MSOpenTech/azure-sdk-for-go"
 	"github.com/MSOpenTech/azure-sdk-for-go/clients/vmClient"
@@ -257,9 +255,7 @@ func (d *Driver) Create() error {
 		return err
 	}
 
-	log.Info("Waiting for SSH...")
-	log.Debugf("Host: %s SSH Port: %d", d.getHostname(), d.SSHPort)
-	return ssh.WaitForTCP(fmt.Sprintf("%s:%d", d.getHostname(), d.SSHPort))
+	return nil
 }
 
 func (d *Driver) GetURL() (string, error) {
@@ -315,10 +311,8 @@ func (d *Driver) Start() error {
 	if err := vmClient.StartRole(d.MachineName, d.MachineName, d.MachineName); err != nil {
 		return err
 	}
-	if err := d.waitForSSH(); err != nil {
-		return err
-	}
-	return d.waitForDocker()
+
+	return nil
 }
 
 func (d *Driver) Stop() error {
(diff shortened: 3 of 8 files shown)


ehazlett:
+1  The SoftLayer driver can actually be updated to remove the install for `curl` as well as we install that as a base dependency (https://github.com/docker/machine/blob/master/libmachine/provision/ubuntu.go#L25).

/cc @ggiamarchi for the openstack question -- I'm not positive we can remove the state check for that one.

nathanleclaire:
Oh excellent!  I want to get this in ASAP (probably will wait til after the release)

> There are separate commits when it is not obvious if the change is correct. For example, in the openstack driver, there is a wait for the machine to be in the correct state. I don't think this is useful since libmachine also does that, but maybe I am mistaken.

This _should_ be fine, but we should definitely test it out / get @ggiamarchi input.

nathanleclaire:
Looking at `drivers/openstack/openstack.go` looks like `GetState` should correctly return `SHUTOFF` so I think we are good to go on this.

LGTM - thanks for doing this @vincentbernat

ggiamarchi:
I've just tested the OpenStack driver. Everything is good :+1:

vincentbernat:
I have added a commit to remove curl installation for SoftLayer. I think there is still room for simplification. Waiting for state "running" should not be needed since this is done by libmachine but then there is a wait for "transaction setup" and I am unsure of what it is. I have left it as is. ping @cpuguy83 for validation.

cpuguy83:
Hmm, I was unaware of this transaction setup.
Here's my guess at what it actually does:

As part of the VM provisioning for softlayer, originally we told softlayer to run an install script directly via the API (basically pointed it at get.docker.com).
Sometimes the VM's SSH would become available sometimes while that installation was still happening so I had all kinds of hackery to make sure the docker binary was available before doing anything else... I'm guessing this transaction wait is there to make sure any install scripts we told softlayer to run for us are finished before continuing.

Since machine is doing all setup now, maybe this is not needed?
Looks like @hairyhenderson added this transaction wait.

ehazlett:
thanks @cpuguy83 .  @hairyhenderson would this have any affect on our provisioning?

hairyhenderson:
I added the "wait for transactions" to make sure that the VM is well and truly set up from SoftLayer's point of view before we start installing Docker. Part of it is to make sure scripts are finished, but there's also some steps that SL may (or may not) perform after that.

Sometimes it's safe to start as soon as SSH is available, but not always. I wanted to avoid any possible conflicts by just waiting. IMO, this step is still needed.

Maybe the elegant thing would be to add something to the `Driver` interface that the provisioner could hook into for driver-specific stuff like this? I dunno...

vincentbernat:
❦ 16 avril 2015 12:12 -0700, Dave Henderson <notifications@github.com> :

> I added the "wait for transactions" to make sure that the VM is well
> and truly set up from SoftLayer's point of view before we start
> installing Docker. Part of it is to make sure scripts are finished,
> but there's also some steps that SL may (or may not) perform after
> that.
>
> Sometimes it's safe to start as soon as SSH is available, but not
> always. I wanted to avoid any possible conflicts by just waiting. IMO,
> this step is still needed.

If it's needed, we should of course keep it.

> Maybe the elegant thing would be to add something to the Driver
> interface that the provisioner could hook into for driver-specific
> stuff like this? I dunno...

The provisioner uses GetState (indirectly) to check if the machine is
running. You could put your code here instead. However, as far as I am
concerned, my goal was to remove unneeded code, so if the code is in
fact needed, we can just stay as is.
-- 
Make sure every module hides something.
            - The Elements of Programming Style (Kernighan & Plauger)

ehazlett:
@hairyhenderson if it's needed for stability let's keep it

nathanleclaire:
Yes I'm OK with keeping it for now - this kind of thing (and there was another related issue lately) makes me think that we should consider some kind of universal way (either in the provisioners or drivers) of checking to see "Hey, are you OK to begin provisioning?" since some providers do custom stuff (e.g. install packages) that we will want to wait for.

ehazlett:
@vincentbernat you ok with those updating to keep those?

vincentbernat:
@ehazlett I didn't remove them, so they are currently still here. I only removed the wait for SSH and the curl installation. Maybe @hairyhenderson could test that softlayer driver still works with this branch?

ehazlett:
@vincentbernat ah thanks!  @hairyhenderson can you confirm this is still working as expected?

hairyhenderson:
@vincentbernat - I just cloned your branch and tested out a `create` with SoftLayer... There's some really weird stuff going on:

```
DEBU[0272] Still waiting - active transaction is SERVICE_SETUP... 
DEBU[0282] Getting to WaitForSSH function...            
DEBU[0406] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectionAttempts=30 -o LogLevel=quiet -p 22 -i /Users/hendersd/.docker/machine/machines/dmtest-ams01/id_rsa root@null exit 0 
DEBU[0406] Error running ssh command 'exit 0' : exit status 255 
DEBU[0409] Getting to WaitForSSH function...            
DEBU[0529] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectionAttempts=30 -o LogLevel=quiet -p 22 -i /Users/hendersd/.docker/machine/machines/dmtest-ams01/id_rsa root@null exit 0 
DEBU[0529] Error running ssh command 'exit 0' : exit status 255 
DEBU[0532] Getting to WaitForSSH function...
```

The weird bit is the hostname is `null` in the ssh command...

I *think* this might be happening because of a bug that was fixed recently. I'm going to rebase and see if that fixes things.

hairyhenderson:
Ok, with a rebase, I can confirm that things still work as expected.

:+1:

vincentbernat:
Great! I have also rebased to latest master.

ehazlett:
LGTM

drivers/google/compute_util.go
```diff
@@ -194,13 +194,6 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 		return err
 	}
 
-	instance, err = c.instance()
```
jeffmendoza: Removing this line causes a panic. Instance still needs to be refreshed.

nathanleclaire: cc @vincentbernat

vincentbernat: @jeffmendoza Sorry, didn't know that. Do you want me to do a PR about this or do you prefer to do it yourself?

ehazlett:
@jeffmendoza would you mind taking a look at #1072 to see if that looks good?  I tested against GCE and it works as expected.  thanks!

Issue #982, opened 2015-04-09
Add support for SoftLayer Local disk on VMs 


huslage:
The SoftLayer API supports Local disk via the `localDiskFlag` boolean parameter. Please expose this so that we can create VMs with Local disk.

```python
client['Virtual_Guest'].createObject({
    'hostname': 'myhostname',
    'domain': 'example.org',
    'startCpus': 1,
    'maxMemory': 1024,
    'hourlyBillingFlag': 'true',
    'operatingSystemReferenceCode': 'UBUNTU_LATEST',
    'localDiskFlag': 'false'
})
```

ehazlett:
Thanks!  This option is supported via the `--softlayer-local-disk` option.  See https://github.com/docker/machine/blob/master/docs/index.md#ibm-softlayer for details.

mbentley:
Upon further investigation, it looks like the driver doesn't actually sent and pass the localdisk option.  The option is there in machine but it essentially isn't doing anything.

PR #1005, opened 2015-04-15
softlayer: enable local disk in hostspec

ehazlett:
This fixes the bug where the local disk option wasn't being passed to the SL API.

Refs #982


hairyhenderson:
LGTM!

ehazlett:
@hairyhenderson thanks!

PR #1006, opened 2015-04-15
Reducing the number of SSH connection attempts

hairyhenderson:
Also setting an explicit timeout of 10 seconds. This way, if the host is unreachable, it'll only block for 30 seconds.


nathanleclaire:
LGTM... Nice catch! cc @ehazlett

ehazlett:
LGTM :+1:

Issue #967, opened 2015-04-06
Proposal: Logging cleanup


nathanleclaire:
The current method of outputting information from Docker Machine commands is a bit of an awkward fit.  The majority of code uses `logrus` to print things to the terminal, and while `logrus` is a great package there are a few issues:

- We'd like finer-grained control over what goes to STDOUT and STDERR
- Current output looks not so great on Windows
- `logrus` is meant more for long-running applications than for the ad-hoc sort of commands invoked using `docker-machine`; while Machine may run as a long-lived daemon or server someday, we aren't quite ready to move to that model yet.

Because we have so much code that uses `logrus` and don't want to chuck this method of outputting information out completely, this proposal is to wrap the existing `log.*` calls with our own package (`github.com/docker/machine/log`?) which will give us more control over the result of these calls.

That way, all of the existing code which uses, say, `log.Warnf` will simply wrap calls to `fmt.Printf`, calls to `log.Errorf` will wrap `fmt.Fprintf(os.Stderr`, etc.

In general, I'm thinking that what I want to do is make an interface with the relevant methods, and have several swappable "backends" e.g. `terminal`, `file`, `logstash` so that we could take advantage of the versatility `logrus` provides at some point down the line here.   Similar to what we do with hosting providers or underlying OSes.  And `terminal` being the simplest, it would be the default.

Sticking points:

- What to do about `log.WithFields`?  The Openstack / Rackspace drivers use this non-trivially and I don't really want to just chuck it out, but it will be difficult to reconcile with this proposal since it seems to mandate including `logrus` directly.  We could wrap that stuff too I guess.
- Might want to consider adding some timing data e.g. "Your machine create took X seconds" since one of the things I like about the existing method is that you can get a feel for how long something's been running or took to run by looking at the second counter.

cc @ehazlett @sthulb @bfirsh

sthulb:
+1 The logging implementation is rather out of place with what I'd expect in a client app.

I guess for now, we should perhaps wrap go's native `log` package with `docker/machine/log`, but this would be a stop-gap for a better solution later.

ehazlett:
+1 for cleaner and "standard" log formats across drivers and platforms.  After glancing into `logrus` I wonder if we should look into the hooks and formatters?  This way we wouldn't have to implement / maintain the various backends ourselves and potentially re-use it if/when a server version happens.  If it doesn't fit then we could roll our own.  Just a thought.

[...]

PR #1008, opened 2015-04-15
Refactor logging to focus on simple STDOUT/STDERR

nathanleclaire:
This also lays the foundation for the possibility of log drivers in the
future, if it is decided that is a direction to pursue.

Closes https://github.com/docker/machine/issues/967.

`log.WithFields` still works pretty similarly ;D

cc @ehazlett @hairyhenderson @sthulb @bfirsh PTAL


nathanleclaire:
### Concerns:

1. `github.com/Sirupsen/logrus` is still in `Godeps` - I don't really mind keeping it there in case we end up wanting to write a "driver" for it or use again for something in the future, but let me know your thoughts.  If I understand correctly removing it won't save us space on a repo clone or anything due to git tracking all of the history.
2.  I haven't implemented _all_ the methods that `logrus` originally supported - just the ones we have used - I definitely think there is an argument for moving over to only using only `log.*`, `log.*f` and killing `log.*ln` entirely.  See the interface if you are confused what I am talking about.

log/log.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+	"strconv"
+)
+
+// Why the interface?  We may only want to print to STDOUT and STDERR for now,
+// but it won't neccessarily be that way forever.  This interface is intended
+// to provide a "framework" for a variety of different logging types in the
+// future (log to file, log to logstash, etc.) There could be a driver model
+// similar to what is done with OS or machine providers.
+type Logger interface {
```
nathanleclaire: this is the interface I am referencing in that "concerns" comment

nathanleclaire:
investigating travis build

sthulb:
I'd like to see logrus be removed. Cleaner working tree and less for travis to do (it uses shallow clones)

nathanleclaire:
> I'd like to see logrus be removed. Cleaner working tree and less for travis to do (it uses shallow clones)

Ah yeah, that's a good reason.

hairyhenderson:
Just a couple nitpicks:
- I would've expected to see `log.go` removed - is that missing from the commit (or am I blind?)
- I also would've expected to see `main.go`'s reference to `initLogging` (~ln19) removed

Otherwise, looks good :)

ehazlett:
Overall LGTM.  +1 for removing logrus -- we can always add it back if needed.

nathanleclaire:
> I would've expected to see log.go removed - is that missing from the commit (or am I blind?)
> I also would've expected to see main.go's reference to initLogging (~ln19) removed

Nope, it's not there - wanted to see the general sentiment around removing logrus completely was and it looks like everyone wants to, so I will put in those changes soon.

nathanleclaire:
Update: I've removed the files which reference logrus in the main package, but unfortunately `github.com/docker/docker/utils` which gets imported by a few of our modules expects logrus to be present, so I haven't removed that from Godep for now.

sthulb:
@nathanleclaire We can port docker/utils to machine, we only use one function out of the package.

main.go
```diff
@@ -4,7 +4,6 @@ import (
 	"os"
 	"path"
 
-	log "github.com/Sirupsen/logrus"
```
hairyhenderson: You'll need to also import `"github.com/docker/machine/log"` here (`log` is referenced on ln 75)

hairyhenderson:
@nathanleclaire & @sthulb - it looks like `docker/utils` is used in 3 places, and in 2 of the places it's referenced by an unreferenced unexpected function (`generateVMName` in azure.go and vcloudair.go). 

The last place is openstack.go, where it references `utils.GenerateRandomID`. I'm pretty sure we could port that usage, though honestly I'm not even sure why the KeyPair _needs_ to have a random suffix... But that's another issue.

nathanleclaire:
OK I've removed `logrus`, moved the two functions that we depend on to `machine/utils/utils.go`, removed `docker/docker/api` and `/docker/docker/utils` deps (we were only using api in once place, and it also depends on `logrus`).  Seems to build and work fine, such a large change makes me a little nervous so maybe in the future we should do it in separate (follow-up) PRs.

nathanleclaire:
Update:

1. Rebased
2. Moved `TerminalLogger` into its own package which `github.com/docker/machine/log` imports.  This will allow us to import different log drivers selectively if we want. For instance, in the drivers, we could change the import to something like `log "github.com/docker/machine/log/driver"`, which would allow the logger for the drivers to act slightly different than the default (prepend the machine name to the output, log to disk if there are errors, etc.).  It's also not infeasible to imagine the default logger as configurable via flag as well, so that for instance there could be an option to log with color in the output.
3. Filed an issue to remove the remaining `log.*ln` usages.

I think I'll make an issue to update the docs so that the output looks correct as well.

@ehazlett PTAL

ehazlett:
reviewed, tested  - works as expected.  oh and yay for the windows side!!!

2015-04-30-154025_882x283_scrot

ehazlett:
LGTM

nathanleclaire:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

PR #1009, opened 2015-04-16
Deleting Machine-generated SSH key from SoftLayer when host is canceled.

### LINKED ISSUE #765: SoftLayer SSH keys should be deleted when the corresponding host is deleted
I was just troubleshooting a bunch of stuff related to the SoftLayer driver, and created a few dozen new hosts in the process. Even though `docker-machine rm -f myhost` did fine at cancelling the host, the associated SSH key wasn't deleted from SoftLayer.

I see in some other drivers the remote keys are trashed when the host is deleted (during `docker-machine rm myhost`).

I can probably provide a PR for this, but want to make sure there isn't any reason we _wouldn't_ want to do this...

hairyhenderson:
All the other drivers seem to do this, so I can't imagine this is a bad
idea ;)

Fixes #765

### DIFF
--- a/drivers/softlayer/softlayer.go
+++ b/drivers/softlayer/softlayer.go
@@ -150,6 +150,19 @@ func (c *sshKey) Create(label, key string) (*SshKey, error) {
 	return &k, nil
 }
 
+func (c *sshKey) Delete(id int) error {
+	var (
+		method = "DELETE"
+		uri    = fmt.Sprintf("%s/%v", c.namespace(), id)
+	)
+
+	_, err := c.newRequest(method, uri, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func (c *Client) VirtualGuest() *virtualGuest {
 	return &virtualGuest{c}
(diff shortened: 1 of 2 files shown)


nathanleclaire:
LGTM

nathanleclaire:
cc @ehazlett

drivers/softlayer/softlayer.go
```diff
@@ -150,6 +150,19 @@ func (c *sshKey) Create(label, key string) (*SshKey, error) {
 	return &k, nil
 }
 
+func (c *sshKey) Delete(id int) error {
+	var (
+		method = "DELETE"
+		uri    = fmt.Sprintf("%s/%v", c.namespace(), id)
+	)
+
+	_, err := c.newRequest(method, uri, nil)
```
ehazlett: fyi you can also do:

```
if _, err := c.newRequest(method, uri, nil); err != nil {
    return err
}
```

ehazlett:
LGTM

Issue #915, opened 2015-03-30
Docker Machine get stuck on IBM Softlayer 


wzymaster:
Here is a derived issue from  #863.

Sorry, I just download the pre-release of 0.2.0 and run docker-machine like below :

./docker-machine create --driver softlayer --softlayer-user leow --softlayer-api-key xxx  --softlayer-cpu 4 --softlayer-memory 8192 --softlayer-disk-size 0 --softlayer-domain imdemocloud.com --softlayer-hostname "leo-demo1" --softlayer-image CENTOS_LATEST_64 --softlayer-region hkg02 leo-demo1
INFO[0000] Creating SSH key...
INFO[0007] Getting Host IP
INFO[0032] Waiting for host to become available
INFO[0154] Waiting for host setup transactions to complete

Unfortunatley, it still get stuck at 'Waiting for host setup transaction to complete" step .

However, I do see this VM machine provisioned on SoftLayer and I can log in. It looks like a clean machine without docker.

Do I run something wrong here ? @hairyhenderson

The below is the build of docker-machine which I am using :

Zhongs-MacBook-Pro:Leo_Demo wzymaster$ ./docker-machine -v
docker-machine version 0.2.0 (45f08d9)

ehazlett:
Currently, only Ubuntu is supported.  Could you try with that and see if you still have the issue?

ehazlett:
I just created a machine on Softlayer fine with the defaults.

hairyhenderson:
@wzymaster - I'm thinking @ehazlett is right - you should use Ubuntu instead (just omit the `--softlayer-image` option and it'll default to the right image). But, can you retry with the `--debug` option (before the `create` subcommand) to make sure it's not something else?

wzymaster:
Here is what I tried but still hanging there after almost 40 minutes :  

Zhongs-MacBook-Pro:Leo_Demo wzymaster$ ./docker-machine create --driver softlayer \
>      --softlayer-user leow \
>      --softlayer-api-key xxxxxxxxxx \
>       --softlayer-cpu 4 \
>      --softlayer-memory 8192 \
>      --softlayer-disk-size  0 \
>      --softlayer-domain imdemocloud.com \
>      --softlayer-hostname "leo-demo"  \
>      --softlayer-region hkg02  \
>     leo-demo
INFO[0000] Creating SSH key...
INFO[0007] Getting Host IP
INFO[0030] Waiting for host to become available
INFO[0195] Waiting for host setup transactions to complete

[...]

PR #1013, opened 2015-04-16
Softlayer driver, correctly store provisioned VM IP address

zekizeki:
Fix for Docker Machine gets stuck on IBM Softlayer #915
The ip address of the provisioned VM was not being stored in the driver IPAddress variable, this variable
was being used to check if the VM ssh service was up. However as the variable was a blank string the actual address being checked defaulted
to localhost:22, as a result the installation steps on the VM only worked if the user was running an ssh daemon on the machine they were using docker-machine from.

Signed-off-by: Rob Smart <smartrob@uk.ibm.com>


ehazlett:
LGTM

hairyhenderson:
:+1: LGTM!

zekizeki:
@ehazlett @hairyhenderson thanks guys ! :)

ehazlett:
@zekizeki thank you!

PR #1023, opened 2015-04-17
Factoring Bats tests

vincentbernat:
Hey!

I noticed that many driver Bats tests are just copy/paste of another one. I wanted to factor them to avoid divergence between tests. The first commits are just a normalization step for the last commit. Unfortunately, I understood how Bats work only at the end (when I wanted to test the changes): it uses a preprocessing phase to replace the `@test` markers, so it is currently not possible to put common tests into another file. I have opened an issue about this: https://github.com/sstephenson/bats/issues/99

Until there is a solution, this PR is useless. I just push it here for reference, in case there is a workaround. Feel free to close it if you don't like unresolvable PR to hang around.


ehazlett:
@vincentbernat yeah i tried the same thing :)  Thanks for the PR -- hopefully we'll get inclusion and can remove the boilerplate. Thanks!

nathanleclaire:
I think that we might be able to make this type of thing work if we wrapped the calls to bats with our own script.  Discussion: https://github.com/sstephenson/bats/issues/100

E.g., `test/integration/run.sh virtualbox`, would set the `NAME` and `DRIVER` environment variables based on `$1`, and then call `bats b2dcommon.bats` which contains common tests.

ehazlett:
@nathanleclaire what does this mean for #1158?

vincentbernat:
@ehazlett don't bother with this PR, I'll check the other PR when it is done.

ehazlett:
Thanks!

Issue #906, opened 2015-03-29
Create a consistent Windows experience


ehazlett:
The current Windows experience with Machine is sub-par from its Mac/Linux counterpart.  This will bring it up to be consistent.  This includes getting Machine running as well as logging and output formatting consistency.

- [x] Use Powershell for Machine interaction (https://github.com/docker/machine/pull/1033)
- [x] Cleanup logging (ascii chars in output)
- [x] Remove dependency on external apps if possible (i.e. ssh, bash)

ahmetalpbalkan:
:+1: following this

znmeb:
A point and click interface for Client Hyper-V is pretty high on my wishlist for Windows. Maybe that belongs under Boot2Docker?

ahmetalpbalkan:
@znmeb care to explain more? What is that?

znmeb:
@ahmetalpbalkan Boot2Docker is a one-click installer. It installs Git for Windows and VirtualBox, then creates a virtual machine running the Boot2Docker ISO. What I'm looking for is something that does the same thing but creates the virtual machine in Windows 8.1 Professional's Client Hyper-V native virtualizer.

The pieces are there - Docker Machine can create Hyper-V machines running Boot2Docker and the Boot2Docker installer can do the same thing with VirtualBox.

garthk:
Isn't the idea to replace the `boot2docker` command line (not VM) with `docker machine` once the latter hits parity? 

If so: this is the right repo, @znmeb, but you might want your own issue for

* making a separate installer available without VirtualBox and
* using `--driver= hyper-v` by default if it's available

… as neither are a matter of bringing the Windows experience of `docker machine` up to parity with Linux and OS X.

ehazlett:
@garthk yes the idea is to replace the boot2docker CLI tool with machine.

@znmeb there are plans to update the current installer mechanism to include machine etc similar to b2d.  we haven't ironed out exactly what that looks like but will have one when machine enters non-beta.

znmeb:
Sounds good to me - I'm chomping at the bit to test something. ;-) I just wish I knew PowerShell and open source Windows installer tooling well enough to build something myself.

ehazlett:
@znmeb i know almost nothing about the installer :)  however, here is the source in case you want to check it out: https://github.com/boot2docker/windows-installer

ehazlett:
[...]

PR #783, opened 2015-03-15
Fixing `config` and `env` bad handling of paths with spaces

hairyhenderson:
Fixes #781

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

PR #1033, opened 2015-04-20
shell selection for env

ehazlett:
This adds PowerShell support and shell selection support (to force selection for `bash`, `fish`, etc) for the `env` command.

Refs #906

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
@@ -658,8 +659,8 @@ func detectShell() (string, error) {
 	// attempt to get the SHELL env var
 	shell := filepath.Base(os.Getenv("SHELL"))
 	// none detected; check for windows env
-	if shell == "." && os.Getenv("windir") != "" {
-		log.Printf("On Windows, please specify either cmd or powershell with the --shell flag.\n\n")
+	if runtime.GOOS == "windows" {
+		log.Printf("On Windows, please specify either 'cmd' or 'powershell' with the --shell flag.\n\n")
 		return "", ErrUnknownShell
 	}
 
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -228,6 +228,10 @@ var Commands = []cli.Command{
 				Name:  "swarm",
 				Usage: "Display the Swarm config instead of the Docker daemon",
 			},
+			cli.StringFlag{
+				Name:  "shell",
+				Usage: "Force environment to be configured for specified shell",
+			},
 			cli.BoolFlag{
 				Name:  "unset, u",
 				Usage: "Unset variables instead of setting them",
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -31,6 +32,10 @@ import (
 	"github.com/docker/machine/libmachine/swarm"
 	"github.com/docker/machine/state"
 	"github.com/docker/machine/utils"
+)
+
+var (
+	ErrUnknownShell = errors.New("unknown shell")
 )
 
 type machineConfig struct {
@@ -647,4 +652,20 @@ func getCertPathInfo(c *cli.Context) libmachine.CertPathInfo {
 		ClientCertPath: clientCertPath,
 		ClientKeyPath:  clientKeyPath,
 	}
+}
+
+func detectShell() (string, error) {
+	// attempt to get the SHELL env var
+	shell := filepath.Base(os.Getenv("SHELL"))
+	// none detected; check for windows env
+	if shell == "." && os.Getenv("windir") != "" {
+		log.Printf("On Windows, please specify either cmd or powershell with the --shell flag.\n\n")
+		return "", ErrUnknownShell
+	}
+
+	if shell == "" {
+		return "", ErrUnknownShell
+	}
+
+	return shell, nil
 }
--- a/commands/create.go
+++ b/commands/create.go
@@ -2,7 +2,6 @@ package commands
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 
 	log "github.com/Sirupsen/logrus"
@@ -89,21 +88,8 @@ func cmdCreate(c *cli.Context) {
 		log.Fatalf("error setting active host: %v", err)
 	}
 
-	info := ""
-	userShell := filepath.Base(os.Getenv("SHELL"))
-
-	switch userShell {
-	case "fish":
-		info = fmt.Sprintf("%s env %s | source", c.App.Name, name)
-	default:
-		info = fmt.Sprintf(`eval "$(%s env %s)"`, c.App.Name, name)
-	}
-
-	log.Infof("%q has been created and is now the active machine.", name)
-
-	if info != "" {
-		log.Infof("To point your Docker client at it, run this in your shell: %s", info)
-	}
+	info := fmt.Sprintf("%s env %s", c.App.Name, name)
+	log.Infof("To point your Docker client at it, run this in your shell: %s", info)
 }
 
 // If the user has specified a driver, they should not see the flags for all
--- a/commands/env.go
+++ b/commands/env.go
@@ -39,7 +39,7 @@ func cmdEnv(c *cli.Context) {
 
 	t := template.New("envConfig")
 
-	usageHint := generateUsageHint(c.Args().First(), userShell)
+	usageHint := generateUsageHint(c.App.Name, c.Args().First(), userShell)
 
 	shellCfg := ShellConfig{
 		DockerCertPath:  "",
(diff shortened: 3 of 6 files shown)


nathanleclaire:
Why not switch to a `text/template`?  It seems likely to be easier to understand and more maintainable given how complex this is getting.

ehazlett:
That works.  Thx

ehazlett:
@nathanleclaire updated PTAL

nathanleclaire:
Cool.  I was thinking something more along the lines of this for the template (to help with readability):

```
{{ .Prefix }}DOCKER_TLS_VERIFY{{ .Delimiter }}{{ .DockerTLSVerify }}{{ .Suffix }}
{{ .Prefix }}DOCKER_HOST{{ .Delimiter }}{{ .DockerHost }}{{ .Suffix }}
{{ .Prefix }}DOCKER_CERT_PATH{{ .Delimiter }}{{ .DockerCertPath }}{{ .Suffix }}
# {{ .UsageHint }}
```

(then the newlines would just be taken out of the suffix)

nathanleclaire:
On second thought, I think we can cut most of the verbosity in the switches down by having a template which does stuff like:

```
{{ if .Shell -eq "fish" }}
set -x DOCKER_TLS_VERIFY {{ .DockerTlsVerify }};
set -x DOCKER_HOST {{ .DockerHost }};
set -x DOCKER_CERT_PATH {{ .DockerCertPath }};
{{ end }}
{{ if .Shell -eq "bash" }}
export DOCKER_TLS_VERIFY={{ .DockerTlsVerify }}
export DOCKER_HOST={{ .DockerHost }}
export DOCKER_CERT_PATH={{ .DockerCertPath }}
{{ end }}
# {{ .UsageHint }}
```

Does it sound good to you?  I think it will be a lot more readable / easier to understand to anyone looking at the code.

ehazlett:
Fixing test.

The problem is the windows shells are very, very picky about whitespace.  For example, if there is a line starting with a blank in powershell it dies.

nathanleclaire:
@ehazlett I see.  That shouldn't be a problem if we're spitting out only the powershell-specific template when we are in powershell, wouldn't it?  The example in that last comment would only output the relevant bits based on the shell.

commands/commands.go
```diff
@@ -636,3 +645,19 @@ func getCertPathInfo(c *cli.Context) libmachine.CertPathInfo {
 		ClientKeyPath:  clientKeyPath,
 	}
 }
+
+func detectShell() (string, error) {
+	// attempt to get the SHELL env var
+	shell := filepath.Base(os.Getenv("SHELL"))
```
nathanleclaire: Is this wrapped in `filepath.Base` so that we get the value we expect in case SHELL contains something like `/usr/bin/zsh`?  Just making sure I understand correctly.

blacktop:
Don't forget **zsh** homies! :sunglasses:

ehazlett:
@blacktop i believe zsh behaves the same as bash -- that's what we do today :)

commands/commands.go
ehazlett: Correct -- this is the current behavior today.

ehazlett:
Tests updated.  @nathanleclaire PTAL

ehazlett:
I think this would be better using some split `text/template` (as @nathanleclaire mentioned) however, there are several nuances regarding whitespace with the various terminals.  I think we should continue on with this and we can revisit if we continue to add / or want to optimize.

blacktop:
@ehazlett, you right, you right. :wink: 

I could have sworn that I used to be able to set the env like with `$(boot2docker shellinit)`.  i.e. `$(docker-machine env dev)`, which gives the error: `export: not valid in this context: shell:`

I see now that I am supposed to type: `eval "$(docker-machine env dev)"` instead.

hairyhenderson:
@blacktop - yeah - `eval` is necessary now (see #795 and #783).

ehazlett:
@nathanleclaire that sound good to you?

nathanleclaire:
@ehazlett Yes, definitely- can you please add some docs?  At least a small paragraph and example explaining that `env` works with `cmd.exe` and Powershell as well.  I will be ready to go on this shortly after I get a chance to actually play around with it a bit, code looks good.

nathanleclaire:
Also, should we maybe not be hardcoding the `docker-machine` binary name?

ehazlett:
@nathanleclaire updated to remove the hardcoded binary.

nathanleclaire:
@ehazlett Nice, thanks - I'm getting some issues when using the command hint on Powershell.  WDYT?

image

nathanleclaire:
Yeah, it works flawlessly if you do:

```console
$ docker-machine env --shell powershell dev | Invoke-Expression
```

nathanleclaire:
Other than that, testing on Powershell and cmd.exe looks good on my end.

ehazlett:
Tests fixed.  Usage hint updated for powershell.

ping @nathanleclaire

nathanleclaire:
LGTM let's do this :+1: 

Created separate issue to track documenting this : https://github.com/docker/machine/issues/1054

commands/commands.go
```diff
@@ -644,3 +653,19 @@ func getCertPathInfo(c *cli.Context) libmachine.CertPathInfo {
...
+	// none detected; check for windows env
+	if shell == "." && os.Getenv("windir") != "" {
+		log.Printf("On Windows, please specify either cmd or powershell with the --shell flag.\n\n")
```
ahmetalpbalkan: might be nice to put cmd, powershell strings in single quotes for readability.

ehazlett: good point -- thx!

commands/commands.go
```diff
@@ -644,3 +653,19 @@ func getCertPathInfo(c *cli.Context) libmachine.CertPathInfo {
 		ClientKeyPath:  clientKeyPath,
 	}
 }
+
+func detectShell() (string, error) {
+	// attempt to get the SHELL env var
+	shell := filepath.Base(os.Getenv("SHELL"))
+	// none detected; check for windows env
+	if shell == "." && os.Getenv("windir") != "" {
```
ahmetalpbalkan: I think this check should be:

    if runtime.GOOS == "windows"

although  `shell == "." && os.Getenv("windir") != ""` implies the same thing, you should be calling filepath.Base after `if windows` b/c it's not used in windows case at all.

also don't rely on `windir` :) People can run docker.exe with unsetting the env. GOOS is safe to use.

ahmetalpbalkan:
@ehazlett if they're working fine, LGTM

It's sort of hard to trace from the code by eye. So maybe if can you please paste the outputs/screenshots from cmd/powershell (without piping to invoke-expression) I can double check it for you.

commands/commands.go
ehazlett: +1

ehazlett:
2015-04-30-110610_906x319_scrot

ehazlett:
2015-04-30-121847_1023x329_scrot

masaeedu:
Could the docs be updated to show how this can be used (as opposed to msysgit)?

nathanleclaire:
@masaeedu Take a look at the `--shell` flag documentation [here](http://docs.docker.com/machine/#env),

StefanScherer:
@nathanleclaire Great, thanks for updating the docs.

One small thing for the cmd.exe is that the `#` sign is not a valid comment sign for the cmd.exe. 

Here is a small proposal how it could look like for cmd.exe with `rem` and a command that sets the environment variables:

```cmd
$ docker-machine.exe env --shell cmd dev
set DOCKER_TLS_VERIFY=1
set DOCKER_HOST=tcp://192.168.99.101:2376
set DOCKER_CERT_PATH=C:\Users\captain\.docker\machine\machines\dev
set DOCKER_MACHINE_NAME=dev
rem Run this command to configure your shell:
rem for /f "tokens=*" %i in ('docker-machine.exe env --shell=cmd dev') do %i
```

And in the powershell sample there is missing the is missing the machine name `dev`:

```powershell
# docker-machine.exe env --shell=powershell dev | Invoke-Expression
```

ahmetalpbalkan:
@stefanscherer cmd.exe output of env cmd was designed for copy-paste.

Your proposal looks interesting. Care to put that in a proposal/pull request?

StefanScherer:
@ahmetalpbalkan Have a look at #1493 ;-)

masaeedu:
@StefanScherer Re: missing machine name in the powershell output; that is odd, because I'm using 0.3.0 and it works correctly for me:

    PS C:\Users\msaeedud> docker-machine env --shell powershell boot2docker
    $Env:DOCKER_TLS_VERIFY = "1"
    $Env:DOCKER_HOST = "tcp://192.168.155.24:2376"
    $Env:DOCKER_CERT_PATH = "C:\Users\msaeedud\.docker\machine\machines\boot2docker"
    $Env:DOCKER_MACHINE_NAME = "boot2docker"
    # C:\ProgramData\chocolatey\lib\docker-machine\bin\docker-machine.exe env --shell=powershell boot2docker | Invoke-Expression

Note how my machine name, `boot2docker` showed up in the comment.

StefanScherer:
@masaeedu I have seen that the sourcecode is correct. It is only missing in the docs.

PR #1037, opened 2015-04-21
Moves the 'stopping machine' log below error log

### LINKED ISSUE #1032: Proposal: Change log output when upgrading an unsupported driver
When upgrading a Docker machine with the `vmwarefusion` driver, you get the following error message:

```
$ docker-machine upgrade      
INFO[0000] Stopping machine to do the upgrade...        
ERRO[0000] Upgrade functionality is currently not supported for these providers, as they use a custom ISO.
```

I'd like to propose to remove the first `INFO` line, as it isn't executed at all (the VM is still running after this), and it doesn't make sense to stop a machine if you know beforehand that you cannot upgrade it.

RickWieman:
As it doesn't make sense to first tell that the machine will be stopped for upgrade, but then throw an error that it isn't possible at all, I moved the 'stopping machine' info log below the error log.
This seems to be a legit change, as the machine isn't stopped at all...

Fixes #1032.

### DIFF
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -50,12 +50,12 @@ func (provisioner *Boot2DockerProvisioner) Service(name string, action pkgaction
 }
 
 func (provisioner *Boot2DockerProvisioner) upgradeIso() error {
-	log.Infof("Stopping machine to do the upgrade...")
-
 	switch provisioner.Driver.DriverName() {
 	case "vmwarefusion", "vmwarevsphere":
 		return errors.New("Upgrade functionality is currently not supported for these providers, as they use a custom ISO.")
 	}
+
+	log.Info("Stopping machine to do the upgrade...")
 
 	if err := provisioner.Driver.Stop(); err != nil {
 		return err


libmachine/provision/boot2docker.go
```diff
@@ -50,13 +50,13 @@ func (provisioner *Boot2DockerProvisioner) Service(name string, action pkgaction
...
 
 func (provisioner *Boot2DockerProvisioner) upgradeIso() error {
-	log.Infof("Stopping machine to do the upgrade...")
-
 	switch provisioner.Driver.DriverName() {
 	case "vmwarefusion", "vmwarevsphere":
 		return errors.New("Upgrade functionality is currently not supported for these providers, as they use a custom ISO.")
 	}
 
+	log.Infof("Stopping machine to do the upgrade...")
```
ehazlett: I know this wasn't from you but we can just do `log.Info(..)` here (no `Infof`).  Would you mind updating?

RickWieman: Ah, you're right. I'll update and rebase ASAP :+1:

RickWieman:
Fixed it @ehazlett :blush:

ehazlett:
Thanks!

LGTM

Issue #1041, opened 2015-04-22
Cannot build 'env' output from 'inspect' output


hamiltont:
I'm trying to interface with `docker-machine` from an external program using `inspect`. Some information seems to be missing. Using the digitalocean driver, I see 

```
$ docker-machine env
export DOCKER_TLS_VERIFY=1
export DOCKER_CERT_PATH="/Users/hamiltont/.docker/machine/machines/staging"
export DOCKER_HOST=tcp://104.236.9.119:2376
```

When I run `inspect`, I don't find the `2376` anywhere in the output, and the value of `TlsVerify` seems to disagree with `DOCKER_TLS_VERIFY` above. I know that [2376 is the recommended port](https://github.com/docker/docker/blob/master/docs/sources/articles/https.md), but without it being output from `inspect` I'm hard-coding an implementation detail of `docker-machine` if I assume it's always 2376. 

```
$ docker-machine inspect
{
    "DriverName": "digitalocean",
    "Driver": {
        "AccessToken": "<snip>",
        "DropletID": <snip>,
        "DropletName": "",
        "Image": "ubuntu-14-04-x64",
        "MachineName": "staging",
        "IPAddress": "104.236.9.119",
        "Region": "nyc3",
        "SSHKeyID": 773910,
        "SSHUser": "root",
        "SSHPort": 22,
        "Size": "512mb",
        "IPv6": false,
        "Backups": false,
        "PrivateNetworking": false,
        "CaCertPath": "/Users/hamiltont/.docker/machine/certs/ca.pem",
        "PrivateKeyPath": "/Users/hamiltont/.docker/machine/certs/ca-key.pem",
        "DriverKeyPath": "",
        "SwarmMaster": false,
        "SwarmHost": "tcp://0.0.0.0:3376",
        "SwarmDiscovery": ""
    },
    "StorePath": "/Users/hamiltont/.docker/machine/machines/staging",
    "HostOptions": {
        "Driver": "",
        "Memory": 0,
        "Disk": 0,
        "EngineOptions": {
            "Dns": null,
            "GraphDir": "",
            "Ipv6": false,
            "Labels": null,
            "LogLevel": "",
            "StorageDriver": "",
            "SelinuxEnabled": false,
            "TlsCaCert": "",
            "TlsCert": "",
            "TlsKey": "",
            "TlsVerify": false,
            "RegistryMirror": null
        },
        "SwarmOptions": {
            "IsSwarm": false,
            "Address": "",
            "Discovery": "",
[...]

PR #1043, opened 2015-04-22
Adding/Renaming IPAddress properties for consistency across drivers

hairyhenderson:
Addresses #1041, and vaguely mentioned during #921...

This is primarily for consistent UX on `docker inspect` commands. Specifically, I should be able to retrieve a host's IP address for any driver with `docker-machine inspect -f '{{.Driver.IPAddress}}'`

For each driver (except probably `fakedriver` and `none`), an `IPAddress` field should be added to the `Driver` struct.

I'll update this list as I get the different drivers updated:
- [x] amazonec2 _(previously done)_
- [x] azure
- [x] digitalocean _(previously done)_
- [x] google
- [x] hyperv
- [x] none
- [x] openstack _(named `Ip` instead)_
- [x] softlayer _(previously done)_
- [x] virtualbox
- [x] vmwarefusion _(previously done)_
- [x] vmwarevcloudair _(named `PublicIP` instead)_
- [x] vmwarevsphere

### DIFF
--- a/commands/inspect_test.go
+++ b/commands/inspect_test.go
@@ -29,7 +29,7 @@ func TestCmdInspectFormat(t *testing.T) {
 	assert.Equal(t, "\"none\"", actual)
 
 	actual, _ = runInspectCommand(t, []string{"--format", "{{prettyjson .Driver}}"})
-	assert.Equal(t, "{\n    \"URL\": \"unix:///var/run/docker.sock\"\n}", actual)
+	assert.Equal(t, "{\n    \"IPAddress\": \"\",\n    \"URL\": \"unix:///var/run/docker.sock\"\n}", actual)
 }
 
 func runInspectCommand(t *testing.T, args []string) (string, *libmachine.Host) {
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -21,6 +21,7 @@ import (
 )
 
 type Driver struct {
+	IPAddress               string
 	MachineName             string
 	SubscriptionID          string
 	SubscriptionCert        string
@@ -312,7 +313,9 @@ func (d *Driver) Start() error {
 		return err
 	}
 
-	return nil
+	var err error
+	d.IPAddress, err = d.GetIP()
+	return err
 }
 
 func (d *Driver) Stop() error {
@@ -329,7 +332,12 @@ func (d *Driver) Stop() error {
 
 	log.Debugf("stopping %s", d.MachineName)
 
-	return vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName)
+	if err := vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName); err != nil {
+		return err
+	}
+
+	d.IPAddress = ""
+	return nil
 }
 
 func (d *Driver) Remove() error {
@@ -364,7 +372,8 @@ func (d *Driver) Restart() error {
 		return err
 	}
 
-	return nil
+	d.IPAddress, err = d.GetIP()
+	return err
 }
 
 func (d *Driver) Kill() error {
@@ -381,7 +390,12 @@ func (d *Driver) Kill() error {
 
 	log.Debugf("killing %s", d.MachineName)
 
-	return vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName)
+	if err := vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName); err != nil {
+		return err
+	}
+
+	d.IPAddress = ""
+	return nil
 }
 
 func generateVMName() string {
--- a/drivers/google/google.go
+++ b/drivers/google/google.go
@@ -14,6 +14,7 @@ import (
 
 // Driver is a struct compatible with the docker.hosts.drivers.Driver interface.
 type Driver struct {
+	IPAddress      string
 	MachineName    string
 	SSHUser        string
 	SSHPort        int
@@ -250,7 +251,11 @@ func (d *Driver) Start() error {
 	if err != nil {
 		return err
 	}
-	return c.createInstance(d)
+	if err = c.createInstance(d); err != nil {
+		return err
+	}
+	d.IPAddress, err = d.GetIP()
+	return err
 }
 
 // Stop deletes the GCE instance, but keeps the disk.
@@ -259,7 +264,11 @@ func (d *Driver) Stop() error {
 	if err != nil {
 		return err
 	}
-	return c.deleteInstance()
+	if err = c.deleteInstance(); err != nil {
+		return err
+	}
+	d.IPAddress = ""
+	return nil
 }
 
 // Remove deletes the GCE instance and the disk.
--- a/drivers/hyperv/hyperv_windows.go
+++ b/drivers/hyperv/hyperv_windows.go
@@ -19,6 +19,7 @@ import (
 )
 
 type Driver struct {
+	IPAddress      string
 	SSHUser        string
 	SSHPort        int
 	storePath      string
@@ -334,7 +335,13 @@ func (d *Driver) Start() error {
 	if err != nil {
 		return err
 	}
-	return d.wait()
+
+	if err := d.wait(); err != nil {
+		return err
+	}
+
+	d.IPAddress, err = d.GetIP()
+	return err
 }
 
 func (d *Driver) Stop() error {
@@ -356,6 +363,7 @@ func (d *Driver) Stop() error {
 			break
 		}
 	}
+	d.IPAddress = ""
 	return nil
 }
 
@@ -406,6 +414,7 @@ func (d *Driver) Kill() error {
 			break
 		}
 	}
+	d.IPAddress = ""
 	return nil
 }
 
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -30,6 +30,7 @@ const (
 )
 
 type Driver struct {
+	IPAddress           string
 	CPU                 int
 	MachineName         string
 	SSHUser             string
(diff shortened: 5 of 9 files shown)


nathanleclaire:
Hmm so existential question... Do you feel like this should always access the user's local "cache", or do you think it should actually calculate the value?  In other words, consistency or availability guarantee?

hairyhenderson:
@nathanleclaire - well, is there any situation where the IP address would change while a machine is running? I mean, technically it _could_...

I suppose cache makes the most sense to me, since it makes the `inspect` command faster...

hairyhenderson:
Ok, this is ready to be poked and prodded... I've tested out `none` and `virtualbox`, but I'd appreciate if some testing could be done on the other drivers I messed with (`azure`, `google`, `hyperv`, `openstack` - probably `rackspace` too, `vmwarevcloudair`, and `vmwarevsphere`)

In `openstack`'s case, I _renamed_ the `Ip` property to `IPAddress`, but in `vmwarevcloudair`'s case I added a new `IPAddress` property that essentially mirrors `PublicIP` - I wasn't really sure there what the right behaviour was.

Also, in some drivers, `IPAddress` will disappear when the machine is stopped, while in others the `IPAddress` persists (but may change once the machine is started again). This inconsistent behaviour could probably use some harmonizing, but I'm not really sure which behaviour is more desirable...

ehazlett:
would you mind rebasing -- after that i'll check it out.  thanks!

hairyhenderson:
@ehazlett - rebased :)

commands/inspect_test.go
```diff
@@ -29,7 +29,7 @@ func TestCmdInspectFormat(t *testing.T) {
 	assert.Equal(t, "\"none\"", actual)
 
 	actual, _ = runInspectCommand(t, []string{"--format", "{{prettyjson .Driver}}"})
```
nathanleclaire: Any reason to use `prettyjson` instead of the usual format here?  Might save you some whitespace headaches

drivers/azure/azure.go
```diff
@@ -312,6 +313,8 @@ func (d *Driver) Start() error {
 		return err
 	}
 
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Why swallow the error?

drivers/azure/azure.go
```diff
@@ -329,7 +332,12 @@ func (d *Driver) Stop() error {
 
 	log.Debugf("stopping %s", d.MachineName)
 
-	return vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName)
+	if err := vmClient.ShutdownRole(d.MachineName, d.MachineName, d.MachineName); err != nil {
+		return err
+	}
+
+	d.IPAddress = ""
```
nathanleclaire: Why not set the `d.IPAddress` field above and leave the `return` line like it was before?

drivers/google/google.go
```diff
@@ -242,7 +243,11 @@ func (d *Driver) Start() error {
 	if err != nil {
 		return err
 	}
-	return c.createInstance(d)
+	if err = c.createInstance(d); err != nil {
+		return err
+	}
+	d.IPAddress, err = d.GetIP()
```
nathanleclaire: Needs error check

drivers/hyperv/hyperv_windows.go
```diff
@@ -334,7 +335,13 @@ func (d *Driver) Start() error {
 	if err != nil {
 		return err
 	}
-	return d.wait()
+
+	if err := d.wait(); err != nil {
+		return err
+	}
+
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Why swallow the error?

drivers/virtualbox/virtualbox.go
```diff
@@ -399,6 +400,8 @@ func (d *Driver) Start() error {
 		log.Infof("VM not in restartable state")
 	}
 
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Same

drivers/virtualbox/virtualbox.go
```diff
@@ -434,6 +440,7 @@ func (d *Driver) Remove() error {
 			return err
 		}
 	}
+	d.IPAddress = ""
```
nathanleclaire: What if the `vbm` command fails and the machine is still around, including with an IP address?

drivers/vmwarevcloudair/vcloudair.go
```diff
@@ -418,8 +419,9 @@ func (d *Driver) Create() error {
 	// Set VAppID with ID of the created VApp
 	d.VAppID = vapp.VApp.ID
 
-	return nil
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Swallowed error

drivers/vmwarevcloudair/vcloudair.go
```diff
@@ -540,8 +543,9 @@ func (d *Driver) Start() error {
 		return err
 	}
 
-	return nil
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Error check

drivers/vmwarevcloudair/vcloudair.go
```diff
@@ -640,6 +645,8 @@ func (d *Driver) Restart() error {
 		return err
 	}
 
+	d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Same

drivers/vmwarevsphere/vsphere.go
```diff
@@ -406,6 +407,8 @@ func (d *Driver) Start() error {
 			return err
 		}
 
+		d.IPAddress, _ = d.GetIP()
```
nathanleclaire: Same

commands/inspect_test.go
hairyhenderson: The point was to make sure the prettyjson formatting worked as expected :)

drivers/azure/azure.go
hairyhenderson: I didn't want to risk leaving the machine in an IP-less state if `vmClient.ShutdownRole` errors.

nathanleclaire: Ah yes, that seems like the correct behavior, as I commented elsewhere.  Gotcha.

commands/inspect_test.go
nathanleclaire: Ah, OK, so it's an existing test for `prettyjson`.  Gotcha.

drivers/google/google.go
hairyhenderson: I think I originally meant to `return err` below, but then I thought it would be weird to return GetIP's error. But that's probably not at all weird, so I'll fix this :)

(this also applies to all the places I did `d.IPAddress, _ = d.GetIP()`)

drivers/virtualbox/virtualbox.go
hairyhenderson: Thanks - missed this one :)

drivers/google/google.go
nathanleclaire: Yep seems reasonable to me since error "bubble up" is the general approach we take.  It should produce output like:

```
Error starting machine: error getting ip: dial TCP timed out
```

drivers/azure/azure.go
hairyhenderson: Fixed

drivers/google/google.go
hairyhenderson: fixed

drivers/hyperv/hyperv_windows.go
hairyhenderson: fixed

drivers/virtualbox/virtualbox.go
hairyhenderson: fixed

drivers/virtualbox/virtualbox.go
hairyhenderson: fixed

drivers/vmwarevcloudair/vcloudair.go
hairyhenderson: fixed

drivers/vmwarevcloudair/vcloudair.go
hairyhenderson: fixed

drivers/vmwarevcloudair/vcloudair.go
hairyhenderson: fixed

drivers/vmwarevsphere/vsphere.go
hairyhenderson: fixed

hairyhenderson:
@nathanleclaire PTAL... Note I plan on squashing this soon... the number of commits is getting ridiculous :)

nathanleclaire:
Indeed!  Please squash down to one commit when you get a chance.

hairyhenderson:
@nathanleclaire ok, it's squashed now

ehazlett:
LGTM and works as expected -- @nathanleclaire would you mind taking a look as well?

ehazlett:
@nathanleclaire sorry -- GH didn't update the comment stream -- yours weren't there -- disregard :)

nathanleclaire:
So, come to think of it, what's the purpose of modifying the value on `Remove` here?  If the `Remove` succeeds, the reference will be cleaned up anyway, so why change the string in the struct ourselves?

ggiamarchi:
LGTM

hairyhenderson:
@nathanleclaire - good point :) I can pull out the changes in `Remove`

hairyhenderson:
Ok, I've pulled out the changes in `Remove()`, rebased, and squashed again. PTAL @nathanleclaire

nathanleclaire:
LGTM

hairyhenderson:
Thanks!

PR #1044, opened 2015-04-22
Enhancing `docker-machine ip` to support multiple host arguments

### LINKED ISSUE #999: `machine ip` chained arguments
## What I do

I use the `ip` command to get the ip list of my two servers :

```bash
~$ docker-machine ls
NAME    ACTIVE   DRIVER         STATE     URL                         SWARM
nodeA            digitalocean   Running   tcp://x.x.x.x:2376   
nodeB   *        digitalocean   Running   tcp://y.y.y.y:2376   

~$ 
```

## What I get

I only get the first IP of the list :

```bash
~$ docker-machine ip $(docker-machine ls -q)
x.x.x.x

~$ 
```

## What I expected

I expected to have the same behaviour as the `rm` command.
So I expected to see the following ouput :

```bash
~$ docker-machine ip $(docker-machine ls -q)
x.x.x.x
y.y.y.y

~$ 
```

hairyhenderson:
Fixes #999

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -234,7 +234,7 @@ var Commands = []cli.Command{
 	{
 		Name:        "ip",
 		Usage:       "Get the IP address of a machine",
-		Description: "Argument is a machine name. Will use the active machine if none is provided.",
+		Description: "Argument(s) are one or more machine names. Will use the active machine if none is provided.",
 		Action:      cmdIp,
 	},
 	{
@@ -342,6 +342,7 @@ func machineCommand(actionName string, host *libmachine.Host, errorChan chan<- e
 		"restart":       host.Restart,
 		"kill":          host.Kill,
 		"upgrade":       host.Upgrade,
+		"ip":            host.PrintIP,
 	}
 
 	log.Debugf("command=%s machine=%s", actionName, host.Name)
--- a/commands/ip.go
+++ b/commands/ip.go
@@ -1,18 +1,13 @@
 package commands
 
 import (
-	"fmt"
-
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/codegangsta/cli"
 )
 
 func cmdIp(c *cli.Context) {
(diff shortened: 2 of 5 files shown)


nathanleclaire:
Why not use  `runActionWithContext` like `start` and `stop` commands do?  This logic is already implemented there.

hairyhenderson:
@nathanleclaire - that's exactly why this is still a WIP ;)

I looked at `runActionWithContext` for inspiration, but there was some reason I thought I couldn't use it. I'll give it another shot with that :)

hairyhenderson:
Oh yeah, I remember now... `runActionWithContext` is geared towards methods that don't return anything  other than an error. But I think I might be able to work around that by wrapping the GetIP() call...

hairyhenderson:
@nathanleclaire - PTAL - have I done this sanely? :)

nathanleclaire:
Looks like you've definitely got the right idea - how about adding a `GetIP` method to `Host` like the way the others are done?  You should be able to take the closure that you have almost verbatim for this.

nathanleclaire:
Oh, and also, let's just call the action "ip" for consistency's sake - unless I'm mistaken, they map directly to the commands that you invoke on the CLI.

hairyhenderson:
Yeah, ok that sounds better... :)

hairyhenderson:
Ok, this time I think I like it. I'm going to work on some tests, and get some documentation worked out now.

nathanleclaire:
Nice.  Code LGTM

hairyhenderson:
@nathanleclaire - ok, I've added some tests and docs, and squashed. I think this is done now - PTAL :)

nathanleclaire:
LGTM!  Appreciate you taking the initiative on the test and doc.

nathanleclaire:
cc @ehazlett for review

ehazlett:
LGTM

PR #1058, opened 2015-04-26
Reporting success of each machine when removing multiples

### LINKED ISSUE #1053: Remove multiple machines with a single command
```
~> docker-machine rm swarm-master swarm-node-01 swarm-node-02
INFO[0000] The machine was successfully removed. 
```
removed only a single machine.

The command should be updated to remove multiple machines. It would be better if it can take a regular expression, this can be very useful for (re)deployment of a cluster.

hairyhenderson:
Fixes #1053

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -34,10 +34,11 @@ func cmdRm(c *cli.Context) {
 		if err := mcn.Remove(host, force); err != nil {
 			log.Errorf("Error removing machine %s: %s", host, err)
 			isError = true
+		} else {
+			log.Infof("Successfully removed %s", host)
 		}
 	}
 	if isError {
 		log.Fatal("There was an error removing a machine. To force remove it, pass the -f option. Warning: this might leave it running on the provider.")
 	}
-	log.Print("The machine was successfully removed.")
 }


nathanleclaire:
Nice

commands/rm.go
```diff
@@ -34,10 +34,11 @@ func cmdRm(c *cli.Context) {
 		if err := mcn.Remove(host, force); err != nil {
 			log.Errorf("Error removing machine %s: %s", host, err)
 			isError = true
+		} else {
+			log.Infof("Successfully removed %s", host)
```
nathanleclaire: Why have this here instead of down on line 42 like it was before?

hairyhenderson: So that success is reported immediately after each host is removed, and so that users know _which_ host(s) got removed successfully.

nathanleclaire: Hm, so in my opinion we do not really need that little `if isError` block at the end since the users will be notified if the removal fails, so maybe we can take that out with the PR too?  cc @ehazlett WDYT?

Also, eventually we will need to move over this function to the new method of doing the actions concurrently (that `runActionWithContext` thing), since this way is old (it was the first multiarg of its kind).  But that can be done in another PR.

hairyhenderson: I dunno. I like the `if isError` block. It reminds users that even though 99 of their 100 specified hosts were removed successfully, there _was_ one that had a problem. And it gives them some good hints on what to do about it.

+1 on porting this to `runActionWithContext`. Concurrency for the win!

nathanleclaire:
> I dunno. I like the if isError block. It reminds users that even though 99 of their 100 specified hosts were removed successfully, there was one that had a problem. And it gives them some good hints on what to do about it.

Hm.  OK.  I mostly think of exit codes as the thing for that.  I don't feel super strongly about it so I'm cool with leaving it in if you like it.

ehazlett:
I think telling which machines were removed and which ones failed is beneficial.

arun-gupta:
+1 on @ehazlett's comment.

nathanleclaire:
LGTM

ehazlett:
LGTM

nathanleclaire:
_NICE_

Issue #1036, opened 2015-04-21
[AmazonEC2 driver] -- make "Autoassign Public IP address" configurable 


andrea-spoldi:
see issue #581 

"Autoassign Public IP address" (https://github.com/docker/machine/blob/master/drivers/amazonec2/amz/ec2.go#L189) should be configurabile in order to create a new host into a VPC's private subnet, like 

--auto-public-ip = "true" >> a public Elastic IP Address is assigned and port 22 is checked against it for monitoring when the new host come up 

--auto-public-ip = "false" >> do not set any public IP address and check port 22 of the new instance's private IP address (retrieved via EC2's APIs)

AS

sthulb:
Mine! I'll look into this tomorrow morning.

CivJ:
:+1: 

I would love to see this!

sthulb:
@CivJ / @andrea-spoldi: I've added this in #1063, mind testing to see if it's what you guys wanted?

CivJ:
I was about to test this when I noticed the [build failure] (https://travis-ci.org/docker/machine/builds/60405134). I'll hold off until that's resolved. @sthulb thanks for getting this in so quickly, I'm pretty excited to see it in action.

sthulb:
Fixed locally.

CivJ:
Creation and SSH are working for me:
```
$ ./docker-machine_darwin-amd64 create -d amazonec2 --amazonec2-access-key=$AWS_ACCESS_KEY_ID --amazonec2-secret-key=$AWS_SECRET_ACCESS_KEY --amazonec2-vpc-id=$AWS_VPC_ID --amazonec2-private-address-only  dm-03-poc
Private: %#v true
INFO[0000] Launching instance...                        
INFO[0182] "dm-03-poc" has been created and is now the active machine. 
INFO[0182] To point your Docker client at it, run this in your shell: eval "$(docker-machine_darwin-amd64 env dm-03-poc)"
```
However, I may have encountered a bug (possibly in the docker client):
```
$ docker -v
Docker version 1.6.0, build 4749651
$ ./docker-machine_darwin-amd64 ls
```
I cannot list my machines, and we hang here for at least 10 minutes. I suspect we're hanging on the API call to AWS that fetches the status. As you can see, I'm running the OXS docker-machine build.

CivJ:
This is working for me now! Maybe it was a problem on Amazon's side?

```
$ ./docker-machine_darwin-amd64 ls
[...]

PR #1063, opened 2015-04-28
Add --amazonec2-private-address-only

sthulb:
For #1036.

I've added `--amazonec2-private-address-only`. This will stop machine from assigning a public address.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -69,6 +69,7 @@ type Driver struct {
 	keyPath             string
 	RequestSpotInstance bool
 	SpotPrice           string
+	PrivateIPOnly       bool
 }
 
 func init() {
@@ -164,6 +165,10 @@ func GetCreateFlags() []cli.Flag {
 			Usage: "AWS spot instance bid price (in dollar)",
 			Value: "0.50",
 		},
+		cli.BoolFlag{
+			Name:  "amazonec2-private-address-only",
+			Usage: "Only use a private IP address",
+		},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


drivers/amazonec2/amazonec2.go
```diff
@@ -221,6 +226,9 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.SwarmDiscovery = flags.String("swarm-discovery")
 	d.SSHUser = flags.String("amazonec2-ssh-user")
 	d.SSHPort = 22
+	d.PrivateIPOnly = flags.Bool("amazonec2-private-address-only")
+
+	fmt.Println("Private: %#v", d.PrivateIPOnly)
```
ehazlett: Could we remove this or throw to debug?

drivers/amazonec2/amazonec2.go
```diff
@@ -372,7 +380,7 @@ func (d *Driver) Create() error {
 			return fmt.Errorf("Error get instance: %s", err)
 		}
 	} else {
-		inst, err := d.getClient().RunInstance(d.AMI, d.InstanceType, d.Zone, 1, 1, d.SecurityGroupId, d.KeyName, d.SubnetId, bdm, d.IamInstanceProfile)
+		inst, err := d.getClient().RunInstance(d.AMI, d.InstanceType, d.Zone, 1, 1, d.SecurityGroupId, d.KeyName, d.SubnetId, bdm, d.IamInstanceProfile, d.PrivateIPOnly)
```
nathanleclaire: Heh.  I don't really expect you to do it in this PR, but if we have so many options it would be better to make an "options" struct for this and have it passed in as a struct literal.

ehazlett: @nathanleclaire i was thinking the same thing the other day.  i think it would be good in a follow up PR.

sthulb: This is low priority for me... I'm working on replacing the entire driver.

ehazlett:
LGTM

PR #1064, opened 2015-04-28
Adding localhost to the list of alt_names

cthulhuology:
When attempting to connect to the docker api from the machine itself,
the TLS verification of the certificate checked against the public
IP address of the primary interface.  This is undesirable on hosts
which have NAT rules that block access to that address by default.

Adding "localhost" to the list of alt_names allows the cert to be
verified and connections to localhost (either 127.0.0.1 or [::1]) to
the port to pass verification. Otherwise one would need to disable
verification just to connect to the local docker instance.

### DIFF
--- a/libmachine/provision/utils.go
+++ b/libmachine/provision/utils.go
@@ -76,7 +76,7 @@ func ConfigureAuth(p Provisioner, authOptions auth.AuthOptions) error {
 	// TODO: Switch to passing just authOptions to this func
 	// instead of all these individual fields
 	err = utils.GenerateCert(
-		[]string{ip},
+		[]string{ip,"localhost"},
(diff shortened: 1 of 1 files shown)


sthulb:
+1

dmp42:
@cthulhuology can you rebase?

@docker/machine-maintainers wdyt?

nathanleclaire:
Seems reasonable and is a needed step to any sort of NAT-only mode for VirtualBox - given a rebase I'm happy to merge.

dgageot:
@docker/machine-maintainers Here's a rebased PR https://github.com/docker/machine/pull/2065

dmp42:
#2065 has been merged.

Issue #1066, opened 2015-04-28
[GCE] Happened to panic runtime error, Therefore not added user, pem file is not copied to the local


zchee:
Happened to panic runtime error when launch google compute engine instance.

This problem happen in the HEAD build.
Doesn't seem occurred in the pre-release 0.2.0 version.

It seems instance is created successfully.
but, **`.pem` is not copied to the local.**
Therefore, docker-machine {ssh, env} commands are not available.

Duplicate https://github.com/docker/machine/issues/771 ?

See result.
```
./docker-machine_darwin-amd64 create docker-machine-google --driver google --google-disk-size "50" --google-disk-type "pd-ssd" --google-machine-ty
pe "f1-micro" --google-project xxxxx-xxxxx-12345 --google-zone "asia-east1-c"
INFO[0000] Opening auth URL in browser.
INFO[0000] https://accounts.google.com/o/oauth2/auth?client_id=....
INFO[0000] If the URL doesn't open please open it manually and copy the code here.
Enter code: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
INFO[0006] Got code: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
INFO[0006] Saving token in /Users/zchee/.docker/machine/machines/docker-machine-google/gce_token
INFO[0006] Creating host...
INFO[0007] Generating SSH Key
INFO[0008] Creating instance.
INFO[0008] Creating firewall rule.
INFO[0013] Waiting for Instance...
INFO[0093] Uploading SSH Key
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x103a1b]

goroutine 16 [running]:
runtime.panic(0x6d3880, 0xb2dd84)
        /usr/src/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/docker/machine/drivers/google.(*ComputeUtil).createInstance(0x208c6b080, 0x208c322d0, 0x0, 0x0)
        /go/src/github.com/docker/machine/drivers/google/compute_util.go:218 +0xe2b
[...]

PR #1067, opened 2015-04-28
Update GCE imageName to v20150316

zchee:
I happened error of https://github.com/docker/machine/issues/1066 when the test launch instance, but this change is not the cause.

/cc @ehazlett


ehazlett:
LGTM

Issue #905, opened 2015-03-29
Expand Provisioning capabilities


ehazlett:
This will enable the provisioning capabilities for other base distributions besides boot2docker and Ubuntu.  For example, RedHat like distributions.

flavio:
Hi, where is this work being done? I would like to help adding SUSE/openSUSE support.

ehazlett:
@flavio not yet -- i was planning on starting on centos/fedora this week.

flavio:
Any update on this topic?

ehazlett:
@flavio #1090

jiangshengwu:
Looking forward to CentOS support, and maybe `os-release` shouldn't be the only way to detect operating system.

ehazlett:
@jiangshengwu what else would you suggest?

jiangshengwu:
@ehazlett As you've been working on the expanding of provision, maybe you've noticed that some linux release versions don't have `os-release`. Such as CentOS6.x or 7, but `centos-release` instead.

Other ways of release version detection can be useful: `redhat-release` or `lsb_release` command which is more universal.

ehazlett:
@jiangshengwu ah i see -- thanks -- we'll keep it in mind :)

Tritlo:
could we get support for the default AWS Linux? Running the AWS Linux AMIs is cheaper than running the ubuntu ones.

PR #1090, opened 2015-05-01
RedHat provisioning

ehazlett:
This enables Machine to provision RedHat (RHEL).  Tested on RHEL 7.0.

Refs #905

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1586,7 +1586,7 @@ The default base operating system for Machine is Boot2Docker on local providers
 (VirtualBox, Fusion, Hyper-V, etc) and the latest Ubuntu LTS supported
 by the cloud provider.  RedHat Enterprise Linux is also supported.  To use
 RHEL, you will need to select the image accordingly with the provider.  For
-example, in Amazon EC2, you could use "ami-12663b7a" as the
+example, in Amazon EC2, you could use a RedHat 7.1 AMI ("ami-12663b7a") as the
 `--amazonec2-ami` option which create an instance using RHEL 7.1 64-bit.
 
 ## Release Notes
--- a/docs/index.md
+++ b/docs/index.md
@@ -1110,6 +1110,7 @@ Options:
 
 The DigitalOcean driver will use `ubuntu-14-04-x64` as the default image.
 
+<<<<<<< HEAD
 Environment variables and default values:
 
 | CLI option                          | Environment variable              | Default  |
@@ -1121,6 +1122,22 @@ Environment variables and default values:
 | `--digitalocean-ipv6`               | `DIGITALOCEAN_IPV6`               | `false`  |
 | `--digitalocean-private-networking` | `DIGITALOCEAN_PRIVATE_NETWORKING` | `false`  |
 | `--digitalocean-backups`            | `DIGITALOCEAN_BACKUPS`            | `false`  |
+=======
+#### exoscale
+Create machines on [exoscale](https://www.exoscale.ch/).
+
+Get your API key and API secret key from [API details](https://portal.exoscale.ch/account/api) and pass them to `machine create` with the `--exoscale-api-key` and `--exoscale-api-secret-key` options.
+
+Options:
+
+ - `--exoscale-api-key`: Your API key.
+ - `--exoscale-api-secret-key`: Your API secret key.
+ - `--exoscale-instance-profile`: Instance profile. Default: `small`.
+ - `--exoscale-disk-size`: Disk size for the host in GB. Default: `50`.
+ - `--exoscale-security-group`: Security group. It will be created if it doesn't exist. Default: `docker-machine`.
+
+If a custom security group is provided, you need to ensure that you allow TCP port 2376 in an ingress rule.
+>>>>>>> 18cfe58... redhat: updated docs to explain base operating systems
 
 #### Generic
 Create machines using an existing VM/Host with SSH.
@@ -1513,6 +1530,7 @@ Options:
 
 The VMware vSphere driver uses the latest boot2docker image.
 
+<<<<<<< HEAD
 Environment variables and default values:
 
 | CLI option                        | Environment variable      | Default                  |
@@ -1562,6 +1580,14 @@ Environment variables and default values:
 | `--exoscale-security-group`     | `EXOSCALE_SECURITY_GROUP`    | `docker-machine` |
 | `--exoscale-availability-zone`  | `EXOSCALE_AVAILABILITY_ZONE` | `ch-gva-2`       |
 | `--exoscale-keypair`            | `EXOSCALE_KEYPAIR`           | -                |
+
+## Base Operating Systems
+The default base operating system for Machine is Boot2Docker on local providers
+(VirtualBox, Fusion, Hyper-V, etc) and the latest Ubuntu LTS supported
+by the cloud provider.  RedHat Enterprise Linux is also supported.  To use
+RHEL, you will need to select the image accordingly with the provider.  For
+example, in Amazon EC2, you could use "ami-12663b7a" as the
+`--amazonec2-ami` option which create an instance using RHEL 7.1 64-bit.
 
 ## Release Notes
 
--- a/libmachine/provision/redhat.go
+++ b/libmachine/provision/redhat.go
@@ -1,0 +1,279 @@
+package provision
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+
+	"github.com/docker/machine/drivers"
+	"github.com/docker/machine/libmachine/auth"
+	"github.com/docker/machine/libmachine/engine"
+	"github.com/docker/machine/libmachine/provision/pkgaction"
+	"github.com/docker/machine/libmachine/swarm"
+	"github.com/docker/machine/log"
+	"github.com/docker/machine/ssh"
+	"github.com/docker/machine/utils"
+)
+
+func init() {
+	Register("RedHat", &RegisteredProvisioner{
+		New: NewRedHatProvisioner,
+	})
+}
+
+func NewRedHatProvisioner(d drivers.Driver) Provisioner {
+	return &RedHatProvisioner{
+		packages: []string{
+			"curl",
+		},
+		Driver: d,
+	}
+}
+
+type RedHatProvisioner struct {
+	packages      []string
+	OsReleaseInfo *OsRelease
+	Driver        drivers.Driver
+	AuthOptions   auth.AuthOptions
+	EngineOptions engine.EngineOptions
+	SwarmOptions  swarm.SwarmOptions
+}
+
+func (provisioner *RedHatProvisioner) Service(name string, action pkgaction.ServiceAction) error {
+	command := fmt.Sprintf("sudo systemctl %s %s", action.String(), name)
+
+	if _, err := provisioner.SSHCommand(command); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) Package(name string, action pkgaction.PackageAction) error {
+	var packageAction string
+
+	switch action {
+	case pkgaction.Install:
+		packageAction = "install"
+	case pkgaction.Remove:
+		packageAction = "remove"
+	case pkgaction.Upgrade:
+		packageAction = "upgrade"
+	}
+
+	switch name {
+	case "docker":
+		name = "docker"
+	}
+
+	command := fmt.Sprintf("sudo -E yum %s -y %s", packageAction, name)
+
+	if _, err := provisioner.SSHCommand(command); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) isAWS() bool {
+	if _, err := provisioner.SSHCommand("curl -s http://169.254.169.254/latest/meta-data/ami-id"); err != nil {
+		return false
+	}
+
+	return true
+}
+
+func (provisioner *RedHatProvisioner) configureRepos() error {
+
+	// TODO: should this be handled differently? on aws we need to enable
+	// the extras repo different than a standalone rhel box
+
+	repoCmd := "subscription-manager repos --enable=rhel-7-server-extras-rpms"
+	if provisioner.isAWS() {
+		repoCmd = "yum-config-manager --enable rhui-REGION-rhel-server-extras"
+	}
+
+	if _, err := provisioner.SSHCommand(fmt.Sprintf("sudo %s", repoCmd)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) installDocker() error {
+	if err := provisioner.Package("docker", pkgaction.Install); err != nil {
+		return err
+	}
+
+	if err := provisioner.Service("docker", pkgaction.Restart); err != nil {
+		return err
+	}
+
+	if err := provisioner.Service("docker", pkgaction.Enable); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) dockerDaemonResponding() bool {
+	if _, err := provisioner.SSHCommand("sudo docker version"); err != nil {
+		log.Warn("Error getting SSH command to check if the daemon is up: %s", err)
+		return false
+	}
+
+	// The daemon is up if the command worked.  Carry on.
+	return true
+}
+
+func (provisioner *RedHatProvisioner) Provision(swarmOptions swarm.SwarmOptions, authOptions auth.AuthOptions, engineOptions engine.EngineOptions) error {
+	provisioner.SwarmOptions = swarmOptions
+	provisioner.AuthOptions = authOptions
+	provisioner.EngineOptions = engineOptions
+
+	// set default storage driver for redhat
+	provisioner.EngineOptions.StorageDriver = "devicemapper"
+
+	if err := provisioner.SetHostname(provisioner.Driver.GetMachineName()); err != nil {
+		return err
+	}
+
+	// setup extras repo
+	if err := provisioner.configureRepos(); err != nil {
+		return err
+	}
+
+	for _, pkg := range provisioner.packages {
+		if err := provisioner.Package(pkg, pkgaction.Install); err != nil {
+			return err
+		}
+	}
+
+	// install docker
+	if err := provisioner.installDocker(); err != nil {
+		return err
+	}
+
+	if err := utils.WaitFor(provisioner.dockerDaemonResponding); err != nil {
+		return err
+	}
+
+	if err := makeDockerOptionsDir(provisioner); err != nil {
+		return err
+	}
+
+	provisioner.AuthOptions = setRemoteAuthOptions(provisioner)
+
+	if err := ConfigureAuth(provisioner); err != nil {
+		return err
+	}
+
+	if err := configureSwarm(provisioner, swarmOptions); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) Hostname() (string, error) {
+	output, err := provisioner.SSHCommand("hostname")
+	if err != nil {
+		return "", err
+	}
+
+	var so bytes.Buffer
+	if _, err := so.ReadFrom(output.Stdout); err != nil {
+		return "", err
+	}
+
+	return so.String(), nil
+}
+
+func (provisioner *RedHatProvisioner) SetHostname(hostname string) error {
+	if out, err := provisioner.SSHCommand(fmt.Sprintf(
+		"sudo hostname %s && echo %q | sudo tee /etc/hostname",
+		hostname,
+		hostname,
+	)); err != nil {
+		log.Info(out)
+		log.Errorf("error setting hostname: %s", err)
+		return err
+	}
+
+	if _, err := provisioner.SSHCommand(fmt.Sprintf(
+		"if grep -xq 127.0.1.1.* /etc/hosts; then sudo sed -i 's/^127.0.1.1.*/127.0.1.1 %s/g' /etc/hosts; else echo '127.0.1.1 %s' | sudo tee -a /etc/hosts; fi",
+		hostname,
+		hostname,
+	)); err != nil {
+		log.Errorf("error setting /etc/hosts: %s", err)
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) GetDockerOptionsDir() string {
+	return "/etc/docker"
+}
+
+func (provisioner *RedHatProvisioner) SSHCommand(args string) (ssh.Output, error) {
+	return drivers.RunSSHCommandFromDriver(provisioner.Driver, args)
+}
+
+func (provisioner *RedHatProvisioner) CompatibleWithHost() bool {
+	return provisioner.OsReleaseInfo.Id == "rhel"
+}
+
+func (provisioner *RedHatProvisioner) GetAuthOptions() auth.AuthOptions {
+	return provisioner.AuthOptions
+}
+
+func (provisioner *RedHatProvisioner) SetOsReleaseInfo(info *OsRelease) {
+	provisioner.OsReleaseInfo = info
+}
+
+func (provisioner *RedHatProvisioner) GenerateDockerOptions(dockerPort int) (*DockerOptions, error) {
+	var (
+		engineCfg  bytes.Buffer
+		configPath = "/etc/sysconfig/docker"
+	)
+
+	// remove existing
+	if _, err := provisioner.SSHCommand(fmt.Sprintf("sudo rm %s", configPath)); err != nil {
+		return nil, err
+	}
+
+	driverNameLabel := fmt.Sprintf("provider=%s", provisioner.Driver.DriverName())
+	provisioner.EngineOptions.Labels = append(provisioner.EngineOptions.Labels, driverNameLabel)
+
+	// systemd / redhat will not load options if they are on newlines
+	// instead, it just continues with a different set of options; yeah...
+	engineConfigTmpl := `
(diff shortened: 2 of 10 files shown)


libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+	command := fmt.Sprintf("sudo -E yum %s -y %s", packageAction, name)
+
+	if _, err := provisioner.SSHCommand(command); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) isAWS() bool {
```
ehazlett: I wasn't sure about this.  On AWS the extras have to be configured with a separate command.  Perhaps there is a better way.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+	return true
+}
+
+func (provisioner *RedHatProvisioner) Provision(swarmOptions swarm.SwarmOptions, authOptions auth.AuthOptions, engineOptions engine.EngineOptions) error {
+	provisioner.SwarmOptions = swarmOptions
+	provisioner.AuthOptions = authOptions
+	provisioner.EngineOptions = engineOptions
+
+	// set default storage driver for redhat
+	provisioner.EngineOptions.StorageDriver = "devicemapper"
```
ehazlett: @nathanleclaire is this the right spot to add specific EngineOptions?

ehazlett:
@nathanleclaire @sthulb @hairyhenderson  PTAL

libmachine/provision/redhat.go
nathanleclaire: Oh really?  Very interesting.  We should ping Nirmal and see what he thinks.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+	return true
+}
+
+func (provisioner *RedHatProvisioner) configureRepos() error {
+
+	// TODO: should this be handled differently? on aws we need to enable
+	// the extras repo different than a standalone rhel box
+
+	repoCmd := "subscription-manager repos --enable=rhel-7-server-extras-rpms"
+	if provisioner.isAWS() {
```
nathanleclaire: Why not check whether we're on AWS by attempting a typecast of the driver to `amazonec2.Driver` like so:

```go
if _, err := provisioner.Driver.(amazonec2.Driver); err != nil {
    // do AWS-specific logic
} else {
    // do logic for everyone else
}
```

Admittedly this won't work for `generic`, so maybe we could fall back on the other method in that case.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+		hostname,
+		hostname,
+	)); err != nil {
+		log.Info(out)
+		log.Errorf("error setting hostname: %s", err)
+		return err
+	}
+
+	if _, err := provisioner.SSHCommand(fmt.Sprintf(
+		"if grep -xq 127.0.1.1.* /etc/hosts; then sudo sed -i 's/^127.0.1.1.*/127.0.1.1 %s/g' /etc/hosts; else echo '127.0.1.1 %s' | sudo tee -a /etc/hosts; fi",
```
nathanleclaire: The `127.0.0.1` line is only on Ubuntu and Debian hosts IIRC.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// remove existing
+	if _, err := provisioner.SSHCommand(fmt.Sprintf("sudo rm %s", configPath)); err != nil {
+		return nil, err
+	}
+
+	driverNameLabel := fmt.Sprintf("provider=%s", provisioner.Driver.DriverName())
+	provisioner.EngineOptions.Labels = append(provisioner.EngineOptions.Labels, driverNameLabel)
+
+	// systemd / redhat will not load options if they are on newlines
```
nathanleclaire: I wonder if any experts in the house might know a way we can get around this.  CC @tianon @vbatts @huslage

libmachine/provision/redhat.go
ehazlett: If you are using RHEL with a subscription you can use the Docker docs to install.  However, on EC2, that command doesn't work -- you have to use the above.  I'm assuming it's subscription related.

libmachine/provision/redhat.go
ehazlett: Yeah I think the typecast is better.  Thx!

libmachine/provision/redhat.go
nathanleclaire: Very interesting question.  It hasn't come up before but this seems fine to me.

nathanleclaire: Actually, we might want to default to only setting this if the option is empty: If the user knows what they are doing and wants to over-ride the default, we shouldn't prevent them (this would).

libmachine/provision/redhat.go
tianon: Ah, systemd + `EnvironmentFile` is a load of hack in my experience.  I believe that systemd upstream even discourages the use of `EnvironmentFile` entirely, but I don't have a source to back that up. :innocent:

http://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile= is definitely relevant, though.  It mentions that backslash-ending options should work.

tianon: Also, the syntax of this file as read by systemd versus sysvinit is actually different.  Systemd does custom processing of it, while sysvinit expects it to be a shell script and sources it directly.

tianon: (this is the biggest reason why [our official `docker.service` file](https://github.com/docker/docker/blob/e1df16b3a3f42c77234825fb504a657458a8abba/contrib/init/systemd/docker.service) does not include `EnvironmentFile`, despite it being proposed more than once)

tianon: See also https://github.com/docker/docker/pull/5141#issuecomment-40940481 for more long-winded explanation for why `/etc/systemd/system` drop-ins should be preferred.

nathanleclaire: Hm, yeah, I'm wondering if we might want to just generate the service file ourselves directly.

libmachine/provision/redhat.go
vbatts: hey hey, so devicemapper is fine, but is this going to be on loopback? or is the partitioning setting up a block device for thin-pool or metadata/data ?

ehazlett: @vbatts we do not do anything with host level block resources.  we simply are specifying the driver just as the docker install does.

libmachine/provision/redhat.go
ehazlett: @tianon thx -- I actually started out generating the entire service file and then ran away :)  I will read those threads as well but perhaps that's the safer way.  thanks!

tianon: Yeah, we should be able to generate just a simplified stub like:
```
[Service]
ExecStart=
ExecStart=/usr/bin/docker -d <whatever flags we need to add here>
```

It might be safer if we figure out some way to ask where the existing `docker.service` file is so we can scrape the `ExecStart` line it has and not break things like socket activation, but I don't know how much that really matters in this case either.  Worth playing around for sure.

libmachine/provision/redhat.go
sthulb: How do we deal with RHEL7 like distros?

ssh/client.go
```diff
@@ -82,6 +83,28 @@ func (client *Client) Run(command string) (Output, error) {
...
+
+	modes := ssh.TerminalModes{
+		ssh.ECHO:          1,
+		ssh.TTY_OP_ISPEED: 14400,
+		ssh.TTY_OP_OSPEED: 14400,
+	}
+
+	// request tty -- fixes error with hosts that use
+	// "Defaults requiretty" in /etc/sudoers - I'm looking at you RedHat
+	if err := session.RequestPty("xterm-256color", termHeight, termWidth, modes); err != nil {
```
sthulb: we should set the terminal to `os.Getenv("TERM")` and default to xterm

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,279 @@
...
+	case pkgaction.Remove:
+		packageAction = "remove"
+	case pkgaction.Upgrade:
+		packageAction = "upgrade"
+	}
+
+	switch name {
+	case "docker":
+		name = "docker"
+	}
```
hairyhenderson: This looks like a no-op... Might want to remove it...

docs/index.md
```diff
@@ -1239,20 +1254,13 @@ Options:
...
- - `--exoscale-api-secret-key`: Your API secret key.
- - `--exoscale-instance-profile`: Instance profile. Default: `small`.
- - `--exoscale-disk-size`: Disk size for the host in GB. Default: `50`.
- - `--exoscale-security-group`: Security group. It will be created if it doesn't exist. Default: `docker-machine`.
-
-If a custom security group is provided, you need to ensure that you allow TCP port 2376 in an ingress rule.
+## Base Operating Systems
+The default base operating system for Machine is Boot2Docker on local providers
+(VirtualBox, Fusion, Hyper-V, etc) and the latest Ubuntu LTS supported
+by the cloud provider.  RedHat Enterprise Linux is also supported.  To use
```
hairyhenderson: Might want to clarify that RHEL/CentOS 7+ is supported... Last I checked (which was a long while ago), the packages for RHEL/CentOS 6.x were super-old.

ibuildthecloud:
I just put in #1096.  It's for RancherOS support, but as a part of that work I refactored the ubuntu provisioner out into a GenericProvisioner and UbuntuProvision.  Seem like this code could also embed the GenericProvisioner I created.

docs/index.md
ehazlett: Good point.  And actually, it's just RHEL 7+ at the moment (CentOS behaves slightly differently).

hairyhenderson: Oh, I guess because of the RHN stuff? It'd probably be a good idea to get CentOS support in here before too long, but maybe as a follow-up, not necessarily this PR.

docs/index.md
```diff
@@ -999,6 +999,21 @@ Options:
 
 The DigitalOcean driver will use `ubuntu-14-04-x64` as the default image.
 
+#### exoscale
```
nathanleclaire: Alphabetical order?

libmachine/provision/redhat.go
nathanleclaire: It's boilerplate in case more get added in the future, makes it easier to track the mapping that way.

RusDavies:
Hi,

I few days ago, I started coding a FedoraProvisioner for docker-machine.  Then I realized this pull request for a (very similar) Redhat provisioner was already in progress.   However, there are some differences. 

So, I adapted by using the code for this pull request (from the redhat-provisioning branch of ehazlett/machine), and modifying it to allow integration of my code for Fedora.  To do so, I split the RedhatProvisioner into two parts, one containing code that is generic to all redhat-family distributions, and another that seems specific to RHEL only.  With that done, it was a trivial matter to add a new Fedora provisioner, which others may find useful. 

Code is here https://github.com/RusDavies/machine/tree/provisioner-fedora.  Important parts are the files redhat_family.go, redhat_family_RHEL.go, and redhat_family_fedora.go, all under libmachine/provision. 

I've been using the resulting FedoraProvisioner on Amazon E2, using the standard Fedora 21 cloud images.  It seems to works fine.  

At this point I'd like to engage in discussion about the acceptability of this adaptation/extension of the RedhatProvisioner code for inclusion in docker-machine.  However, given the RedhatProvisioner code itself hasn't yet been merged, then I'm not sure of  the best way to proceed.  Which of the following is preferable:
1) Make a new pull request against the ehazlett/machine repository
2) Make a new pull request against the docker/machine repository
3) Wait for the RedhatProvision to be merged, and then make a pull request against docker/machine.
4) Something else?

Thanks :o)

sthulb:
@ehazlett at this point, can this provisioner be a generic for both redhat + centos + fedora? or are they very different?

docs/index.md
ehazlett: Yes it was listed out of order from one of the rebases.

RusDavies:
@sthulb , there are some items in this RedhatProvisioner that seem very specific to RHEL only, such as the configureRepos().   Likewise for installOfficialDocker().  These are both hard coded into the Provision() procedure.

ehazlett:
@RusDavies i have actually been working on Fedora and CentOS provisioners based off of RHEL. 

@sthulb yeah there are some minor differences.

@RusDavies you can see the branch here: https://github.com/ehazlett/machine/compare/redhat-provisioning...ehazlett:fedora-provisioner

It's pretty lightweight.

RusDavies:
@ehazlett , I thought I may be too late to the game.  Good to know it's coming, anyway.  :o)

Some questions: 
1) How will you prevent the FedoraProvisioner from attempting to run your RHEL subscription stuff.  Will you gate it only for OsReleaseId="rhel"?  For example, this will fail on Fedora: "subscription-manager repos --enable=rhel-7-server-extras-rpms".  Similarly, what's the effect on Fedora of, "yum-config-manager --enable rhui-REGION-rhel-server-extras"?    

2) What other configuration nuances will other redhat-like distributions have?  Are you going to include every nuance in the general code, and gate each case? 

3) Is it not better to take such distribution specifics out of the generalized code, and put them in a higher layer?

Thanks

ehazlett:
@RusDavies thanks for the feedback.  It's not done (I've experienced the same issues in Fedora as well) but I pushed to show the work.  It most likely won't make it in for 0.3.0 as we have a bunch of testing for RHEL yet.

The problem with the RedHat distros is each one configures things slightly different then with various repos.  This multiplies when you add more flavors.

RusDavies:
@ehazlett , no problem.  I'm a happy to help. 

The slight variance between Redhat friends is precisely my concern. As other friends are added, it may tend towards a spaghetti of gates and exceptions.  

I've updated my own code to more closely match your branch with the FedoraProvisioner.  Take another look: using an interface, as I have, to place hooks in the process really doesn't add much complexity, but does help to create separation between distribution specifics.

ehazlett:
@RusDavies actually after looking we don't need `configureRepo` at all since we use our own RPMs.  I will look into yours but I'm not sure we need all of the hooks -- the Fedora provisioner is now 39 loc.  I would lean towards adding the hooks only if needed as it's more overhead and potential code paths.

RusDavies:
@ehazlett , ok, fair enough.

RusDavies:
@ehazlett, more feedback ...

RedhatProvisioner.installOfficialDocker() fails if docker is already installed from a different source.  

This can happen when using "--driver general" to connect to an existing machine, or otherwise using a custom image, wherein docker is already installed from e.g. distribution repos. 

Worth noting, at least in my case, and likely for others, machines are deployed against a BOM which includes software versions.  If I have an AMI that includes a particular version of docker, then I don't necessarily want to change the version or source.  In my case, this important not just for consistency & stability, but also for compliance reasons.  

If docker is already present on the target machine, can we just use it?

ehazlett:
@RusDavies I understand the concern, however, this requires more configuration.  If you already have docker and skip the install, then we aren't guaranteed to be able to configure Docker as well (docker options, TLS, etc).  I'm OK with adding the check (we do the same thing in other drivers) however, there could still be errors so we don't want to recommend doing this.

RusDavies:
@ehazlett , if an provisioning fails  on account of a docker-machine rejecting the installed version of docker, then that gives me clear and timely advise that my BOM needs revising.  However, if the docker version gets automagically changed without warning, and then causes some unexpected breakage in another software & seemingly disconnected from the actions of docker-machine, then I'll have people from ops waving pitchforks at me.  ;o) 

Perhaps a version check, with an error thrown if too old, would be appropriate?

ehazlett:
@RusDavies completely understand :)  at one point, i was the one carrying the pitchfork ;)

RusDavies:
@ehazlett, possibly daft question (I'm not a systemd expert): 

Context: 
I notice that docker-machine writes the docker config file to /lib/systemd/system/docker.service.  To the best of my knowledge, that's where packagers should place their service files, which get clobbered if the package gets updated.

Yet, docker-machine isn't the packager for docker.  Instead, it's an administrator tool.  To the best of my knowledge, the place for the administrator to override service files is /etc/systemd/system/, where they won't get clobbered by upgrades.  

The location makes no difference, either way, for short lived images, or where software versions are static.  However, it's possible that some images/machines may expect upgrades during their lifecycle .  This may be especially true of hosts with static IPs when using the generic driver, which don't need re-provisioning between "image changing" events.  

Of course, the administrator may place the service file into the correct location... yet, so could the tool. 

Question:
Given docker-machine is an administrator tool, then why isn't the service file placed in /etc/systemd/system/ ?

RusDavies:
@ehazlett , just a heads up... RedhatProvisioner seems to break flanneld on Fedora 21.

Flanneld makes use of DOCKER_NETWORK_OPTIONS from /etc/sysconfig/docker-network, which isn't set by docker-machine.  However, even when I hand transfer the configuration, and do a `systemdctl systemctl stop docker && systemctl daemon-reload && ip link delete docker0 && systemd start flanneld && sleep 5 && systemd start docker`, it still doesn't work, in that it fails to assign docker0 an address from within the range configured for flanneld.

If I first do a `systemctl reinstall docker-io`, to revert back to the distribution package, and then reset the docker0 interface as above, then it works perfectly.

So, maybe it's to do with the configuration files, or maybe it's to do with the version of docker?

ehazlett:
@RusDavies this is a difference in how RedHat chooses to setup their version of Docker vs. the official.  The RedHat system files use environment files where as the official docker version won't.  This doesn't look like it's specific to machine.

There was a reason we chose not to use env files in the systemd service files.  Perhaps @tianon can explain better than myself :)

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,235 @@
...
+	Register("RedHat", &RegisteredProvisioner{
+		New: NewRedHatProvisioner,
+	})
+}
+
+func NewRedHatProvisioner(d drivers.Driver) Provisioner {
+	return &RedHatProvisioner{
+		GenericProvisioner{
+			DockerOptionsDir:  "/etc/docker",
+			DaemonOptionsFile: "/lib/systemd/system/docker.service",
```
tianon: Yeah, this is definitely not a place we should be writing to; this should definitely go in `/etc/systemd/system/docker.service` instead (as a stub, not the full file, as I described in https://github.com/docker/machine/pull/1090/files#r29536500).

tianon:
@ehazlett yeah, the `EnvironmentFile` discussion above (https://github.com/docker/machine/pull/1090/files#r29532734) has some links to useful `docker/docker` issues about that

What I think @RusDavies is getting at though is that after running this, flannel no longer works, and it's probably due to some kind of mismatch.  `/etc/sysconfig/docker-network` is definitely not a docker-official place for anything at all, so I'm not sure what the issue is, but it does sound like it's got something to do with Docker versions -- does the provisioner force a newer or older version of Docker somehow?

RusDavies:
@ehazlett  @tianon, I completely understand that docker uses a service file approach, and without EnvironmentFiles.  Makes sense that docker shouldn't be concerned with distribution specifics.  That's for the distribution guys to worry about.   I'd expect (although I don't actually know) that the generic version of flanneld also doesn't concern itself with `/etc/sysconfig/docker-machine`.

There was really two points:
1) With docker-machine, we are getting into distribution-specifics when provisioning for a Fedora host, for example.  It seems to me that the differentiated provisioner is an adaptation layer, intended to make the configuration compliant with the target distribution.  If the Fedora provisioner doesn't update `/etc/sysconfig/docker` and friends, and yet software in the fedora repos expects as much, then it's not really provisioning for Fedora. 

2) It's possible (TBC) the version of docker being installed by docker-machine doesn't play nicely with flanneld in the repos for Fedora 2;  that is to so, it doesn't seem to work even when `/etc/sysconfig/docker` and friends do match the configuration in the docker service file (even if the docker service file then doesn't make use of those files).

libmachine/provision/redhat.go
ehazlett: thanks @tianon

ehazlett:
@RusDavies the new pushes to the branch should fix the issue.  We now use `/etc/systemd/system/docker.service` with just a stub (thx @tianon).  This should not affect any of the packaged config files.  This is done for all of the RedHat type distros.

ehazlett:
@tianon re: flannel with docker, the package installs the latest stable of docker.  These are the RPMs from Docker.  We currently do not have a way to specify versions of Docker.

@RusDavies we have discussed the idea of allowing the install of specific versions but it's not in yet.

docs/index.md
```diff
@@ -1281,20 +1296,13 @@ Options:
...
- - `--exoscale-disk-size`: Disk size for the host in GB. Default: `50`.
- - `--exoscale-security-group`: Security group. It will be created if it doesn't exist. Default: `docker-machine`.
-
-If a custom security group is provided, you need to ensure that you allow TCP port 2376 in an ingress rule.
+## Base Operating Systems
+The default base operating system for Machine is Boot2Docker on local providers
+(VirtualBox, Fusion, Hyper-V, etc) and the latest Ubuntu LTS supported
+by the cloud provider.  RedHat Enterprise Linux is also supported.  To use
+RHEL, you will need to select the image accordingly with the provider.  For
+example, in Amazon EC2, you could use "ami-12663b7a" as the
```
nathanleclaire: Hm, if this is a real AMI ID, is it in `us-east-1`?  Might be worth calling out either way.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if _, err := provisioner.SSHCommand(command); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) isAWS() bool {
+	if _, err := provisioner.SSHCommand("curl -s http://169.254.169.254/latest/meta-data/ami-id"); err != nil {
```
nathanleclaire: Can we please put a comment here and turn the magic endpoint into a `const`?  :)

ehazlett:
@nathanleclaire updated PTAL

nathanleclaire:
Nice - reviewing this is in my queue - FYI it might need a small update when https://github.com/docker/machine/pull/1138 gets merged in, or vice versa.

ehazlett:
ok rebased and updated to support external ssh with `tty` allocation.

@nathanleclaire PTAL

nathanleclaire:
Do you want to add the `-o PasswordAuthentication=no` fix here or in a separate PR?

Also, you probably should squash all those commits, and also this needs a rebase.

nathanleclaire:
I made a PR to add the SSH option mentioned here : https://github.com/ehazlett/machine/pull/8

With that patch, provisioning on CentOS works now.

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func (provisioner *RedHatProvisioner) GenerateDockerOptions(dockerPort int) (*DockerOptions, error) {
+	var (
+		engineCfg  bytes.Buffer
+		configPath = provisioner.DaemonOptionsFile
+	)
+
+	// remove existing
```
nathanleclaire: If we're not going to use this code, why not take it out?

nathanleclaire:
Tested:

- basic creation on CentOS, Fedora, and RHEL
- engine options
- swarm functionality

I'm ready to go on this once ehazlett#8 gets merged in.

denverdino:
@ehazlett Evan 
After the 26432b7 the Native SSH client is broken. I got the following error when executing the following lines of "ssh/client.go"

```
	if err := session.RequestPty("xterm-256color", termHeight, termWidth, modes); err != nil {
		return string(output), err
	}
```

The error output for provisioning my ubuntu image is 
```
SSH cmd err, output: EOF: 
```

And I think the logic of the Output method also has some problem for the command will be executed twice. Does it work as expected?
```
func (client NativeClient) Output(command string) (string, error) 
	...
	output, err := session.CombinedOutput(command)
	...
	return string(output), session.Run(command)
}
```

Thanks

ehazlett:
@sthulb do you know what is causing the Native to have issues?

@nathanleclaire is the `Output` func correct or did I munge a merge?

nathanleclaire:
Mmm, yeah that `Output` function definitely looks wrong.  I'll take a look at it.

ehazlett:
@nathanleclaire thx

libmachine/provision/redhat.go
```diff
@@ -0,0 +1,226 @@
...
+
+	for _, pkg := range provisioner.Packages {
+		log.Debugf("installing base package: name=%s", pkg)
+		if err := provisioner.Package(pkg, pkgaction.Install); err != nil {
+			return err
+		}
+	}
+
+	// update OS -- this is needed for libdevicemapper and the docker install
+	if _, err := provisioner.SSHCommand("sudo yum -y update"); err != nil {
```
sampowers: Is there a way to make this step be conditional/take a shorter amount of time?  I already have a mirror on the local network, but this is taking up most of the setup time.

ehazlett:
Unfortunately this depends on the provider and mirror.  If you have a local mirror then that's about all you can do.  We must update as some images do not contain the proper device mapper libs that are needed.

marcellodesales:
Hi @ehazlett, I'm trying to use the generic driver on RHEL 7.1... I have the following setup:

SSK keys already authorized on the host
-------

```
~ ⌚ 1:04:08
$ ssh 10.132.52.146

Last login: Tue Jun 23 07:59:07 2015 from 172.17.193.49
[mdesales@pppdc9prd8ok ~]$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.1 (Maipo)
```

Command to create host fails while importing keys
--------

After a few seconds, the command to create and use the generic driver fails..

```
~ ⌚ 0:58:57
$ docker-machine create --driver "generic" --generic-ip-address "10.132.52.146" --generic-ssh-user "mdesales" dotci
Creating CA: /home/mdesales/.docker/machine/certs/ca.pem
Creating client certificate: /home/mdesales/.docker/machine/certs/cert.pem
Importing SSH key...
Error creating machine: exit status 1
```

What should I do?

PR #1100, opened 2015-05-04
Fixing bug in GetIP where Public IP would always be returned 

### LINKED ISSUE #1099: softlayer: can't create when the --softlayer-private-net-only flag is set
This happens because the `GetIP()` function in `drivers/softlayer/driver.go` always calls `VirtualGuest().GetPublicIp()` regardless of the private-net-only setting.

hairyhenderson:
...despite use of `--softlayer-private-net-only`

Fixes #1099


nathanleclaire:
LGTM

ehazlett:
LGTM

PR #1103, opened 2015-05-04
Read active host from DOCKER_HOST instead of from file

nathanleclaire:
This is a re-worked version of #777.  For a feel of the motivations and issues related to it, see that proposal.

THIS DEPRECATES THE "SHORT FORM" MACHINE COMMANDS WHICH AUTOMAGICALLY RUN AGAINST THE ACTIVE HOST, e.g. `docker-machine start`, `docker-machine stop` without any arguments.

This is quite near completion, but I still need to update the integration tests.

cc @ehazlett @sthulb @hairyhenderson @tiborvass @SvenDowideit @fredlf 


ehazlett:
When trying to reset the environment I get the following:

```
$> eval "$(docker-machine env -u)"
Error: Expected one machine name as an argument.
```

fredlf:
Docs LGTM, but you might want to scan through existing docs to make sure we don't have any examples doing things the old way.

nathanleclaire:
> When trying to reset the environment I get the following:

Ah, right, forgot about that usage.  I'll update the "validation" to fix.

nathanleclaire:
@ehazlett Ha, I fixed and I just discovered a bug.  If the IP of the machine is `""` because it is stopped, we will mistakenly think it is the active host.  I will update to check this case too.

nathanleclaire:
OK, should be fixed now.

ehazlett:
LGTM

PR #1108, opened 2015-05-05
Fix issue where GetIP in VirtualBox failes due to no SSH

ibuildthecloud:
This patch consolidates most of the SSH command and wait logic to the
drivers package.  The existing functions in host are just wrappers to the
real implementation in drivers.

Signed-off-by: Darren Shepherd <darren@rancher.com>

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -8,7 +8,6 @@ import (
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/log"
 	"github.com/docker/machine/provider"
-	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/state"
 )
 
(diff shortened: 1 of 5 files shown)


ibuildthecloud:
@nathanleclaire This is my attempt at fixing the VirtualBox issue I saw today where things randomly fail with RancherOS on VirtualBox because SSH is not always available immediately.

One random thing I changed was that WaitForTCP before would read 1 byte.  I changed it to just do a connect.  I was having issue where it would never succeed because it would just get an EOF on read.

ibuildthecloud:
I think I broke travis.  The build associated with this PR has the wrong commit ID.

nathanleclaire:
> @nathanleclaire This is my attempt at fixing the VirtualBox issue I saw today where things randomly fail with RancherOS on VirtualBox because SSH is not always available immediately.

Nice.

> One random thing I changed was that WaitForTCP before would read 1 byte. I changed it to just do a connect. I was having issue where it would never succeed because it would just get an EOF on read.

Yeah, I noticed that.  I wonder what the original motivation was.  I'm fine with this change, it makes sense to me.

nathanleclaire:
> I think I broke travis. The build associated with this PR has the wrong commit ID.

Odd.  Can you try adding a comment or something + rebasing this into a different commit SHA + force pushing to try and kick off a new build in case something went weird on their side?  Meanwhile I'll check it out locally.  Looks pretty good to me.

ibuildthecloud:
Fixed.  It was actually my problem.  I misunderstood what travis was doing.  I guess travis merges before the build (which is good).  There was a compilation error.  Anyhow fixed.

nathanleclaire:
Nice.  LGTM cc @ehazlett

ehazlett:
LGTM

PR #1109, opened 2015-05-05
Default storage driver to "" and let provisioner choose default

ibuildthecloud:
The storage driver will be "" by default an the provisioner must choose
their default provisioner.  If the user chosen storage driver is not
supported the provisioner should return an error.

I changed boot2docker to throw an error if something besides aufs is chosen.  I have no clue what boot2docker supports so that may be wrong.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -188,7 +188,6 @@ var sharedCreateFlags = []cli.Flag{
 	cli.StringFlag{
 		Name:  "engine-storage-driver",
 		Usage: "Specify a storage driver to use with the engine",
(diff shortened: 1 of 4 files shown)


ibuildthecloud:
Looks like b2d supports all storage drivers.  I will have to update that.

ibuildthecloud:
Updated to allow drivers for b2d.

nathanleclaire:
LGTM

nathanleclaire:
cc @ehazlett

ehazlett:
LGTM

PR #1110, opened 2015-05-05
CPU count fixes for `vmwarefusion` driver

### LINKED ISSUE #1095: [VMWare] Why auto set Processors to 8 core?
I created docker machine use VMWare Fusion.

I was specify set to
```
--driver vmwarefusion
--vmwarefusion-disk-size "1000000"
--vmwarefusion-memory-size "4098"
```

There is no set of cpu in it.

I check setting after launch vmware, 8 core is specified.
It gives all of the number of CPU, I think that is go too far.

/cc @ehazlett, @frapposelli 

ps, My mac spec

```
  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro11,3
  Processor Name:	Intel Core i7
  Processor Speed:	2.3 GHz
  Number of Processors:	1
  Total Number of Cores:	4
  L2 Cache (per Core):	256 KB
  L3 Cache:	6 MB
  Memory:	16 GB
```

frapposelli:
- Added `--vmwarefusion-cpu-count` parameter
- Defaults to number of **physical** cores available (calling `sysctl`)

Fixes #1095 

Signed-off-by: Fabio Rapposelli <fabio@vmware.com>

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -11,10 +11,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
@@ -40,7 +42,7 @@ type Driver struct {
 	IPAddress      string
 	Memory         int
 	DiskSize       int
-	CPUs           int
+	CPU            int
 	ISO            string
 	Boot2DockerURL string
 	CaCertPath     string
@@ -70,6 +72,12 @@ func GetCreateFlags() []cli.Flag {
 			EnvVar: "FUSION_BOOT2DOCKER_URL",
 			Name:   "vmwarefusion-boot2docker-url",
 			Usage:  "Fusion URL for boot2docker image",
+		},
+		cli.IntFlag{
+			EnvVar: "FUSION_CPU_COUNT",
+			Name:   "vmwarefusion-cpu-count",
+			Usage:  "number of CPUs for the machine (-1 to use the number of CPUs available)",
+			Value:  -1,
 		},
 		cli.IntFlag{
 			EnvVar: "FUSION_MEMORY_SIZE",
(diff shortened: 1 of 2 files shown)


ehazlett:
Does performance really gain by limiting it to physical CPUs as opposed to cores?

znmeb:
@ehazlett I don't have a Mac so I can't benchmark it but my guess is that Macs have hyperthreaded Intel chips so you have one chip, n cores and 2*n threads. You should at least limit the number of virtual processors to the number of *cores* - allocating all the *threads* will probably slow you down.

frapposelli:
What @znmeb said, in addition, Fusion has a way to determine if it's running on a hyperthread or on a physical core and can schedule cpu time accordingly, firing all the logical cores would work on a machine that is running a light "native" workload and a single VM but will degrade when you spin up multiple "all-cores" VMs, I think allocating only physical cores is a good tradeoff.

ehazlett:
@frapposelli @znmeb thanks for the explanation.

ehazlett:
I'm still not crazy about the hacky call out.  Since we know the platform we are on, would it be safe to just numcpu/2?

/cc @nathanleclaire

frapposelli:
@ehazlett I initially wrote a simple `numcpu/2` version but I'm not 100% confident that all Mac platforms have hyperthreading, that's why I resorted to `sysctl`.

ehazlett:
@frapposelli ok i think that's fine.

@nathanleclaire thoughts?

ehazlett:
@frapposelli after thinking about this for a while, I think we should default it to a single core and let the user decide.  this keeps things like resource utilization at the host level to a min but allow them to configure it as high as they want.  this also doesn't require the hacky call out.  we also get the benefit of not overusing the host in the event of multiple VMs by default.

thoughts?

nathanleclaire:
> after thinking about this for a while, I think we should default it to a single core and let the user decide. this keeps things like resource utilization at the host level to a min but allow them to configure it as high as they want. this also doesn't require the hacky call out. we also get the benefit of not overusing the host in the event of multiple VMs by default.

I'm +1 on this.  I'd like to keep the default behavior consistent with VBox, and also not overload users who may not have a particularly powerful computer by default.

frapposelli:
OK, sounds good to me too.

ehazlett:
@frapposelli thanks!  closing.

PR #1117, opened 2015-05-05
windows: fix shell detection to not bail on bash envs

ehazlett:
This fixes a bug where machine bails if you are using a bash env on windows (i.e. msysgit, etc).


nathanleclaire:
LGTM

PR #1121, opened 2015-05-05
Add environment variable option to skip docker build in script/build

nathanleclaire:
cc @ehazlett WDYT?  This would be useful to me.


ehazlett:
This should be fine.  LGTM

PR #1122, opened 2015-05-05
Fix Windows binary path being broken and add -L to remaining commands

ehazlett:
Fix windows install docs.


nathanleclaire:
LGTM

PR #1125, opened 2015-05-06
Adding --filter flag to ls command

### LINKED ISSUE #1076: Proposal: Filters for "docker-machine ls"
## Abstract

Right now it is not possible to query only a subset of machines for their state using `docker-machine ls`.  For a variety of reasons, including the ability to integrate nicely with other commands such as `rm` that accept multiple machine names as arguments, I propose that we support a `docker-machine ls --filter` flag which will only query a subset of those machines.  It is similar to how `docker images` has a `--filter dangling=true` option.

Then we can do things like:

```console
$ # Stop the "foobar" swarm with one command!
$ docker-machine stop $(docker-machine ls --filter swarm=foobar)
```

## Implementation

If you look in `commands/ls.go` you can see that we have a little block like this:

```go
        hostList, err := mcn.List()
        if err != nil {
                log.Fatal(err)
        }
```

Right after this block is where the filtering should take place.  `List` will return all of the hosts that we have a reference to locally.  The approach I would take is to add a `FilterOptions` struct to this file, which contains the settings that we might want to filter by (`--filter` should be a `StringSlice` flag, so we will have to add one or two functions to parse this into the struct as well).

We can start simple with just "SwarmName" and "DriverName" parameters and expand from there if we want more, so let's not get too elaborate on the first go:

```go
type FilterOptions struct {
    SwarmName string
    DriverName string
}
```

Add a function, maybe named `filterHosts`, which accepts a `[]*Host` and `FilterOptions` and returns a `[]*Host` with the ones that we don't want removed.  Then everything in the rest of the function will work the same as it ever has since we will just be modifying the value of what `hostList` is today.

cc @hairyhenderson @sthulb @ehazlett

hairyhenderson:
Initially supporting `swarm=`, `driver=` and `state=` filters.

Fixes #1076 

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -2,22 +2,36 @@ package commands
 
 import (
 	"fmt"
-	"log"
 	"os"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/log"
 )
+
+// FilterOptions -
+type FilterOptions struct {
+	SwarmName  []string
+	DriverName []string
+	State      []string
+}
 
 func cmdLs(c *cli.Context) {
 	quiet := c.Bool("quiet")
+	filters, err := parseFilters(c.StringSlice("filter"))
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	mcn := getDefaultMcn(c)
 	hostList, err := mcn.List()
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	hostList = filterHosts(hostList, filters)
 
 	// Just print out the names if we're being quiet
 	if quiet {
@@ -67,4 +81,103 @@ func cmdLs(c *cli.Context) {
 	}
 
 	w.Flush()
+}
+
+func parseFilters(filters []string) (FilterOptions, error) {
+	options := FilterOptions{}
+	for _, f := range filters {
+		kv := strings.SplitN(f, "=", 2)
+		key, value := kv[0], kv[1]
+
+		switch key {
+		case "swarm":
+			options.SwarmName = append(options.SwarmName, value)
+		case "driver":
+			options.DriverName = append(options.DriverName, value)
+		case "state":
+			options.State = append(options.State, value)
+		default:
+			return options, fmt.Errorf("Unsupported filter key '%s'", key)
+		}
+	}
+	return options, nil
+}
+
+func filterHosts(hosts []*libmachine.Host, filters FilterOptions) []*libmachine.Host {
+	if len(filters.SwarmName) == 0 &&
+		len(filters.DriverName) == 0 &&
+		len(filters.State) == 0 {
+		return hosts
+	}
+
+	filteredHosts := []*libmachine.Host{}
+	swarmMasters := getSwarmMasters(hosts)
+
+	for _, h := range hosts {
+		if filterHost(h, filters, swarmMasters) {
+			filteredHosts = append(filteredHosts, h)
+		}
+	}
+	return filteredHosts
+}
+
+func getSwarmMasters(hosts []*libmachine.Host) map[string]string {
+	swarmMasters := make(map[string]string)
+	for _, h := range hosts {
+		swarmOptions := h.HostOptions.SwarmOptions
+		if swarmOptions != nil && swarmOptions.Master {
+			swarmMasters[swarmOptions.Discovery] = h.Name
+		}
+	}
+	return swarmMasters
+}
+
+func filterHost(host *libmachine.Host, filters FilterOptions, swarmMasters map[string]string) bool {
+	swarmMatches := matchesSwarmName(host, filters.SwarmName, swarmMasters)
+	driverMatches := matchesDriverName(host, filters.DriverName)
+	stateMatches := matchesState(host, filters.State)
+
+	return swarmMatches && driverMatches && stateMatches
+}
+
+func matchesSwarmName(host *libmachine.Host, swarmNames []string, swarmMasters map[string]string) bool {
+	if len(swarmNames) == 0 {
+		return true
+	}
+	for _, n := range swarmNames {
+		if host.HostOptions.SwarmOptions != nil {
+			if n == swarmMasters[host.HostOptions.SwarmOptions.Discovery] {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func matchesDriverName(host *libmachine.Host, driverNames []string) bool {
+	if len(driverNames) == 0 {
+		return true
+	}
+	for _, n := range driverNames {
+		if host.DriverName == n {
+			return true
+		}
+	}
+	return false
+}
+
+func matchesState(host *libmachine.Host, states []string) bool {
+	if len(states) == 0 {
+		return true
+	}
+	for _, n := range states {
+		s, err := host.Driver.GetState()
+		if err != nil {
+			log.Warn(err)
+		}
+		if n == s.String() {
+			return true
+		}
+	}
+	return false
 }
--- a/test/integration/ls.bats
+++ b/test/integration/ls.bats
@@ -1,0 +1,27 @@
+#!/usr/bin/env bats
+
+load helpers
+
+teardown() {
+  echo "$BATS_TEST_NAME
+----------
+$output
+----------
+
+" >> ${BATS_LOG}
+  machine rm -f testmachine
+}
+
+@test "ls: filter on driver" {
+  run machine create -d none --url tcp://127.0.0.1:2375 testmachine
+  run machine ls --filter driver=none
+  [ "$status" -eq 0  ]
+  [[ ${lines[1]} =~ "testmachine" ]]
+}
+
+@test "ls: filter on swarm" {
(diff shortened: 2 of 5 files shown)


commands/ls.go
```diff
@@ -67,3 +80,48 @@ func cmdLs(c *cli.Context) {
...
+			}
+		} else if filters.DriverName != "" {
+			if h.DriverName == filters.DriverName {
+				filteredHosts = append(filteredHosts, h)
+			}
+		} else {
+			filteredHosts = append(filteredHosts, h)
+		}
+	}
+	return
```
nathanleclaire: We don't return anything here?

commands/ls.go
```diff
@@ -67,3 +80,48 @@ func cmdLs(c *cli.Context) {
...
+		switch key {
+		case "swarm":
+			options.SwarmName = value
+		case "driver":
+			options.DriverName = value
+		default:
+			err = fmt.Errorf("Unsupported filter key '%s'", key)
+		}
+	}
+	return
```
nathanleclaire: We don't return anything here?

commands/ls.go
hairyhenderson: I was using a "naked return" (so, equivalent to `return filteredHosts`), but I see that's not very clear. I'll make it explicit :)

commands/ls.go
hairyhenderson: I was using a "naked return" (so, equivalent to `return options, err`), but I see that's not very clear. I'll make it explicit :)

nathanleclaire:
Oh, you can name return parameters.  That's confusing.

nathanleclaire:
I think I'd really prefer to keep the style consistent with the rest of the codebase and explicitly return local variables, please.

commands/ls.go
nathanleclaire: Nice, thanks.

nathanleclaire:
Hmm.... this PR makes me think we really should just have a `SwarmName` field in the `SwarmOptions` struct for easy return.  Maybe we can move over to that model.

nathanleclaire:
You're going to need to keep: https://github.com/docker/machine/pull/1130/files in mind.  Also, I think we should probably add a `SwarmOptions.SwarmName` struct to make the filtering and display logic easier to understand.  That will require a bit of migration song and dance, so let's discuss it a bit in IRC.

hairyhenderson:
@nathanleclaire - fixed the naked returns.

Still need to rebase.

hairyhenderson:
Ok, rebased & squashed.

@nathanleclaire, @ehazlett PTAL :)

commands/ls.go
```diff
@@ -68,3 +81,49 @@ func cmdLs(c *cli.Context) {
...
+	filteredHosts := []*libmachine.Host{}
+	swarmMasters := make(map[string]string)
+	for _, h := range hosts {
+		swarmOptions := h.HostOptions.SwarmOptions
+		if swarmOptions != nil && swarmOptions.Master {
+			swarmMasters[swarmOptions.Discovery] = h.Name
+		}
+	}
+
+	for _, h := range hosts {
```
nathanleclaire: Hm, to me this really seems like it should go in its own function.  What do you think?

nathanleclaire: To clarify, I generally try to avoid deeply nested blocks of logic, and also I have a hard time reasoning about what this little chunk does just by looking at it, so it'd be good to have it in a little function called `getX` and I can look at the assignment and say, "Oh, this is the bit that gets X from these pieces of information passed in".

nathanleclaire:
Also, what's the game plan for filtering by multiple drivers and swarms?

commands/ls.go
hairyhenderson: Oh, totally. Will change that.

hairyhenderson:
@nathanleclaire - good point. I'm thinking it'll require changing FilterOptions's fields to `[]string`s.

And, as I think about it more, I'm thinking of changing `FilterOptions` to a `map[string][]string` instead - more flexible that way, I think...

I'll hack at it tonight :)

nathanleclaire:
> And, as I think about it more, I'm thinking of changing FilterOptions to a map[string][]string instead - more flexible that way, I think...

Why a map?  If the fields are known before-hand (which they are in our case) it is type-safer.

hairyhenderson:
@nathanleclaire - yeah, I suppose that's a good point... Especially with ls, we'll always know the fields beforehand.

I've added support for filtering multiples - PTAL

commands/ls.go
```diff
@@ -68,3 +81,67 @@ func cmdLs(c *cli.Context) {
...
+		swarmOptions := h.HostOptions.SwarmOptions
+		if swarmOptions != nil && swarmOptions.Master {
+			swarmMasters[swarmOptions.Discovery] = h.Name
+		}
+	}
+	return swarmMasters
+}
+
+func filterHost(host *libmachine.Host, filters FilterOptions, swarmMasters map[string]string) bool {
+	if len(filters.SwarmName) > 0 {
```
nathanleclaire: Are these `len()` checks really needed?  If you attempt a `range` over a nil slice, I think it will just skip it.

hairyhenderson: Yeah - the check is to find out if the filters have been specified at all

commands/ls.go
```diff
@@ -68,3 +82,76 @@ func cmdLs(c *cli.Context) {
...
+			options.State = append(options.State, value)
+		default:
+			return options, fmt.Errorf("Unsupported filter key '%s'", key)
+		}
+	}
+	return options, nil
+}
+
+func filterHosts(hosts []*libmachine.Host, filters FilterOptions) []*libmachine.Host {
+	filteredHosts := []*libmachine.Host{}
```
nathanleclaire: Why not just have one check at the beginning of this function, which does something like:

```go
if len(filters.SwarmName) == 0 && len(filters.DriverName) == 0 && len(filters.State) == 0 {
    return hosts
}
```

It would allow you to get rid of the `len` checks / deeply nested logic below, as well as the final `else` condition.

nathanleclaire:
Looking really good - one nitpicky code style thing as noted in the diff.  This needs docs, and I'd prefer to see an integration test as well.  Thanks for taking the initiative on this, it's much appreciated.

commands/ls.go
```diff
@@ -68,3 +82,76 @@ func cmdLs(c *cli.Context) {
...
+		}
+	} else if len(filters.DriverName) > 0 {
+		for _, n := range filters.DriverName {
+			if host.DriverName == n {
+				return true
+			}
+		}
+	} else if len(filters.State) > 0 {
+		for _, n := range filters.State {
+			s, _ := host.Driver.GetState()
```
nathanleclaire: Let's make sure to check this error, please.  Probably a `log.Warn` should do.

commands/ls.go
hairyhenderson: Thanks - this is a good idea.

commands/ls.go
hairyhenderson: Ok, done

hairyhenderson:
@nathanleclaire - I think that does it for the code.

I'll work on a bats test and docs and commit them later.

nathanleclaire:
Nice!  Looking very good.

hairyhenderson:
@nathanleclaire - docs and integration test are in.

So, the way I implemented this, `--filter driver=virtualbox --filter swarm=foo` means `driver=virtualbox OR swarm=foo`. In Docker, this would be an `AND` instead.

I think I got mixed up when I was dealing with the `--filter driver=virtualbox --filter driver=generic` case, which should be an OR (and is in Docker).

So, I guess what I'm saying is I have some more work to do to change the original to an `AND` :)

hairyhenderson:
@nathanleclaire - I think I'm happy now with the way the code/docs/tests work now... PTAL :)

ehazlett:
Reviewed and tested with various filters -- awesome work :)

LGTM

hairyhenderson:
Thanks @ehazlett! I want to squash before this goes in, so please wait for that before merging :)

nathanleclaire:
Still need to give this a final whirl but looking in really great shape.

hairyhenderson:
Squashed!

test/integration/ls.bats
```diff
@@ -0,0 +1,27 @@
...
+
+@test "ls: filter on driver" {
+  run machine create -d none --url tcp://127.0.0.1:2375 testmachine
+  run machine ls --filter driver=none
+  [ "$status" -eq 0  ]
+  [[ ${lines[1]} =~ "testmachine" ]]
+}
+
+@test "ls: filter on swarm" {
+  run machine create -d none --url tcp://127.0.0.1:2375 --swarm --swarm-master --swarm-discovery deadbeef testmachine
```
nathanleclaire: LOL @ deadbeef!

hairyhenderson: :cow2:

hairyhenderson: oh, and I just realized that was missing a `token://`. I suppose it works because it's the `none` driver, but I'll fix it anyway just in case someone uses it as an example some day :wink:

hairyhenderson: Ok, fixed now.

nathanleclaire:
Nice.

I really appreciate this, it will be most helpful for people who are having their `ls` blocked / b0rked by troubled machines.

LGTM :+1:

PR #1127, opened 2015-05-06
Fix resize of terminal window

sthulb:
This fixes the issue where magical (control) codes will appear after a resize of the terminal window.

cc: @ehazlett @ahmetalpbalkan

### DIFF
--- a/ssh/client.go
+++ b/ssh/client.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"time"
 
 	"github.com/docker/docker/pkg/term"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
Windows note: such magical codes won't appear on windows. However if you want to respond to the client terminal size changes, see "monitorTtySize" function in docker cli codebase. I think we should reuse that. 

Apparently that method also should be moved to the term package and should be provided a callback function as an argument.

I believe this change won't be of any use in windows and by the forever polling loop you're duplicating code with docker cli. Unix programs get a SIGWINCH signal when term size changes and docker cli handles that; for windows we do polling (for now, although that can be also event based). 

Therefore currently I'm -1 for this solution. Let's make monitorTtySize better by moving it into docker/pkg/term and reuse here.

ehazlett:
I agree with @ahmetalpbalkan.  I think we should re-use what is in the docker cli code.

PR #1132, opened 2015-05-06
provisioner: update metadata before install or upgrade

### LINKED ISSUE #1114: 'machine create' with Ubuntu provisioner fails with "Error creating machine: Process exited with: 100. Reason was:  ()"
Creating a machine (I happened to be using the openstack driver) using an Ubuntu image fails if the os needs updates. 

The apt-get step to install curl is the point of failure, so to work around it, you can add 'sudo apt-get update && ' to the command in libmachine/provision/ubuntu.go::Package().

@ehazlett had seen this before, and he's going to fix it ASAP. I wanted to get it documented in the meantime.

ehazlett:
This issues an `apt-get update` for the Ubuntu provisioner so installs and updates do not fail.  This also helps if you package lists are out of date on the base OS.

Closes #1114

### DIFF
--- a/libmachine/provision/ubuntu.go
+++ b/libmachine/provision/ubuntu.go
@@ -49,11 +49,14 @@ func (provisioner *UbuntuProvisioner) Service(name string, action pkgaction.Serv
 func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.PackageAction) error {
 	var packageAction string
 
+	updateMetadata := true
+
 	switch action {
 	case pkgaction.Install:
 		packageAction = "install"
 	case pkgaction.Remove:
 		packageAction = "remove"
+		updateMetadata = false
 	case pkgaction.Upgrade:
 		packageAction = "upgrade"
 	}
@@ -62,6 +65,13 @@ func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.Pack
 	switch name {
 	case "docker":
 		name = "lxc-docker"
+	}
+
+	if updateMetadata {
+		// issue apt-get update for metadata
+		if _, err := provisioner.SSHCommand("sudo -E apt-get update"); err != nil {
+			return err
+		}
 	}
 
 	command := fmt.Sprintf("DEBIAN_FRONTEND=noninteractive sudo -E apt-get %s -y  %s", packageAction, name)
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Ah, nice.  I think we need some robustness around this kind of thing in general, for instance there was an issue about waiting for existing package stuff to finish before attempting our own.

libmachine/provision/ubuntu.go
```diff
@@ -49,11 +49,14 @@ func (provisioner *UbuntuProvisioner) Service(name string, action pkgaction.Serv
 func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.PackageAction) error {
 	var packageAction string
 
+	updateMetadata := true
```
nathanleclaire: Is the idea that eventually this will be configurable / something the users can turn off?

libmachine/provision/ubuntu.go
```diff
@@ -64,6 +67,13 @@ func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.Pack
 		name = "lxc-docker"
 	}
 
+	if updateMetadata {
+		// issue apt-get update for metadata
+		if _, err := provisioner.SSHCommand("sudo -E apt-get update"); err != nil {
```
nathanleclaire: Do we need the same `DEBIAN_FRONTEND` env var here?

libmachine/provision/ubuntu.go
ehazlett: No the idea is that we shouldn't update the metadata when doing a remove.

libmachine/provision/ubuntu.go
ehazlett: No -- it should just pull the latest package lists.

libmachine/provision/ubuntu.go
nathanleclaire: Ah, it's a default, I see.

Probably a little nit-picky, but do you mind moving it into a var block like so at the top of the function?

```go
var (
    packageAction   string
    updateMetadata bool
)
```

clnperez:
Anything I can do to help with this one?

nathanleclaire:
@clnperez I think it's about ready to go, thanks though.

Evan other than my `var` block nit LGTM

libmachine/provision/ubuntu.go
ehazlett: Yep -- good catch

ehazlett:
updated -- @nathanleclaire PTAL

nathanleclaire:
LGTM

clnperez:
Thanks @ehazlett and @nathanleclaire!

nathanleclaire:
> Thanks @ehazlett and @nathanleclaire!

:+1:

ehazlett:
@clnperez yw :D

PR #1137, opened 2015-05-07
Fix formatting and non-returning of error

nathanleclaire:
This PR fixes a few issues that I identified:

1. Error wasn't being checked from `mcn.Get`, so things were blowing up if user tried to operate on a machine name that didn't exist
2. Formatting in the daemon dial was wonky

cc @ehazlett 


ehazlett:
Good catch!  LGTM

PR #1138, opened 2015-05-07
Implement configurable Swarm options

nathanleclaire:
This allows the user to specify a scheduling strategy, and any arbitrary swarm master flag (e.g. heartbeat) they want, to a created machine.  There aren't many `--join` options so I haven't bothered with configuration of the worker bees.

Still to do : 

~~1. Docs~~
~~2. Integration tests~~

However, I wanted to get the code in ASAP, so PTAL.

:bee: :bee: :bee: :bee: :bee: 

cc @ehazlett @vieux @aluzzardi @hairyhenderson @sthulb 

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -204,6 +204,16 @@ var sharedCreateFlags = []cli.Flag{
 		Value: "",
 	},
 	cli.StringFlag{
+		Name:  "swarm-strategy",
+		Usage: "Define a default scheduling strategy for Swarm",
+		Value: "spread",
+	},
+	cli.StringSliceFlag{
+		Name:  "swarm-opt",
+		Usage: "Define arbitrary flags for swarm",
+		Value: &cli.StringSlice{},
+	},
+	cli.StringFlag{
 		Name:  "swarm-host",
 		Usage: "ip/socket to listen on for Swarm master",
 		Value: "tcp://0.0.0.0:3376",
--- a/commands/create.go
+++ b/commands/create.go
@@ -77,11 +77,13 @@ func cmdCreate(c *cli.Context) {
 			TlsVerify:        true,
 		},
 		SwarmOptions: &swarm.SwarmOptions{
-			IsSwarm:   c.Bool("swarm"),
-			Master:    c.Bool("swarm-master"),
-			Discovery: c.String("swarm-discovery"),
-			Address:   c.String("swarm-addr"),
-			Host:      c.String("swarm-host"),
+			IsSwarm:        c.Bool("swarm"),
+			Master:         c.Bool("swarm-master"),
+			Discovery:      c.String("swarm-discovery"),
+			Address:        c.String("swarm-addr"),
+			Host:           c.String("swarm-host"),
+			Strategy:       c.String("swarm-strategy"),
+			ArbitraryFlags: c.StringSlice("swarm-opt"),
 		},
 	}
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -592,6 +592,40 @@ $ docker-machine create -d virtualbox \
     gdns
 ```
 
+##### Specifying Swarm options for the created machine
+
+In addition to being able to configure Docker Engine options as listed above,
+you can use Machine to specify how the created Swarm master should be
+configured).  There is a `--swarm-strategy` flag, which you can use to specify
+the [scheduling strategy](https://docs.docker.com/swarm/scheduler/strategy/)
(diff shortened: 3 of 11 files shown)


sthulb:
Can we hijack this a bit and remove the swarm options from the drivers? i.e. EC2. 
There's not any real reason for them to know about swarm.

nathanleclaire:
@sthulb Hm, there seems to be some cases where we can take them out e.g. `rackspace`. I think in some cases unfortunately they need to be swarm-aware (e.g. for `amazonec2` we have to open `3376` on the security group if the node is a swarm master), although I would definitely be interested in trying to find a more elegant solution.

sthulb:
The drivers have a mechanism to open ports. We could utilise it from hosts.go maybe.

ehazlett:
I think down the road we could use the `AuthorizePort` functionality to help remove this dependency.

nathanleclaire:
Yeah, at least for the immediate future, `AuthorizePort` seems like the right way to handle this.

ehazlett:
Overall looks good.  I think after docs / tests this will be good.

nathanleclaire:
We should make an issue to track removing swarm logic from providers directly.

aluzzardi:
Minor nit:

I'd call it `--swarm-opt` rather than `--swarm-flag` just to be consistent with Docker:

```
docker -d --storage-opt dm.basesize=20G
```

hairyhenderson:
@aluzzardi `--swarm-flag` is consistent with the existing `--engine-flag` option in Machine (from #1040), but I agree they should be consistent with Docker as much as makes sense.

Maybe `--swarm-flag` should be `--swarm-opt`, and `--engine-flag` should be renamed to `--engine-opt` (in a separate PR, of course)? IMO this is totally doable now since `--engine-flag` will be brand-new in 0.3.0.

/cc @nathanleclaire @ehazlett

ehazlett:
+1 for `--engine-opt` and `--swarm-opt`.

nathanleclaire:
Yes, it will be changed to `--engine-opt` and `--swarm-opt`.

nathanleclaire:
Updated with docs and fixed an issue that I identified where the agent wasn't getting created on master nodes.  Now all that remains for this is some tests.

ehazlett:
Is this ready or are we waiting on more tests?

nathanleclaire:
I'm going to add a few more tests, hopefully today!

nathanleclaire:
Note to self: This should probably check if the swarm containers are running before running them.

cpuguy83:
How come there is a `--swarm-strategy` flag? Based on the docs `--swarm-opt` should work for all `manage` flags.

nathanleclaire:
> How come there is a --swarm-strategy flag? Based on the docs --swarm-opt should work for all manage flags.

It will, some flags we are supporting first class though.  For instance, in the engine options we are supporting `--engine-storage-driver` as its own flag for easy reference.  Now, it's definitely a good question whether or not that's the right approach.  

As far as I can see it, the biggest argument in favor of supporting some flags directly is that they are directly "self-documenting" and available to look at in the CLI help text.  The biggest argument against is uniformity - it makes setting the options slightly inconsistent.

@ehazlett WDYT?

nathanleclaire:
Added an integration test.  I'd say I'm ready to merge this whenever you feel comfortable @ehazlett.

ehazlett:
Swarm isn't starting.  Are we mapping the certs properly for the manager?

```
time="2015-05-21T13:20:47Z" level=fatal msg="Couldn't load X509 key pair (, ): open : no such file or directory. Key encrypted?"
```

ehazlett:
Looks like an issue with the Swarm options.  `tlscacert`, `tlscert` and `tlskey` are all blank when sent to the swarm container:

```
       "Cmd": [
            "manage",
            "--tlsverify",
            "--tlscacert=",
            "--tlscert=",
            "--tlskey=",
            "-H",
            "tcp://0.0.0.0:3376",
            "--strategy",
            "binpacking",
```

nathanleclaire:
Hm, I must have flubbed something.  What's the driver and OS you're testing on @ehazlett ?

nathanleclaire:
It's an Ubuntu issue... I thought I updated a line which I hadn't... I got this

nathanleclaire:
Fixed, PTAL

nathanleclaire:
Rebased

nathanleclaire:
Ah, fixing this issue.

ehazlett:
LGTM

vieux:
:+1:

PR #1140, opened 2015-05-07
Implement docker-machine scp

nathanleclaire:
Still to do : test, test, test.  But here it is for playing around with if you are curious.  Not having this feature was making me miserable as I was having to carefully craft my `scp` commands to the server manually.

cc @sthulb @ehazlett @hairyhenderson @SvenDowideit @tianon WDYT

### DIFF
--- a/commands/scp.go
+++ b/commands/scp.go
@@ -1,0 +1,141 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/codegangsta/cli"
+	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/log"
+)
+
+var (
+	ErrMalformedInput = fmt.Errorf("The input was malformed")
+)
+
+var (
+	// TODO: possibly move this to ssh package
+	baseSSHArgs = []string{
+		"-o", "IdentitiesOnly=yes",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=quiet", // suppress "Warning: Permanently added '[localhost]:2022' (ECDSA) to the list of known hosts."
+	}
+)
+
+func getInfoForScpArg(hostAndPath string, mcn libmachine.Machine) (*libmachine.Host, string, []string, error) {
+	// TODO: What to do about colon in filepath?
+	splitInfo := strings.Split(hostAndPath, ":")
+
+	// Host path.  e.g. "/tmp/foo"
+	if len(splitInfo) == 1 {
+		return nil, splitInfo[0], nil, nil
+	}
+
+	// Remote path.  e.g. "machinename:/usr/bin/cmatrix"
+	if len(splitInfo) == 2 {
+		path := splitInfo[1]
+		host, err := mcn.Get(splitInfo[0])
+		if err != nil {
+			return nil, "", nil, fmt.Errorf("Error loading host: %s", err)
+		}
+		args := []string{
+			"-i",
+			host.Driver.GetSSHKeyPath(),
+		}
+		return host, path, args, nil
+	}
+
+	return nil, "", nil, ErrMalformedInput
+}
+
+func generateLocationArg(host *libmachine.Host, path string) (string, error) {
+	locationPrefix := ""
+	if host != nil {
+		ip, err := host.Driver.GetIP()
+		if err != nil {
+			return "", err
+		}
+		locationPrefix = fmt.Sprintf("%s@%s:", host.Driver.GetSSHUsername(), ip)
+	}
+	return locationPrefix + path, nil
+}
+
+func getScpCmd(src, dest string, sshArgs []string, mcn libmachine.Machine) (*exec.Cmd, error) {
+	cmdPath, err := exec.LookPath("scp")
+	if err != nil {
+		return nil, errors.New("Error: You must have a copy of the scp binary locally to use the scp feature.")
+	}
+
+	srcHost, srcPath, srcOpts, err := getInfoForScpArg(src, mcn)
+	if err != nil {
+		return nil, err
+	}
+
+	destHost, destPath, destOpts, err := getInfoForScpArg(dest, mcn)
+	if err != nil {
+		return nil, err
+	}
+
+	// Append needed -i / private key flags to command.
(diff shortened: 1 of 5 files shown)


tcnksm:
:+1:

zchee:
+1

voluntas:
:+1:

sthulb:
+1 in theory. :)

Can we utilise `golang.org/x/crypto/ssh` to do this?

nathanleclaire:
> Can we utilise golang.org/x/crypto/ssh to do this?

I think long term we can and should, perhaps with the same model we are going to use for `ssh` now.  However, it seems like it would require a lot of "rolling from scratch" with low-level bits (at a cursory glance I can't find any well-established libraries to help), and I think the user need for copy files from place to place is pretty immediate.  So, I'd be OK with this for now and a plan to write our own layer eventually.

hairyhenderson:
:+1:

wired8:
:+1:

commands/scp.go
```diff
@@ -0,0 +1,136 @@
...
+		"-o", "IdentitiesOnly=yes",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=quiet", // suppress "Warning: Permanently added '[localhost]:2022' (ECDSA) to the list of known hosts."
+	}
+)
+
+func getInfoForScpArg(hostAndPath string, mcn libmachine.Machine) (*libmachine.Host, string, []string, error) {
+	// TODO: What to do about colon in filepath?
+	splitInfo := strings.Split(hostAndPath, ":")
```
hairyhenderson: Probably easiest to just do:
```go
splitInfo := strings.SplitN(hostAndPath, ":", 2)
```

commands/scp.go
```diff
@@ -0,0 +1,136 @@
...
+	// TODO: possibly move this to ssh package
+	baseSSHOptions = []string{
+		"-o", "IdentitiesOnly=yes",
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+		"-o", "LogLevel=quiet", // suppress "Warning: Permanently added '[localhost]:2022' (ECDSA) to the list of known hosts."
+	}
+)
+
+func getInfoForScpArg(hostAndPath string, mcn libmachine.Machine) (*libmachine.Host, string, []string, error) {
```
hairyhenderson: Could you name the return values here? It'd clarify what the 4 result params are
```go
func getInfoForScpArg(hostAndPath string, mcn libmachine.Machine) (host *libmachine.Host, path string, args []string, err error) {
```

commands/scp.go
nathanleclaire: Nice... that's definitely better

ehazlett:
Reviewed and tested.  Works as expected.

LGTM

nathanleclaire:
Couple things:

1. Think I am going to change the default to _not_ `-r` and make a `-r` flag to keep things consistent with the binary, and safer (won't accidentally cause large copy operations).  WDYT?
2.  Need to add a few unit tests
3.  Need to add integration tests

hairyhenderson:
+1 on the `-r` flag :)

ehazlett:
i agree, however, that does bring up a question: should we support `scp` args?  personally, if i don't get recursion, i don't see myself ever using this command.  i agree i wouldn't want it on by default so i don't accidentally upload something, but not having it would be tough.

nathanleclaire:
> i agree, however, that does bring up a question: should we support scp args? personally, if i don't get recursion, i don't see myself ever using this command. i agree i wouldn't want it on by default so i don't accidentally upload something, but not having it would be tough.

I think we should support a limited subset - I also can't imagine having this without `-r` (it's turned on by default in the current implementation), but I don't think there will be too much demand for extra flags in addition to that.  We could play it by ear if it turns out there are users who want additional flags, additionally I'm pretty sure some of the things such as compression can be enabled in their SSH configuration file if they want it really badly, or by default.

hairyhenderson:
>  I'm pretty sure some of the things such as compression can be enabled in their SSH configuration file if they want it really badly, or by default.

Yeah, the ssh config file should be mentioned in the docs for this. It'd be _nice_ if compression were always-on, but there would then need to be a way to turn it off.

Speaking of which, WDYT about a flag to turn the `-3` option off? If I'm scping something enormous between machines that can talk to each other, the `-3` is going to make things unnecessarily slow.

nathanleclaire:
> Speaking of which, WDYT about a flag to turn the -3 option off? If I'm scping something enormous between machines that can talk to each other, the -3 is going to make things unnecessarily slow.

I'm not opposed, but don't plan to cover it in the original implementation - maybe in a followup PR.

nathanleclaire:
Update: Added some unit tests, made it not default to recursive (there's a flag for it).

commands/scp.go
nathanleclaire: Is it not clear enough from the return at the end?

commands/scp.go
nathanleclaire: Actually, I'm going to punt on this for now.  I think it will be OK since e.g. `$PATH` uses `:` as a delimiter, so I'm pretty sure it's not very common to have in UNIX filepaths, which are pretty much assumed to be what we're handling here at this point..

commands/scp.go
hairyhenderson: Well, I did figure it out just fine when I read the full function. And I can't speak for what the average coder does, but when I'm reading code I generally read a function's signature first to try to figure out what it's doing, then I look for a `return` at the end of the function. In this case it's `return nil, "", nil, ErrMalformedInput`. This is a super-minor thing, and I'm picking nits, but I do think it's nice to make code as easy to read as possible :)

nathanleclaire: I thought that maybe a comment would help, but my experience with comments like that is they usually get out of date and confusing.  Perhaps a more general comment would help clear it up a bit.

hairyhenderson: I'm not a fan of too many comments either - they get insta-stale. But why can't you just name the return values? Or am I missing something?

ehazlett:
LGTM

nathanleclaire:
:tada: :tada: :tada: :tada: :confetti_ball:  :confetti_ball:  :confetti_ball:  :confetti_ball:

PR #1143, opened 2015-05-07
Fix panic when TCP timeout happens on SSH connect

ibuildthecloud:
Now with more running and less panicking.

When a TCP dial errored but the maxDialAttempts was not reached, the conn was left nil and the loop exited.


ehazlett:
LGTM

PR #1146, opened 2015-05-07
If there was an SSH error, log the output

clnperez:
I struggled to figure out why a machine create was failing, and
eventually asked if anyone else had seen the vague error I was getting:
"Error creating machine: Process exited with: 100. Reason was: ()"

It is immensely helpful if we log the command output instead of
just the rc.

It would also have been useful if the driver providers used the output, but
this covers most scenarios since they don't.

### DIFF
--- a/drivers/utils.go
+++ b/drivers/utils.go
@@ -30,7 +30,9 @@ func RunSSHCommandFromDriver(d Driver, command string) (ssh.Output, error) {
 		return output, err
 	}
 
-	return client.Run(command)
+	output, err = client.Run(command)
+	log.Debug(fmt.Sprintf("SSH cmd err, output: %v: %s", err, output))
+	return output, err
 }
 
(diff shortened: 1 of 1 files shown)


clnperez:
Issue 1114 was created for the failure I was seeing.

nathanleclaire:
Oh, we should just `log.Debug` every time.  Even healthy things I prefer to see in debug.

ehazlett:
+1 for logging all to debug as well.  there is no such thing as "too much debug" for me :)

zchee:
+1

clnperez:
Sure thing. I deleted the 'if' so it will always log the ssh output (with -D), and changed the format specifier since most of the time err is nil, not a string.

nathanleclaire:
LGTM

nathanleclaire:
Thanks for this.

ehazlett:
LGTM

PR #1158, opened 2015-05-12
Refactor integration tests

nathanleclaire:
The integration tests before had a ton of code duplication and were generally really confusing.  There's definitely some stuff I'm not super comfortable with in this refactor, but it's a start towards a more modular way of doing things.

Deliberate callouts:

1. I'm not sure if `load ../../helpers` is a good pattern, but I'm not sure if there's a better way to handle it.
2. Maybe `DRIVER` should be set as an environment variable instead of an argument, i.e. `DRIVER=virtualbox ./run-bats.sh core/`
3. The documentation describes the new "interface" and structure pretty well, so take a look at that if you want the bird's eye view.
4. I wonder if there's a way to make it so I don't have to do relative path stuff / invoke the script from a specific directory.
5. There were some environment variables tests for `virtualbox`, which I have taken out, we may or may not want to add them back in. 

cc @ehazlett @hairyhenderson @sthulb for review / feedback
cc @tianon @SvenDowideit @tiborvass for bash auditing if you're willing, it'd be appreciated ;D


hairyhenderson:
I like this!

> I'm not sure if load ../../helpers is a good pattern, but I'm not sure if there's a better way to handle it.

> I wonder if there's a way to make it so I don't have to do relative path stuff / invoke the script from a specific directory.

You can handle this with `$(dirname $0)`. Something like:

```sh
TEST_DIR=$(dirname $0)
```

at the top of `run_bats.sh` should do it. Then you can invoke the script from anywhere, and I think you should be able to `load ${TEST_DIR}/helpers`.

CONTRIBUTING.md
```diff
@@ -87,60 +87,96 @@ the coverage for the VirtualBox driver's package, browse to `/drivers/virtualbox
...
-you have the binary, you can run test against a specified driver:
+This runs tests against the generated binary.  To use, first make sure to
+install BATS (use that link).  Then run `./script/build` to generate the binary
+for your system.
 
-```
-$ bats test/integration/driver-virtualbox.bats
+### Basic Usage
+
+Once you have the binary, change directories into the `script/integration`
```
ehazlett: This should be `test/integration`.

ehazlett:
I like this too!  I do not like changing directories.  An update like @hairyhenderson said would be great.

/cc @aluzzardi i would love your opinions on this?  have you guys done anything with loading, etc in swarm?

CONTRIBUTING.md
nathanleclaire: Ah, nice catch.

test/integration/drivers/virtualbox/upgrade.bats
```diff
@@ -0,0 +1,20 @@
...
+
+@test "$DRIVER: upgrade" {
+  run machine upgrade $NAME
+  echo ${output}
+  [ "$status" -eq 0  ]
+}
+
+@test "$DRIVER: upgrade is correct version" {
+  SERVER_VERSION=$(docker $(machine config $NAME) version | grep 'Server version' | awk '{ print $3; }')
+  [[ "$SERVER_VERSION" != "1.4.1" ]]
```
tianon: Do you want to verify before `upgrade` that `SERVER_VERSION` is `1.4.1` so that this test's assumption is verified too?

nathanleclaire: Yeah, that's a good idea!

tianon:
Nothing seems too insane here IMO. :+1:

ehazlett:
thx @tianon !

nathanleclaire:
Update: I've fixed the `../../helpers` style to be `${BASE_DIR}/helpers.bash` and also now the script works to invoke from anywhere.

nathanleclaire:
PTAL

test/integration/drivers/virtualbox/upgrade.bats
nathanleclaire: Updated!

aluzzardi: We just load helpers in relative form, such as `load ../helpers`. bats already takes care of the rest

aluzzardi: (`.bash` omitted on purpose - bats does that for you also)

nathanleclaire: Omitting the `.bash` doesn't work with this notation, I wanted to avoid the `../../` style in case we ever moved the tests around

aluzzardi:
As a general feedback, on our side we try to be able to run with vanilla `bats` (and we have our `run.sh` script on the side).

The reason is that it forces us to keep bats file simples (you can't rely on an external script doing dirty work for you).

But that may not work for you

test/integration/drivers/virtualbox/upgrade.bats
```diff
@@ -0,0 +1,26 @@
...
+
+export OLD_ISO_URL="https://github.com/boot2docker/boot2docker/releases/download/v1.4.1/boot2docker.iso"
+
+@test "$DRIVER: create for upgrade" {
+  run machine create -d virtualbox --virtualbox-boot2docker-url $OLD_ISO_URL $NAME
+}
+
+@test "$DRIVER: verify that docker version is old" {
+  # Have to run this over SSH due to client/server mismatch restriction
+  SERVER_VERSION=$(machine ssh $NAME docker version | grep 'Server version' | awk '{ print $3; }')
```
tianon: If you're using `awk` anyhow, why not do something more interesting?
```bash
SERVER_VERSION=$(machine ssh $NAME docker version | awk -F ': ' '$1 == "Server version" { print $2 }')
```

test/integration/run-bats.sh
```diff
@@ -0,0 +1,93 @@
...
+if [ "$ARCH" = "x86_64" ]; then
+    ARCH="amd64"
+else
+    ARCH="386"
+fi
+
+# Set this ourselves in case bats call fails
+EXIT_STATUS=0
+
+# TODO: Should the script bail out if these are set already?
```
aluzzardi: We set the equivalent of those directly in our `helpers.bash`:

```bash
# Root directory of integration tests.
INTEGRATION_ROOT=$(dirname "$(readlink -f "$BASH_SOURCE")")

# Test data path.
TESTDATA="${INTEGRATION_ROOT}/testdata"

# Root directory of the repository.
SWARM_ROOT=${SWARM_ROOT:-$(cd "$INTEGRATION_ROOT/../.."; pwd -P)}

# Path of the Swarm binary.
SWARM_BINARY=${SWARM_BINARY:-${SWARM_ROOT}/swarm}
```

Our `run.sh` script may need some of those, so our run script starts with:
```bash
#!/usr/bin/env bash
set -e

cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"

# Load the helpers.
. helpers.bash
```

test/integration/drivers/virtualbox/upgrade.bats
nathanleclaire: O_O

tianon: "Not as clumsy or random as a blaster"

nathanleclaire: hahaha

aluzzardi:
Not sure about the current state of things, I noticed on the removed part that basically the tests were pretty much the same with a different driver.

An alternative to that would be to have test fixtures that test whatever any driver is supposed to do, then each driver test would just set up the environment, set the right DRIVER environment and run the fixtures.

`bats` own tests have an example of this: They do some setting up, then shell out bats which runs the same suite again.

A practical example would probably be better than my gibberish above:

**test/integration/fixtures/driver.bats**
```
@test "$DRIVER: machine should not exist" {
 run machine inspect $NAME		
 [ "$status" -eq 1  ]		
}
```

**test/integration/virtualbox.bats**:
```
setup() {
  # vbox setup and stuff
}

teardown() {
 # same
}

@test "VirtualBox" {
  bats -p "${FIXTURES_ROOT}/driver.bats"
}
```

test/integration/run-bats.sh
nathanleclaire: Interesting -- even if we don't do everything exactly the same I'd like to keep it as consistent as possible, so perhaps I'll rename our variables to be more congruent.

ehazlett:
@aluzzardi +1 i like that too

nathanleclaire:
> Not sure about the current state of things, I noticed on the removed part that basically the tests were pretty much the same with a different driver.

I'm quite confused.  That's what this PR does.  You can invoke a specific test or subset of tests for a specific driver by:

```console
$ ./run-bats.sh virtualbox core/
```

An environment variable for the driver seems more flexible (and I may well move over to that model), but the idea is the same.

I can't tell if there's something in your suggestion which I've missed here?  What is different?  Is the desire to be able to directly invoke `bats virtualbox.bats` and it just works with the correct driver?  I guess I can see the utility of that, since then you could invoke bats directly on the directory and you could run all tests for all drivers with one command (something not really possible with this PR) -- but it still requires us to maintain a file for each driver, which albeit small ones in this case is what I was trying to move away from in this PR.  Running _all_ of the integration tests is a pretty uncommon case anyway and a way to do it concurrently is something we should be orienting towards IMO.

The issue with setup and teardown in our case is mostly that:

1. setup (machine creation) is really expensive time-wise, and naturally we want to encourage people to run the tests frequently, so in some cases we just want to run one create and run ~15-20 tests on that created machine.
2. some tests require a custom create, e.g. engine options, but are shared between drivers, so the usual `setup` / `teardown` sequence doesn't really work well for us since they run between every single assertion in a `bats` file.
3. We _definitely_ need to be able to test driver-specific stuff as well, e.g. I created an instance on AWS with these parameters and I want to verify that it was created correctly with those AWS-specific things, so we have to account for that as well.

> @aluzzardi +1 i like that too

@ehazlett  Which part?

@aluzzardi What does the example you have listed accomplish which is not already covered by this PR?  I don't understand.

nathanleclaire:
> The reason is that it forces us to keep bats file simples (you can't rely on an external script doing dirty work for you).

As far as this goes, I agree that you shouldn't have too much dirty work, but I don't see how else we can clean up all machines after creation without having `machine rm -f $(machine ls -q)` at the end of every BATS file, and it seems silly to me to duplicate that everywhere, especially since it's likely to get forgotten in some cases and leave machines running.

ehazlett:
to the env var and using `bats` direct instead

nathanleclaire:
@ehazlett The issue I have with making the tests directly invokable, instead of always going through a wrapper script, is that it requires each test to do its own cleanup instead of making it automatic, so almost every test will have the boilerplate:

```bash
@test "$DRIVER: remove machine after upgrade test" {
  run machine rm -f $NAME
}

# Cleanup of machine store should always be the last 'test'
@test "$DRIVER: cleanup" {
  run rm -rf $MACHINE_STORAGE_PATH
  [ "$status" -eq 0  ]
}
```

Although if we use bats to wrap bats like @aluzzardi mentioned, maybe we could just chuck the cleanup logic in a `teardown` in helpers.  Actually, I think I might be able to move everything over to the method mentioned, with the one caveat that since all of that boilerplate will be in helpers, we will have to rely on relative paths (and thus update them manually if we move files around).  What do you think?  Want to do things like this?

ehazlett:
Yeah the only thing I worry about with the `teardown` is the invocation on each test.  I think a wrapper would be fine for now (since we currently do it that way) and then continue to iterate / improve as we use it.

nathanleclaire:
Update: Revised this to use a new form which accepts the driver name as an environment variable instead of an argument.

nathanleclaire:
This is pretty much ready to go whenever.

nathanleclaire:
Ah, forgot to update the `scp.bats` test to the new method.  1 sec

nathanleclaire:
Update: Fixed `scp.bats` test.

CONTRIBUTING.md
```diff
@@ -87,60 +87,105 @@ the coverage for the VirtualBox driver's package, browse to `/drivers/virtualbox
...
 ## Integration Tests
+
+### Setup
+
 We utilize [BATS](https://github.com/sstephenson/bats) for integration testing.
-This runs tests against the generated binary.  To use, make sure to install
-BATS (use that link).  Then run `./script/build` to generate the binary.  Once
-you have the binary, you can run test against a specified driver:
+This runs tests against the generated binary.  To use, first make sure to
+install BATS (use that link).  Then run `./script/build` to generate the binary
```
hairyhenderson: I know you didn't change this, but here's a vaguely-related nitpick:

```diff
-install BATS (use that link).  Then run `./script/build` to generate the binary
+[install BATS](https://github.com/sstephenson/bats).  Then run `./script/build` to generate the binary
```

Makes it a little less ambiguous. :)

hairyhenderson:
:+1:

CONTRIBUTING.md
nathanleclaire: Ha, seems like a good update.

nathanleclaire:
I think the Virtualbox saved state test just got lost in the aether, I'll re-add it.

nathanleclaire:
I wonder if it's maybe not a good idea to clean up the newly created store between `.bats` file runs, but rather only once at the end of the wrapper script.  The main reason I am thinking this is : (a) It is more common to have a persistent store and I'd rather catch issues if it were to get "polluted" _but_ more importantly (b) For local providers such as `virtualbox`, the "cached" ISO gets nuked every time.  That's no fun since we end up re-downloading it for every single `.bats` file run.

nathanleclaire:
OK, I think I'm good with this as a final iteration.

Some rough edges have been revised, I've added a section documenting how to change up the creation args, and now the store persists per-`run-bats.sh`-run.  Also re-added the mysterious missing test.

Please take a look @ehazlett

ehazlett:
Awesome.  I'll check this out today!

test/integration/core/core-commands.bats
```diff
@@ -71,6 +67,11 @@ function setup() {
   [[ ${output} == *"not running. Please start this with"* ]]
 }
 
+@test "$DRIVER: machine should now allow upgrade when stopped" {
```
ehazlett: Should this be `machine should not allow upgrade when stopped` -- "not" instead of "now"?

nathanleclaire: Yes

ehazlett:
very awesome. much delete.

LGTM

PR #1165, opened 2015-05-13
Debian provisioner

chantra:
Add support to provision Debian hosts
Refactor the UbuntuProvisioner code into DebianFamilyProvisioner

### DIFF
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -135,7 +135,11 @@ func (provisioner *Boot2DockerProvisioner) GetAuthOptions() auth.AuthOptions {
 	return provisioner.AuthOptions
 }
 
-func (provisioner *Boot2DockerProvisioner) GenerateDockerOptions(dockerPort int) (*DockerOptions, error) {
+func (provisioner *Boot2DockerProvisioner) GetInitSubsystem() int {
+	return SYSVINIT
+}
+
+func (provisioner *Boot2DockerProvisioner) GenerateDockerOptions(dockerPort int, initSystem int) (*DockerOptions, error) {
 	var (
 		engineCfg bytes.Buffer
 	)
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -1,15 +1,7 @@
 package provision
 
 import (
-	"fmt"
-
 	"github.com/docker/machine/drivers"
-	"github.com/docker/machine/libmachine/auth"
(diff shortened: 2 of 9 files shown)


ehazlett:
Thanks for the contribution!  However, in this state this will not work.  Debian uses systemd for it's configuration and the docker install sets it up.  I have a working branch here https://github.com/ehazlett/machine/blob/debian-provisioner/libmachine/provision/debian.go

ehazlett:
Just opened https://github.com/docker/machine/pull/1178 for feedback.

chantra:
@ehazlett it does work on debian 7.8 (which is what I tested it on)

chantra:
@ehazlett ok, I hacked together compatibility for Debian 7 and Debian 8.

I am a go n00b so feedback is more than welcome as I may have approach the solution from the wrong angle.

The end goal is to refactor as much as possible. Debian/Ubuntu should virtually use almost the same driver, only few changes should be required. Same goes to CentOS/RHEL...

The diff currently fail at passing: ./script/validate-dco I am sorting this out on my side.

ehazlett:
Since Ubuntu differs quite a bit from Debian, I'm not sure why we need to deviate from just the Generic provisioner as in my #1178.  That only has the bits that are specific to Debian.  As for the older versions, I would say we only support the latest stable on each.  We don't go back for any other so I'm fine saying it's Debian jessie+.  It's just more code paths and issues to worry about with the various init config systems.

I would like to have a single PR for this.  I'm not saying mine is right -- I am just not sure we need the Debian "family" abstraction as they are quite different (see the diffs between the RedHat PRs -- ~30 lines of code for the others) and there are so many nuances between Ubuntu and Debian, to me they are their own provisioners, only sharing package systems.

Thoughts?

chantra:
@ehazlett I seems there isa lot of code duplication overall. debian and ubuntu provisioner are pretty much the same but for some special cases (checking if sudo is installed, using systemd instead of upstart).
Maybe most of it should go into the generic provisioner though. Anyhow, there seem to be some common patterns that many distros fall into, those could be re-used across the codebase rather than be duplicated IMHO.

ehazlett:
@chantra which duplication?  we could perhaps pull the stuff out of the `Provision` func but the ordering matters and we need to override certain things.  I will see if we can re-org the generic and reduce some of the duplication in both cases.

ehazlett:
Thanks for the contribution! We have merged the Debian provisioner.  As such, closing this one.

PR #1172, opened 2015-05-14
Move over to real retries method

nathanleclaire:
I noticed that on creates it would just burn through the retries very quickly, and also the errors weren't getting logged to `Debug` (saw it in @botchagalupe cast).  I'm still not sure why there are handshake / timeout issues so commonly after we've established a TCP dial works, but this is definitely a cleanup of this section if nothing else (includes sleeps etc. too now).

cc @ehazlett @ibuildthecloud @sthulb 

### DIFF
--- a/ssh/client.go
+++ b/ssh/client.go
@@ -2,7 +2,6 @@ package ssh
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -10,6 +9,7 @@ import (
 
 	"github.com/docker/docker/pkg/term"
 	"github.com/docker/machine/log"
+	"github.com/docker/machine/utils"
 	"golang.org/x/crypto/ssh"
 )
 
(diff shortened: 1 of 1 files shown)


sthulb:
Looks good.

ssh/client.go
```diff
@@ -63,23 +63,30 @@ func NewConfig(user string, auth *Auth) (*ssh.ClientConfig, error) {
 	}, nil
 }
 
+func dialSuccess(client *Client) func() bool {
+	return func() bool {
+		_, err := ssh.Dial("tcp", fmt.Sprintf("%s:%d", client.Hostname, client.Port), client.Config)
```
ehazlett: `if _, err := ...` ?

ehazlett:
Minor nit otherwise looks good.

ssh/client.go
nathanleclaire: Works for me

nathanleclaire:
@ehazlett Updated!

ehazlett:
LGTM

PR #1173, opened 2015-05-14
virtualbox: change --virtualbox-boot2docker-url to --virtualbox-iso

### LINKED ISSUE #1147: Change --*boot2docker* flags and filenames etc. to be OS-agnostic
Right now we have flags and code such as `--virtualbox-boot2docker-url` which specifically mention boot2docker, but they are being used for other, more OS-agnostic things now.  The names should be changed to be more general.  For instance, to `--virtualbox-image`.  Docs, tests, and things will need to be updated as well.

cc @ibuildthecloud @ehazlett @SvenDowideit

ehazlett:
This changes `--virtualbox-boot2docker-url` to `--virtualbox-iso`.

Closes #1147.

For example:

```
docker-machine create -d virtualbox \
    --virtualbox-iso https://github.com/rancherio/os/releases/download/v0.3.1-rc2/machine-rancheros.iso \
    test-rancher
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -492,7 +492,7 @@ Usage: docker-machine create [OPTIONS] [arg...]
 Create a machine
 
 Options:
-   --virtualbox-boot2docker-url                                                                         The URL of the boot2docker image. Defaults to the latest available version [$VIRTUALBOX_BOOT2DOCKER_URL]
+   --virtualbox-iso                                                                         The URL / path of the VirtualBox ISO image. Defaults to the latest available version of Boot2Docker [$VIRTUALBOX_ISO]
    --virtualbox-cpu-count "1"                                                                           number of CPUs for the machine (-1 to use the number of CPUs available) [$VIRTUALBOX_CPU_COUNT]
    --virtualbox-disk-size "20000"                                                                       Size of disk for host in MB [$VIRTUALBOX_DISK_SIZE]
    --virtualbox-import-boot2docker-vm                                                                   The name of a Boot2Docker VM to import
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -953,8 +953,8 @@ INFO[0008] Starting machine back up...
 INFO[0008] Waiting for VM to start...
 ```
 
-> **Note**: If you are using a custom boot2docker ISO specified using
-> `--virtualbox-boot2docker-url` or an equivalent flag, running an upgrade on
+> **Note**: If you are using a custom VirtualBox ISO specified using
```
nathanleclaire: This isn't _strictly_ true any more as it is handled by the provisioner, I wonder if / when we should update it.

drivers/virtualbox/virtualbox.go
```diff
@@ -77,9 +77,9 @@ func GetCreateFlags() []cli.Flag {
 			Value:  20000,
 		},
 		cli.StringFlag{
-			EnvVar: "VIRTUALBOX_BOOT2DOCKER_URL",
-			Name:   "virtualbox-boot2docker-url",
-			Usage:  "The URL of the boot2docker image. Defaults to the latest available version",
+			EnvVar: "VIRTUALBOX_ISO",
```
nathanleclaire: This also needs to be updated in the "de-marshalling" bit which translate flags into struct parameters in the driver's `SetConfigFromFlags` method.

nathanleclaire:
It's a good start, but it also needs to be updated for the `hyperv`, `vmwarefusion`, `vmwarevsphere` drivers.

nathanleclaire:
Mmm, also, all of those have struct parameters which are called some variant of `d.Boot2DockerURL` which needs to be updated as well.

docs/index.md
ehazlett: So when we issue "upgrade" it doesn't pull from latest?

ehazlett:
I purposely didn't update them because I'm not sure they are all supported.  I guess it would be up to the provisioners to make sure to support the various bits.  I originally just did this for vbox as that was what the issue was for and didn't want to make sweeping changes across the drivers.  Perhaps we should just back out until we really know what this all touches.

Issue #1091, opened 2015-05-01
Proposal: Abstract SSH "backends"


nathanleclaire:
The recent SSH changes are great for users who we cannot count on having the `ssh` binary around for (mostly Windows users), but for Mac and Linux users I'd argue that we should default to shelling out to `ssh` instead of the new stuff for two main reasons:

1. `ssh` binary has been around much, much, longer.  It is battle tested and includes many provisions for things that we would otherwise be rolling on our own, e.g. timeout and retry logic.
2.  Debugability.  I've been noticing that it's much harder to get a grasp on what is happening with the native code `ssh` stuff and fix it.

For these reasons I don't think we should drop shelling out to `ssh` entirely like we have in current master.  We should provide a common interface, check for the `ssh` binary (`exec.LookPath`) and if it's there use it, otherwise default to the native code implementation.

Similarly, we _badly_ need `scp` type functionality and it'd be much easier to get by just shelling out than by re-writing scp ourselves (none of us have time for that, really, and I'd rather just send an error saying "scp is not supported" if it is not available until we can get it together with native code).

cc @ehazlett @sthulb @ehazlett

sthulb:
Whilst I agree that the ssh term we have isn't great, I think we should stick with it.

hairyhenderson:
I'm not _totally_ against, this, but I worry that it would add extra maintenance, and it would also mean trying to maintain parity between two different SSH backends...

ehazlett:
So I think we should do the following:

Implement the default to the ssh binary if it's present -- mainly for the following:  we are noticing some quirks (timeouts, retries, handshake errors, etc).  I'm fearful of what else may come and us not being able to fix it fast enough causing stability issues in the release.  We rely heavily on SSH and need it to be solid.  This would allow us an "interim" phase to test out our `crypto/ssh` implementation and get it solid and then switch to it when it's stable.

@hairyhenderson I don't there will be much maintenance as we had it pretty solid and didn't require much maintenance in the recent weeks.  I think we should do the light maintenance in favor of stability with the end goal being a single portable ssh.

Thoughts?

nathanleclaire:
[...]

Issue #534, opened 2015-02-12
Turn StrictHostKeyChecking on for SSH


nathanleclaire:
If you run something like `docker-machine -D ssh dev ls` you'll see the actual SSH command we are running behind the scenes:

```console
$ docker-machine -D ssh dev ls
DEBU[0000] executing: ssh -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -p 65069 -i /Users/nathanleclaire/.docker/machines/dev/id_rsa docker@localhost ls
```

Some of this looks OK (although we probably don't need the port forward to `localhost` any more), but instead of turning off strict host key checking (a security feature to prevent MITM attacks) we should probably be automatically adding the host to the user's `known_hosts` file instead.  The `-o StrictHostKeyChecking=no` was probably OK for local VMs, but the whole game changes when we're SSHing out to machines on the public internet.

ehazlett:
+1.  I had the known_hosts stuff adding when we were using IdentityAuth.
I'll add this to the list.  Thanks

sthulb:
We could also drop shelling out to SSH.

ehazlett:
Yes I'm planning this as part of the core refactor.

sthulb:
Me too. Although it's a large bit of work in terms of security.

sthulb:
Short term should be to implement this issue and then long term remove it in favour of our own client.

nathanleclaire:
Wait, what would we do instead?  I don't think we want to include a whole SSH / readline implementation in the machine code, why not use what's there?

sthulb:
The crypto/ssh package will do 90% of the job.

SvenDowideit:
or, pull ssh out entirely, and use `docker run`

md5:
@nathanleclaire how were you thinking the host's SSH key would be retrieved to be added to `known_hosts`?

I was thinking it might be interesting to add SSH host keys to new `machine` hosts instead of allowing random keys to be generated and accepting them blindly. It looks like cloud-init supports this via the `ssh_keys` setting.

ehazlett:
@md5 yes cloudinit does this.  as a matter of fact, i used this in my cloudinit branch.

md5:
:+1:

sthulb:
My SSH branch will undo any changes to the existing implementation. I have host key checking in my branch.

ehazlett:
[...]

PR #1174, opened 2015-05-14
Move to using SSH shellout by default, and Go native impl as a backup

nathanleclaire:
As discussed in https://github.com/docker/machine/issues/1091

Closes https://github.com/docker/machine/issues/1091 and a variety of stability issues, at least for users who have an `ssh` binary around locally (OSX/Linux/Windows with mysgit and/or Cygwin).

Depends on https://github.com/docker/machine/pull/1172, mostly to make the inevitable rebase a little easier.

Callouts:

- Moves from the `Output` model to simply returning `string` in most cases where it is relevant, in a lot of cases the returned `output` was either getting thrown away with `_` or converted into a string with a little song and dance anyway, so I figured this way was simpler.  We can revisit streaming output if this turns out to be over-simplistic in the future.
- There are some docs to cover how to still use the Go native solution even if you have `ssh` locally and also explain a little behind the scenes logic

Please review this carefully, I have tested every `docker-machine ssh` angle I can think of and a variety of creates etc. but you never know.

Thanks

cc @ehazlett @hairyhenderson @sthulb

### DIFF
--- a/commands/ssh.go
+++ b/commands/ssh.go
@@ -1,28 +1,23 @@
 package commands
 
 import (
-	"io"
-	"os"
+	"fmt"
 	"strings"
 
 	"github.com/docker/machine/log"
+	"github.com/docker/machine/state"
 
 	"github.com/codegangsta/cli"
-	"github.com/docker/machine/drivers"
-	"github.com/docker/machine/ssh"
 )
 
 func cmdSsh(c *cli.Context) {
-	var (
-		output ssh.Output
-		err    error
-	)
+	args := c.Args()
+	name := args.First()
+	cmd := ""
 
-	if len(c.Args()) == 0 {
+	if name == "" {
 		log.Fatal("Error: Please specify a machine name.")
 	}
-
-	name := c.Args().First()
 
 	certInfo := getCertPathInfo(c)
 	defaultStore, err := getDefaultStore(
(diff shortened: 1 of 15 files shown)


ssh/client.go
```diff
@@ -1,53 +1,119 @@
...
 	"github.com/docker/docker/pkg/term"
 	"github.com/docker/machine/log"
+	"github.com/docker/machine/utils"
 	"golang.org/x/crypto/ssh"
 )
 
-type Client struct {
-	Config   *ssh.ClientConfig
+type Client interface {
+	Output(command string) (string, error)
```
sthulb: Do we no longer care about StdErr?

commands/ssh.go
```diff
@@ -1,21 +1,18 @@
...
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/drivers"
-	"github.com/docker/machine/ssh"
 )
 
 func cmdSsh(c *cli.Context) {
 	var (
-		output ssh.Output
-		err    error
+		err error
```
cpuguy83: I don't think you need this declaration here since err is defined many times below.
Or at least can do `var err error` instead of doing the block.

main.go
```diff
@@ -58,6 +59,15 @@ func main() {
 	app.Email = "https://github.com/docker/machine"
 	app.Before = func(c *cli.Context) error {
 		os.Setenv("MACHINE_STORAGE_PATH", c.GlobalString("storage-path"))
+		switch c.GlobalString("ssh-client-type") {
+		case "gonative":
```
cpuguy83: what about just `native` and `distro`? (don't like distro either, but...)

commands/ssh.go
nathanleclaire: Good catch, I think I can get by without this `var` completely.

ssh/client.go
nathanleclaire: Nice catch, maybe we should use `CombinedOutput`?

main.go
```diff
@@ -58,6 +59,15 @@ func main() {
...
+			ssh.SetDefaultClient(ssh.GoNative)
+		case "shellout":
+			ssh.SetDefaultClient(ssh.Shellout)
+		case "":
```
cpuguy83: Seems like this whole bit should just checking for "gonative" and using shellout option otherwise.

main.go
```diff
@@ -100,6 +110,12 @@ func main() {
 			Usage:  "Private key used in client TLS auth",
 			Value:  "",
 		},
+		cli.StringFlag{
+			EnvVar: "MACHINE_SSH_CLIENT_TYPE",
+			Name:   "ssh-client-type",
+			Usage:  "Type of SSH client to use.  Options: gonative, shellout",
+			Value:  "",
```
cpuguy83: See above about the switch, no default value is set here.

ssh/client.go
```diff
@@ -1,53 +1,119 @@
...
+const (
+	NoDefault SSHClientType = ""
+	Shellout  SSHClientType = "shellout"
+	GoNative  SSHClientType = "gonative"
+)
+
+var (
+	baseSSHArgs = []string{
+		"-o", "IdentitiesOnly=yes",
+		"-o", "StrictHostKeyChecking=no",
```
cpuguy83: Bear in mind that with this off, you are open to MITM

nathanleclaire: Yeah.  It's known issue unfortunately: https://github.com/docker/machine/issues/534

Hopefully we can come up with a better solution in the future.

nathanleclaire:
Hm, looks like I've accidentally introduced some merge weirdness.

main.go
nathanleclaire: Hm, so, the reason I didn't just default this to `shellout`, was because we do a check to see if there is an `ssh` binary around to determine if we should use `shellout` or `gonative`.  If the user doesn't have `ssh` I would really like running the commands to be transparent to them anyway without them having to worry about setting the value.  Maybe I could do the check for the binary in the `Before` block instead of in the `ssh` package?

main.go
nathanleclaire: Hm, I think it would be difficult for me to divine what either of those means just by looking at them.  I tried to name them to be very explicit about what they are.  @ehazlett WDYT?

nathanleclaire:
Cool, fixed the merge weirdness / rebased.

main.go
cpuguy83: Honestly the option almost seems like it should be a one-way thing, if it's set then use native, if not then check if ssh bin is available or not.

nathanleclaire: Hm, like a bool flag (e.g. `--go-native-ssh`) instead of a string flag?

cpuguy83: precisely.

nathanleclaire: Yeah, I like that- I guess I had some kind of hazy idea that maybe at some point other SSH backends might be implemented, but that seems kind of like a YAGNI thing to me at this point since I can't think of any obvious examples.

main.go
ehazlett: I think perhaps `native` and `external`.

main.go
ehazlett: +1 for the bool flag

ehazlett:
This needs a rebase.

nathanleclaire:
Rebased.  Need to update a few things before merge though.

nathanleclaire:
Updated:

1. `--native-ssh` bool flag instead of string flag
2. call it `external` and `native` instead of `shellout` and `gonative`
3. Use `CombinedOutput` instead of `Output` (include STDERR)

PTAL

ssh/client.go
nathanleclaire: Updated to use `CombinedOutput`.

nathanleclaire:
Also, updated docs and added integration test.

ehazlett:
LGTM

PR #1175, opened 2015-05-15
Adding new DOCKER_MACHINE_NAME environment variable

### LINKED ISSUE #925: Proposal: add DOCKER_MACHINE_NAME in machine env
Goal: add following line when execute `machine env dev`

    export DOCKER_MACHINE_NAME=dev

Why?
When we want to control more than more machine in the same time, we need to open several terminal tabs, set the different env for each tabs, and switch between these tabs. It's easy to forget where we are. Currently, we can use `export | grep DOCKER_HOST` to see which host we are using. However, sometimes they are only the ipv4/ipv6 addresses and hard to remember.

To release this pain, I propose that we can add a new environment variable `DOCKER_MACHINE_NAME` so that we can get it by `export | grep DOCKER_MACHINE_NAME`.
Moreover, we can set `$PS1` as something like 

    export PS1="(\$DOCKER_MACHINE_NAME) \W \u$ "

Then the machine name will always be displayed in the bash promt. For example:

    (dev) ~ willy $ eval `machine env staging`
    (staging) ~ willy$

hairyhenderson:
Fixes #925 

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	envTmpl = `{{ .Prefix }}DOCKER_TLS_VERIFY{{ .Delimiter }}{{ .DockerTLSVerify }}{{ .Suffix }}{{ .Prefix }}DOCKER_HOST{{ .Delimiter }}{{ .DockerHost }}{{ .Suffix }}{{ .Prefix }}DOCKER_CERT_PATH{{ .Delimiter }}{{ .DockerCertPath }}{{ .Suffix }}{{ .UsageHint }}`
+	envTmpl = `{{ .Prefix }}DOCKER_TLS_VERIFY{{ .Delimiter }}{{ .DockerTLSVerify }}{{ .Suffix }}{{ .Prefix }}DOCKER_HOST{{ .Delimiter }}{{ .DockerHost }}{{ .Suffix }}{{ .Prefix }}DOCKER_CERT_PATH{{ .Delimiter }}{{ .DockerCertPath }}{{ .Suffix }}{{ .Prefix }}DOCKER_MACHINE_NAME{{ .Delimiter }}{{ .MachineName }}{{ .Suffix }}{{ .UsageHint }}`
 )
 
 var (
@@ -30,6 +30,7 @@ type ShellConfig struct {
 	DockerHost      string
 	DockerTLSVerify string
 	UsageHint       string
+	MachineName     string
 }
 
 func cmdEnv(c *cli.Context) {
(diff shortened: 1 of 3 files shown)


sthulb:
LGTM.

ehazlett:
LGTM

nathanleclaire:
I'm not really opposed per se... but, if the user removes the machine, the environment variables will still be populated, potentially leading to confusion.

@hairyhenderson Have you tested this PR on all those different shells?

ehazlett:
@nathanleclaire that happens already today:

```
ehazlett ~> docker-machine ls
NAME      ACTIVE   DRIVER       STATE     URL                         SWARM
rancher            virtualbox   Running   tcp://192.168.99.100:2376   
ehazlett ~> eval $(docker-machine env rancher)
ehazlett ~> echo $DOCKER_HOST
tcp://192.168.99.100:2376
ehazlett ~> docker-machine rm rancher
Successfully removed rancher
ehazlett ~> echo $DOCKER_HOST
tcp://192.168.99.100:2376
ehazlett ~> docker-machine ls
NAME   ACTIVE   DRIVER   STATE   URL   SWARM
ehazlett ~> echo $DOCKER_HOST
tcp://192.168.99.100:2376
```

hairyhenderson:
> I'm not really opposed per se... but, if the user removes the machine, the environment variables will still be populated, potentially leading to confusion.

What @ehazlett said... It's already potentially confusing so this isn't adding anything new. If people do something like `docker-machine ssh ${DOCKER_MACHINE_NAME}` after `rm`ing the host, it'll fail in a not-too-surprising way. :)

> @hairyhenderson Have you tested this PR on all those different shells?

Nope. I don't have access to the Windows shells especially, but I presume that all the other env vars work on all the shells... If the template is handled properly for the different shells already, this should be fine.

But, I will give it a shot with `fish` when I get some time. :fish:

thaJeztah:
@hairyhenderson no need to install fish, here you go: https://registry.hub.docker.com/u/jess/fish/dockerfile/

hairyhenderson:
Thanks @thaJeztah ;)

hairyhenderson:
Ok, I've tested it with `fish`, and it works fine. Incidentally, I didn't end up using `jess/fish` because even though I could easily mount the machine binary in, Machine refused to load the vbox or vmwarefusion drivers (presumably due to missing the `VBoxManage` and `vmx` binaries). Anyway, nothing wrong with `brew install fish` ;)

thaJeztah:
> Anyway, nothing wrong with brew install fish ;)

Heh, alright. It was worth a try :)

ehazlett:
@nathanleclaire any objections?

nathanleclaire:
I'm OK with this, but we need to be certain to test on Powershell and `cmd.exe` before the release.

LGTM :+1:

PR #1189, opened 2015-05-18
Clean up logs package

peter-edge:
Signed-off-by: Peter Edge <peter.edge@gmail.com>

This does a few things:

- Adds missing `*ln` methods to Logger
- Makes Logger settable
- Takes care of concurrency issues
- Exposes NewStandardLogger (aka TerminalLogger from before), allowing same logger but with different writers
- Fixes `*ln` methods to match golang `fmt,log` packages

Some logging within docker-machine might now not produce newlines when the caller expected it to in the past, since now this is idomatic use of `*ln`. I'm happy to do an audit of all log messages to check for this if asked :)


ehazlett:
@nathanleclaire wasnt' there a reason you left the `*ln` out?

nathanleclaire:
> @nathanleclaire wasnt' there a reason you left the *ln out?

Simplicity.  The current implementation is mimicking `logrus`'s API, and I hadn't updated the code which uses the extra `*ln` methods yet to remove those calls, but I'd rather take the `*.ln` methods out in favor of simply `log.Warn` etc. and leave the `*f` methods as they are, without a newline (I think it may be a bit troublesome for contributors to remember every time).  

It _is_ a bit troublesome that this has some inconsistency with the Go standard libraries, so perhaps we need some additional design and discussion around the matter.  The ultimate goal of the `log` package is for the interface to logging the information to be decoupled from the end result, so that more custom things could be done with the information being passed to the logger, e.g. writing it to a file as well as to STDOUT.

@peter-edge I definitely appreciate the effort, however perhaps I could persuade you to open an issue / proposal instead so we could suss out some of the intricacies of the problem?

ngauthier:
But logrus's interface has all three method types, like in this PR: https://github.com/Sirupsen/logrus/blob/master/logrus.go#L82-L94

Additionally, logrus's ln methods actually strip the ln off here: https://github.com/Sirupsen/logrus/blob/master/entry.go#L249-L252

So that it's only when logrus finally runs the logging through the specific TextFormatter (as opposed to any other remote formatter or json or anything) that a newline is added specifically for terminal output:

https://github.com/Sirupsen/logrus/blob/master/text_formatter.go#L92

This is because logrus is designed to ship logs to many systems, and a terminal is just one possible location, and the other protocols have other styles of delimiting log lines.

So, really docker machine isn't logging in the same sense that logrus or go's logger log. It's not to syslog or a remote source, it's pretty much always going to a terminal. So really the logger here is much more akin to fmt.

If we really want to make this a better logger we could make it either use go's standard logger to log to stdout/stderr, or we could just plug in logrus. Then we could remove all the custom logging code in machine.

peter-edge:
+1 on Nick's commentary

This allows other loggers to be developed and plugged in in at will, with the rest of the codebase logging per expected behavior. Sticking with golang's idioms means there are no surprises. I also just did a pass on every single log statement - about 3/4 of them were relying on the newlines, and another 1/4 were not, so there were actually double newlines being logged in a lot of the code :)

peter-edge:
And for the record, that was a really fun series of `vim $(ag --nogroup --ignore Godeps -Q log.Info\( | cut -f 1 -d : | sort | uniq)` commands so my head kind of hurts :)

peter-edge:
Also note re: nick's commentary: this allows other implementations of Logger passed in in the future to do things like decide whether to strip an expected newline. Ie how do you differentiate whether a newline is part of the actual input log message, or just added as part of logging? I would want to know the difference if I were shipping these logs in a structured format.

peter-edge:
Let me know how you want to proceed, I'm open to anything.

peter-edge:
Not to blow up your inbox, but some of the motivating factor behind this starting PR:

I did this over the weekend for fun https://github.com/peter-edge/go-machine and usually when I log things right now, I use this structured logger that can encode logs in a JSON-RPC-able way (well that is if the MSB is not set on a byte, ie >127, fix for that coming), and having this in there will make my life easier.

Also I'm sure I'm not the only one who will want custom logging in the future as docker-machine grows :)

nathanleclaire:
> So, really docker machine isn't logging in the same sense that logrus or go's logger log. It's not to syslog or a remote source, it's pretty much always going to a terminal. So really the logger here is much more akin to fmt.

Yes, it is a bit of a misnomer the way it is set up today, but it is designed to allow us the maximum amount of flexibility in the future (e.g. if we want to add optional color, we can just change the logger when that flag is enabled).  It certainly could use better documentation, but I think mostly people get the idea that `log.Info` is normal output, `log.Debug` is hidden unless debug mode is on, `log.Fatal` bails with a message, and so on.

> If we really want to make this a better logger we could make it either use go's standard logger to log to stdout/stderr, or we could just plug in logrus. Then we could remove all the custom logging code in machine.

We had `logrus` in place before and the reason for this new code was deliberately to take it out and replace it with something that is more suited to our specific needs on this project.  We need more granularity over what we log to STDOUT/STDERR and more control over what happens when someone calls the logging functions, additionally logrus' behavior on Windows shells (where we are attempting to cover the cases of msysgit, Cygwin, cmd.exe and Powershell alike) is tricky to get under control / consistent (frequently you end up with visible escape codes or alternatively formatted logs which are hard to read etc.).  It is intentionally minimalistic for the time being.

If we require that users append `\n` to every `log.*f` call, I think it will be forgotten often and become a point of frustration.  If it is being appended to every case, I would rather just make it the default, and just use `fmt.Print` as an "escape hatch" if finer granularity is needed.  Additionally, it is an eyesore.  I would much rather define our standard interface better.

Does that help to explain the motivation for why things are architected this way currently?

> I did this over the weekend for fun https://github.com/peter-edge/go-machine and usually when I log things right now, I use this structured logger that can encode logs in a JSON-RPC-able way (well that is if the MSB is not set on a byte, ie >127, fix for that coming), and having this in there will make my life easier.

Ah, I see.  Getting a feel for stuff like this is why I suggested a proposal.  In general I like to discuss the motivations for a change before jumping straight into PRs, this sounds like a candidate for a "log driver" if you will.

> Also I'm sure I'm not the only one who will want custom logging in the future as docker-machine grows :)

This is the motivation for why `Logger` is an interface, so that we could have customizability without having to change the rest of the code.

peter-edge:
So I think the entire golang community is pretty used to using "\n", and this would break that, no? I am a big fan of sticking to conventions :) This Logger breaks that convention, so it would be less easy to predict what it does at first glance. Plus the motivations described above (before this PR, machine was double-newlining, it's not possible to differentiate between a log that newlines or not, it's not possible to log something without a newline, etc). But mainly that this is the golang convention.

nathanleclaire:
Oh, and btw @peter-edge, one of the big goals of the 0.4.0 release (next after this one) will be the decouple the Go bindings / `libmachine` from the `docker-machine` CLI as much as possible (so that it's more reusable and not bound to this implementation), so it'd be great to have you stay tuned and contribute to discussions around the design, usage, etc. of the API.

ngauthier:
Is there a public location where 0.4.0 is being discussed?

nathanleclaire:
@ngauthier Not yet, it's kind of just loose talk right now, but eventually there will be a wiki entry, as well as probably separate proposal issues for the various features.

peter-edge:
What is the timeline for 0.3 and 0.4, if you don't mind me asking? :)

nathanleclaire:
@peter-edge No problem.  0.3.0 will follow the release cadence of Docker 1.7 and, while I can't give you exact dates because so many factors go into the mix, should be out in approximately 4-6 weeks from now.  Likewise, 0.4.0 will correspond with the Docker 1.8 release and, as far as we know right now, should be out approximately 2 months after that.

peter-edge:
Hey @ehazlett @nathanleclaire re: https://github.com/docker/machine/issues/1211, can we actually remove Fatal* and Panic*, which I personally dislike a lot in general? Maybe compromise on the other stuff? Ie always add a newline or always don't. @ngauthier

ngauthier:
I think always adding a newline is OK *if* we separate that responsibility from the log message generation responsibility.

So, you have one portion of the logging that has the Info, Infof that never add newlines, but then you link that together with the IOWriterOutput to build a logger, and the IOWriterOutput takes entries and outputs them to an io object (like stdout) with a newline after each one. This is logrus's structure and I like the pluggability.

peter-edge:
Ah, so in this case, I might:

- Remove all *ln methods - a log statement is just a log statement.
- the default logger has an Encoder wrapper that adds a newline (and strips a newline if one is present?).

```go
type Encoder interface {
    Encode(p []byte) ([]byte, error)
}
```

ngauthier:
Yeah, so since right now the only Encoder is a LineBasedIOEncoder then
effectively we get a newline of every log message, but in the future if we
log in other ways the newline isn't there when we do the encoding. I think
it solves both issues.

peter-edge:
Hey, any thoughts on the above? @ehazlett @nathanleclaire

nathanleclaire:
Hmm, it seems reasonable (I haven't had a chance to dig in and play around with the idea though), but we're going to be stalled actually implementing it until after the release, probably.

nathanleclaire:
As far as I can see it my biggest priorities for Machine logging in the immediate future are:

1. Extensibility - being able to separate the "content" from where the logs end up like you mention, so we can multiplex the output to various places such as a file on disk _and_ STDOUT, that way they can be accessed later.
2. Being able to log in various places / situations concurrently and clearly demarcate where the log is coming from.  The immediate need for this would be if we are creating multiple machines at the same time: we should be able to, say, prefix them based on the name of the machine who is logging information, so that we can tell "who is saying what".  The current arch doesn't support this well (or at all) and I'm definitely all ears on suggestions how to make it work well.

nathanleclaire:
I don't think this is the direction we want to go and the pull request is really quite large.  I'm going to close this but thanks for the effort @peter-edge and hope to see you around again soon!

PR #1192, opened 2015-05-19
add user-defined installation of docker

TrumanLing:
add user-defined installation of docker

### DIFF
--- a/main.go
+++ b/main.go
@@ -58,6 +58,7 @@ func main() {
 	app.Email = "https://github.com/docker/machine"
 	app.Before = func(c *cli.Context) error {
 		os.Setenv("MACHINE_STORAGE_PATH", c.GlobalString("storage-path"))
+		os.Setenv("DOCKER_INSTALL_COMMAND", c.GlobalString("docker-install-command"))
 		return nil
 	}
 	app.Commands = commands.Commands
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -100,6 +101,12 @@ func main() {
 			Usage:  "Private key used in client TLS auth",
 			Value:  "",
 		},
+		cli.StringFlag{
```
ehazlett: This is not needed.  If the env var is set, then it will be retrieved in the `installDockerGeneric`.  The reason we use `Before` to set the `MACHINE_STORAGE_PATH` is for testing and how the tests detect.

We shouldn't need this extra flag at all.

ehazlett:
I'm torn about whether to allow the entire command to be passed.  I think this could grow way out of scope and cause more issues.  Perhaps we just limit it to what URL to pass?

thoughts @sthulb @hairyhenderson @nathanleclaire

nathanleclaire:
@TrumanLing, I appreciate the work and the sentiment but I'm -1 on this in general.  Part of the goal of Machine is to standardize these types of things and make them easier to control and reason about.  Could you elaborate on how you envision this being used and/or what problem it is solving?

TrumanLing:
@nathanleclaire @ehazlett Thx for your repsone. Because we make script to configure and install docker, and this script is ```tar```ed in package, we should ```wget``` it and unpack, then run it, so we hope machine have ```command input``` interface.

ehazlett:
Yeah I'm also -1 on this.  We want to give the most consistent experience across the board and having custom built binaries adds potential for provisioning errors.  There is nothing stopping you from using the `ssh` and `scp` commands post-install to customize afterwords.

hairyhenderson:
@TrumanLing - Interesting... Can you explain in more detail what your docker install script accomplishes that isn't accomplished by the current behaviour? It seems to me there might an underlying _other_ missing feature that may be more appropriate to put in.

I'm also wondering if this sort of thing would be best done by some sort of cloud-config support in the future...

sthulb:
I'm not too sure about this either

Issue #886, opened 2015-03-26
Add option to skip SSH provisioning


twhiteman:
Currently the "none" driver name is used to skip the SSH provisioning steps:
https://github.com/docker/machine/blob/45f08d9815734d88d25e9c49873cf67b6699a5fd/libmachine/host.go#L110

I'd like to be able to skip the SSH provisioning too, for a [SDC Docker](https://github.com/joyent/sdc-docker) driver I'm writing.

My initial thoughts are to add *SupportsSSH* or *SupportsSSHProvisioning()* to the Driver interface - but that requires changes to all existing drivers - which I'd prefer to avoid.

Any thoughts for best approach?

nathanleclaire:
@twhiteman What specifically are you wanting to skip?  Installation of Docker?  Setting hostname, configuration of the certs etc.?

twhiteman:
@nathanleclaire All of it :P - I wanted to create a driver (which behaves similar to the *none* driver), but where I can specify the SDC Docker settings, like:
```
   --sdc-account "twhiteman"		SDC Account [$SDC_ACCOUNT]
   --sdc-key "path-to-cert"			SDC SSH key [$SDC_KEY]
   --sdc-region "us-east-3b" 		SDC data center (DC) [$SDC_REGION]
```
This is in the aim of then being able to use *docker-machine env* to populate the docker client settings.

Perhaps another take on this would be imaging the provisioning steps occurring over a REST API?

Note: The SDC Docker already handles the *remote* docker machine setup in other ways, so the other docker-machine commands (start, stop, upgrade, rm, ...) are not applicable for SDC Docker.

nathanleclaire:
What OS is SDC behind the scenes?  If SDC supports `/etc/os-release` I think your easiest way would be to implement a custom provisioner.

twhiteman:
OS: SmartOS / Illumos (aka Solaris) - no /etc/os-release (it's Unix but not Linux).

For custom provisioner, it looks like it still requires SSH, thus my question about an alternative merthod (if I'm reading the code right). C.f.:
https://github.com/docker/machine/blob/45f08d9815734d88d25e9c49873cf67b6699a5fd/libmachine/provision/provisioner.go#L71
https://github.com/docker/machine/blob/45f08d9815734d88d25e9c49873cf67b6699a5fd/libmachine/host.go#L110

twhiteman:
[...]

PR #1197, opened 2015-05-19
Drivers: add joyent triton driver

twhiteman:
Adds support for the Joyent SDC docker platform - fixes 1196

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,7 +2,7 @@
 +++
 title = "Overview of Docker Machine"
 description = "Introduction and Overview of Machine"
-keywords = ["docker, machine, amazonec2, azure, digitalocean, google, openstack, rackspace, softlayer, virtualbox, vmwarefusion, vmwarevcloudair, vmwarevsphere, exoscale"]
+keywords = ["docker, machine, amazonec2, azure, digitalocean, google, openstack, rackspace, softlayer, triton, virtualbox, vmwarefusion, vmwarevcloudair, vmwarevsphere, exoscale"]
 [menu.main]
 parent="smn_workw_machine"
 +++
--- a/drivers/triton/triton.go
+++ b/drivers/triton/triton.go
@@ -1,0 +1,577 @@
+package triton
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"math/big"
+	"net/http"
+	"net/url"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/codegangsta/cli"
+	"github.com/docker/machine/drivers"
+	"github.com/docker/machine/log"
+	"github.com/docker/machine/state"
+	"github.com/docker/machine/utils"
+)
+
+const (
+	driverName                  = "triton"
+	TritonDefaultDockerPort     = 2376
+	TritonDefaultCloudapiDomain = "api.joyent.com"
+)
+
+type Driver struct {
+	CloudApiURL   string
+	DockerApiURL  string
+	DataCenter    string
+	Account       string
+	PrivateKey    string
+	SkipTlsVerify bool
+	MachineName   string
+	storePath     string
+}
+
+func init() {
+	drivers.Register(driverName, &drivers.RegisteredDriver{
+		New:            NewDriver,
+		GetCreateFlags: GetCreateFlags,
+	})
+}
+
+func GetCreateFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:   "triton-url",
+			Usage:  "Triton cloudapi URL",
+			Value:  "",
+			EnvVar: "SDC_URL",
+		},
+		cli.StringFlag{
+			Name:   "triton-datacenter",
+			Usage:  "Triton datacenter name",
+			Value:  "us-east-1",
+			EnvVar: "SDC_DC",
+		},
+		cli.StringFlag{
+			Name:   "triton-account",
+			Usage:  "Triton account name",
+			Value:  "",
+			EnvVar: "SDC_ACCOUNT",
+		},
+		cli.StringFlag{
+			Name:   "triton-key",
+			Usage:  "SSH private key for Triton authentication",
+			Value:  "",
+			EnvVar: "SDC_KEY",
+		},
+		cli.BoolFlag{
+			Name:   "triton-skip-tls-verify",
+			Usage:  "Skip tls verification 'true' or 'false' (defaults to 'false')",
+			EnvVar: "SDC_SKIP_TLS_VERIFY",
+		},
+	}
+}
+
+func NewDriver(machineName string, storePath string, caCert string, privateKey string) (drivers.Driver, error) {
+	return &Driver{
+		MachineName: machineName,
+		storePath:   storePath,
+	}, nil
+}
+
+/* --------------------------------------------------------- */
+/* Implement the drivers.Driver interface.                   */
+/* --------------------------------------------------------- */
+
+// AuthorizePort authorizes a port for machine access
+func (d *Driver) AuthorizePort(ports []*drivers.Port) error {
+	return nil
+}
+
+// Create a host using the driver's config
+func (d *Driver) Create() error {
+	log.Infof("Generating %s user certificates - you will be prompted for", driverName)
+	log.Infof("your SSH private key password (if it's password protected).")
+
+	err := d.RegisterWithSdcCloudApi()
+	if err != nil {
+		return err
+	}
+
+	err = d.GenerateCertificates()
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Success!")
+
+	return nil
+}
+
+// DeauthorizePort removes a port for machine access
+func (d *Driver) DeauthorizePort(ports []*drivers.Port) error {
+	return nil
+}
+
+// DriverName returns the name of the driver as it is registered
+func (d *Driver) DriverName() string {
+
+	/**
+	 * Overriding the driver name to avoid SSH provisioning - issue #886
+	 *
+	 * We only need override the driver name in the 'create' step, so this
+	 * approximately checks if this is a create.
+	 */
+	for _, v := range os.Args {
+		if v == "create" {
+			return "none"
+		}
+	}
+
+	return driverName
+}
+
+// GetIP returns an IP or hostname that this host is available at
+// e.g. 1.2.3.4 or docker-host-d60b70a14d3a.cloudapp.net
+func (d *Driver) GetIP() (string, error) {
+	// DockerApiURL looks like: 'tcp://foo.bar:2376'
+	u, err := url.Parse(d.DockerApiURL)
+	if err != nil {
+		return "", err
+	}
+	return strings.Split(u.Host, ":")[0], nil
+}
+
+// GetMachineName returns the name of the machine
+func (d *Driver) GetMachineName() string {
+	return d.MachineName
+}
+
+// GetSSHHostname returns hostname for use with ssh
+func (d *Driver) GetSSHHostname() (string, error) {
+	return "", fmt.Errorf("SSH is not available for the triton driver")
+}
+
+// GetSSHPort returns port for use with ssh
+func (d *Driver) GetSSHPort() (int, error) {
+	return 22, nil
+}
+
+// GetSSHUsername returns username for use with ssh
+func (d *Driver) GetSSHUsername() string {
+	return d.Account
+}
+
+// GetSSHKeyPath returns key path for use with ssh
+func (d *Driver) GetSSHKeyPath() string {
+	return path.Join(d.storePath, "id_rsa")
+}
+
+// GetURL returns a Docker compatible host URL for connecting to this host
+// e.g. tcp://1.2.3.4:2376
+func (d *Driver) GetURL() (string, error) {
+	return d.DockerApiURL, nil
+}
+
+// GetState returns the state that the host is in (running, stopped, etc)
+func (d *Driver) GetState() (state.State, error) {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+	u := d.GetHttpsURL()
+
+	resp, err := client.Get(fmt.Sprintf("%s/_ping", u))
+	if err != nil {
+		return state.Error, nil
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+
+	if string(body[:]) == "OK" {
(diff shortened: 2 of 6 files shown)


hairyhenderson:
I _literally just_ finished watching [the demo video](https://www.joyent.com/developers/videos/docker-and-the-future-of-containers-in-production) not 10 minutes ago, and I was thinking _man, someone should write a docker-machine driver for SDC, this stuff looks cool!_

No lie!

Anyway, thanks for the PR, @twhiteman!  Looks like you need to [sign your commit](https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work), then the CI build should pass.

Also, looks like some weird mergeness going on - maybe you just need to squash the commit?

/cc @ehazlett @nathanleclaire

twhiteman:
Yeah, my first attempt failed due to go fmt, then I squashed and pushed again... now it's failing due to something that's not my commit... I guess I messed up the squash.

hairyhenderson:
@twhiteman - actually it's just failing because of the lacking signature. The build checks for that :)

twhiteman:
@hairyhenderson but the failed build is complaining about c840be829b3bf9cbd4fed2b0215f5236da351df7, which is not my commit

hairyhenderson:
@twhiteman OH - sorry, I see that now. Just squash it down to a single commit and force-push, then in theory it should be fine :)

ehazlett:
Yeah a rebase / squash should be fine.

hairyhenderson:
@twhiteman - I wonder if something like `joyentsdc` would be a better name for the driver... `sdc` is a little cryptic to people who aren't already familiar with it... It would also match more closely the convention with other drivers (such as `amazonec2` or `vmwarefusion`). WDYT?

twhiteman:
Nod, I could change to that easily enough. The one thing against using "joyent" in the name (and this is a minor point) was that since SDC is open-source, there could be other operators out there (not joyent) that run SDC.

Anyway, I'll go ahead and update to joyentsdc and we can cross that bridge (sdc/joyent naming) if/when we get to it.

hairyhenderson:
This is a good point. What might make sense, then is to have an sdc driver and then add another one (named joyenttriton maybe?) which is based on sdc but sets certain joyent specific things (API endpoint for one). This might be similar to the rackspace and openstack drivers.

docs/index.md
```diff
@@ -1121,6 +1121,25 @@ Options:
...
 
+#### Joyent SmartDataCenter
+
+Create machines on [Joyent's SmartDataCenter](http://joyent.com/).
+
+Before using the Docker service on Joyent's SmartDataCenter, you need to have completed the signup process and generated a set of SSH keys (make sure you know the name and location of your SSH key). Please visit [getting started](https://www.joyent.com/developers/getting-started) for more information.
+
+If you want to try to run SDC on your laptop please visit [the getting started with Cloud on a Laptop section](https://github.com/joyent/sdc#getting-started) or if you already are running SmartDataCenter [here](https://github.com/joyent/sdc-docker) is what you need to do in order to enable the Docker service
+
+        $ docker-machine create --driver sdc --sdc-region=$REGION --sdc-account=$ACCOUNT --sdc-key=$PATH_TO_SSH_KEY
```
misterbisson: Perhaps this would be preferable?

```bash
docker-machine create --driver triton —triton-dc=$DATA_CENTER —triton-account=$ACCOUNT —triton-key=$PATH_TO_PRIVATE_SSH_KEY

**Comment by twhiteman** [comment]
Okay, the driver name and flags have changed to "Triton" and the readme/docs mention it as "Joyent Triton".

As for the SDC (open source platform) driver - there are no differences between SDC and Triton (as yet), so any SDC users will be able to use this Triton driver (or we'll refactor the driver to share/inherit in the future).

File: drivers/triton/triton.go [hunk 2]
```diff
@@ -0,0 +1,573 @@
...
+		GetCreateFlags: GetCreateFlags,
+	})
+}
+
+func GetCreateFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:   "triton-url",
+			Usage:  "Triton Cloudapi URL",
+			Value:  "",
```
hairyhenderson: API URLs generally have defaults in other drivers - I wonder if there's a sane default you could set here?

drivers/triton/triton.go
```diff
@@ -0,0 +1,573 @@
...
+		cli.StringFlag{
+			Name:   "triton-account",
+			Usage:  "Triton account name",
+			Value:  "",
+			EnvVar: "SDC_ACCOUNT",
+		},
+		cli.StringFlag{
+			Name:   "triton-key",
+			Usage:  "SSH private key for Triton authentication",
+			Value:  "",
```
hairyhenderson: Does it make sense to default to `~/.ssh/id_rsa` maybe?

drivers/triton/triton.go
```diff
@@ -0,0 +1,573 @@
...
+	if d.Account == "" {
+		return fmt.Errorf("You must specify the account name using --triton-account")
+	}
+	if d.PrivateKey == "" {
+		return fmt.Errorf("You must specify the SSH key using --triton-key")
+	}
+
+	if d.CloudApiURL != "coal" && strings.IndexAny(d.CloudApiURL, ":./") == -1 {
+		// Shortend format for the cloudapi name, e.g. "us-east-3b"
+		d.CloudApiURL = fmt.Sprintf("https://%s.%s", d.CloudApiURL, TritonDefaultCloudapiDomain)
```
hairyhenderson: This is a little too magical here:
1. There's no mention of `coal` elsewhere (especially in the `--help` message for `--triton-url`)
2. The `--help` message doesn't mention anything about being able to specify a region instead of a full URL

IMO you should add a `--triton-region` flag instead, and require either `--triton-url` or `--triton-region` to be set (but not both). Maybe the CoaL use-case could be handled with `--triton-region coal`.

drivers/triton/triton.go
```diff
@@ -0,0 +1,573 @@
...
+	log.Debugf("Downloading ca.pem file from %s", caUrl)
+
+	var resp *http.Response
+	var err error
+
+	if d.CloudApiURL == "coal" {
+		// Have to use an insecure https request for coal.
+		client := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
hairyhenderson: I'm thinking this could be necessary for other (non-coal) custom deployments of SDC. IMO this should be extracted to a flag (`--triton-tls-skip-verify` or something), which could perhaps default to `true` for the `coal` region.

hairyhenderson:
@twhiteman - there's an awful lot of shelling out going on in the driver. For the places where you shell out to `openssl`, could you use anything from `utils/certs.go` instead? And for the `ssh` shell-outs, could you use the stuff in the `ssh` package instead?

drivers/triton/triton.go
misterbisson: Region is not an addressable entity. Joyent has named data centers with predictable URLs, but open source and private cloud users each have URLs that are less predictable. "Coal" is a special and very specific circumstance for users who've installed the data center management software on their laptops (coal = "cloud on a laptop").

There are three cases here:

1. A URL for the CloudAPI endpoint that assists with the configuration. This can be in a Joyent-hosted data center, a private cloud, or even a coal implementation.
2. A named data center in Joyent's public cloud.
3. A data center named "coal" on a local laptop.

hairyhenderson: Thanks for the clarification @misterbisson :)

I'm thinking maybe a generic `sdc` driver and a Joyent-specific `triton` driver would be least complicated, given all this. The `triton` driver would then take a `--triton-region` (or `--triton-datacenter`) flag, and the `sdc` driver would take a `--sdc-url` argument, and _that_ would be the place for CoaL I think.

It would be a good idea to stay somewhat in line with other CLI utilities, but it's also important for this to look and feel like the rest of the docker-machine drivers. To be clear, my main concern is that it's a little surprising for a flag named `--triton-url` to accept anything other than a real URL.

misterbisson: Another clarification: SDC, like SmartOS, is a component of Triton, but SDC doesn't provide Docker services.

twhiteman:
Thanks for feedback @hairyhenderson.

For command line default values - yes, agreed, I think I can provide better ones.

> There's no mention of coal

I'll update the docs to better reflect the "coal" naming and I'll combine it with a ```--triton-datacenter``` (short name) such that the URL is not double-used (confusing).

> --triton-tls-skip-verify 

Yes, I think a ```--triton-tls-skip-verify``` flag is a good idea, at it could be useful for other situations.

> shelling out

Shell is used for two cases:
* ```coal``` automatic url service discovery - I will try to use the ```ssh``` library, though my first attempt has it prompting me for a password (for a password-less key)
* ```openssl``` key/cert generation - I did look at using ```utils/certs``` and also ```crypto/x509```, but I didn't find how to do the same as what openssl does (and this may be my lack of go-lang familiarity)

Will update with new PR once I finish these items.

arnos:
+1

twhiteman:
A few things changed in this latest PR:
* dropped the `coal` specific handling - this also removed need to shell out for SSH
* added --triton-skip-tls-verify argument
* added --triton-datacenter argument, as easy way to use a shorthand version of the url
* added default values for arguments
* updated documentation

Note: I could not figure out how to replace the shell out for `openssl` and `ssh-keygen`, so I've marked them as dependencies in the documentation.

misterbisson:
@hairyhenderson, @ehazlett: any feedback here?

corbinu:
Any progress here? I am constantly having to switch between different Tritom locations and boot2docker

hairyhenderson:
Sorry for the inattention on my part here... I started a new job and had a new kid in the past 4 weeks, so I haven't had a lot of personal time lately ;)

I'm going to give this a try tonight.

drivers/triton/triton.go
hairyhenderson: @twhiteman - any opinion on this?

hairyhenderson:
```console
$ docker-machine create -d triton --triton-account hairyhenderson --triton-key ~/.ssh/id_rsa triton     
Generating triton user certificates - you will be prompted for
your SSH private key password (if it's password protected).
Success!
To see how to connect Docker to this machine, run: machine env triton
$ eval "$(machine env triton)"
$ docker run --rm hello-world
Unable to find image 'hello-world:latest' locally
latest: Pulling from hello-world (abf65824-5d7a-4cc2-b0a7-834c9cf34013)
91c95931e552: Already exists. 
a8219747be10: Already exists. 
Status: Downloaded newer image for docker.io/hello-world:latest
Hello from Docker.
This message shows that your installation appears to be working correctly.
...
```

Seems to be working!

Code-wise, things look fine (except for one super-minor outstanding suggestion that I'd like an opinion on).

The BATS integration tests don't pass with the driver, but I'm not sure they should, since a bunch of the functionality isn't supported (`stop`/`kill`/`upgrade`/etc... all make no sense for triton).

So from a quick pass this is a :+1: from me. Will need a review from @ehazlett and @nathanleclaire before it can be merged.

hairyhenderson:
This likely is unrelated to the driver, but I'm encountering this while I'm testing it out, so - it looks like Triton ignores the `--memory` flag, and always gives me the 1GB package.

```console
$ docker run -d -m 128m --name foo debian:jessie
e7107503e0434b40a02801282b6a3926057268bbad01495ea35396fdb5e581d0
$ docker inspect -f '{{.HostConfig.Memory}}' foo
1073741824
```

And if I look in the Triton UI (https://my.joyent.com/main/#!/docker/containers), I see it's using the `t4-standard-1G` package, where it _should_ be using `t4-standard-128M`.

But, as I said, this probably isn't related to this PR!

trentm:
@hairyhenderson The problem that you hit with "-m 128m" is a bug with Triton + docker 1.7 (it works with docker <= 1.6) that is on me to fix. The ticket can be viewed here: https://smartos.org/bugview/DOCKER-458

And yah, that issue isn't related to this PR. :)

misterbisson:
@hairyhenderson thanks for giving this a look and :+1:, and congratulations on all the exciting new changes in your life!

drivers/triton/triton.go
twhiteman: Yes, it does actually default to *~/.ssh/id_rsa* in the code when no value is set (but not the flag value seen here). I couldn't (shouldn't?) actually use tilde as the default flag value, as that is a bash'ism, which then means the code would need to look/treat tilde as a special case.

twhiteman:
@hairyhenderson 

> The BATS integration tests don't pass with the driver, but I'm not sure they should,
> since a bunch of the functionality isn't supported (stop/kill/upgrade/etc... all make
> no sense for triton).

I noticed the tests infrastructure has changed a few times (since I started this), what is the recommended way to run the tests against a particular driver?

dekobon:
+1 Please merge this!

misterbisson:
@ehazlett or @nathanleclaire is there anything we can do to move this forward?

corbinu:
Anything I can do to help? I have been having to build my own version every time a new docker-machine is released.

bahamat:
+1 I'd love this.

ehazlett:

@ehazlett & @nathanleclaire

hairyhenderson:
@twhiteman @misterbisson - FYI, there's now a list of available driver plugins in the Docker Machine docs: https://github.com/docker/machine/blob/master/docs/AVAILABLE_DRIVER_PLUGINS.md

Not sure if you guys have a working driver plugin yet, but if so it'd be great to see it listed there.

misterbisson:
Thanks for the followup, @hairyhenderson. @tianon did some investigation in https://github.com/tianon/docker-machine-driver-triton and got the driver plugin working. Sadly, however, the driver plugin isn't enough.

Machine's workflow is to provision a VM using the driver plugin, then configure it (including setting certs) by SSHing into it. That's the part that doesn't work on Triton, since there is no VM. All containers on Triton run on multi-tenant bare metal, so there's nothing that a customer can SSH into to set certs. 

Instead, the certs are set at account creation time when the user uploads an SSH public key. We use [the user's SSH key to generate a TLS cert](https://github.com/tianon/docker-machine-driver-triton/blob/master/src/triton/driver.go#L69-L73), so there's no additional configuration for Machine to do there.

So, while the driver works, Docker Machine doesn't work on Triton because of that incompatibility. Do you have any thoughts on how we can move forward?

corbinu:
@hairyhenderson I would also like to know if there is anyway I can help move this forward. At the moment I am stuck on my own custom fork of the original implementation.

tianon:
My only possible thought was to use the Go SSH package to pretend we're an
SSH server that happens to only implement the exact commands Machine uses
to grab certs, etc, but that feels like going way too far into the wrong
direction. :cry:

Having Machine be a generic interface for managing all possible
configurations of my DOCKER_* environment variables to talk to multiple
Docker API endpoints is my own personal dream -- even things like
Rackspace's Carina could benefit from loosening the coupling of SSH <->
Docker host.

hairyhenderson:
@tianon - In concept it seems pretty straightforward to be able to let drivers override the provisioner, though it probably would require a `Driver` API change. @twhiteman logged #886 for this sort of thing a while back...

PR #1208, opened 2015-05-21
exoscale: destroy the SSH keypair when removing the machine

vincentbernat:
Otherwise, when recreating a machine with the same name, we would get an
error about this SSH keypair already existing.

To avoid any confusion, let's say that machine will handle all keypairs
and the user cannot provide its own. Therefore, move the keypairs into
their own "namespace" and remove the option to specify a keypair.

This is also what is done by Digital Ocean.

First commit is totally unrelated but I didn't want to do a PR just for it. Second commit is needed to get the appropriate method to destroy keypair.

### DIFF
--- a/Godeps/_workspace/src/github.com/pyr/egoscale/src/egoscale/types.go
+++ b/Godeps/_workspace/src/github.com/pyr/egoscale/src/egoscale/types.go
@@ -1,205 +1,208 @@
 package egoscale
 
 import (
-	"net/http"
 	"encoding/json"
+	"net/http"
 )
 
 type Client struct {
-	client *http.Client
-	endpoint string
-	apiKey string
+	client    *http.Client
+	endpoint  string
+	apiKey    string
 	apiSecret string
 }
 
-
 type Error struct {
-	ErrorCode int `json:"errorcode"`
-	CSErrorCode int `json:"cserrorcode"`
-	ErrorText string `json:"errortext"`
+	ErrorCode   int    `json:"errorcode"`
+	CSErrorCode int    `json:"cserrorcode"`
+	ErrorText   string `json:"errortext"`
+}
+
+type StandardResponse struct {
+	Success     string `json:"success"`
+	DisplayText string `json:"displaytext"`
 }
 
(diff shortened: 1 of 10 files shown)


Godeps/_workspace/src/github.com/pyr/egoscale/src/egoscale/types.go
```diff
@@ -12,13 +12,17 @@ type Client struct {
 	apiSecret string
 }
 
-
 type Error struct {
 	ErrorCode int `json:"errorcode"`
```
ehazlett: This looks like it needs a `go fmt`.

drivers/exoscale/exoscale.go
```diff
@@ -383,6 +376,14 @@ func (d *Driver) Stop() error {
 
 func (d *Driver) Remove() error {
 	client := egoscale.NewClient(d.URL, d.ApiKey, d.ApiSecretKey)
+
+	// Destroy the SSH key
+	_, err := client.DeleteKeypair(d.KeyPair)
```
ehazlett: Minor: fyi you can do this as:
```
if _, err := client.DeleteKeypair(d.KeyPair); err != nil {
        return err
}
```

Godeps/_workspace/src/github.com/pyr/egoscale/src/egoscale/types.go
vincentbernat: That's "upstream". `go fmt` would also change more than that. I'll ask.

vincentbernat:
Updated for both comments. Integration tests are successful.

Godeps/_workspace/src/github.com/pyr/egoscale/src/egoscale/types.go
ehazlett: Ah ok -- I didn't check the path.  Thanks.

ehazlett:
LGTM

vincentbernat:
Thanks!

ehazlett:
yw!

PR #1215, opened 2015-05-21
exoscale: fix driver hanging on failed VM operation

janeczku:
Fixes the following issue:   
The exoscale driver does not evaluate correctly the job status of API commands. 
If any job has failed on exoscale's side (e.g. creating new instance) it will not raise the error and instead continue polling the status for the next 2 minutes until `d.waitForVM()` times out.

/cc @pyr

### DIFF
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -419,14 +419,21 @@ func (d *Driver) waitForVM(client *egoscale.Client, jobid string) (*egoscale.Dep
 	i := 0
 	var resp *egoscale.QueryAsyncJobResultResponse
 	var err error
+WaitLoop:
 	for ; i < maxRepeats; i++ {
 		resp, err = client.PollAsyncJob(jobid)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


ehazlett:
@janeczku this needs a rebase.  thanks!

vincentbernat:
The change is quite invasive just to raise an error if `resp.JobStatus == 2`. If you want to cleanup the code style, maybe use separate commits?

janeczku:
@vincentbernat it was just to tempting seeing we could reduce API queries by a factor of 2 (beware the rate limit) ;) Seeing you are uncomfortable with it, i will revert the refactoring and just push the job status eval.

vincentbernat:
Oh, I didn't realize that the number of API queries was divided by 2. My point was only that the original commits contained many changes: logging changes, refactoring and the bug fix. I was just wishing to have a separate commit for the bug fix. Feel free to resubmit the remaining bits in another PR.

@ehazlett LGTM

janeczku:
@vincentbernat :+1:

Issue #230, opened 2015-01-06
UX Suggestion: warn users when commands will take more than 10 seconds or so


nathanleclaire:
Generally I think that people expect commands to run almost instantaneously so I think it'd be good to warn them in commands like `create` that "This might take a few minutes...".

ehazlett:
+1

ehazlett:
+1 to @ahmetalpbalkan

janeczku:
![activityindicator](http://g.recordit.co/XxdLPtDBEc.gif)

Unfinished work. I think it would be best to integrate this in the logger. For that the logger should be a global instance so we can keep track of the state of the progress/activity indicator (e.g. clear the indicator when a new log event occurs and on os.exit).

ahmetalpbalkan:
@janeczku :clap:

nathanleclaire:
@ahmetalpbalkan Will this work compatibly with Windows shells?  `cmd.exe`, msysgit, Powershell?  The Docker client, for instance, currently has a lot of issues with the "real-time" output from `docker pull` etc. in my experience.  I think I might prefer the boot2docker-cli way of just outputting dots.  Still big +1 on telling the user it might take a minute though.

ahmetalpbalkan:
@nathanleclaire if we integrate `docker/pkg/term` correctly to the machine it will work on windows shells as well. We're currently in process of rewriting the ANSI emulation package (and thus the windows parts in `docker/pkg/term`) you can see here: https://github.com/docker/docker/pull/13224.

Once that's done I can help you guys integrate fancy ANSI animations into windows consoles.

Otherwise as is this (and `machine ssh`) won't work on windows right away today.

janeczku:
@nathanleclaire that's because docker's progress indicator is implemented by using ansi escape sequences, which aren't really supported by win shells. The code i am testing should work on any terminal. E.g. msysgit on Windows XP:
http://![msysgit](http://g.recordit.co/tJ7KlbBEsS.gif)

ahmetalpbalkan:
@janeczku true. The ansi emulator we're writing basically will allow stdout/stderr streams that you can set as os.Stdout/os.Stderr and it will capture ANSI sequences and make win32 API calls to achieve the same visual behavior. Please go ahead with your change and once we're done developing the ANSI emulator and integrate it to docker/docker,we're gonna integrate it to machine as well and make windows consoles an officially supported target as well.

janeczku:
@ahmetalpbalkan :+1:

PR #1216, opened 2015-05-21
CLI progress/activity indicator

janeczku:
This implements a progress/activity indicator for commands that take longer than a couple of seconds.

Please feel free to comment or push against this PR's branch if you think the code needs improvement in some way or another.

[asciicast](https://asciinema.org/a/agfb5vwtr0szwsw5p8rkcwo2t)

Related to #230
cc: @ahmetalpbalkan @ehazlett @nathanleclaire

### DIFF
--- a/commands/inspect_test.go
+++ b/commands/inspect_test.go
@@ -52,8 +52,6 @@ func runInspectCommand(t *testing.T, args []string) (string, *libmachine.Host) {
 	shell := os.Getenv("SHELL")
 	r, w, _ := os.Pipe()
 
-	os.Stdout = w
-	os.Stderr = w
 	os.Setenv("MACHINE_STORAGE_PATH", TestStoreDir)
 	os.Setenv("SHELL", "/bin/bash")
 
@@ -95,6 +93,8 @@ func runInspectCommand(t *testing.T, args []string) (string, *libmachine.Host) {
 		t.Fatal(err)
 	}
 
+	os.Stdout = w
+	os.Stderr = w
 	outStr := make(chan string)
 
 	go func() {
--- a/commands/kill.go
+++ b/commands/kill.go
@@ -7,7 +7,9 @@ import (
 )
 
 func cmdKill(c *cli.Context) {
+	log.Spinner.Start("Killing machine(s)")
 	if err := runActionWithContext("kill", c); err != nil {
 		log.Fatal(err)
 	}
+	log.Spinner.Stop()
 }
--- a/commands/restart.go
+++ b/commands/restart.go
@@ -7,8 +7,10 @@ import (
 )
 
 func cmdRestart(c *cli.Context) {
+	log.Spinner.Start("Restarting machine(s)")
 	if err := runActionWithContext("restart", c); err != nil {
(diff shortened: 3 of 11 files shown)


nathanleclaire:
Pretty interesting patch.  I will check it out soon.

ahmetalpbalkan:
I think we should keep the intermediate steps instead of removing them and using only one line. "Docker pull" also has an animation but at the end you see what's happened on each line (i.e. what layers are downloaded). In this case providing more info about the steps can be better IMO.

nathanleclaire:
+1 @ahmetalpbalkan ... the more info, the merrier.

janeczku:
@ahmetalpbalkan Yeah, we can keep those lines and prefix them with a "+" or similar.

ehazlett:
:heart_eyes: 

+1 @ahmetalpbalkan

zchee:
+1
Awesome user experience!

janeczku:
@zchee  :stuck_out_tongue:

ahmetalpbalkan:
@janeczku any plans to move this forward? looks really cool.

janeczku:
@ahmetalpbalkan, the patch is functional as is. Whether it gets moved forward is not really my call: @nathanleclaire @ehazlett

nathanleclaire:
Sorry to be the difficult one here, I appreciate the work, but I think it is best for us if we keep our logging / terminal handling as simple as possible.  However, the good news is that a lot of the work I've been doing with `libmachine` should make creating your own Go wrapper to do whatever fancy tricks you want with the output a lot easier.

ahmetalpbalkan:
:+1: :clap:

janeczku:
@nathanleclaire, i appreciate your explanation and where you are taking `libmachine` with the new API capabilities :+1: 
I do  however disagree with the characterization of this PR as "fancy trick". I don't think this is a fair characterization of what this patch actually implements: Satisfying a basic principle of UI design.
Currently there *is* a lack of feedback during long running commands in `docker-machine`: Often the CLI will just sit there for 10 seconds or even half a minute and longer after the user issued a command -  leaving the user to wonder whether the app has hung-up. 

[UI Design principle No. 1 according to Jakob Nielsen](http://www.designprinciplesftw.com/collections/10-usability-heuristics-for-user-interface-design)

> **Visibility of system status**
The system should always keep users informed about what is going on, through appropriate feedback within reasonable time.

That being said, I am sure there are different and maybe better ways to tackle this UX issue than what constitutes this particular PR. I just wanted to stress that there is need for something like this.

nathanleclaire:
> That being said, I am sure there are different and maybe better ways to tackle this UX issue than what constitutes this particular PR. I just wanted to stress that there is need for something like this.

I agree, and I think it would be better addressed by a simpler method such as printing a `.` periodically while waiting for SSH and a `o` while waiting for the Docker daemon (similar to how b2d-cli worked), and/or printing messages to indicate "This may take a few minutes" where relevant.  Because the Machine CLI and log need to work flexibly in a variety of different situations and OSes, I'm wary of introducing too much complexity.

PR #1217, opened 2015-05-21
logger: fix error logger writing to stdout instead of stderr

janeczku:


nathanleclaire:
Ah, oops.  LGTM cc @ehazlett

ehazlett:
LGTM

PR #1218, opened 2015-05-22
vbox: use intel driver for nic

### LINKED ISSUE #986: virtualbox: Hangs on Waiting for VM to start
Tried starting a VM via `docker-machine start` and it hung on:
````
INFO[0000] Waiting for VM to start...
````
Check the VirtualBox GUI and the VM state was `aborted`.
The underlying cause here was that my VM image was corrupted, but as a user I'd expect machine to be aware of this and provide an appropriate error message.

ehazlett:
This changes the driver in vbox to use intel.  This adds a bunch of network stability to the VM.  See https://github.com/docker/machine/issues/986#issuecomment-103963797

Closes #986

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -273,7 +273,7 @@ func (d *Driver) Create() error {
 
 	if err := vbm("modifyvm", d.MachineName,
 		"--nic1", "nat",
-		"--nictype1", "virtio",
+		"--nictype1", "82540EM",
 		"--cableconnected1", "on"); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


ehazlett:
@nathanleclaire PTAL

nathanleclaire:
LGTM

nathanleclaire:
Eventually it might be nice to be configurable as a flag, however I don't see a pressing need to have it and we can simply add it if it becomes apparent that users want it.

marioidival:
Somebody merge this T_T

ibuildthecloud:
IIRC doesn't using the Intel driver instead of virtio cause a big change in network performance in vbox?

nathanleclaire:
> IIRC doesn't using the Intel driver instead of virtio cause a big change in network performance in vbox?

We're definitely a bit nervous about it, but in general our thinking is to err on the side of stability for now until we can get some benchmarks together for this.  The VBox docs (https://www.virtualbox.org/manual/ch06.html) say:

>  Performance-wise the virtio network adapter is preferable over Intel PRO/1000 emulated adapters, which are preferred over the PCNet family

I definitely would like to make it configurable via a flag, for users who aren't hitting the kind of issues that this is intended to fix.

ehazlett:
Merging this for now for stability.  We can add a flag later if we find it's necessary.  Worst case scenario, users can change it in the VirtualBox UI.

PR #1220, opened 2015-05-22
Adding SSHKeyPath to drivers so it's listed in the `inspect` output

### LINKED ISSUE #1207: Print out entire ssh command issued in debug mode?
Version 0.2 of docker prints out the entire `ssh` command that is being issued as part of the debug output. This behaviour seems to have been removed from the coming 0.3. Can it be put back?

Being able to access the `ssh` command is handy when automating machine, which I do from [machinery](https://github.com/efrecon/machinery). For example, machinery can use `rsync` to copy files between the host and the machine that is being created. To copy, `rsync` needs to know the exact `ssh` command to execute and picking up the command that is used by machine itself guarantees that it will continue to work in future versions.

hairyhenderson:
Fixes #1207

### DIFF
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -157,7 +157,3 @@ func (d *Driver) Kill() error {
 	_, err := drivers.RunSSHCommandFromDriver(d, "sudo shutdown -P now")
 	return err
 }
-
-func (d *Driver) publicSSHKeyPath() string {
-	return d.GetSSHKeyPath() + ".pub"
-}


nathanleclaire:
I'm not quite sold on this.  I don't really want to keep expanding the scope of `inspect`, it was originally added simply as a way to quickly debug and figure out what had been marshalled to disk after a create.  More fields in the structs == more state to keep track of == more things which are likely to go wrong IMO.  There's a lot of code duplication as well.

It sounds like a `--ssh-dry-run` option like I mentioned might solve @efrecon's use case better, no?  IIRC they are interested in getting access to an `ssh` command with the proper arguments to use with `rsync`.

efrecon:
`--ssh-dry-run` (or perhaps just `--dry-run`?) would solve my use case. As long as it is possible to access the `ssh` command, any way would do. In [machinery](https://github.com/efrecon/machinery), I have already made use of this twice:

1. Once to guess the `scp` command when there was no `docker-machine scp` yet.
2. Now to give away the command to `rsync` for copying trees.

efrecon:
On the other hand, @nathanleclaire, even though I fully understand your point, it feels like the identity information is the only missing part from `inspect`. There is already information about the port number for example... I let you decide, since you all have the big picture, as said before `--dry-run` solves my use case(s).

ehazlett:
Yeah I'm on the fence as well.  Although the key info is static so it does seem to make sense in `inspect` -- but I agree I don't want to keep creeping things in to it.

hairyhenderson:
Been mulling over @nathanleclaire's comments for a little while, and I've been flip-flopping between both sides of the fence... I think I still want to put `SSHKeyPath` in `inspect`'s output, though.

I guess the underlying problem here is `inspect` was originally meant as a debugging tool, but there seems to be growing demand for people to be able to use it to integrate Machine into other tools. For my part, I think this is a good thing... `inspect` is essentially public API now (well, as much as it can be in a 0.x release ;)), and I think it'd be good for it to be useful beyond just debugging purposes.

I can understand the reluctance to track more state, but IMO in this case it makes sense. What we're doing right now is creating a key pair (i.e. more state), but we're not tracking that key pair's location except by the assumption that the driver's `GetSSHKeyPath()` function will always return the right path. IMO this isn't a good assumption, because if the currently-hard-coded-in-all-drivers value of `"id_rsa"` ever changes for a driver (or if a user wants to use their own key pair for some reason), existing machines will be broken.

I also totally agree that all the code duplication is a bad thing. But I think that's something that needs to be addressed separately (there's lots of other fields that get duplicated).

Oh, and one last point - the `generic` driver already exposes its `SSHKey` field to `inspect`, so there's already a precedent.

nathanleclaire:
I should clarify that upon reflection, I'm not really opposed to showing the SSH key path in `inspect`, in fact showing things like this, driver status, etc. could be quite useful, but I think we have two main hurdles we need to deal with before adding more functionality to `inspect` like this:

1. Hardcoded path values in the machine config are problematic in general.  If we stored them in a fashion which was more akin to `"SSHKeyPath": "${STORE_PATH}/machines/${MCN_NAME}"` (just spitballing on this syntax) and did custom de-marshalling (remember Windows in this puzzle as well), we could be a step closer to portable config.  Putting this on disk just exacerbates the problem.  We've been seeing issues lately where, say, users try to move their `~/.docker/machine` directory to another computer, and can't use their machines because the paths are all wrong in the config (even though it could probably work with cloud providers).  To resolve these problems we need a better way of saying, "Go look for this value before you show it to the user".
2. Code / functionality duplication in the drivers.  We're long overdue for consolidating common fields like SSH user, memory, etc. in general into an embedded base struct, and things like this will just make it harder to do.  I think we need to fix our situation with that first before adding more stuff.  Unfortunately, it's been my experience that if things don't get fixed up now, they never will and they will cause problems down the line.

> I can understand the reluctance to track more state, but IMO in this case it makes sense. What we're doing right now is creating a key pair (i.e. more state), but we're not tracking that key pair's location except by the assumption that the driver's GetSSHKeyPath() function will always return the right path. IMO this isn't a good assumption, because if the currently-hard-coded-in-all-drivers value of "id_rsa" ever changes for a driver (or if a user wants to use their own key pair for some reason), existing machines will be broken.

Accurately returning the SSH key's location is a promise that the struct makes when it fulfills the interface.  If that method misbehaves, it is a bug.

> Oh, and one last point - the generic driver already exposes its SSHKey field to inspect, so there's already a precedent.

That driver _has to_ do that: the others don't.

hairyhenderson:
@nathanleclaire - thanks :)

> 1. Hardcoded path values in the machine config are problematic in general. If we stored them in a fashion which was more akin to `"SSHKeyPath": "${STORE_PATH}/machines/${MCN_NAME}"` (just spitballing on this syntax) 

Yeah - I like that idea. How about using templates and storing something like `"SSHKeyPath": "{{StorePath}}/machines/{{MachineName}}"`? Not sure if this would be any better than the sh-like syntax, but it's an idea ;)

> 2. Code / functionality duplication in the drivers. We're long overdue for consolidating common fields like SSH user, memory, etc. in general into an embedded base struct, and things like this will just make it harder to do.

Yes, absolutely :) I started messing around with this yesterday at https://github.com/hairyhenderson/machine/tree/default-driver-impl - I'll turn that into a PR at some point soon.

> Accurately returning the SSH key's location is a promise that the struct makes when it fulfills the interface. If that method misbehaves, it is a bug.

Agreed, but my thinking was more that the current behaviour makes it difficult to ever change the location or allow users to specify a custom location. But de-duplicating the drivers will mitigate this.

efrecon:
From an outside point-of-view and without knowing the internals of `docker-machine`, I think that using templates is a very good idea, especially since Go seems to have [good support for templates](http://golang.org/pkg/text/template/). That would open for future more complex (internal?) scenarios without bringing much more complexity to the code. As long as `docker-machine inspect` return the state at the time of the call and given the current configuration, it is all good for all external tools that wish to automate `docker-machine` in one way or another.

ehazlett:
@nathanleclaire +1 for templating -- being more portable is a great thing

hairyhenderson:
Just rebased this to merge in the changes from #1273 - makes this _significantly_ simpler... But also breaks stuff ;) I'll get back to this... eventually...

dmp42:
@hairyhenderson is this still actual, or do you feel the latest master fixes the problem? (in light of https://github.com/docker/machine/issues/1207#issuecomment-118990751 )

hairyhenderson:
@dmp42 - I'd still be interested in getting this merged. I addressed one of @nathanleclaire's concerns (about duplication) in #1273, and I think the other main concern (hard-coding paths) absolutely needs to be addressed (we can't have portable configs without addressing that). I _don't_ think that needs to be a blocker of this PR though, since we have a whole pile of other hard-coded paths already. IMO one more won't kill us, and a solution needs to be generic enough to handle all the paths anyway, not just this one.

dmp42:
@hairyhenderson looking at some of the code here, it's something I'm already doing as part of `generic` cleanup.
I'll submit it soon and see you over there in a moment ;)

dgageot:
@hairyhenderson would you still want to this PR merged? If so, can you take a look at why the tests are failing? Thanks a lot!

dmp42:
@dgageot unfortunately we rolled back #2033 which was covering that as well.
Let's keep this open for the time being while I re-PR.

dgageot:
@hairyhenderson The PR @dmp42 is talking about is https://github.com/docker/machine/pull/2117

hairyhenderson:
Ok, thanks. Once #2117 lands I'll rebase and see what's left.

hairyhenderson:
Heh. So I rebased after #2117, and there's one piddly change left - really just a removal of an unreferenced function. It's no longer relevant at all to the original point of the PR. So I'm going to :skull: this.

There's probably other unreferenced functions that we could ferret out later as a clean-up PR ;)

dmp42:
Thanks @hairyhenderson !

More cleanup incoming indeed.

PR #1228, opened 2015-05-25
Implemented '--tls-san' global arg for adding extra SANs

### LINKED ISSUE #1227: Server-side certs are too restrictive, need ability to add extra SANs
Right now, each generated server-side certificate (both for the Swarm master and the docker daemon) contain only one SAN (Subject Alt Name) which points to the Host IP (usually localhost).

This makes it too restrictive, it's not possible to connect to my Swarm API from my internal network...

I propose a global option `--tls-san' which allows to specify extra SANs.

samalba:
Added the ability to add extra SANs to all server-side certificates generated by machine.

Fixes #1227

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -67,6 +67,7 @@ func cmdCreate(c *cli.Context) {
 			ClientKeyPath:  certInfo.ClientKeyPath,
 			ServerCertPath: filepath.Join(utils.GetMachineDir(), name, "server.pem"),
 			ServerKeyPath:  filepath.Join(utils.GetMachineDir(), name, "server-key.pem"),
+			ServerCertSANs: c.GlobalStringSlice("tls-san"),
 		},
 		EngineOptions: &engine.EngineOptions{
 			ArbitraryFlags:   c.StringSlice("engine-opt"),
--- a/libmachine/auth/auth.go
+++ b/libmachine/auth/auth.go
@@ -11,4 +11,5 @@ type AuthOptions struct {
 	ServerKeyRemotePath  string
 	PrivateKeyPath       string
 	ClientCertPath       string
+	ServerCertSANs       []string
 }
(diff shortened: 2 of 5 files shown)


samalba:
btw, I still need to add an integration test and make sure the regenerate-certs works (just realized it)...

ehazlett:
cool thanks!

samalba:
Added integration tests. It also tests that `regenerate-certs' works as expected.

ehazlett:
This looks good.  Should we have the ability to specify `--tls-san` during regeneration?  For example:

```
docker-machine --tls-san foo.local --tls-san localhost regenerate-certs test-vbox
```

That could be a follow up PR.

samalba:
That sounds like a good idea to me. I'll keep it for this week-end. I can add it to this PR

ehazlett:
:+1:

evq:
Hey I was wondering if there were any plans to merge this soon?

I found it useful while trying out the new experimental docker 1.8 features on AWS where I wanted the certificates to be signed for both the private and public interfaces. Also I couldn't find any commit by @samalba which allowed it to be respecified during regeneration so I that implemented it on my local branch. I can create a PR for it if you guys are interested.

Thanks!

samalba:
Can you submit your changeset against my fork so I can incorparate it to this PR (still under your name), so we have both commits at the same location?

This is where it leaves: https://github.com/samalba/machine/tree/tls-san

evq:
Sure will do! I added an integration test case and will push once I get the tests running locally.

ChrisPearce:
:+1:
This would be a very helpful addition.

stephanbuys:
@ehazlett any ideas if this will be merged soon?

dmp42:
@samalba do you still plan to carry this one through? :-)

ChrisPearce:
Is this PR still being considered?

It would help for the scenario of trying to access the docker host from within a container.

Currently, the default address assigned to access the docker host from within a container is 172.17.42.1. Unfortunately this IP address does not seem to be included in the common name of the default certificate. The --tls-san option/argument functionality would provide an easy way to allow the IP address to be accepted by the certificate checking.

PR #1231, opened 2015-05-26
Moved VMware vSphere driver to main b2d release tree

frapposelli:
vSphere driver can be moved to the b2d main branch as it only depends on VMware tools that have been merged already.


ehazlett:
LGTM

PR #1235, opened 2015-05-26
add static build script

### LINKED ISSUE #1234: Add static compile to build script
In order to use machine in a stripped environment (i.e. `FROM scratch` or `busybox` etc) it would be nice to statically compile as part of the build script.

ehazlett:
Adds a script to statically build Machine.

Closes #1234.

### DIFF
--- a/script/build-static
+++ b/script/build-static
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -e
+
+if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Mmm maybe we could just make this the default?

dmp42:
@ehazlett I'm closing this since the current build system lets you specify `STATIC` (and cross-compilation implies static).
Let me know if I missed something here.

PR #1247, opened 2015-05-27
redhat: fix tty provision bug

### LINKED ISSUE #1245: RedHat Provisioning tty allocation
As part of #1242 the general `-t` was removed from the base args for the external SSH client.  This breaks provisioning on RedHat.

ehazlett:
This fixes #1245.

### DIFF
--- a/drivers/utils.go
+++ b/drivers/utils.go
@@ -8,15 +8,15 @@ import (
 	"github.com/docker/machine/utils"
 )
 
-func RunSSHCommandFromDriver(d Driver, command string) (string, error) {
+func GetSSHClientFromDriver(d Driver) (ssh.Client, error) {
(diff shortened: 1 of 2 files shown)


ehazlett:
@nathanleclaire this fixes the bug.  there is some duplication as we need to use the RedHat provisioner in the `SetHostname` func.  For now, it works.  PTAL

nathanleclaire:
LGTM

nathanleclaire:
Would be good to get test(s) in for this somehow, not exactly sure how though.  Maybe a test specifically using `amazon` driver with a custom AMI.

ehazlett:
Yeah good idea.  I will do that in a follow up (#1250).

PR #1255, opened 2015-05-28
exoscale: ensure we sleep a bit while waiting for the VM to be ready

vincentbernat:
Otherwise, the timeout is reached far too soon: instead of waiting for
the VM for two minutes, we wait only for a couple of seconds.

### DIFF
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -433,13 +433,11 @@ WaitLoop:
 		}
 		switch resp.Jobstatus {
 		case 0: // Job is still in progress
-			continue
 		case 1: // Job has successfully completed
 			break WaitLoop
 		case 2: // Job has failed to complete
 			return nil, fmt.Errorf("Operation failed to complete")
 		default: // Some other code
-			continue
 		}
 		time.Sleep(2 * time.Second)
 	}
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/machine/drivers"
 	"github.com/docker/machine/log"
 	"github.com/docker/machine/state"
+	"github.com/docker/machine/utils"
 	"github.com/pyr/egoscale/src/egoscale"
 )
 
@@ -421,28 +422,24 @@ func (d *Driver) Kill() error {
 
 func (d *Driver) waitForVM(client *egoscale.Client, jobid string) (*egoscale.DeployVirtualMachineResponse, error) {
 	log.Infof("Waiting for VM...")
-	maxRepeats := 60
-	i := 0
 	var resp *egoscale.QueryAsyncJobResultResponse
 	var err error
-WaitLoop:
-	for ; i < maxRepeats; i++ {
+	if err = utils.WaitForSpecificOrError(func() (bool, error) {
 		resp, err = client.PollAsyncJob(jobid)
 		if err != nil {
-			return nil, err
+			return true, err
 		}
 		switch resp.Jobstatus {
 		case 0: // Job is still in progress
 		case 1: // Job has successfully completed
-			break WaitLoop
+			return true, nil
 		case 2: // Job has failed to complete
-			return nil, fmt.Errorf("Operation failed to complete")
+			return true, fmt.Errorf("Operation failed to complete")
 		default: // Some other code
 		}
-		time.Sleep(2 * time.Second)
-	}
-	if i == maxRepeats {
-		return nil, fmt.Errorf("Timeout while waiting for VM")
+		return false, nil
+	}, 60, 2*time.Second); err != nil {
+		return nil, err
 	}
 	vm, err := client.AsyncToVirtualMachine(*resp)
 	if err != nil {
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -347,8 +347,7 @@ func (d *Driver) Start() error {
 	if err != nil {
 		return err
 	}
-	_, err = d.waitForVM(client, svmresp)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Couple of things:

1. Why does the driver do this kind of waiting in the first place?  Everything of this nature should be handled in the core `libmachine` code for consistency.  `GetState`, IIRC, should be the method which indicates to `libmachine` that it is OK to proceed with provisioning etc.
2. As a side note, `utils.WaitFor` is the canonical way to handle these kind of try-something-then-sleep-then-try-it-again loops.  Please use it if you need to do this kind of thing.

vincentbernat:
The API is asynchronous. The machine doesn't even exist until the job is completed. We are not waiting for the machine to start but for the machine to be created. If we try to use `GetState`, we would get an error, the machine doesn't exist. Once the job completes, the machine is starting and we leave libmachine handles that.

I'll look to use `utils.WaitFor`.

vincentbernat:
About `utils.WaitFor`, there is no error handling possible. I have added variants with ability to return errors in a separate commit. Tell me if that's OK for you or if we keep the original code.

janeczku:
> About utils.WaitFor, there is no error handling possible. I have added variants with ability to return errors in a separate commit. Tell me if that's OK for you or if we keep the original code.

:+1: If `utils.WaitFor` is to be the all purpose wait method then it needs error handling (e.g. breaking the wait loop if the wrapped function returns an error)
```
func WaitFor(f func() (bool, error)) error {
```

janeczku:
> Why does the driver do this kind of waiting in the first place? Everything of this nature should be handled in the core libmachine code for consistency. GetState, IIRC, should be the method which indicates to libmachine that it is OK to proceed with provisioning etc.

I agree. Except, libmachine currently does not check GetState at all after machine has been created. It immediately starts the TCP check via WaitForSSH without knowing if the machine has even started yet.

vincentbernat:
I have updated the PR by using `utils.WaitForSpecificOrError` which was submitted a few hours ago and do the same stuff that the functions I previously introduced.

vincentbernat:
@nathanleclaire any issue left with this PR?

nathanleclaire:
`WaitForSpecificOrError` seems fine to me.  

> @nathanleclaire any issue left with this PR?

Just one more thing -- Can you move the closure into its own (named) function please?  :)

Thanks!

vincentbernat:
I need to access the last response obtained in the loop to be able to get the VM object and there is no way to return a value with `WaitForSpecificOrError`. Should I introduce another function `WaitForSpecificOrErrorAndReturnResult`?

vincentbernat:
I cannot pass a parameter either, so I'll have to keep an anonymous function. I can move the main body of the loop into a separate function and use only a minimal anonymous function.

nathanleclaire:
@vincentbernat If you pass in a pointer to `client` as a parameter to a function which returns a `func() (bool, error)` as the argument for `utils.WaitFor`, will that work?

janeczku:
I would just put the second API call in its own function and have the waitForVm method not return any values. It's much more economical that way since all but one call to waitForVm discard the return value anyway so the second API call is just overhead.

vincentbernat:
@nathanleclaire is that really an improvement to use pointers to return values?

pyr:
What's the rationale behind moving to a named function ? It probably won't increase readability and passing around pointers defeats the purpose of having a clear separation of concerns. Are there other concerns that are trying to be adressed ?

As a more generic point, it seems awkward to try and force this specific use case out of a loop and into the `utils.Wait*` family of functions, the driver will probably be at risk of running behind the probable updates of this API and looking at the diff output, moving out of a loop doesn't gain much, again, maybe I'm not seeing the greater overarching point, care to elaborate ?

vincentbernat:
@nathanleclaire so, I have pushed another commit with some refactoring (but this is quite independent of the issue this PR was trying to fix). To avoid complexity of passing arguments through a pointer but keeping the ability to use `utils.WaitForSpecificOrError`, a new `waitForJob` function is introduced and used by most of the code. The job checking is done in another function, like requested. A closure is still present to pass the appropriate jobid we are waiting for. The `waitForVM` is still using this function and will do an extra API call to get the VM from the job status. This keeps the code simple enough.

nathanleclaire:
> What's the rationale behind moving to a named function ? It probably won't increase readability and passing around pointers defeats the purpose of having a clear separation of concerns. Are there other concerns that are trying to be adressed ?

> As a more generic point, it seems awkward to try and force this specific use case out of a loop and into the utils.Wait* family of functions, the driver will probably be at risk of running behind the probable updates of this API and looking at the diff output, moving out of a loop doesn't gain much, again, maybe I'm not seeing the greater overarching point, care to elaborate ?

Consistency.  As the project is large and continues to grow, it is critical to be able to context switch with low overhead, as well as to be able to expect uniformity with common behavior patterns such as "wait until this condition is met, with a check every X seconds".

> To finish off, you initialy had two points and the WaitFor one was a side-note. Since this obviously spans a longer discussion, maybe committing the fix for the bug would be a good start then we could move on to a PR discussing the merits of loops and loop-mimicking functions ?

Generally it's been my experience that it is much more likely for things to be cleaned up before they get merged upstream than after.  So, even though it may seem picky, it really is with the best interests of the project at heart with respect to readability and maintainability.  I won't always be perfect, but I want to get an understanding of the problem before agreeing to merge the code, as well as encourage code re-use where possible.

nathanleclaire:
@vincentbernat Looks clean -- I like it -- thanks for your persistence.

LGTM cc @ehazlett

pyr:
@nathanleclaire Thanks for clarifying. I take your point on enforcing cleanups before. I don't think the consistency argument holds water for things where the language is expressive enough to clearly express intent. That being said, when in Rome...

PR #1257, opened 2015-05-28
redhat: use engine version 1.7rc1

ehazlett:
This upgrades the Docker Engine to the latest RC.  This is a temporary bump until we get an official install process (either via the install script for via an official repo).

We will need to update the final URLs to the 1.7 RPM before releasing the 0.3.0 as well.  This at least will get us testing on 1.7.

### DIFF
--- a/libmachine/provision/centos.go
+++ b/libmachine/provision/centos.go
@@ -8,7 +8,7 @@ const (
 	// TODO: eventually the RPM install process will be integrated
 	// into the get.docker.com install script; for now
 	// we install via vendored RPMs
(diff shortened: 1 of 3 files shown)


ehazlett:
I don't think we will have a repo / install script integration for 1.7  (talked to Jess).  We will probably have to pin this until we get the install script.

ehazlett:
@nathanleclaire PTAL

nathanleclaire:
LGTM

nathanleclaire:
Pinning it is fine for now IMO especially as this feature is considered highly experimental

ehazlett:
Cool.  Sounds good.  Thanks!

PR #1262, opened 2015-05-28
Don't allow remove on non-stopped machines

### LINKED ISSUE #1179: docker-machine rm should not work on non-stopped machines
It's too dangerous IMO.  You could easily blow away machines which you did not mean to, and also breaks existing idioms e.g. Docker and Vagrant won't allow you to remove containers and machines which are running.

cc @ehazlett @hairyhenderson @sthulb WDYT?

I've also been considering a "pinning" mechanism which would allow you to deliberately protect machines from deletion even if they are stopped, but first things first.

nathanleclaire:
Closes #1179

Please review and after first code review I will add some tests cc @ehazlett @hairyhenderson 

There is also some general cleanup of messaging on actions, etc. to make it more consistent and informative.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -42,6 +42,16 @@ var (
 	ErrExpectedOneMachine = errors.New("Error: Expected one machine name as an argument.")
 )
 
+type ErrorMachineAction struct {
+	actionName    string
+	machineName   string
+	originalError error
+}
+
+func (e ErrorMachineAction) Error() string {
+	return fmt.Sprintf("Error running action %q on machine %q: %s", e.actionName, e.machineName, e.originalError)
+}
+
 type machineConfig struct {
 	machineName    string
 	machineDir     string
@@ -412,13 +422,16 @@ func machineCommand(actionName string, host *libmachine.Host, errorChan chan<- e
 		"ip":            host.PrintIP,
 	}
 
-	log.Debugf("command=%s machine=%s", actionName, host.Name)
-
 	if err := commands[actionName](); err != nil {
-		errorChan <- err
+		errorChan <- ErrorMachineAction{
+			actionName:    actionName,
+			machineName:   host.Name,
+			originalError: err,
+		}
 		return
 	}
 
+	log.Info(host.Name)
 	errorChan <- nil
 }
 
@@ -451,7 +464,7 @@ func runActionForeachMachine(actionName string, machines []*libmachine.Host) {
 		serialChan := make(chan error)
 		go machineCommand(actionName, machine, serialChan)
 		if err := <-serialChan; err != nil {
-			log.Errorln(err)
+			log.Error(err)
 		}
 		close(serialChan)
 	}
@@ -461,7 +474,7 @@ func runActionForeachMachine(actionName string, machines []*libmachine.Host) {
 	// rate limit us.
 	for i := 0; i < numConcurrentActions; i++ {
 		if err := <-errorChan; err != nil {
-			log.Errorln(err)
+			log.Error(err)
 		}
 	}
 
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -32,13 +32,13 @@ func cmdRm(c *cli.Context) {
 
 	for _, host := range c.Args() {
 		if err := mcn.Remove(host, force); err != nil {
-			log.Errorf("Error removing machine %s: %s", host, err)
+			log.Errorf("Error running action %q on machine %q: %s", "remove", host, err)
 			isError = true
 		} else {
-			log.Infof("Successfully removed %s", host)
+			log.Info(host)
 		}
 	}
 	if isError {
-		log.Fatal("There was an error removing a machine. To force remove it, pass the -f option. Warning: this might leave it running on the provider.")
+		log.Fatal("There was an error removing one or more machines.")
 	}
 }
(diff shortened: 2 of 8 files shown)


drivers/virtualbox/virtualbox.go
```diff
@@ -381,12 +379,10 @@ func (d *Driver) Start() error {
 		if err := vbm("startvm", d.MachineName, "--type", "headless"); err != nil {
 			return err
 		}
-		log.Infof("Starting VM...")
```
nathanleclaire: IMO these types of messages (ones which might overlap / duplicate what the core is saying) should stay out of the main output, maybe we could put them in `Debug` if we _really_ want to keep them.

libmachine/host.go
```diff
@@ -263,7 +272,9 @@ func (h *Host) Remove(force bool) error {
 	}
 
 	if err := h.SaveConfig(); err != nil {
-		return err
+		if !force {
```
nathanleclaire: As discussed on IRC @ehazlett

commands/rm.go
```diff
@@ -31,14 +31,15 @@ func cmdRm(c *cli.Context) {
...
-			log.Errorf("Error removing machine %s: %s", host, err)
+			log.Errorf("Error running action %q on machine %q: %s", "remove", host, err)
 			isError = true
 		} else {
 			log.Infof("Successfully removed %s", host)
 		}
 	}
 	if isError {
-		log.Fatal("There was an error removing a machine. To force remove it, pass the -f option. Warning: this might leave it running on the provider.")
+		log.Fatal("There was an error removing one or more machines.")
```
nathanleclaire: This change may or may not be a little controversial, but I'm mostly of the opinion we shouldn't really be tipping users off to / encouraging the use of `-f` -- and we should warn that it may still be running on the provider there, not here as a hypothetical.

ehazlett: +1

commands/commands.go
```diff
@@ -428,6 +441,7 @@ func runActionForeachMachine(actionName string, machines []*libmachine.Host) {
 			serialMachines = append(serialMachines, machine)
 		default:
 			numConcurrentActions++
+			log.Infof("Running action %q on machine %q...", actionName, machine.Name)
```
ehazlett: Should this be debug?  Seems a bit chatty.

ehazlett:
What do we do about providers that do not support stop?  I.e. RAX?

ehazlett:
Also, not sure what is happening, but a `stop` action on DO just hangs.

commands/commands.go
nathanleclaire: Mmm, personally I like to see the actions which are being taken, in what order they are running, and so on.  Without it, the drivers fend for themselves with messages like "Starting VM...", "Stopping VM...", some of them don't have them, it's inconsistent, etc.

nathanleclaire:
> What do we do about providers that do not support stop? I.e. RAX?

Argh, I didn't know that was a thing unfortunately.  A `SupportsStop` method on the driver seems kludgey, maybe we will have to selectively allow them based on the driver type i.e. a switch in the `host.Remove` method.

hairyhenderson:
> >What do we do about providers that do not support stop? I.e. RAX?

>Argh, I didn't know that was a thing unfortunately.

And it's likely to be more of a thing in the future, too - there was a PR for a Triton/SDC driver recently, and it doesn't support stop and a few other operations too...

commands/rm.go
hairyhenderson: :+1:

libmachine/host.go
```diff
@@ -256,6 +256,15 @@ func (h *Host) Upgrade() error {
 }
 
 func (h *Host) Remove(force bool) error {
+	machineState, err := h.Driver.GetState()
+	if err != nil {
+		return err
+	}
+
+	if machineState != state.Stopped {
```
hairyhenderson: What if the state is `Error`?

Also, would it maybe make sense to skip this check if `force` is set? If I do a `docker rm -f mycontainer`, it kills then removes the container, so it follows that doing a `docker-machine rm -f mymachine` should stop then remove the machine.

commands/commands.go
nathanleclaire: Maybe we can just spit out the name, like `docker stop`, etc. do

libmachine/host.go
nathanleclaire: I agree with skipping the check on force, but making machines automatically remove-able on `Error` state seems risky to me.

hairyhenderson: Well, I'm not sure what you mean by "automatically", but if a machine is in a permanent `Error`-state (maybe because it was removed from the provider, or if the account on the provider is gone, or whatnot), there needs to be a way to remove the host from docker-machine... Well, other than deleting it from `~/.docker/machine/machines/`, that is :P

nathanleclaire:
Re-visiting this, I think we need to at least prompt before destroying instances.  Our current method is extremely dangerous.

dmp42:
@nathanleclaire ok if we close it for bookkeeping for now?

nathanleclaire:
Yeah let's close, I at least want to have a prompt (`You are going to destroy these machines.  Proceed? [y|n]`) in the future though.

PR #1263, opened 2015-05-28
Remove flat (0.1.0) => nested (0.2.0) migration code

nathanleclaire:
This code was added in to ease the migration from 0.1.0 machines to 0.2.0.

If the user has been using 0.2.0 steadily for their machines, it is unlikely that they will run into issues if this code is removed in 0.3.0.

It will need some thorough review but I would really prefer that we be aggressive about taking things like this out, lest we end up with a bunch of cargo culted code that we're afraid to take out further down the line.

In general I want to guarantee that there won't be breaking changes in machine configuration from one release to the next (i.e. we will always safely migrate), _but_ that there will be no such guarantees if a version is skipped.

cc @ehazlett @hairyhenderson 


ehazlett:
I agree in wanting to keep the codebase clean.  I will check it out.  Thanks!

nathanleclaire:
:+1:

hairyhenderson:
:+1:

It may also be prudent to put a note about forward compatibility guarantees in the docs, especially while we're in a 0.x state.

nathanleclaire:
@ehazlett Does it LGTY?

ehazlett:
LGTM

PR #1264, opened 2015-05-28
Update Go version to 1.4.2

nathanleclaire:
There's not really any great hurry on this AFAIK, but it might be nice to have soon, or after the release if we want to play it super super safe.

This style of build optimizes for bandwidth at the cost of some additional time on first compile / environment build.  The `golang:1.4.2-cross` image is quite large, and I have a feeling that bandwidth is a more limited resource than CPU cycles for most.  It looks like first build with this method takes quite a while comparatively on a fast connection, though (~1 minute for pull vs. ~4 for bootstrap on my Ethernet at the office), so we should consider the tradeoff carefully and do some additional benchmarking.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
-FROM golang:1.3-cross
-RUN apt-get update && apt-get install -y --no-install-recommends openssh-client
+FROM golang:1.4.2-cross
 
 # TODO: Vendor these `go get` commands using Godep.
 RUN go get github.com/mitchellh/gox
 RUN go get github.com/aktau/github-release
 RUN go get github.com/tools/godep
-RUN go get code.google.com/p/go.tools/cmd/cover
+RUN go get golang.org/x/tools/cmd/cover
 
 ENV GOPATH /go/src/github.com/docker/machine/Godeps/_workspace:/go
 ENV MACHINE_BINARY /go/src/github.com/docker/machine/docker-machine


nathanleclaire:
cc @ehazlett @hairyhenderson @sthulb

hairyhenderson:
I'm +1 on the 1.4.2 bump.

The pull-time optimization is nice, and I think it's faster overall for me (I recall `golang:1.3-cross` taking 25mins to pull on my home internet connection a while back). I'm timing the build from this branch right now...

hairyhenderson:
Results:

```
script/build  1.40s user 0.46s system 0% cpu 11:43.16 total
```

This is on my MacBook Pro from 2011 with 16GB RAM, with boot2docker running in VMWare Fusion.

To be honest, I was a little surprised how long it took. If the first-build time is going to be on this order of magnitude, I think I _might_ actually prefer just going `FROM golang:1.4.2-cross`.

The other thing to consider is that this is a somewhat temporary problem, because Go 1.5 will (in theory) land in August, and the `-cross` image for that will be able to be much smaller (or so I've been told). So, it'd probably be better to use `-cross` when we move to 1.5, so maybe it isn't such a big deal to have the extra bandwidth cost with 1.4.2.

One big caveat though - I've actually gotten much more used to just running `godep go build` when I'm working on Machine lately, mostly because it's _so_ much faster. But, I know I should be `script/build`ing :grin:

ehazlett:
Yeah +1 from me too.  I think we can get this in after the 0.3 release and see how it works with the understanding of moving to 1.5 in Aug.  Sound good?

nathanleclaire:
Yikes at those results @hairyhenderson.  That's what I'd worry about with this method.  We might actually be better off just using `golang:1.4.2-cross`.  The other option would be to pre-bake our "base" on our own, which I think would cut down the image size pretty well.

> Yeah +1 from me too. I think we can get this in after the 0.3 release and see how it works with the understanding of moving to 1.5 in Aug. Sound good?

Yes!

sthulb:
+1, we should do it after the release :)

sthulb:
1.5 should be interesting in the summer, I'm looking forward to it, despite the runtime slowdown

nathanleclaire:
@ehazlett WDYT about this?  I changed to just use `golang:1.4.2-cross` image, which I think is ultimately the better path in spite of the large size, although I don't think we should kick out the idea of just generating our own base (it would cut the size by >50% off `golang:1.4.2-cross`).

nathanleclaire:
@ehazlett Does it LGTY?

ehazlett:
LGTM

PR #1268, opened 2015-05-29
Check that instance has started before polling SSH

janeczku:
After creating a new instance, we should check that this instance has been started before invoking the `WaitForSSH` method. We should not attempt to TCP connect to a host that is in state stopped/pending.
Typically the console is also spammed with dozens of identical debug messages when that happens:
`Error waiting for TCP waiting for SSH`
...

See also discussion in #1255.


ehazlett:
Looks like this needs a `gofmt`

janeczku:
oops :)

nathanleclaire:
LGTM

PR #1273, opened 2015-05-29
refactor: Extracting a base Driver struct

hairyhenderson:
This is mainly a de-duplication PR. I've extracted all of the common struct fields and put them into a new `DefaultDriver` struct.

There's been some mention of needing this before, for example in #1220 (and others).

I've tested this manually, but not extensively. There are a few other tweaks I want to make too, but I want to get feedback early to see if I'm on the right trail.

### DIFF
--- a/commands/inspect_test.go
+++ b/commands/inspect_test.go
@@ -29,7 +29,21 @@ func TestCmdInspectFormat(t *testing.T) {
 	assert.Equal(t, "\"none\"", actual)
 
 	actual, _ = runInspectCommand(t, []string{"--format", "{{prettyjson .Driver}}", "test-a"})
-	assert.Equal(t, "{\n    \"IPAddress\": \"\",\n    \"URL\": \"unix:///var/run/docker.sock\"\n}", actual)
+	type ExpectedDriver struct {
+		CaCertPath     string
+		IPAddress      string
+		MachineName    string
+		PrivateKeyPath string
+		SSHPort        int
+		SSHUser        string
+		SwarmDiscovery string
+		SwarmHost      string
+		SwarmMaster    bool
+		URL            string
+	}
+	expected, err := json.MarshalIndent(&ExpectedDriver{MachineName: "test-a", URL: "unix:///var/run/docker.sock"}, "", "    ")
+	assert.NoError(t, err)
+	assert.Equal(t, string(expected), actual)
 }
 
 func runInspectCommand(t *testing.T, args []string) (string, *libmachine.Host) {
--- a/docs/DRIVER_SPEC.md
+++ b/docs/DRIVER_SPEC.md
@@ -38,7 +38,7 @@ level maintenance.
 
 ## Maintainer
 To be supported as an official driver, it will need to be maintained.  There
-can be multiple driver maintainers and they will be identified in the 
+can be multiple driver maintainers and they will be identified in the
 maintainers file.
 
 # Provider Operations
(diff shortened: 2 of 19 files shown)


hairyhenderson:
(also, I'm breaking tests here - will follow up soon with a fix for that)

nathanleclaire:
Ah, very nice, thanks -- looking like a good start -- I'll review shortly.

drivers/default.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+// GetSSHKeyPath -
+func (d *DefaultDriver) GetSSHKeyPath() string {
+	return d.GetStorePath("id_rsa")
+}
+
+// GetStorePath -
+func (d *DefaultDriver) GetStorePath(path string) string {
```
hairyhenderson: I don't like the name of this function, but I couldn't think of anything better at the time... Now I'm thinking something like `ResolveStorePath`... any thoughts?

commands/inspect_test.go
```diff
@@ -29,7 +29,20 @@ func TestCmdInspectFormat(t *testing.T) {
...
+		MachineName    string
+		PrivateKeyPath string
+		SSHPort        int
+		SSHUser        string
+		SwarmDiscovery string
+		SwarmHost      string
+		SwarmMaster    bool
+		URL            string
+	}
+	expected, _ = json.MarshalIndent(&ExpectedDriver{MachineName: "test-a", URL: "unix:///var/run/docker.sock"}, "", "    ")
```
hairyhenderson: This is what I get for changing the none driver...

hairyhenderson:
Ok, I think I'm done messing around with this. I'm happy with it now.

sthulb:
I like this :) We should have done it from the start!

nathanleclaire:
When you get a chance, can you please squash it to one commit @hairyhenderson ?

hairyhenderson:
@nathanleclaire - ok, rebased and squashed. PTAL :)

commands/inspect_test.go
nathanleclaire: Why swallow the error?

nathanleclaire:
I'm getting errors attempting a create on DigitalOcean.

```console
Creating SSH key...
Creating Digital Ocean droplet...
Created droplet ID 5527853, IP address 45.55.150.61
Getting to WaitForSSH function...
Error getting IP address waiting for SSH: GetIP not implemented for unknown
Getting to WaitForSSH function...
Error getting IP address waiting for SSH: GetIP not implemented for unknown
```

It seems that it uses the default's `GetIP` function, and not the `GetIP` function from the `digitalocean.Driver`.

hairyhenderson:
@nathanleclaire - err... weird. I'm pretty sure I tested with the DO driver. Maybe I screwed something up along the way. I'll dig into it.

hairyhenderson:
@nathanleclaire - Ok, I can reproduce the issue. I'm not totally sure why it's happening with the DO driver, and not with the virtualbox driver...

hairyhenderson:
Ok, this is happening because of my poor understanding of how polymorphism works in Go. The default implementation of `GetSSHHostname()` is:

```go
func (d *DefaultDriver) GetSSHHostname() (string, error) {
	return d.GetIP()
}
```

My _assumption_ was that the receiver here would magically change to the outer type, except of course it doesn't in Go. So I'm going to have to re-think this part.

commands/inspect_test.go
hairyhenderson: Yeah ok, handled now.

hairyhenderson:
@nathanleclaire - ok, should be OK now. PTAL.

nathanleclaire:
This is looking good...  I want to review and get in probably right after the release

clnperez:
From how I understand this, the "Default Driver" doesn't work stand-alone, right? It seems like naming it "Default" kind of implies that it will work that way, and maybe do some looking around on your system and pick whatever environment you've got available (VirtualBox, qemu, etc.). So I'm thinking it could be named BaseDriver instead to reduce confusion (although users won't see the name of it at all, but, for developers new to the code).

hairyhenderson:
@clnperez - right, `BaseDriver` is more descriptive. I'll rename it :)

hairyhenderson:
@nathanleclaire @ehazlett - anything left to do before this can be merged?

clnperez:
I was assuming you were going to rename the struct, etc. to BaseDriver instead of DefaultDriver. Maybe I misunderstood and you were just renaming the PR?

hairyhenderson:
@clnperez - d'oh - I thought I did that but I guess I didn't... Sorry! Doing that now :)

hairyhenderson:
Ok, renamed! :)

clnperez:
Woot! Thanks @hairyhenderson

nathanleclaire:
Intend to review this quite soon.

nathanleclaire:
Nice work @hairyhenderson ... Testing looks pretty good on my end

LGTM

ehazlett:
Yay!  LGTM

hairyhenderson:
Thanks! 😁

PR #1276, opened 2015-05-31
Amazon EC2 - Fixing how the Monitoring.Enabled field is set for spot requests

hairyhenderson:
Fixing a bug I ran into earlier where I get this error when using the `--amazonec2-request-spot-instance` option:

```
Error creating machine: Error request spot instance: Problem with AWS API call: Non-200 API response: code=400 message=The parameter Monitoring is not recognized
```


ehazlett:
LGTM

PR #1282, opened 2015-06-01
enable overrides for engine and swarm install

ehazlett:
This adds the ability to override the Docker engine install URL and the Swarm image used during provisioning.  This makes it easy to test new versions.

For example:

`MACHINE_SWARM_IMAGE=swarm:0.3.0-rc1 docker-machine create ...`

or Engine:

`MACHINE_DOCKER_INSTALL_URL=https://test.docker.com docker-machine create ...`

### DIFF
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -188,10 +188,6 @@ func (provisioner *Boot2DockerProvisioner) Provision(swarmOptions swarm.SwarmOpt
 		return err
 	}
 
-	if err := installDockerGeneric(provisioner); err != nil {
-		return err
-	}
-
 	ip, err := provisioner.GetDriver().GetIP()
 	if err != nil {
 		return err
--- a/libmachine/swarm/swarm_options.go
+++ b/libmachine/swarm/swarm_options.go
@@ -1,1 +1,0 @@
-package swarm
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -166,6 +166,12 @@ var sharedCreateFlags = []cli.Flag{
 		),
 		Value: "none",
 	},
+	cli.StringFlag{
+		Name:   "engine-install-url",
+		Usage:  "Custom URL to use for engine installation",
+		Value:  "https://get.docker.com",
+		EnvVar: "MACHINE_DOCKER_INSTALL_URL",
+	},
 	cli.StringSliceFlag{
 		Name:  "engine-opt",
 		Usage: "Specify arbitrary flags to include with the created engine in the form flag=value",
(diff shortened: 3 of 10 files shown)


hairyhenderson:
LGTM!

nathanleclaire:
I am very +1 on this concept, however I think we should implement it "properly" with flags / environment variables and `EngineOptions` / `SwarmOptions` if we're going to move it upstream.  I am concerned that these kinds of ad-hoc toggles could easily grow out of hand.  Could we please update the PR to do so @ehazlett ?

hairyhenderson:
You know, @nathanleclaire's totally right. I hadn't even thought of it ;)

Some flags like `--swarm-image` and `--engine-docker-install-url` would probably do the trick.

These wouldn't _just_ useful for testing, they'd also be useful for users who need custom swarm images for some reason, or a custom docker install script. There have been some issues asking for the latter recently, too.

ehazlett:
@nathanleclaire works for me.

ehazlett:
I was originally headed in that direction but the custom URL didn't seem like a proper "EngineOption"
so I went this way.

nathanleclaire:
> I was originally headed in that direction but the custom URL didn't seem like a proper "EngineOption"
so I went this way.

I agree it's a slightly odd fit given the nature of most of the other struct fields, but is OK with me.

ehazlett:
@nathanleclaire updated PTAL

libmachine/provision/boot2docker.go
```diff
@@ -188,10 +188,6 @@ func (provisioner *Boot2DockerProvisioner) Provision(swarmOptions swarm.SwarmOpt
 		return err
 	}
 
-	if err := installDockerGeneric(provisioner); err != nil {
```
ehazlett: I wasn't sure why this was even there.  B2D has Docker already and this seems like an unnecessary call to me.

libmachine/swarm/swarm_options.go
```diff
@@ -1 +0,0 @@
-package swarm
```
ehazlett: I wasn't sure why this existed.

nathanleclaire: Just flotsam I think (this was what `swarm.go` used to be called)

libmachine/provision/boot2docker.go
nathanleclaire: Fair enough - way have just been knee jerk.  Good looking out.

nathanleclaire:
LGTM

libmachine/swarm/swarm_options.go
ehazlett: cool thx!

icecrime:
Great feature thanks! So I'm glad that it takes a URL (which means we'll get https://experimental.docker.com for free), and yet I'm slightly concerned that it's *so* permissive: I can basically give it any URL (https or not, docker related or not) to pipe into `/bin/sh`.

Maybe it's a non-issue, I'm not sure.

ehazlett:
Yeah I thought about this but was trying to prepare for URLs other than "get" and "test" (i didn't know what "experimental" would end up as).  It does mean that anyone could use any script but as you mentioned, they have ssh too so you could `docker-machine ssh foo -- bad things here` as well.

PR #1292, opened 2015-06-03
virtualbox: more flexibility with networking

### LINKED ISSUE #24: Make VirtualBox network configurable
At the moment the subnet used by VirtualBox is hard coded. It would be neat if this was configurable, like in boot2docker-cli.

### LINKED ISSUE #1290: VirtualBox: re-create hostonly interface on Machine start
If you are troubleshooting vbox machines, it can be useful to remove existing host only interfaces and have them re-created.  The only way Machine creates the interface is on `create`.  It would be beneficial to have them checked and re-created on `start` as well.

h/t to @fxdgear for the trouble working through this ;)

ehazlett:
This checks for the hostonly interface upon machine start and creates if needed.

This also adds the ability to specify the hostonly CIDR.

To customize the host only adapter, you can use the `--virtualbox-hostonly-cidr`
flag.  This will specify the host IP and Machine will calculate the VirtualBox
DHCP server address (a random IP on the subnet between `.1` and `.25`) so 
it does not clash with the specified host IP.
Machine will also specify the DHCP lower bound to `.100` and the upper bound
to `.254`.  For example, a specified CIDR of `192.168.24.1/24` would have a
DHCP server between `192.168.24.2-25`, a lower bound of `192.168.24.100` and 
upper bound of `192.168.24.254`.

Closes #1290
Closes #24

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -3,6 +3,7 @@ package virtualbox
 import (
 	"archive/tar"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -27,6 +28,10 @@ import (
 
 const (
 	isoFilename = "boot2docker.iso"
+)
+
+var (
+	ErrUnableToGenerateRandomIP = errors.New("unable to generate random IP")
 )
 
 type Driver struct {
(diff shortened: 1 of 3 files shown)


drivers/virtualbox/virtualbox.go
```diff
@@ -572,6 +573,44 @@ func (d *Driver) generateDiskImage(size int) error {
 	return createDiskImage(d.diskPath(), size, raw)
 }
 
+func (d *Driver) setupHostOnlyNetwork(machineName string) error {
+	ip, network, err := net.ParseCIDR(d.HostOnlyCIDR)
+	nAddr := network.IP.To4()
+
+	var dhcpAddr net.IP
+	// select pseudo-random DHCP addr; make sure not to clash with the host
+	for {
```
nathanleclaire: Could we maybe move this code into a function, say, `getAvailableDHCPAddr`,  which returns `dhcpAddr`?

drivers/virtualbox/virtualbox.go
```diff
@@ -572,6 +573,44 @@ func (d *Driver) generateDiskImage(size int) error {
...
+			log.Debugf("using %s for dhcp address", dhcpAddr)
+			break
+		}
+	}
+
+	hostOnlyNetwork, err := getOrCreateHostOnlyNetwork(
+		ip,
+		network.Mask,
+		dhcpAddr,
+		net.IPv4(nAddr[0], nAddr[1], nAddr[2], byte(100)),
```
nathanleclaire: Hm, just for clarity's sake, maybe we could assign these last two arguments to variables named `lowerBound` and `upperBound` and then pass those into this function?  It'd make it a bit easier to read, I think.

nathanleclaire:
@ehazlett Couple of code review nits, mostly looks pretty good, want to update those and we'll merge this in?

drivers/virtualbox/virtualbox.go
ehazlett: +1

ehazlett:
@nathanleclaire updated PTAL

nathanleclaire:
Cool.

LGTM

PR #1296, opened 2015-06-03
libmachine: check state before returning GetURL() result

vincentbernat:
The new integration tests check if `env` and `url` return an error when
the machine is stopped. Currently, only virtualbox driver checks
that. Most drivers can return an URL without a problem while the machine
is not running.

We delegate the task of checking if the machine is running to libmachine
instead. libmachine will check the state and if the machine is not
running will return an empty string and
`drivers.ErrHostIsNotRunning`. This is consistent with what is expected
in other parts of the code but may break some drivers.

Tested with the exoscale driver.

### DIFF
--- a/libmachine/host/host.go
+++ b/libmachine/host/host.go
@@ -146,7 +146,14 @@ func (h *Host) Upgrade() error {
 }
 
 func (h *Host) GetURL() (string, error) {
-	return h.Driver.GetURL()
+	s, err := h.Driver.GetState()
+	if err != nil {
+		return "", err
+	}
+	if s == state.Running {
+		return h.Driver.GetURL()
+	}
+	return "", nil
 }
 
 func (h *Host) ConfigureAuth() error {


vincentbernat:
It seems that the behaviour of `GetURL()` returning an error when the host is not running is not well established in the code base. The modification could be CLI-only or an empty string could be returned without `drivers.ErrHostIsNotRunning` error. Thoughts?

ehazlett:
This is failing Travis.

vincentbernat:
Yes, hence my previous comment. Integration tests enforce that `GetURL()` must return an error (and most drivers don't) but most of the code seems to think otherwise. A middle ground would be to return `"", nil`, but I would prefer the interface to be clarified. Maybe only the CLI must return an error.

ehazlett:
@vincentbernat ok sorry I misunderstood.

vincentbernat:
If he didn't notice that, maybe @nathanleclaire could tell who should handle the error when machine is stopped. Currently, only virtualbox driver handles that directly. I don't think that each driver should have to do that. So, either the CLI or libmachine?

nathanleclaire:
Good looking out on this @vincentbernat - I just got bit by it yesterday.  I will need to reflect on this a bit and decide what I feel the best approach is.

nathanleclaire:
Hey @ehazlett, do you have any strongly held opinions on this matter?  I think it seems to make sense to always return nothing for the IP / URL of an instance when it is powered off, since it is not addressable unless turned on.  I _do_ know that several of the integration tests are failing with master today since they assume there should be no IP address when the instance is turned off, but frequently just return the "cached" copy, e.g. https://github.com/docker/machine/blob/master/drivers/digitalocean/digitalocean.go#L262

(Side note: `GetIP` probably should always make a call to get the actual IP address of the instance, lest it changes.  It doesn't in a lot of drivers today IIRC)

ehazlett:
Yeah I'm not strong either way.  I think returning nothing for the IP if it is powered off makes sense.  I also agree that we should be calling to get the actual IP as well.

vincentbernat:
❦  9 juin 2015 09:27 -0700, Nathan LeClaire <notifications@github.com> :

> (Side note: GetIP probably should always make a call to get the actual
> IP address of the instance, lest it changes. It doesn't in a lot of
> drivers today IIRC)

For some drivers, it is guaranteed that the IP won't change.
-- 
Tell the truth or trump--but get the trick.
		-- Mark Twain, "Pudd'nhead Wilson's Calendar"

vincentbernat:
@nathanleclaire Any thought on this item?

nathanleclaire:
Ah, sorry to take so long to review, I've been on holiday.

I definitely want this type of functionality, however I think it should probably work at the `IP` level rather than the `URL` level (since `URL`, as well as other functions, calls `IP`).  Do you agree @vincentbernat @ehazlett ?

vincentbernat:
I think that we need to do both `GetIP()` and `GetURL()` in this case. We cannot just do `GetIP()`, because when calling `GetURL()`, we call driver's `GetURL()` which is likely to call `GetIP()` from the driver directly. If the goal is to minimize the code put in each driver, we don't want each driver to implement the error check.

dmp42:
LGTM though it needs a rebase.

@nathanleclaire it seems like this will help with quality, and has been pending there for a while.

wdyt?

nathanleclaire:
I'm OK to merge this if @vincentbernat will rebase.  Returning `""` for non-running machines makes sense to me.  Sorry for the delay Vincent.

vincentbernat:
PR has been updated (and tested with exoscale driver) to return `""` but no error. Unfortunately, this is not the behaviour expected for the CLI (through the integration tests). On the other hand, if I return `""` and an error, this is not expected by some other tests. You can compare d312e6f and c72867f. You'll have either the integration tests failing, either the unittests (well, the ones from Travis) failing.

The main problem is that drivers are not consistent. Some drivers like VirtualBox will fail with ErrHostIsNotRunning (like expected by the CLI integration tests) while the fakedriver used by other tests will not. So, do we raise an error on `GetURL()` on a not running host (so we need to fix some tests) or we don't (but we need to fix the integration tests or the CLI stuff to raise the error itself).

`GetURL()` could also be move to `base.go` to avoid duplication.

nathanleclaire:
Oof.  Thanks for following up @vincentbernat .  We'll have to discuss a bit about what the best approach is and make everything uniform.

nathanleclaire:
Modifying the milestone to 0.5.3 -- definitely want to implement a solution to this ASAP though.

dgageot:
@vincentbernat It's been a long time and the codebase has changed quite a bit. Do you think you could update this PR?

vincentbernat:
@dgageot To the best of my knowledge, there is still no decision on the consistent behavior that is wanted. Currently, each driver is expected to implement the check itself. If we try to not do that, there is an inconsistency through the code. Only the CLI expects an error. Other parts expect an empty string.

However, I didn't recheck the code. So, this may be outdated. Unfortunately, I am quite short in time currently.

nathanleclaire:
@dgageot Do you like Vincent's solution here?  I would be in favor of it and don't think it would cause problems (although the drivers likely should be updated as well).  Maybe we could carry his PR.

nathanleclaire:
BTW, thanks for the persistence and continuing to respond on this one @vincentbernat !

vincentbernat:
Note that my PR was incomplete (it failed Travis) due to the inconsistency that I pointed out after commiting the PR for the first time. Since it needed modifications outside the driver scope, I didn't carry on to propose something more complete.

This is a bit fuzzy now.

dgageot:
@vincentbernat Some driver did `drivers.ErrHostIsNotRunning` if the VM was stopped. Some didn't. Implementing the contract in `host.GetUrl` was a smart choice to make sure all drivers are aligned. However half the driver are able to do the check in a more efficient way in their `GetUrl` implementation. Because performance in important in `ls`, I prefer to fix the driver one by one.
This is done in this PR https://github.com/docker/machine/pull/3067

dgageot:
Fixed with https://github.com/docker/machine/pull/3067

nathanleclaire:
Thanks @dgageot

PR #1299, opened 2015-06-03
Allowing generic driver to use ssh-agent to get identities

### LINKED ISSUE #1291: Generic driver should use identities offered by ssh-agent when not specified
When using the `generic` driver, you're forced to pass in an SSH key. IMO that should be optional, and in the absence of an explicit identity, the SSH client (external at least, if not native) should be able to look at identities registered to my ssh-agent.

This would allow the use of passphrase-protected SSH keys, perhaps avoiding problems such as https://github.com/docker/machine/issues/1229#issuecomment-106790757

I _think_ this is simply a case of making the flag optional in the generic driver and modifying the external SSH client support to skip the `ssh -i` flag if `--generic-ssh-key` wasn't set.

Thoughts? @ehazlett @nathanleclaire @sthulb

hairyhenderson:
Fixes #1291

### DIFF
--- a/docs/drivers/generic.md
+++ b/docs/drivers/generic.md
@@ -33,9 +33,20 @@ Options:
 
 Environment variables and default values:
 
-| CLI option                 | Environment variable | Default             |
-| -------------------------- | -------------------- | ------------------- |
-| **`--generic-ip-address`** | `GENERIC_IP_ADDRESS` | -                   |
-| `--generic-ssh-key`        | `GENERIC_SSH_KEY`    | `$HOME/.ssh/id_rsa` |
-| `--generic-ssh-user`       | `GENERIC_SSH_USER`   | `root`              |
-| `--generic-ssh-port`       | `GENERIC_SSH_PORT`   | `22`                |
+| CLI option                 | Environment variable | Default                   |
+|----------------------------|----------------------|---------------------------|
+| **`--generic-ip-address`** | `GENERIC_IP_ADDRESS` | -                         |
+| `--generic-ssh-key`        | `GENERIC_SSH_KEY`    | _(defers to `ssh-agent`)_ |
+| `--generic-ssh-user`       | `GENERIC_SSH_USER`   | `root`                    |
+| `--generic-ssh-port`       | `GENERIC_SSH_PORT`   | `22`                      |
+
+##### Interaction with SSH Agents
+
+When an SSH identity is not provided (with the `--generic-ssh-key` flag),
+the SSH agent (if running) will be consulted. This makes it possible to
+easily use password-protected SSH keys.
+
+Note that this usage is _only_ supported if you're using the external SSH client,
+which is the default behaviour when the `ssh` binary is available. If you're
+using the native client (with `--native-ssh`), using the SSH agent is not yet
+supported.
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"path/filepath"
 	"strconv"
 	"time"
 
@@ -25,10 +24,6 @@ const (
 	defaultTimeout = 1 * time.Second
 )
 
-var (
-	defaultSourceSSHKey = filepath.Join(mcnutils.GetHomeDir(), ".ssh", "id_rsa")
-)
-
 // GetCreateFlags registers the flags this driver adds to
 // "docker hosts create"
 func (d *Driver) GetCreateFlags() []mcnflag.Flag {
@@ -46,8 +41,8 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 		},
 		mcnflag.StringFlag{
 			Name:   "generic-ssh-key",
-			Usage:  "SSH private key path",
-			Value:  defaultSourceSSHKey,
+			Usage:  "SSH private key path (if not provided, identities in ssh-agent will be used)",
+			Value:  "",
 			EnvVar: "GENERIC_SSH_KEY",
 		},
 		mcnflag.IntFlag{
@@ -66,7 +61,6 @@ func NewDriver(hostName, storePath string) drivers.Driver {
 			MachineName: hostName,
 			StorePath:   storePath,
 		},
-		SSHKey: defaultSourceSSHKey,
 	}
 }
 
@@ -83,6 +77,17 @@ func (d *Driver) GetSSHUsername() string {
 	return d.SSHUser
 }
 
+func (d *Driver) GetSSHKeyPath() string {
+	if d.SSHKey == "" {
+		return ""
+	}
+
+	if d.SSHKeyPath == "" {
+		d.SSHKeyPath = d.ResolveStorePath("id_rsa")
+	}
+	return d.SSHKeyPath
+}
+
 func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.IPAddress = flags.String("generic-ip-address")
 	d.SSHUser = flags.String("generic-ssh-user")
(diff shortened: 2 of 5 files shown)


ssh/client.go
```diff
@@ -48,7 +48,6 @@ const (
 var (
 	baseSSHArgs = []string{
 		"-o", "PasswordAuthentication=no",
-		"-o", "IdentitiesOnly=yes",
```
ehazlett: Does this have any effect on the other drivers / providers?  /cc @nathanleclaire @sthulb

ssh/client.go
```diff
@@ -261,9 +260,17 @@ func NewExternalClient(sshBinaryPath, user, host string, port int, auth *Auth) (
...
 	args := baseSSHArgs
 
+	// If no identities are explicitly provided, also look at the identities
+	// offered by ssh-agent
+	if len(auth.Keys) > 0 {
+		args = append(args, "-o", "IdentitiesOnly=yes")
+	}
+
 	// Specify which private keys to use to authorize the SSH request.
 	for _, privateKeyPath := range auth.Keys {
```
ehazlett: This would work in most cases, but most operating systems (debian, ubuntu, etc) will fail after a few attempts.  I don't think this will affect us as I believe we only add the single key we generate but just to note.

nathanleclaire:
So this is just for the `external` client right?

How about docs?

hairyhenderson:
> So this is just for the `external` client right?

Yep, that's the idea for now. Since `native` is really only used on Windows where there isn't generally an active `ssh-agent`, I'm not sure it's a big deal. However, it might be a good idea to do that in the future.

> How about docs?

Yeah - I'll whip some up. thanks for the reminder ;)

nathanleclaire:
> Yep, that's the idea for now. Since native is really only used on Windows where there isn't generally an active ssh-agent, I'm not sure it's a big deal. However, it might be a good idea to do that in the future.

Mm, yeah, I'd prefer to at least note it in the documentation.

ssh/client.go
hairyhenderson: Not sure what you mean here... are you talking about the case when `auth.Keys` is `0`?

I figure the normal use case will be for users to just provide the identity on the command line, which will be the same as before...

hairyhenderson:
Ok, docs added. I think this is ready now.

hairyhenderson:
@nathanleclaire @ehazlett - PTAL?

dmp42:
@hairyhenderson this needs a rebase, and probably better to wait for #2033 to merge (btw, if you want to review :)).

nathanleclaire:
I'm still not convinced that this is the right approach.  @hairyhenderson Do you think that having a Docker Machine SSH configuration file might work equally as well?  That way you could add a stanza such as:

```
Host *
    ForwardAgent yes
```

if desired.

hairyhenderson:
@nathanleclaire - you mean adding an SSH configuration inside the machine's store, and then calling `ssh` with `-F`? Not sure I'm fully following...

The main goal of this PR is really to make `--generic-ssh-key` optional. This is important, since forcing users to specify a path to a key only works if the user a) _has_ the key in a file and b) has a non-password-protected key.

Without the ability to fall back to using the agent, it's hard (or impossible) to support users who SSH (with `ForwardAgent yes`) into an intermediate system to run `docker-machine` on, and it's also hard (and super-awkward) to support users who have password-protected keys.

I'm not really sure what other approach could support those use-cases...

hairyhenderson:
_Squashed and rebased._

dgageot:
Hi @hairyhenderson Can you rebase? (I know, once again...)

hairyhenderson:
@dgageot - rebased :)

dgageot:
Thanks @hairyhenderson I'll take a look today

dgageot:
Hi @hairyhenderson I've had troubles testing this PR. Because the generic driver is not always behaving well, I don't know if the problem comes from the PR.
I'll check again tomorrow with a fresh brain.

dgageot:
Hi @hairyhenderson I like the PR! I annotated a couple of things to be fixed.

Also, I get a lot of warning at creation time and one each time I ssh to the machine:
```
Warning: Identity file /Users/dgageot/.docker/machine/machines/foo/id_rsa not accessible: No such file or directory.
```
Any idea?

vskubriev:
Hi @dgageot, well feature. I want to test it. But how to install this version before next release?

I should fork a base master branch, then merge you feature and what's next ?

Thanks for fresh fix )

dgageot:
@vskubriev To test it, you should work on your fork and run those commands:

```
git checkout -b hairyhenderson-ssh-agent-support-1291 master
git pull https://github.com/hairyhenderson/machine.git ssh-agent-support-1291
```

Warning: the code has issues that need to be fixed by @hairyhenderson right now.

hairyhenderson:
@dgageot - thanks for the review. I'll look into the `PreCreateCheck` thing soon... I haven't tested this in depth recently so I want to dig back into this a bit more :)

vskubriev:
@dgageot How to build on ubuntu 14.04 and standart golang package installed ?

```
make && make install
```
Does not work. Where I can find manual to build docker machine ? Or may be it is so hardly and I will wait for new version bump ?

dgageot:
@vskubriev PTAL at https://github.com/docker/machine/blob/master/CONTRIBUTING.md

vskubriev:
@dgageot Ok, I was succesfully build and install fixed version. But it seems not work. 

```bash
docker-machine create -d generic --generic-ip-address 192.168.1.1 --generic-ssh-user admin server
Running pre-create checks...
Creating machine...
(gate) Importing SSH key...
Machine is running, waiting for SSH to be available...
Enter passphrase for key '/home/username/.docker/machine/machines/server/id_rsa': 
```
Identity already un encrypted with ssh-add and I can login to server without any interaction.

It should work, but doesn't work.

vskubriev:
Excusme, I made a mistake while installing new version ((( Excusme for spam too. To proper install I used followoing command:

```bash
$ git clone git@github.com:docker/machine.git
$ cd machine
$ git checkout -b hairyhenderson-ssh-agent-support-1291 master
$ git pull https://github.com/hairyhenderson/machine.git ssh-agent-support-1291
$ export USE_CONTAINER=true
$ sudo make install
$ docker-machine -v
docker-machine version 0.5.3-dev ( 95c7893 )
```

And get the same error as @hairyhenderson:

```bash
username@ws:~/docker/somefolder$ docker-machine create -d generic --generic-ip-address 192.168.1.1 --generic-ssh-user admin server
Running pre-create checks...
Error creating machine: Error with pre-create check: Ssh key does not exist: ""
```

hairyhenderson:
Ok, I've finally had some time to sit down and really dig back into this :)

I've fixed the issues that crept in over the past months and this feature seems to work as expected again...

@dgageot @vskubriev - care to try again? Just as a warning, I force-pushed to rebase against `master`, so you'll need to hard-reset first :)

dgageot:
Thank you very much @hairyhenderson I'm going to test the PR.
Can you squash the commits? Also the documentation needs to be linted like here https://gist.github.com/dgageot/6351236f162893527709

dgageot:
Otherwise, the feature is great. LGTM
@docker/machine-maintainers wdyt?

vskubriev:
@hairyhenderson I cannot setup proper revision on my local machine without a branch. I not good git ) Will wait for a separate branch.

hairyhenderson:
@dgageot - squashed and re-rebased :)

hairyhenderson:
@vskubriev - you could clone it in a brand new location to try it out:

```console
$ git clone git@github.com:hairyhenderson/machine.git
...
$ cd machine
$ git checkout ssh-agent-support-1291
$ USE_CONTAINER=true sudo make install
...
```

dgageot:
LGTM
It would be nice to have unit tests for the code that was changed, also `driver.GetSSHClientFromDriver()` and `host.CreateSSHClient()` look awfully similar, but this can come in another PR.

vskubriev:
@hairyhenderson Thank you for a build tip. I couldn't build per one step) There is my command to success build and install:

```
$ USE_CONTAINER=true make build; sudo make install
$ docker-machine -v
docker-machine version 0.5.3-dev, build cf6136f
```
All done, all works as expected now.

hairyhenderson:
> It would be nice to have unit tests ...

@dgageot - yeah, I totally agree. I can add some in a follow-up PR at some point...

> also driver.GetSSHClientFromDriver() and host.CreateSSHClient() look awfully similar

:laughing: I noticed that - one of them could probably be replaced with the other at some point :smile:

dgageot:
ping @nathanleclaire PTAL

jeanlaurent:
LGTM

nathanleclaire:
LGTM

nathanleclaire:
Thanks for your persistence on this one @hairyhenderson :clap: :clap: :clap:

hairyhenderson:
:tada: @nathanleclaire - thanks! Glad this finally got in :smile:

PR #1300, opened 2015-06-03
Warning user that the active machine was restarted and may have a new IP

### LINKED ISSUE #1298: Restart command doesn't re-export DOCKER_HOST, breaks Swarm
`docker-machine version 0.3.0-dev (fec8b4a)`

Testing with GCE. Restarting a machine in the cloud changes its IP address. `docker-machine restart xxx` command doesn't update the DOCKER_HOST env variable, causing swarm connections to break. Note in the session below how `docker-machine env` correctly lists the updated IP, while DOCKER_HOST is still pointing to the (now) non-existent IP from before the restart.

(I have a firewall rule to allow swarm-master traffic to a machine tagged `docker-swarm-master`, used to illustrate it's not a firewall issue).

```
09:39:40::docker-964 ♨ >  docker-machine ls
NAME           ACTIVE   DRIVER       STATE     URL                         SWARM
gce1                    google       Running   tcp://104.154.84.165:2376   gce1 (master)
gce7                    google       Running   tcp://130.211.126.35:2376   gce1
09:40:07::docker-964 ♨ >  docker-machine restart gce1
Deleting instance.
Waiting for instance to delete.
Creating instance.
Waiting for Instance...
Uploading SSH Key
Waiting for SSH Key
09:41:37::docker-964 ♨ >  docker-machine ls
NAME           ACTIVE   DRIVER       STATE     URL                         SWARM
gce1                    google       Running   tcp://104.154.82.233:2376   gce1 (master)
gce7                    google       Running   tcp://130.211.126.35:2376   gce1
09:42:55::docker-964 ♨ >  gcloud compute instances add-tags gce1 --tags docker-swarm-master --zone us-central1-f
Updated [https://www.googleapis.com/compute/v1/projects/docker-964/zones/us-central1-f/instances/gce1].
09:43:35::docker-964 ♨ >  docker info
FATA[0032] An error occurred trying to connect: Get https://104.154.84.165:3376/v1.18/info: dial tcp 104.154.84.165:3376: i/o timeout

###
### NOTE THE INCORRECT IP, IT NO LONGER EXISTS
###

09:44:12::docker-964 ♨ >  echo $DOCKER_HOST
tcp://104.154.84.165:3376

09:44:38::docker-964 ♨ >  docker-machine env --swarm gce1
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://104.154.82.233:3376"
export DOCKER_CERT_PATH="xxx"
export DOCKER_MACHINE_NAME="gce1"
# eval "$(docker-machine env gce1)"
09:45:58::docker-964 ♨ >  docker-machine -v
docker-machine version 0.3.0-dev (fec8b4a)

###
### And now let's prove it works if configured properly
###

09:48:45::docker-964 ♨ >  eval $(docker-machine env --swarm gce1)
09:54:22::docker-964 ♨ >  docker info
Containers: 1
Strategy: spread
Filters: affinity, health, constraint, port, dependency
Nodes: 1
 gce7: 130.211.126.35:2376
  └ Containers: 1
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 0 B / 7.671 GiB
09:54:24::docker-964 ♨ >
```

P.S.: looks like there's a potential issue here with the number of machines too (it lists 1 node, gce7, but gce1 is omitted) - will keep an eye on it and file an issue once I understand what's going on.

hairyhenderson:
Fixes #1298 

### DIFF
--- a/commands/restart.go
+++ b/commands/restart.go
@@ -1,6 +1,10 @@
 package commands
 
 import (
+	"os"
+	"strings"
+
+	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/log"
 
 	"github.com/codegangsta/cli"
@@ -10,4 +14,58 @@ func cmdRestart(c *cli.Context) {
 	if err := runActionWithContext("restart", c); err != nil {
 		log.Fatal(err)
 	}
+
+	warnIfActiveIPChanged(c)
+}
+
+func warnIfActiveIPChanged(c *cli.Context) {
+	if activeInList(c.Args()) {
+		activeURL := os.Getenv("DOCKER_HOST")
+		log.Debugf("checking to see if active host's URL has changed from %s", activeURL)
+
+		h := getActiveHost(c)
+		u, err := h.GetURL()
+		if err != nil {
+			log.Fatal(err)
+		}
+		if activeURL != u {
+			// TODO: hardcoded port here is a giant kludge...
+			var swarmOpt string
+			if strings.Contains(activeURL, ":3376") {
+				swarmOpt = "--swarm "
+			}
+			log.Warnf("Active machine was restarted, and has a new IP address.\nRun 'eval \"$(docker-machine env %s%s)\"' again.", swarmOpt, h.Name)
+		}
+	}
+}
+
+func activeInList(machines []string) bool {
+	active := os.Getenv("DOCKER_MACHINE_NAME")
+	return active != "" && inArray(active, machines)
+}
+
+func getActiveHost(c *cli.Context) *libmachine.Host {
+	certInfo := getCertPathInfo(c)
+	store, err := getDefaultStore(
+		c.GlobalString("storage-path"),
+		certInfo.CaCertPath,
+		certInfo.CaKeyPath,
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+	active, err := store.GetActive()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return active
+}
+
+func inArray(s string, a []string) bool {
+	for _, cur := range a {
+		if cur == s {
+			return true
+		}
+	}
+	return false
 }
--- a/commands/restart.go
+++ b/commands/restart.go
@@ -1,10 +1,6 @@
 package commands
 
 import (
-	"os"
-	"strings"
-
-	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/log"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Cool idea.

Why not store the value of `DOCKER_HOST` before any action is taken (you can also use `IsActive` to check if the machine you are restarting is active IIRC), and then check to see if it has been invalidated after the action is taken?  I feel like if it is printed every time users will grow immune to it and it will defeat the purpose of having it.

hairyhenderson:
@nathanleclaire - Good idea. I don't see any `IsActive` function... were you thinking of `Machine.GetActive`?

I've mocked something up but there's some work to do before it'll work properly.

nathanleclaire:
Yep, `GetActive` is what I was thinking of.  Then you'd actually check the URL against what is existing in `DOCKER_HOST`.

commands/restart.go
```diff
@@ -1,13 +1,53 @@
...
+	}
+
 	if err := runActionWithContext("restart", c); err != nil {
 		log.Fatal(err)
 	}
+
+	if activeInList(c.Args()) && activeUrl != getMachineUrl(c) {
+		// TODO: hardcoded port here is a giant kludge...
+		var swarmOpt string
+		if strings.Contains(activeUrl, ":3736") {
```
ehazlett: This should be `3376`.  More and more reason we need to remove the hardcode. :(  I think that just might be a goal of 0.4 :)

ehazlett:
+1

commands/restart.go
hairyhenderson: lol - indeed ;)

hairyhenderson:
Ok, I've fixed this all up so it's actually _working_ now, and squashed/rebased.

nathanleclaire:
So revisiting this issue, perhaps we could simply not get too clever and suggest to the user on `start` and `restart` a note that the IP may have changed and they may need to re-eval ?

hairyhenderson:
@nathanleclaire - I'm inclined to agree... that'd certainly have less moving parts to break ;P Plus, I just realized that my change didn't have any support for `start`, so that'd be missed too.

I'll de-complicate this soon...

hairyhenderson:
@nathanleclaire PTAL

nathanleclaire:
Nice.  Nitpick, but the `Infof` is superflous- you only need `log.Info`.

hairyhenderson:
Heh. Was thinking "I need to change this to `Info`" as I was writing that... Looks like I forgot :P

Will fix when I get back to a computer :)

hairyhenderson:
@nathanleclaire - ok, fixed :)

nathanleclaire:
Nice LGTM thanks!

Issue #1308, opened 2015-06-05
Add shell completion for commands and flags


booyaa:
I noticed that codegangsta's cli package is being used to parse command line flags and arguments. This package has shell completion (bash/zsh) and can be enabled with a single line.

I think this would be a handy feature to enable given the number of options you can have to create, configure and manage a machine. 

[docker](https://github.com/docker/docker/tree/master/contrib/completion) and [compose](https://github.com/docker/compose/tree/master/contrib/completion) have completion scripts.

jeanlaurent:
AFAIK this is done. Feel free to re-open if needed.

PR #1309, opened 2015-06-05
Add shell completion

booyaa:
This is my attempt to implement #1308.

Out of the box codegangsta's cli package supports command completion, but not flags (global or command specific) there's a pending enhancement request codegangsta/cli#188. So I've decided to implement this and maybe give back to the cli package too!

I've been playing around with the [config](https://github.com/booyaa/machine/blob/add_shell_completion/commands/commands.go#L245) and [create](https://github.com/booyaa/machine/blob/add_shell_completion/commands/commands.go#L262) commands.

So far I've been unable to access the Flags of commands. I can display the global flags (in app.Flags), but this also has the undesired effect of display aliases and (I think) descriptions of the flags.

```shell
[booyaa@hodor machine GO  (add_shell_completion) ] $ ./docker-machine_darwin-amd64 config<tab>
(Go-based)                     -s,                            auth                           path
--debug,                       -v                             cert                           print
--generate-bash-completion     /Users/booyaa/.docker/machine  certificates                   remotes
--help,                        CA                             client                         show
--native-ssh                   Client                         debug                          storage
--storage-path                 Configures                     for                            the
--tls-ca-cert                  Enable                         generate                       to
--tls-ca-key                   Private                        help                           use
--tls-client-cert              SSH                            implementation.                used
--tls-client-key               TLS                            in                             verify
--version,                     Use                            key                            version
-D                             []                             mode
-h                             against                        native
```
I've added the stock shell completion script in the [script](https://github.com/booyaa/machine/blob/add_shell_completion/script/machine) folder, but whilst writing the issue #1308 I discovered the completion scripts in docker and compose live in a contrib folder, so I'll move it there.

Usage of the completion script is as follows:

```
PROG=docker-machine_os_arch source=script/machine
```
The script can be tailored to predefine PROG as `docker-machine`

Finally any suggestions on any part of this PR (in particular testing) are appreciated. I'm still fairly new to Go, but always willing to learn.


ehazlett:
Thanks for the contribution!  The output looks a little odd.  For example, there is `[]` and `Use`, `for`, `generate`, `implementation`, etc.  Those are not valid commands.  I'm also not sure about having the flags in -- it makes the output a little confusing but that's just me.

booyaa:
Yeah definitely think the [] and other bits are artifacts from the flags array. I'll scale it back to commands initially and move the script to a contrib directory.

Testing should be really easy then, just need to compare the output from the --enable-bash-completion with the commands array.

hairyhenderson:
hey @booyaa - you'll need to sign your commits and re-push for the build to pass

booyaa:
heya @hairyhenderson i'm probably going to rollback the changes to commands/commands.go to the commit before mine. what's the prefer way to sync my changes? close and raise new pr, or just push updates back to master?

nathanleclaire:
We've merged in https://github.com/docker/machine/pull/1993 so I'm going to close this.

PR #1310, opened 2015-06-05
exoscale: ability to specify multiple security groups on instance creation

vincentbernat:
Without this feature, when a user needs a specific security group (i.e
most of the time if they want to expose a service to Internet), they
need to copy the rules needed for Docker itself into each custom
group. With this feature, they can spawn new instances with
`--exoscale-security-group docker-machine,my-custom-group`.

Since the `Driver` struct has been modified (`SecurityGroup string` removed, `SecurityGroups []string` added), machine dies horribly if any exoscale machine has been spawned in the past. How should this be handled? We just ignore it for the moment?

### DIFF
--- a/commands/commands_test.go
+++ b/commands/commands_test.go
@@ -112,6 +112,10 @@ func (d DriverOptionsMock) String(key string) string {
 	return d.Data[key].(string)
 }
 
+func (d DriverOptionsMock) StringSlice(key string) []string {
+	return d.Data[key].([]string)
+}
+
 func (d DriverOptionsMock) Int(key string) int {
 	return d.Data[key].(int)
 }
--- a/docs/index.md
+++ b/docs/index.md
@@ -1174,7 +1174,7 @@ Options:
  - `--exoscale-instance-profile`: Instance profile.
  - `--exoscale-disk-size`: Disk size for the host in GB.
  - `--exoscale-image`: exoscale disk size. (10, 50, 100, 200, 400)
- - `--exoscale-security-group`: Comma-separated list of security groups. Non-existent groups will be created.
+ - `--exoscale-security-group`: Security group. Non-existent groups will be created. Can be repeated.
  - `--exoscale-availability-zone`: exoscale availability zone.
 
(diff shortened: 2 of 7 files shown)


vincentbernat:
Also, I could use a `StringSlice` instead of a comma separated list. What would you prefer?

ehazlett:
@vincentbernat I personally prefer the StringSlice.  I think it's OK to break now as we haven't had a release with the exoscale driver.  We would want to do this before the release.

vincentbernat:
I have added another commit for that.

Currently, this is the only driver to use `StringSlice`. Two other candidates are Openstack own security groups and GC scopes. Both of them are using a comma-separated list. I can convert them (but not test them).

Also, AFAIK, it is not possible to use environment variables to specify multiple security groups. This could be worked around by doing also comma-splitting.

I am going to fix the unittests in a minute.

vincentbernat:
@ehazlett Note that if you are delaying this PR for post-0.3.0, we will have to deal with the migration issue. If you are just too busy, nevermind. :)

ehazlett:
@vincentbernat no -- i didn't see the last commit.  I was waiting on your updates.  I will take a look.

ehazlett:
LGTM

/cc @nathanleclaire

vincentbernat:
Now that 0.3.0 has been published, how could the migration be handled?

nathanleclaire:
@vincentbernat Which is the migration you refer to?

vincentbernat:
@nathanleclaire Yes. Since I am changing the type of the field, any user with existing machine will get an error. If there is nothing yet to handle this kind of migration, I could just switch back to `String` + `split()`.

vincentbernat:
Trying to move forward on this. I have appended another commit which goal is to keep the `SecurityGroup` member in `Driver` struct. The array from `StringSlice` is joined with commas into this variable and when we want to use it, we split again on comma. A bit ugly but will allow seamless upgrade from 0.3.0.

vincentbernat:
@nathanleclaire Any insight?

nathanleclaire:
LGTM

nathanleclaire:
Thanks for your persistence @vincentbernat !

PR #1322, opened 2015-06-07
Lookup VMware fusion location using mdfind

### LINKED ISSUE #685: Cannot create vmwarefusion machine if Fusion installed outside of /Applications
I use [Homebrew Cask](http://caskroom.io) to install many binary applications like VMware Fusion. It places the .app bundles in `/opt/homebrew-cask` and symlinks them into `~/Applications`. As such, the `vmrun` command referenced by machine is not at the expected path.

The issue is that this command fails silently, and machine continues to wait for the VM to spin up, providing no useful feedback.

Ideally, machine could find `vmrun` on its own, but simply acknowledging/reporting the reason for the failure would be a good first step.

docker-machine version 0.1.0
Mac OS X 10.10.2

Debug output below

```
$ docker-machine -D create  -d vmwarefusion fusion
DEBU[0000] executing: /usr/bin/ssh-keygen ssh-keygen -t rsa -N  -f /Users/username/.docker/machine/machines/fusion/id_rsa
 
Your identification has been saved in /Users/username/.docker/machine/machines/fusion/id_rsa.
Your public key has been saved in /Users/username/.docker/machine/machines/fusion/id_rsa.pub.
The key fingerprint is:
xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx username@localhost
The key's randomart image is:
+--[ RSA 2048]----+
|                 |
|                 |
|                 |
|                 |
|                 |
|                 |
|                 |
|                 |
|                 |
+-----------------+
INFO[0000] Creating VM...                               
DEBU[0000] executing: /Applications/VMware Fusion.app/Contents/Library/vmrun start /Users/username/.docker/machine/machines/fusion/fusion.vmx nogui 
INFO[0000] Waiting for VM to come online...             
DEBU[0000] Not there yet 1/60, error: couldn't find MAC address in VMX file /Users/username/.docker/machine/machines/fusion/fusion.vmx 
DEBU[0002] Not there yet 2/60, error: couldn't find MAC address in VMX file /Users/username/.docker/machine/machines/fusion/fusion.vmx 
...
DEBU[0118] Not there yet 60/60, error: couldn't find MAC address in VMX file /Users/username/.docker/machine/machines/fusion/fusion.vmx 
ERRO[0120] Error creating machine: Machine didn't return an IP after 120 seconds, aborting 
WARN[0120] You will want to check the provider to make sure the machine and associated resources were properly removed. 
FATA[0120] Error creating machine                       
```

dleehr:
Fixes #685

### DIFF
--- a/drivers/vmwarefusion/vmrun_darwin.go
+++ b/drivers/vmwarefusion/vmrun_darwin.go
@@ -16,8 +16,10 @@ import (
 )
 
 var (
-	vmrunbin    = "/Applications/VMware Fusion.app/Contents/Library/vmrun"
-	vdiskmanbin = "/Applications/VMware Fusion.app/Contents/Library/vmware-vdiskmanager"
+	out, err = exec.Command("mdfind", "kMDItemContentType == \"com.apple.application-bundle\" && kMDItemFSName = \"VMware Fusion.app\"").Output()
+	fusionpath  = strings.TrimSpace(string(out))
+	vmrunbin    = fusionpath + "/Contents/Library/vmrun"
+	vdiskmanbin = fusionpath + "/Contents/Library/vmware-vdiskmanager"
 )
 
 var (


zchee:
+1

ehazlett:
-1 

We should use `os.LookPath` in Go instead.

nathanleclaire:
Yes, if `exec.LookPath` will work this use case (it will if the binary is in `PATH`, but we've had some issues with binaries outside of `PATH` with Fusion in the past IIRC), we should use it.

sthulb:
This has some merit. Not everyone puts VMware fusion in the default place or uses home brew to install. Although, for the users that don't install it in either place, I'm unsure if we should bother, instead we should advise them to add it to their path manually.

ehazlett:
@sthulb I understand that some would put things outside of `PATH` but IMO I think it gets clunky trying to look through the filesystem on each platform for the tools.  That's what `PATH` is for :)

I do think we could add a note if people are using tools outside of their `PATH` but we probably should do it globally for all other tools as well (ssh on windows, etc).

dmp42:
Closing since maintainers are not really fans on the approach.

PR #1333, opened 2015-06-08
update roadmap for 0.4.0

ehazlett:
The next release we want to focus on stability and improvements to the local providers.  These are the focus points:

* Finalize Boot2Docker Migration (OS)
* Provisioning Stability (error reporting, re-provisioning existing hosts)
* Local Provider Improvements (make VirtualBox, Fusion, Hyper-V extremely stable)

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -4,23 +4,15 @@ Machine currently works really well for development and test environments. The
 goal is to make it work better for provisioning and managing production
 environments.
 
-This is not a simple task -- production is inherently far more complex than
-development -- but there are three things which are big steps towards that goal:
-**client/server architecture**, **swarm integration** and **flexible
-provisioning**.
-
 (Note: this document is a high-level overview of where we are taking Machine.
 For what is coming in specific releases, see our [upcoming
 milestones](https://github.com/docker/machine/milestones).)
 
-### Docker Engine / Swarm Configuration
-Currently there are only a few things that can be configured in the Docker Engine and Swarm.  This will enable more operations such as Engine labels and Swarm strategies.
+### Boot2Docker OS Migration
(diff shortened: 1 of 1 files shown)


sthulb:
+1 I can get behind this :)

hairyhenderson:
Nice and focused - LGTM!

nathanleclaire:
:+1: 

Some specific callouts I'd like to see (might fill in more as I remember them):

- OSX/Windows installer (preferably unified with KM / Compose etc.), like what b2d has today
- Improved per-driver logging.  I'd like to see a structure where machine operations log their debug output to the relevant folder on disk, whether or not they are displayed to the terminal when they are happening, and can be queried later with something like a `docker-machine logs dev` command.

ehazlett:
@nathanleclaire +1

nathanleclaire:
Another:

- A command (or script) which will generate information about the host system and environment for debugging purposes, similar to the one that Docker has.

efrecon:
This focuses on stability and this is good. If you were to consider new features, I would suggest a way to specify labels to attach to created machines so that swarm can make use of these labels for orchestration.

nathanleclaire:
> This focuses on stability and this is good. If you were to consider new features, I would suggest a way to specify labels to attach to created machines so that swarm can make use of these labels for orchestration.

@efrecon I'm not sure if this is exactly what you're after, but in Machine master today you can use `--engine-label` to do this, e.g.:

```console
$ docker-machine create -d amazonec2 --engine-label disk=ssd --engine-label size=micro foobar
```

It will set the labels on the Docker daemon, you should be able to see them using `docker info`.

efrecon:
Thank you @nathanleclaire ! It is what I was looking for, I had missed that option, so disregard my comment...

Issue #1289, opened 2015-06-02
Env swarm parameter lost on the eval help message


jrottenberg:
The helper for the docker-machine env command is not consistent with the parameters passed :

````
$ docker-machine env --swarm swarm-master
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.110:3376"
export DOCKER_CERT_PATH="~/.docker/machine/machines/swarm-master"
export DOCKER_MACHINE_NAME="swarm-master"
# Run this command to configure your shell:
# eval "$(docker-machine env swarm-master)"
`````

Note : it doesn't say 

````
eval "$(docker-machine env --swarm swarm-master)"
````

So running that eval command will get me to the docker api of the swarm master, not exactly what I want.

See the port :
````
$ docker-machine env swarm-master|grep DOCKER_HOST
export DOCKER_HOST="tcp://192.168.99.110:2376"
$ docker-machine env --swarm swarm-master|grep DOCKER_HOST
export DOCKER_HOST="tcp://192.168.99.110:3376"
````

nathanleclaire:
Yeah this should probably be consistent for the release.  cc @ehazlett

UserTaken:
`Swarm` is working on dynamic master election (introduced in 0.3.0-rc1) among other features for fault tolerant cluster so how will that be reflected with `machine` if at all?

aperepel:
Just confirming this is an issue with the current dev release `docker-machine version 0.3.0-dev (fec8b4a)`.

The `env` command output suggestion must include `--swarm` option if one was provided to `docker-machine env`, and omit it otherwise. Right now it will suggest to copy/paste an incorrect `eval` command which will not configure client for swarm-master properly.

nathanleclaire:
> Swarm is working on dynamic master election among other features for fault tolerant cluster so how will that be reflected with machine?

Basically, we're going to have to decouple the swarm name from the swarm master -- there's bit too many assumptions flying right now.

aperepel:
> Basically, we're going to have to decouple the swarm name from the swarm master -- there's bit too many assumptions flying right now.

I understand. Do you know planned timeframes for when swarm ha/failover is expected? I figured this trivial issue could be fixed quickly while there's a big ticket item coming in the future.

ehazlett:
@aperepel the mechanics in Swarm are still being created / decided.  Currently it's an experimental feature and requires specific discovery systems.  Support for it won't come in this release.

[...]

PR #1339, opened 2015-06-08
#1289 Env swarm parameter lost on the eval help message

aperepel:
A clean version of PR1326 - I've since redid the branch from CLI, the tool messed up the origin tracking.

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -95,7 +95,13 @@ func cmdCreate(c *cli.Context) {
 		log.Fatal("You will want to check the provider to make sure the machine and associated resources were properly removed.")
 	}
 
-	info := fmt.Sprintf("%s env %s", c.App.Name, name)
+	swarmCmd := ""
+	if c.Bool("swarm") {
+		// leading space required
+		swarmCmd = " --swarm"
+	}
+
+	info := fmt.Sprintf("%s env %s%s", c.App.Name, name, swarmCmd)
 	log.Infof("To see how to connect Docker to this machine, run: %s", info)
 }
 
--- a/commands/env.go
+++ b/commands/env.go
@@ -48,7 +48,8 @@ func cmdEnv(c *cli.Context) {
 
 	t := template.New("envConfig")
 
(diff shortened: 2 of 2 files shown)


ehazlett:
LGTM

ehazlett:
Would you mind squashing these commits?  Thanks!

aperepel:
Done and done :+1:

ehazlett:
LGTM

aperepel:
Hi, giving this PR a bump, it's been over a month, and all checks are green - anything else one should do to get it in?

nathanleclaire:
I'm sorry to be so late to say this, but I'm pretty hesitant about this.  I'd prefer to keep the logic around this section as simple as possible since it's already kind of hairy and needs to work on several different platforms.  Do you feel like it's confusing for users who want to activate their swarms that they need to remember `--swarm` flag?

nathanleclaire:
Also note that with this PR, you have to use `--swarm` flag with env to get the hint in that text, which naturally implies that you already know about it.

I think that if the concern is around notifying users how to connect to a newly created swarm master (which is what `--swarm` option for env is used for), we might be better off simply adding a new comment line to a created master, which says something like:

```console
$ docker-machine env kitematic
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.101:2376"
export DOCKER_CERT_PATH="/Users/nathanleclaire/.docker/machine/machines/kitematic"
export DOCKER_MACHINE_NAME="kitematic"
# eval "$(docker-machine env kitematic)"
# If you want to talk to the Swarm master, remember to use the --swarm flag with env.
```

dgageot:
An updated PR can be found [here](https://github.com/docker/machine/pull/2027). It fixes the problem with `--swarm` and `--no-proxy` and fixes the hint printed for fish shell.

nathanleclaire:
I think this is fixed in master with a @dgageot PR.  Please re-open if that is not correct.

PR #1341, opened 2015-06-09
windows: error message if using powershell/cmd and none specified

### LINKED ISSUE #1327: windows: powershell not properly detected when using env
When using `env` in PowerShell, the environment setup instructions are incorrect:

```
PS C:\Users\ehazlett\bin> .\docker-machine.exe --native-ssh env dev
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.100:2376"
export DOCKER_CERT_PATH="C:\Users\ehazlett\.docker\machine\machines\dev"
export DOCKER_MACHINE_NAME="dev"
# eval "$(C:\Users\ehazlett\bin\docker-machine.exe env dev)"
```

This should prompt the user to specify the `--shell` argument.

ehazlett:
This will error if using Windows and powershell/cmd and `--shell` is not specified.  The `SHELL` env var is not available to processes in msysgit.  In msysgit, the `TERM` environment variable is set to "cygwin" so this PR checks for that.  If it exists, it continues with the normal detection process.  If not set, it errors with a message asking the user to specify which shell they would like to use.

Fixes #1327

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -652,14 +652,18 @@ func getCertPathInfo(c *cli.Context) libmachine.CertPathInfo {
 }
 
 func detectShell() (string, error) {
-	// attempt to get the SHELL env var
-	shell := filepath.Base(os.Getenv("SHELL"))
-	// none detected; check for windows env and not bash (i.e. msysgit, etc)
-	if runtime.GOOS == "windows" && shell == "" {
+	// check for windows env and not bash (i.e. msysgit, etc)
+	// the SHELL env var is not set for processes in msysgit; we check
+	// for TERM instead
+	if runtime.GOOS == "windows" && os.Getenv("TERM") != "cygwin" {
 		log.Printf("On Windows, please specify either 'cmd' or 'powershell' with the --shell flag.\n\n")
 		return "", ErrUnknownShell
 	}
 
+	// attempt to get the SHELL env var
+	shell := filepath.Base(os.Getenv("SHELL"))
+
+	log.Debugf("shell: %s", shell)
 	if shell == "" {
 		return "", ErrUnknownShell
 	}


ehazlett:
@nathanleclaire PTAL

nathanleclaire:
Where's the bit that checks for `TERM` on msysgit?  I only see checking for `TERM` on Cygwin.

ehazlett:
@nathanleclaire well it's both.  In msysgit, the `TERM` variable is `cygwin`.  Yeah. :)

nathanleclaire:
> In msysgit, the TERM variable is cygwin. Yeah. :)



This might seem a bit pedantic, but could you please update the comment to include that info?  It strikes me as quite odd.

Oliboy50:
Sorry for asking here, but I just wanted to know if this PR will let msysgit (not Cygwin) users set the environment variables.

Currently, instead of `eval "$(docker-machine env MACHINE_NAME)"`, I have to use this workaround (in a batch file):
```
docker-machine env MACHINE_NAME | sed 's/export/set/g' > temp.cmd
sed -i 's/\"//g' temp.cmd
sed -i 's/#/REM/g' temp.cmd
call temp.cmd
del temp.cmd
```
Because neither `export` nor `eval` commands are available with msysgit only.

Thanks for the great work anyway. 
Hope Windows users will enjoy docker-machine like I do.

ehazlett:
@nathanleclaire done

ehazlett:
@Oliboy50 if you use the RC that should already be fixed I believe.

nathanleclaire:
LGTM

Oliboy50:
@ehazlett yep, thanks, I just realize there was a `--shell` option for `docker-machine env`, but you still should replace `#` comments with `REM` to be more "windows compliant" ;)

ehazlett:
@Oliboy50 ah thanks -- i'll take a look thanks!

PR #1343, opened 2015-06-10
Closes #1329 when merged

### LINKED ISSUE #1329: Migrate from B2D
This will finalize the migration from boot2docker.  This is will migrate away from the b2d OS and to a similar more supported base OS.

moxiegirl:
Migration material per the issue; also added example as requested by Nathan.  Nathan, I don't think a whole new section is required.  That file is overloaded as is with sections --- I intend to break it out after this release.

Signed-off-by: Mary Anthony <mary@docker.com>


moxiegirl:
ping @ehazlett @nathanleclaire See what you think of this gents.

moxiegirl:
Just made a small tweak to the links. I'll rebase after comments.

experimental/README.md
```diff
@@ -0,0 +1,20 @@
+# Docker Machine Experimental Features 
+
+This page contains a list of features in the Docker Machine which are
+experimental. Experimental features are **not** ready for production. They are
```
ehazlett: We are currently saying Machine is not ready for production.  This seems to imply that Machine outside of these are ready for production.

ehazlett:
Just making a note: we need to add the `experimental` dir excluded in the coverage check: https://travis-ci.org/docker/machine/builds/66139286#L605

ehazlett:
@moxiegirl why is the "experimental" directory not a subdir of docs?

moxiegirl:
@ehazlett So, the thinking with experimental is to put it outside docs.  This does a couple of things:

- Keeps experimental features out of the "production ready feature" docs; hypertexting can land users anywhere on the page and we don't want them to land on docs for a feature and have them thinking the feature is ready
- Allows you to add to experimental features on the fly --- or pull them --- or add quick warnings.  If we had this material in docs we would have to do a formal publish cycle which can be costly

ehazlett:
@moxiegirl ok thanks.

experimental/README.md
```diff
@@ -0,0 +1,20 @@
...
+provided for test and evaluation in your sandbox environments.  
+
+The information below describes each feature and the Github pull requests and
+issues associated with it. If necessary, links are provided to additional
+documentation on an issue.  As an active Docker user and community member,
+please feel free to provide any feedback on these features you wish.
+
+## Current experimental features
+
+Currently, you can experiment with [migrating from Boot2Docker CLI to Docker Machine](b2d_migration.md). Also, consider reviewing our [rough plan for the migration](b2d_migration_tasks.md).
```
nathanleclaire: Should this be a list? i.e.: 

"Our current list of experimental features is:

- Migrating boot2docker-cli created VMs to Machine
- Support for Red Hat, Debian, CentOS, Fedora, and RancherOS as base OSes
- etc.
"

Any guidance on which new features should be considered experimental and why?  Generally I think of all new features as fairly experimental for one release, so it seems a bit odd to call out particular ones unless we have a specific area we'd worry about such as confusing users, or the implementation is really hacky but badly needed, etc.

nathanleclaire:
Not having a separate section seems OK, although I worry a bit the information will become buried.

moxiegirl:
So, the experimental features are there as a way to get feedback before releasing a feature as final.  The docker/docker team is doing a separate binary but they are a bigger project.    So, I think of BETA as something we are ready to ship but want tested.  Experimental is something we want to call out for UX --- it is a way of saying: "We think this feature is what you want, is it?"

The Experimental section will be called out in the Release Notes.   And, having each project handle it similarly will reinforce the utility of the section.

nathanleclaire:
@moxiegirl Also, as @ehazlett mentioned, you'll need to add `${DIR}/experimental` to the string here: https://github.com/docker/machine/blob/master/script/generate-coverage#L44

moxiegirl:
@nathanleclaire Understood. I'll update/rebase shortly. I know we want to close this quick.

nathanleclaire:
No rush, just letting you know :)

nathanleclaire:
> So, I think of BETA as something we are ready to ship but want tested. Experimental is something we want to call out for UX --- it is a way of saying: "We think this feature is what you want, is it?"

OK, that helps clarify a bit, thanks.

moxiegirl:
ping @ehazlett @nathanleclaire recheck please

nathanleclaire:
LGTM

PR #1348, opened 2015-06-11
Add check firewall message to WaitForDocker

clnperez:
If an image has a firewall in place, currently machine doesn't do
anything to open the docker port. It tries forever to connect to
the docker engine and repeats a message that it is still unable
to connect:

    "Daemon not responding yet: dial tcp 129.41.145.211:2376: connection timed out"

While machine comes up with a good strategy for handling the myriad
of provisioners' firewall packages, at least provide a hint about
what might be happening at this point in the machine create:

    "Daemon not responding yet: dial tcp 129.41.145.211:2376: connection timed out
    Could a firewall be blocking the docker port? Try 'machine ssh [machine]' to investigate if this message continues."

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -123,6 +123,7 @@ func WaitForDocker(ip string, daemonPort int) error {
 		conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", ip, daemonPort))
 		if err != nil {
 			log.Debugf("Daemon not responding yet: %s", err)
+			log.Debug("Could a firewall be blocking the docker port? Try 'machine ssh [machine]' to investigate if this message continues.")
 			return false
 		}
 		conn.Close()


clnperez:
@dave-tucker had suggested adding some tests and a new message to help with the issue I was trying to solve with PR #1156. I think that since there's a WaitForDocker function, we can be sure that provisioning up to this point has gone well, so no other checks need to be done -- and a small message is enough.

ehazlett:
Thanks!  I like the idea of this especially when you are just stuck waiting for it to respond :)  I'm wondering if we should show this message on `Info` after a certain number of tries.  If the user doe not launch machine with debug they will never see this message.  But at the same time, I don't want to show it right away -- only after a number of tries perhaps.

/cc @nathanleclaire thoughts?

nathanleclaire:
Yes, my preference would be to only show that once we've dialed 20 times or so and still gotten nowhere.  This is made a little more difficult by the implementation of `WaitFor`, though, it doesn't really account for such a thing.

clnperez:
@ehazlett, yep. I thought about that, but then went the easy route b/c of what @nathanleclaire stated. :)

clnperez:
Also note that this doesn't happen in quick succession like the wait for SSH function. This one prints every minute, it looks like (although the code looks like it would print every 3 seconds).

How about https://gist.github.com/clnperez/6011bc91579da843869d ?

clnperez:
@ehazlett, @nathanleclaire -- If we go with PR #1446 we can probably nix this message.

nathanleclaire:
As noted in https://github.com/docker/machine/pull/1446, let's move to a discussion of scoping this / feature proposal in the issues.  Thanks!

PR #1359, opened 2015-06-14
Documentation tooling for 1.7 release build using Hugo

moxiegirl:
Tweaking for tooling
Adding in the sed for markdown in Dockerfile
Should be machine not engine

Signed-off-by: Mary Anthony <mary@docker.com>

ping @nathanleclaire @ehazlett Can I get an LGTM on these please gents.  This is for the 1.7 release


moxiegirl:
### The "docs RC plan"

This is the first step to an "docs RC build" if you will.  

After I pull this PR and the other 6, I'm going to do a build to a staging bucket on S3.  That build will include the new theme which is dropping around noon Monday. I'll page through every page on the RC build:

- look for layout issues
- checking all the links
- making sure images appear

Each problem will be noted and then I'll do a set of follow on PRs to fix issues.  That's why I'm wanting to pull this set today, it allows for an RC cycle.  And then, also, if necessary one more before the release this week.

ehazlett:
I see content that is referencing other projects.  Is this expected?  For example, managing images, old docs for Docker 1.5 and 1.4, invalid command reference, etc.

2015-06-15-103209_313x801_scrot

moxiegirl:
Hi Evan,

Yes that is expected. We define a site-wide menu in our theme. When a user builds an individual project, they build only the content for that project but the server instance returns that content "in-place" of the larger organization.  This enables the user to see when they misplace or cause a break in the site menu which is possible from a down-stream project.

Future tooling we plan to look at ways to build projects standalone. This makes particular sense for projects like Kitematic or Hub where these projects might want product specific Help. With Hugo, it is really easy to single-source our docs.  That is have the doc sources live in one place but display in different output media in different locations. I have had conversations with teams interested in this type of tooling but so far no real demand.

Mary

ehazlett:
@moxiegirl cool thanks.  LGTM

PR #1363, opened 2015-06-15
vbox: fix vbox race with unregistervm

ehazlett:
There is a random race with VirtualBox where it will not release its internal lock in time before we issue the `unregistervm` command.  This adds a sleep as a workaround.

I've been hitting this more and more lately.  I think we need it in for 0.3.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -462,6 +462,8 @@ func (d *Driver) Remove() error {
 			return err
 		}
 	}
+	// vbox will not release it's lock immediately after the stop
+	time.Sleep(1 * time.Second)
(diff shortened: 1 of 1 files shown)


ehazlett:
@nathanleclaire PTAL

nathanleclaire:
It is not guaranteed that 1 second sleep will fix the problem completely... Do we have a gameplan on detecting it more elegantly in the future?

nathanleclaire:
Would `utils.WaitFor` a `Stopped` state work?

nathanleclaire:
Hm, we already check it in `Stop` method :(

This seems to fix the issue so, for now at least until we can find a more elegant solution, LGTM

PR #1370, opened 2015-06-16
Fix defaulting to native SSH

nathanleclaire:
cc @ehazlett 

I noticed in some of the KM Windows builds that when SSH is not around, it blows up instead of defaulting to the native SSH client.  That shouldn't happen, this fixes the issue.


ehazlett:
LGTM

PR #1371, opened 2015-06-16
Add more useful error message when possibly getting rate limited by Github

nathanleclaire:
cc @ehazlett 

More feedback from KM - at one point they got an error message back which seems to be related to getting rate limited by Github (complaining about demarshaling not working).  This makes the source of the issue more clear.


ehazlett:
LGTM

PR #1426, opened 2015-06-25
Add logo to docs

nathanleclaire:
cc @ehazlett I think it would look nice.


README.md
```diff
@@ -1,5 +1,7 @@
 # Docker Machine
 
+![](/docs/img/logo.png)
```
thaJeztah: Perhaps add an "alt-text" here, e.g. `![Docker Machine Logo](/docs/img/logo]`

thaJeztah:
I absolutely love the logos, so yeah! Couple of things that need to be checked/verified;

- Are the logos available in a vector format (SVG?), if so, those may be preferred (at the chance it's not supported on older browsers)
- Given that all the the docs files of the various projects are merged, I *think* the files should have unique names, so `logo.png` should be something like `machine-logo.png` (cc @moxiegirl to confirm I'm right here)
- Should we add credits for the logo's / artwork somewhere?

moxiegirl:
@thaJeztah I've heard different things about image formats.  We probably need to define a policy and follow it.  https://quickleft.com/blog/choosing-svg-jpeg-png/ might be a start.  For this, I wouldn't hold the PR up on the image format.

The project README.md file is not part of the docs.docker.com source set.  @thaJeztah  is correct, we are trying to standardize on  `images` as a directory.  In production, these appear in `project/images`.  Result, is you could leave it as `logo.png` as the risk for name collision is pretty low.  @thaJeztah I'm finally getting to the task of writing up docs workflow etc this week, so everyone has a better understanding.

Docker owns the rights to the images that Laurel creates for us.  She is credited as their creator all over.  So, I don't think we need to credit her specially here.

moxiegirl:
LGTM

thaJeztah:
LGTM

Thanks for the extensive explanation; I wanted to be sure we were doing the right thing here (e.g. credits are easily overlooked)

Thanks @nathanleclaire! (sorry, can't press merge for you as I don't have permissions here)

vieux:
:+1:

hairyhenderson:
Nice. :+1:

thaJeztah:
ping @moxiegirl @ehazlett for the merge :-)

ehazlett:
Yay :+1:

PR #1436, opened 2015-06-29
Add SUSE Enterprise Linux and openSUSE provision

flavio:
Add support for SUSE Enterprise Linux and openSUSE

Signed-off-by: Flavio Castelli <fcastelli@suse.com>

### DIFF
--- a/libmachine/provision/suse.go
+++ b/libmachine/provision/suse.go
@@ -1,0 +1,228 @@
+package provision
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+
+	"github.com/docker/machine/drivers"
+	"github.com/docker/machine/libmachine/auth"
+	"github.com/docker/machine/libmachine/engine"
+	"github.com/docker/machine/libmachine/provision/pkgaction"
+	"github.com/docker/machine/libmachine/swarm"
+	"github.com/docker/machine/log"
+	"github.com/docker/machine/utils"
+)
+
+func init() {
+	Register("openSUSE", &RegisteredProvisioner{
+		New: NewOpenSUSEProvisioner,
+	})
+	Register("SUSE Linux Enterprise Desktop", &RegisteredProvisioner{
+		New: NewSLEDProvisioner,
+	})
+	Register("SUSE Linux Enterprise Server", &RegisteredProvisioner{
+		New: NewSLESProvisioner,
+	})
+}
+
+func NewOpenSUSEProvisioner(d drivers.Driver) Provisioner {
+	return &SUSEProvisioner{
+		GenericProvisioner{
+			DockerOptionsDir:  "/etc/docker",
+			DaemonOptionsFile: "/etc/sysconfig/docker",
+			OsReleaseId:       "opensuse",
+			Packages: []string{
+				"curl",
+			},
+			Driver: d,
+		},
+	}
+}
+
+func NewSLEDProvisioner(d drivers.Driver) Provisioner {
+	return &SUSEProvisioner{
+		GenericProvisioner{
+			DockerOptionsDir:  "/etc/docker",
+			DaemonOptionsFile: "/etc/sysconfig/docker",
+			OsReleaseId:       "sled",
+			Packages: []string{
+				"curl",
+			},
+			Driver: d,
+		},
+	}
+}
+
+func NewSLESProvisioner(d drivers.Driver) Provisioner {
+	return &SUSEProvisioner{
+		GenericProvisioner{
+			DockerOptionsDir:  "/etc/docker",
+			DaemonOptionsFile: "/etc/sysconfig/docker",
+			OsReleaseId:       "sles",
+			Packages: []string{
+				"curl",
+			},
+			Driver: d,
+		},
+	}
+}
+
+type SUSEProvisioner struct {
+	GenericProvisioner
+}
+
+func (provisioner *SUSEProvisioner) Service(name string, action pkgaction.ServiceAction) error {
+	reloadDaemon := false
+	switch action {
+	case pkgaction.Start, pkgaction.Restart:
+		reloadDaemon = true
+	}
+
+	// systemd needs reloaded when config changes on disk; we cannot
+	// be sure exactly when it changes from the provisioner so
+	// we call a reload on every restart to be safe
+	if reloadDaemon {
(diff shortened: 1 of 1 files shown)


libmachine/provision/suse.go
```diff
@@ -0,0 +1,211 @@
...
+		},
+	}
+}
+
+func NewSUSEProvisioner(d drivers.Driver) Provisioner {
+	return &SUSEProvisioner{
+		GenericProvisioner{
+			DockerOptionsDir:  "/etc/docker",
+			DaemonOptionsFile: "/etc/sysconfig/docker",
+			OsReleaseId:       "sled",
```
ablock84: Should't this be "sles"?

flavio: Fixed by adding support also for `sles`.

flavio:
Ping. I would have loved to see this code merged in machine 0.4 :(

rtrauntvein:
+1 for SUSE support :)

ablock84:
+1 from me as well

But there is one problem with the current approach how docker is installed on SLES/SLED (probably also applies to openSUSE).

The provisioner uses the installer script provided by https://get.docker.com/, which does rely on the RPM repositories provided by SUSE. At least in the case of the enterprise variants, these packages are already too outdated. Currently this would install docker 1.6. Even if you try to specify the experimental installer URL to machine, the script still installs the RPM's provided by SUSE.

Not sure what the correct solution would be. Either not rely on the installer script inside machine and provide a native-machine solution, or change the installer script to use a repository provided by docker (as already done for other distros with https://yum.dockerproject.org/repo/). This would however mean that the docker team will have to maintain one more distribution.

I thought about opening an Issue for this, but I was not sure were it belongs...depending on the solution, it would be an Issue for machine or docker or whatever project manages the installer script. Any suggestions where to put this?

flavio:
@ablock84 we provide bleeding edge packages both for openSUSE and SLE inside of the [Virtualization:containers](https://build.opensuse.org/project/show/Virtualization:containers) project on OBS.

How would you, as an user, tell machine to install bleeding edge packages? If you explain me the workflow you would like to have I can provide the code.

ablock84:
@flavio I was not aware of the Virtualization:containers repo for SLE12 and will definitely try this out next time I need to update/install docker on one of our machines. Thanks for the pointer. Maybe this info should be added to the SUSE installation instructions found at https://docs.docker.com/installation/SUSE/. Currently only the openSUSE repos are listed there.

I would expect the same workflow as with other distros. As I understand machine, it reuses the installer script found at https://get.docker.com/. A user can also provide an alternative URL to an installer script with --url, for example "--url https://experimental.docker.com/". In case of other distros (e.g. debian/ubuntu), these scripts also setup the correct APT/Yum repositories which are maintained by the docker team before installing docker. This always results in the current release of docker (or the experimental one if the experimental url was provided).

For SLES/SLED, I would expect the same behavior. The scripts should add the Virtualization:containers repo and then install docker. As an alternative, the machine provisioner could setup this repo, but I would assume that the script solution would be the preferred one.

I'd already be very happy when this PR could get merged. As a workaround until the installer script is fixed/updated, I can setup the Virtualization:containers repo manually before I use machine to do the remaining work.

flavio:
@ablock84 thanks for the useful feedback, really appreciated!

I'll make a PR against the official experimental script to install docker from obs://Virtualization:containers.

Good point also about the documentation, I'll update it as well.

ehazlett:
I wonder if we could get our new Yum repos to have SUSE support. I think that would be the best option.

ablock84:
Just tested the SUSE provisioner on one of our machines, this time with the Virtualization:containers added manually before calling machine. It worked fine, until I used the new --engine-env option to setup the needed proxy configuration. Looks like the provisioner is currently ignoring the --engine-env option.

flavio:
@ehazlett  wrote:
> I wonder if we could get our new Yum repos to have SUSE support. I think that would be the best option. 

We (SUSE) are already maintaining packages for SLE and openSUSE. We publish them for both latest upstream version and for the "supported by SUSE" Docker versions. We do that for all the flavours of openSUSE (13.2, Factory, Tumbleweed and soon also 42) and SLE (12 and soon 12SP1).

If requested by the community we could even setup a [OBS](http://openbuildservice.org/) project to automatically build the docker daemon straight from the code inside of git HEAD.

I think extending the https://experimental.docker.com/ script to pull packages from the "bleeding edge" repository maintained by SUSE, as suggested by @ablock84, is a nice solution.

I don't understand the advantages of having openSUSE/SUSE packages built by your infrastructure.

flavio:
> Just tested the SUSE provisioner on one of our machines, this time with the Virtualization:containers added manually before calling machine. It worked fine, until I used the new --engine-env option to setup the needed proxy configuration. Looks like the provisioner is currently ignoring the --engine-env option.

I'll look into that. Thanks

gimoh:
:+1: 
just tested with openSUSE 13.2, worked great

flavio:
I've updated the installation script provided at get.docker.com and experimental.docker.com with [this](https://github.com/docker/docker/pull/16221) pull request.

Everything is working fine: it's possible to install the stable docker package both on openSUSE and SUSE Linux Enterprise by using the get.docker.com installation url. Otherwise it's possible to install the bleeding edge packages by using the experimental.docker.com installation url.

flavio:
Hello, the installation script has been merged into docker/master. Can you please take care of this PR now? :heart:

flavio:
Ping @nathanleclaire: I see machine 0.5.2 rc(s) are out. Would you mind to merge this pull request into it? The openSUSE and SUSE communities would be really glad :)

dmp42:
LGTM

dmp42:
@flavio I reverted the merge, since you were apparently not on the latest master and the build broke.
Can you rebase and PR again?

Thanks.

flavio:
I updated the code. @dmp42 sorry about the issue

nathanleclaire:
It looks like Github is confused about the situation, so @flavio perhaps you should open a brand new PR.

dmp42:
@nathanleclaire that's #2035

nathanleclaire:
OK, let's close this then

nathanleclaire:
Oh, I see.  Please disregard

PR #1446, opened 2015-07-01
Add Firewall Handling to Provisioners

clnperez:
If, for example, a cloud image has a strict firewall in place,
machine will not complete provisioning because it will not be
able to communicate with the Docker port (2376).

This patch adds a few firewallaction package, which is modeled
after the packageaction package. It also includes the Unbuntu
provisioner logic.

The decision to use different logic for each provisioner was
made after realizing that manipulating iptables would have to
be done slightly differently for each provisioner (since they
store their rules in different files) -- so there would already
have to be some provisioner awareness. Additionally, using the
provisioners' firewall manipulation packages (ufw, firewall-cmd,
etc.) will provide the extra benefit of persisting rules across
reboots more easily.

### DIFF
--- a/libmachine/provision/ubuntu.go
+++ b/libmachine/provision/ubuntu.go
@@ -6,6 +6,7 @@ import (
 	"github.com/docker/machine/drivers"
 	"github.com/docker/machine/libmachine/auth"
 	"github.com/docker/machine/libmachine/engine"
+	"github.com/docker/machine/libmachine/provision/firewallaction"
 	"github.com/docker/machine/libmachine/provision/pkgaction"
 	"github.com/docker/machine/libmachine/swarm"
 	"github.com/docker/machine/log"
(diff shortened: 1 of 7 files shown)


clnperez:
This picks up where PR #1156 left off, but expands it a bit. @ehazlett, it turns out the pure iptables approach we discussed [I don't think] will work after all.

The CI build checks failed b/c I didn't add a test file yet -- but I can if we think this is a good way to go.

/cc @nathanleclaire

clnperez:
@ehazlett, @nathanleclaire Ping.

libmachine/provision/ubuntu.go
```diff
@@ -113,6 +135,16 @@ func (provisioner *UbuntuProvisioner) Provision(swarmOptions swarm.SwarmOptions,
 		}
 	}
 
+	// @TODO: Like everwhere else, update this when ports are configurable.
+	if err := provisioner.Firewall(firewallaction.Allow, 2376); err != nil {
+		log.Warnf("Error allowing the Docker daemon TCP port: %s", err)
+	}
+	if swarmOptions.IsSwarm {
+		if err := provisioner.Firewall(firewallaction.Allow, 2375); err != nil {
```
nathanleclaire: This should be `3376` I think, unless I'm misunderstanding.

nathanleclaire:
I'm pretty keen on this basic idea, but would be hesitant to move forward on it without supporting at least Debian and Red Hat as well.

@ibuildthecloud Do you have any comment or sentiments on this basic idea from a RancherOS perspective?

@davetucker @mrjana @mavenugo @nerdalert Do you have any opinions about how this sort of thing should be handled by Machine?  I think it's pretty relevant to your interests ;D

clnperez:
@nathanleclaire, thanks for the look! I've got the RH provisioner done (and tested) locally. With feedback from everyone you just pinged, I can add others as well.

nathanleclaire:
>  I've got the RH provisioner done (and tested) locally. 

Oh nice!

nathanleclaire:
One thing to take a look at is that `libnetwork` has an `iptables` module, I'm not sure if it's relevant here but I would really prefer to re-use code if we can.

clnperez:
@nathanleclaire I think I looked at `libnetwork`'s `iptables` module, but I'll look again. It has been a while since I started all this!

clnperez:
ah, @nathanleclaire, I didn't see a way to persist a rule there. I think that is fine for docker (IIUC), but we'd need this rule to always be there across reboots. Can someone from the libnetwork confirm? Maybe @mrjana?

nathanleclaire:
@clnperez Thanks for all the good work on this one, however the scope of Machine is exploding as-is and I think we need to re-group and/or define the API barriers / roles of the tool better before moving forward with things like this.  However, would love to see some discussion going on how to resolve the various networking / port forwarding / firewall issues.

Issue #1444, opened 2015-07-01
OpenStack driver should accept multiple --openstack-net-id switches


robertjustjones:
The OpenStack driver accepts only a single network option (id or name) which overrides the default connection(s).

robertjustjones:
The PR accepts a comma separated list of net-id's or net-name's for openstack and rackspace drivers.

nathanleclaire:
cc @ggiamarchi

fsoppelsa:
@nathanleclaire what is status of this issue?

nathanleclaire:
@fsoppelsa Not actively being worked on, but happy to take PRs.

Would be great to see some suggestions for developing against the OpenStack driver as well, I am not too familiar with it.

fsoppelsa:
@nathanleclaire seems the PR is abandoned.

I am in favor to add this to drivers/openstack, but it's not clear which one of the networks will be the one where docker-machine will communicate. From the code I see that a list of networks is loaded, but I don't understand which one will become default. I would like to hear from @robertjustjones

robertjustjones:
I no longer have access to the environment to test against.  My main environment was Rackspace then. The idea was that (outside docker-machine) a machine could be set up with network interfaces for Public/eth0, Rackspace/Private/eth1 and Custom/Private/eth2.  The new config switches implemented in the PR enable docker-machine to do the same. I don't know but suspect that the first interface would be the default for docker-machine. I hope that helps.

fsoppelsa:
Thanks @robertjustjones do you think you will have a chance to rebase and resubmit the PR?

robertjustjones:
@fsoppelsa I've rebased the code. Can you try it out from https://github.com/robertjustjones/machine.git from branch rackspace-net-switches ?  (It includes openstack)

fsoppelsa:
```
$ neutron net-list
| 44ead515-da4b-443b-85cc-a5d13e06ddc8 | mynet05 | 14466141-3a5e-4355-b75f-a3c957abe9a3 192.168.111.0/24 |
$ docker-machine create \
> --driver openstack \
> --openstack-image-id 98011e9a-fc46-45b6-ab2c-cf6c43263a22 \
> --openstack-flavor-id 3 \
> --openstack-floatingip-pool public \
> --openstack-net-id 44ead515-da4b-443b-85cc-a5d13e06ddc85 \
> --openstack-sec-groups machine \
> --openstack-ssh-user ubuntu \
> u1
[...]

Issue #1440, opened 2015-06-30
Rackspace driver doesn't list or recognize openstack switches like --openstack-net-name


robertjustjones:
The Rackspace driver PR #73 mentions support for adding networks by name and adds that functionality to the underlying openstack driver, but the rackspace driver doesn't report or recongnize such a switch
```
docker-machine create -d rackspace --openstack-net-name mynet ...
```
ignores the net directive. And there is no `--rackspace-net-name` option.

nathanleclaire:
cc @ggiamarchi

guoqiao:
+1 for this issue.

PR #1451, opened 2015-07-02
Rackspace/Openstack - Add/enhance --*-net-* create switches.

robertjustjones:
This PR addresses #1440 and #1444.

I'm having trouble with the version of gophercloud. When using `--rackspace-net-name`, I get an error with a url that has `.../v2.0/v2.0/...` which looks to be fixed in the latest gophercloud. When I try to run godep update, I get "not know version control system" on many other packages. Suggestions? @hairyhenderson suggested that it might be a conflict with docker/docker dependencies.

Other than that, `--rackspace-net-id`, `--openstack-net-id` and `openstack-net-name` all work with multiple comma separated networks.

### DIFF
--- a/drivers/rackspace/rackspace.go
+++ b/drivers/rackspace/rackspace.go
@@ -2,6 +2,7 @@ package rackspace
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/drivers"
@@ -60,6 +61,16 @@ func GetCreateFlags() []cli.Flag {
 			Usage:  "Rackspace flavor ID. Default: General Purpose 1GB",
 			Value:  "general1-1",
 			EnvVar: "OS_FLAVOR_ID",
+		},
+		cli.StringFlag{
+			Name:  "rackspace-net-id",
+			Usage: "Rackspace comma separated network ids the machine will be connected on",
+			Value: "",
+		},
+		cli.StringFlag{
+			Name:  "rackspace-net-name",
+			Usage: "Rackspace comma separated network names the machine will be connected on (eg, PublicNet,ServiceNet,MyNet)",
+			Value: "",
 		},
 		cli.StringFlag{
 			Name:  "rackspace-ssh-user",
@@ -121,6 +132,12 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.EndpointType = flags.String("rackspace-endpoint-type")
 	d.ImageId = flags.String("rackspace-image-id")
 	d.FlavorId = flags.String("rackspace-flavor-id")
+	if flags.String("rackspace-net-id") != "" {
+		d.NetworkIds = strings.Split(flags.String("rackspace-net-id"), ",")
+	}
+	if flags.String("rackspace-net-name") != "" {
+		d.NetworkNames = strings.Split(flags.String("rackspace-net-name"), ",")
+	}
 	d.SSHUser = flags.String("rackspace-ssh-user")
 	d.SSHPort = flags.Int("rackspace-ssh-port")
 	d.SwarmMaster = flags.Bool("swarm-master")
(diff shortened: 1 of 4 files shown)


robertjustjones:
Don't merge this yet, I'm working on the double v2 issue now. BTW, it had nothing to do with dependencies.

robertjustjones:
This PR is ready to go for review.

The issue above was that the rackspace driver was trying to use some gophercloud/openstack code and needed to be the gophercloud/rackspace -specific code.

As noted, I've updated the Godep rackspace/gophercloud dependency, but that was not actually required, just good measure.

hairyhenderson:
Thanks for this! Since the dependency update is tangential, could you please split it into a separate PR?

robertjustjones:
@hairyhenderson, this one is done, just the drivers/* changes. The dep update PR to follow.

drivers/rackspace/rackspace.go
```diff
@@ -62,6 +63,16 @@ func GetCreateFlags() []cli.Flag {
 			EnvVar: "OS_FLAVOR_ID",
 		},
 		cli.StringFlag{
+			Name:  "rackspace-net-id",
+			Usage: "Rackspace comma separated network ids the machine will be connected on",
```
hairyhenderson: Hrm... A `StringFlag` isn't really meant to be used for specifying multiple options - would be better to use a `StringSliceFlag`, then the flag can be repeated a number of times (it's more in line with how other options work in docker-machine):

```go
...
    cli.StringSliceFlag{
        Name:  "rackspace-net-id",
        Usage: "Rackspace network id the machine will be connected on. Can be specified multiple times.",
        Value: &cli.StringSlice{},
    },
...
```

drivers/rackspace/rackspace.go
```diff
@@ -62,6 +63,16 @@ func GetCreateFlags() []cli.Flag {
...
+			Value: "",
+		},
+		cli.StringFlag{
```
hairyhenderson: _(see above - use a `StringSliceFlag`)_

robertjustjones: Funny, I actually had that and changed it to match the implementation of `openstack-sec-groups`. Should we make that change to also include it?  Is there anyone who can test that switch if we do?

robertjustjones:
@hairyhenderson, it looks to me like the drivers use a different `drivers.DriverOptions` structure that include a `[]string` options. Turns out that none of the drivers have slice style args/flags.

Here's where it's used
https://github.com/robertjustjones/machine/commit/49ab6eb60441ba9778a38d1f4069a0d90eca2a83#diff-532c8189ade775e8b44f93d6a08402faL257

And here's where it traces back to
https://github.com/docker/machine/blob/master/drivers/drivers.go#L165-L169

Thoughts?

hairyhenderson:
@robertjustjones - ahh... you're absolutely right :) Sorry about that. Some more work would be required to allow `DriverOptions` to contain string slices, and I think that'd be a good idea eventually, but not a big deal right now.

Thanks :)

robertjustjones:
I've rebased to current in hopes that this can get merged.

dmp42:
@robertjustjones sorry for this will have to be rebased again.

@ggiamarchi @hairyhenderson and other openstack/rackspace interested people, can you help review this?

drivers/rackspace/rackspace.go
rgbkrk: I think as long as you're consistent with your implementation, we should use the `StringSliceFlag` leaving `openstack-sec-groups` to get changed by someone else who needs it.

dgageot:
Hi @robertjustjones any chance you could rebase this PR. Otherwise we might have to close it

dgageot:
@robertjustjones No news for >20 days. Please feel free to reopen if you'd like to carry on working on this PR

robertjustjones:
Reopening with a new rebase.

nathanleclaire:
@robertjustjones You could rebase and force push on this existing branch, then re-open the PR, or submit a brand new PR.

robertjustjones:
@nathanleclaire #2974 is the new PR. Thanks!

nathanleclaire:
Thanks @robertjustjones

PR #1453, opened 2015-07-03
Amazon Ec2 - Add the option amazonec2-use-private-address

### LINKED ISSUE #1442: AWS with private addresses
I'm having a problem with docker machine creating non-public facing instances at AWS/EC2. If I add --amazonec2-private-addresses-only it doesn't add a public IP to the instance. However when there is no public IP present, the instance has no Internet access (this is normal for EC2 I think - no public Internet if you don't have a public IP).

What I really want is for docker-machine to spin up a VM with a public IP but connect to it through the private IP (since I have a VPN between my office and our VPC). Right now if I don't specify --amazonec2-private-address-only, docker-machine will default to connecting to it via the public IP.

It does create the instance and everything fine (I can ssh to it), but docker-machine says "Launching instance..." then hangs forever when I execute this command, and I assume that is because it trying to download docker to the Ubuntu AMI and not able to do so without public internet access:

    docker-machine create --driver amazonec2 --amazonec2-private-address-only=true --amazonec2-security-group=MySecurityGroup dtest002

Is what I'm looking for possible? I think it probably *would* work if I were to use docker-machine to create this VM without putting --amazonec2-private-address-only, then go in and modify my ~/.docker/machine/machines/dtest002/config file, change the IP from the public IP to the local IP, then maybe ssh into the dtest002 and change the docker config to bind only to the private IP. But that is some extra work. Any ideas?

rdgreis:
Fixes #1442 

Adding the option --amazonec2-use-private-address to force the usage of private IP address instead of the public one. Specially useful at cases like described in #1442.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -69,6 +69,7 @@ type Driver struct {
 	RequestSpotInstance bool
 	SpotPrice           string
 	PrivateIPOnly       bool
+	UsePrivateIP		bool
 	Monitoring          bool
 }
 
@@ -171,6 +172,10 @@ func GetCreateFlags() []cli.Flag {
 			Usage: "Only use a private IP address",
 		},
 		cli.BoolFlag{
+        			Name:  "amazonec2-use-private-address",
+        			Usage: "Force the usage of private IP address",
+        		},
+		cli.BoolFlag{
 			Name:  "amazonec2-monitoring",
 			Usage: "Set this flag to enable CloudWatch monitoring",
 		},
@@ -228,6 +233,7 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.SSHUser = flags.String("amazonec2-ssh-user")
 	d.SSHPort = 22
 	d.PrivateIPOnly = flags.Bool("amazonec2-private-address-only")
+	d.UsePrivateIP = flags.Bool("amazonec2-use-private-address")
 	d.Monitoring = flags.Bool("amazonec2-monitoring")
 
 	if d.AccessKey == "" {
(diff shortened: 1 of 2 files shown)


rdgreis:
Fixing gofmt issues.

l15k4:
@rdgreis Hey, do I understand it correctly that it's primary purpose is creating machines from ec2 instances? Because the outcome is the same as If I didn't use this flag, it is indefinitely `Waiting for SSH to be available...`

This is shell of an ec2 instance : 
```
ubuntu@adhoc:~$     docker-machine create \
>     --driver amazonec2 \
>     --amazonec2-access-key ${AWS_ACCESS_KEY_ID} \
>     --amazonec2-secret-key ${AWS_SECRET_ACCESS_KEY} \
>     --amazonec2-region eu-west-1 \
>     --amazonec2-vpc-id vpc-fcf8e29e \
>     --amazonec2-private-address-only \
>     --amazonec2-instance-type t2.micro \
>     --amazonec2-root-size 16 \
>     --amazonec2-security-group druid-p \
>     --amazonec2-zone a \
>     testing-amazonec2-private-address-only
Running pre-create checks...
Creating machine...
(testing-amazonec2-private-address-only) Launching instance...
```

rdgreis:
@l15k4 , Sorry for the delay.

The intention of the -use-private-address option, is to force docker-machine to use the instance's private address, so I can access the instance trough my vpn ( office->VPC ).

The option --amazonec2-use-private-address will force the usage of private address, but you still have a public ip address on the instance. 

If you use the option --amazonec2-private-address-only, you will use the private address to connect, but the instance will not get a public id address. If you need internet access on the instance, you will need to provide a gateway.

l15k4:
@rdgreis 
Thank you, I think it was working all the time, one just needs to use `--native-ssh` when creating machine from within the ec2 instance. Without it it gets stuck on ssh connection.

PS I have agent forwarding enabled ... I think it;s like when you must do `docker-machine ssh foo -A` but docker-machine `create` doesn't have this flag...

PR #1454, opened 2015-07-03
Amazon Ec2 - Add the option amazonec2-use-private-address

### LINKED ISSUE #1442: AWS with private addresses

rdgreis:
Fixes #1442

Adding the option --amazonec2-use-private-address to force the usage of private IP address instead of the public one. Specially useful at cases like described in #1442.


robertjustjones:
Hey @rdgreis. To get the Travis check to pass, you'll need to use "commit -s" to add the "Signed-off-by" line on your commit messages. Just saw the same myself on my recent PR. Look there for an example. It only means that you wrote the code. See docker/docker contributing.md for more

robertjustjones:
You'll also be asked the rebase to squash those commits... easy to do both at the same time.

claudeatsafe:
@rdgreis any plans on merging this?

ehazlett:
@rdgreis would you mind rebasing and making sure your commits are signed with the DCO (https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work).  Otherwise, I think this looks good.

/cc @nathanleclaire

rdgreis:
Thanks @robertjustjones, my first time doing this sign/squash stuff.

@ehazlett rebased, I think it's ok now. :)

claudeatsafe:
:+1: Thanks!

ehazlett:
Closing as this was carried and merged in #1668.  Thanks!

PR #1465, opened 2015-07-06
list openstack remote instance

TrumanLing:
list openstack remote instance

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -338,6 +338,85 @@ var Commands = []cli.Command{
 		},
 	},
 	{
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "quiet, q",
+				Usage: "Enable quiet mode",
+			},
+			cli.StringSliceFlag{
+				Name:  "filter",
+				Usage: "Filter output based on conditions provided",
+				Value: &cli.StringSlice{},
+			},
+			cli.StringFlag{
+				Name: "driver, d",
+				Usage: fmt.Sprintf(
+					"Driver to create machine with. Available drivers: %s",
+					strings.Join(drivers.GetDriverNames(), ", "),
+				),
+				Value: "none",
+			},
+			cli.StringFlag{
+				EnvVar: "OS_AUTH_URL",
+				Name:   "openstack-auth-url",
+				Usage:  "OpenStack authentication URL",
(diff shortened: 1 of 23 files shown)


hairyhenderson:
@TrumanLing thanks for the PR! I'm not sure what the rationale behind this new command is - is there some discussion in a separate issue about this? Can you explain why you think this is a command that should be in docker-machine?

Thanks!

TrumanLing:
@hairyhenderson originally, I want to implement it by expanding options for command ```ls```,but later time it is more convenient to implement another separate command ```rem-vm-ls```;
In addition, I want to monitor the instance status, so I made this PR.

hairyhenderson:
@TrumanLing - ah. It looks to me very similar to the `nova list` command, with nothing really docker-specific. Is that right?

We want to keep the scope of Docker Machine pretty focused, and not try to overlap too much with platform-specific tools, otherwise it will get _very_ bloated :)

TrumanLing:
@hairyhenderson right!!! It is same as ```nova list```, but generally, normal user can not log into openstack to type command ```nova list```. More importantly, if openstack fails to create instance after receiving request from ```docker machine```, then openstack system will  continue to keep resources allocated for this instance, however ```docker machine```'s command ```rm``` can not remove this instance, even if with option ```-f```, it just removes it's local storage. So I made another PR @1482. Even PR @1482 can make consistency between local storage and openstack system, I also hope to provide user one interface to monitor their instances.

hairyhenderson:
@TrumanLing - not sure I understand... You can run `nova list` from any system that has network access to OpenStack - you just install it with `pip install python-novaclient`. It's the same API that docker-machine uses, so if you can reach it with `docker-machine`, then you can reach it with `nova list`.

ggiamarchi:
@TrumanLing I understand your need but i agree with @hairyhenderson on the fact that it overlaps the `nova list` command. OpenStack instances not managed by docker-machine should not be visible using a docker-machine command. About your last point "instance leak", i agree there is probably something to do to make `docker-machine rm` command more reliable after a create error but IMO this is a very different problem.

ehazlett:
Thanks for the contribution!  I also agree with @hairyhenderson and @ggiamarchi.  While I see your reasoning behind this, this is out of the scope of machine.

PR #1475, opened 2015-07-06
Add RESPONSE time column to ls output

nathanleclaire:
If `docker-machine ls` lags, I like to know which call took the longest to respond so I can know which machines are problematic.

This adds support for a "RESPONSE" column which shows how long each goroutine took.

I'm not sure if this is the exact form it should take, but I definitely want some functionality like this somewhere, even if we have to move it into a `--extra` option for `ls` or something.

WDYT @ehazlett @sthulb @hairyhenderson @clnperez ??

image

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -46,7 +46,7 @@ func cmdLs(c *cli.Context) {
 	swarmInfo := make(map[string]string)
 
 	w := tabwriter.NewWriter(os.Stdout, 5, 1, 3, ' ', 0)
-	fmt.Fprintln(w, "NAME\tACTIVE\tDRIVER\tSTATE\tURL\tSWARM")
+	fmt.Fprintln(w, "NAME\tACTIVE\tDRIVER\tSTATE\tURL\tSWARM\tRESPONSE")
 
 	for _, host := range hostList {
 		swarmOptions := host.HostOptions.SwarmOptions
@@ -77,8 +77,9 @@ func cmdLs(c *cli.Context) {
 				swarmInfo = fmt.Sprintf("%s (master)", swarmInfo)
 			}
 		}
-		fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%s\n",
-			item.Name, activeString, item.DriverName, item.State, item.URL, swarmInfo)
+
+		fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
+			item.Name, activeString, item.DriverName, item.State, item.URL, swarmInfo, item.ApiResponseTime)
 	}
 
 	w.Flush()
--- a/libmachine/host.go
+++ b/libmachine/host.go
@@ -55,12 +55,13 @@ type HostMetadata struct {
 }
 
 type HostListItem struct {
-	Name         string
-	Active       bool
-	DriverName   string
-	State        state.State
-	URL          string
-	SwarmOptions swarm.SwarmOptions
+	ApiResponseTime time.Duration
+	Name            string
+	Active          bool
+	DriverName      string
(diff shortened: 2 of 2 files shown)


hairyhenderson:
:+1: I totally like this. But this does make the output even wider... How about a `ls -l` option? IMO the `ACTIVE` column might hide behind that too...

And maybe `LAG` might be more descriptive than `RESPONSE`?

ehazlett:
:+1: We've discussed a similar "health" command to show this info.  I like the idea of showing host related info.

ehazlett:
This needs a rebase.

nathanleclaire:
Rebased!

ehazlett:
@nathanleclaire another (hopefully last!) rebase please!  i <3 this :D

ehazlett:
Also we probably need to update the docs.

dmp42:
@nathanleclaire rebase needed :-)
@docker/machine-maintainers review needed :)

dgageot:
@nathanleclaire Would you like to rebase?

nathanleclaire:
Do we feel like we have a reasonable enough amount of horizontal space to add another column?  I do enjoy this feature, but don't 100% feel like it's necessary in `ls`.

Maybe:

- Put the column before `ERROR` (since it's shorter), or:
- Plan some kind of `docker-machine health` command which would report various health statistics

askb:
:+1: this is definitely a good feature to have, all columns may not be required which could be categorized into different views, at higher level view (default with `machine ls`, good to include the `RESPONSE` here) and have more-detailed view (output when extra flag is used ... ex: `ls ---errors` or `ls --extra`).  Why? Errors strings can overrun the row and break the column formatting, could potentially happen when longer error string is returned. Therefore, if possible keep the `ERRORS` column in more detailed view (we already have `STATE` which displays is machine is on `errors`). To get actual error string{s} admin/users could use an additional flag, which displays fewer or only required columns (`NAME`, `ERRORS`)

pandrew:
@hairyhenderson @nathanleclaire how about `-v` for verbosity level. Add more -v or just --verbose to get deeper info for debugging. `ls` starts to get a bit cluttered. I absolutely think the both response time and remote daemon version is great but when i list machines i want to know which list i want to know what machines is in my inventory so i know which i can connect to, next can i talk with my client to these. If not do they have an outdated version of docker and need an upgrade? `ls` seems slow. Hm do i have a network problem, how long does it take for my api calls to get there?

hairyhenderson:
@pandrew - I totally agree. We already have a `-q/--quiet` flag, so it seems reasonable to have a `-v/--verbose` (or `-l/--long` as I suggested earlier :grin:).

I think the ideal long-term option is probably to pare down the `ls` subcommand and create a new `health` subcommand as @ehazlett and @nathanleclaire have been talking about, but I think this is useful to just get in to the existing `ls` output too.

thaJeztah:
How about having a `-f`/`--format` flag, like `docker ps`? This column could be omitted by default, but added by passing a format (or possibly a default format, like the docker client allows)

dgageot:
@thaJeztah @hairyhenderson See a PR that add `-f / --format` flag https://github.com/docker/machine/pull/2702

thaJeztah:
Thanks @dgageot!

jeanlaurent:
@nathanleclaire this needs to be updated to work with #2702.

dgageot:
@nathanleclaire do you plan to rebase/fix this one?

nathanleclaire:
@dgageot Sure I'll give it a whirl.

dgageot:
Merged with https://github.com/docker/machine/pull/2927

nathanleclaire:
Oh, nice job @dgageot !

PR #1482, opened 2015-07-07
save configuration even if vm creation failure

TrumanLing:
1.save configuration even if vm creation failure
 2.delete local storage without -f option if server create return err

### DIFF
--- a/commands/scp_test.go
+++ b/commands/scp_test.go
@@ -38,6 +38,10 @@ func (d ScpFakeDriver) GetURL() (string, error) {
 	return "", nil
 }
 
+func (d ScpFakeDriver) GetMachineCreateReturnErr() (bool, error) {
+	return false, nil
+}
+
 func (d ScpFakeDriver) GetIP() (string, error) {
 	return "12.34.56.78", nil
 }
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -268,6 +268,10 @@ func (d *Driver) DriverName() string {
 	return driverName
 }
 
+func (d *Driver) GetMachineCreateReturnErr() (bool, error) {
+	return false, fmt.Errorf("Not implemented yet")
+}
+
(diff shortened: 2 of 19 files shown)


ehazlett:
Can you explain the issue this is solving?

TrumanLing:
@ehazlett add field ```CreateCmdReturnErr``` to check status of command ```server.Create```, if command return error, OpenStack nova service doesn't issue instruction to create instance, but it creates local storage for this creation, so add code ```h.SaveConfig``` to store ```error return``` of creation, it is convenient to make ```force delete```  when issuing ```rm``` command in source code, even though it can be manually ```rm``` with option ```-f```.
In addition,  code ```h.SaveConfig``` can store ```instance id``` when command ```server.Create``` return successfully but OpenStack nova service failed at period of implementing code ```d.waitForInstanceActive```, in such case, if no ```instance id``` is stored, it will return ```itemNotFound (404)``` when issuing ```ls``` and ```rm``` or other command, at least storing ```instance id``` makes consistence between local storage and remote OpenStack.

TrumanLing:
@ehazlett PTAL

ehazlett:
@nathanleclaire wdyt?

nathanleclaire:
-1

1. We can't merge anything in which isn't fully implemented
2. The `Driver` interface is way overloaded already
3. Saving config is already over-done and I don't want to keep piling on more stuff.  Config should be saved when the instance is created, and it's why we have made a concentrated effort to exit from the `Create` call as soon as possible (when the relevant bits e.g. instance ID have been filled in).  We haven't updated all the drivers, though, and some still do things like poll for state or IP address that they don't need to.
4. An `error` implies that `Create` failed, not succeeded.  If it succeeded but didn't exit `Create` for whatever reason, the `Create` method should exit sooner.

nathanleclaire:
I'm going to close this, but feel free to open a proposal in the issues detailing the specific problem you are trying to solve and how.

Issue #1319, opened 2015-06-06
machine 0.3 does not work if docker daemon needs http_proxy


lasdecoeur:
Hi,

I saw that docker-machine now delete the initial /etc/default/docker file (ubuntu with openstack driver) and put his DOCKER_OPTS in a new one.

What if there is some "export http_proxy" in this file (from a previous installation) ? The daemon started by machine hasn't got internet and is not able to pull images.

I searched some issues about that and found nothing.

A solution is to do docker-machine ssh after and put new lines in this file and restart the daemon, but it could be nice to be able to give to docker-machine (with a file or CLI options) to set those variables ?. 

Any help/advices ? Thanks.

hairyhenderson:
This has been brought up a few other times, but from different angles...

I think all we need is some kind of `--proxy` setting on `create` which would be delegated to the provisioners to set `http_proxy`/`https_proxy`in the environment when downloading stuff, and also in the engine config file.

A bunch of other requests like these have been closed due to plans to move to cloudinit, but those don't seem to be materializing as quickly as previously thought ;)

/cc @nathanleclaire @ehazlett

sthulb:
If we could, we should detect the proxy and then pass it through, if it's a local machine.

hairyhenderson:
@sthulb +1, though that's definitely only relevant for the local drivers

ehazlett:
@hairyhenderson good point re: cloudinit.  I think we could add a `--proxy` option.  I'll look into this.

nathanleclaire:
> I think all we need is some kind of --proxy setting on create which would be delegated to the provisioners to set http_proxy/https_proxyin the environment when downloading stuff, and also in the engine config file.

I'm in favor of the general idea (we've had enough requests for setting the proxy for daemon in particular that I think we should support it), I am not sure if we should conflate provisioning-time proxy with daemon run-time proxy or not though.  Having a proxy for the daemon would imply `--engine-proxy` or similar.

> If we could, we should detect the proxy and then pass it through, if it's a local machine.

Hm, maybe an option/flag to do so, but I'm hesitant on automatic behaviors such as this.  The user should be in the mindset of thinking of created machines as independent entities with their own behaviors.  Conflating this has already caused us some pain before, e.g. with volumes.

sirlatrom:
@nathanleclaire I can imagine multiple scenarios when the proxy used for `docker-machine` would be a different one than that used by the engines it creates and configures

stephanbuys:
[...]

PR #1497, opened 2015-07-09
added {http,https,no} proxy support to docker engines

udryan10:
Adds HTTP_PROXY, HTTPS_PROXY and NO_PROXY support to docker engines per issue #1319

### DIFF
--- a/docs/reference/create.md
+++ b/docs/reference/create.md
@@ -60,6 +60,7 @@ Options:
    --engine-registry-mirror [--engine-registry-mirror option --engine-registry-mirror option]           Specify registry mirrors to use
    --engine-label [--engine-label option --engine-label option]                                         Specify labels for the created engine
    --engine-storage-driver "aufs"                                                                       Specify a storage driver to use with the engine
+   --engine-env                                                                                         Specify environment variables to set in the engine
    --swarm                                                                                              Configure Machine with Swarm
    --swarm-master                                                                                       Configure Machine to be a Swarm master
    --swarm-discovery                                                                                    Discovery service to use with Swarm
@@ -144,6 +145,17 @@ $ docker-machine create -d virtualbox \
     --engine-opt dns=8.8.8.8 \
     --engine-opt log-driver=syslog \
     gdns
+```
+
+Additionally, Docker Machine supports a flag, `--engine-env`, which can be used to
+specify arbitrary environment variables to be set within the engine with the syntax `--engine-env name=value`. For example, to specify that the engine should use `example.com` as the proxy server, you could run the following create command:
+
+```
+$ docker-machine create -d virtualbox \
+    --engine-env HTTP_PROXY=http://example.com:8080 \
+    --engine-env HTTPS_PROXY=https://example.com:8080 \
+    --engine-env NO_PROXY=example2.com \
+    proxbox
 ```
 
 ## Specifying Docker Swarm options for the created machine
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -196,6 +196,11 @@ var sharedCreateFlags = []cli.Flag{
 		Name:  "engine-storage-driver",
 		Usage: "Specify a storage driver to use with the engine",
 	},
+	cli.StringSliceFlag{
+		Name:  "engine-env",
+		Usage: "Specify environment variables to set in the engine",
+		Value: &cli.StringSlice{},
+	},
 	cli.BoolFlag{
 		Name:  "swarm",
 		Usage: "Configure Machine with Swarm",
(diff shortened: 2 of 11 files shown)


nathanleclaire:
Hm, cool, thanks for the contribution. 

I'm wondering if it might be better to have just one flag, which will then apply to any environment variable, instead of three (and possibly more in the future) different flags?

That way, you would : 

```console
$ docker-machine create -d provider --engine-env HTTP_PROXY=http://foo.com --engine-env HTTPS_PROXY=bar.com proxybox
```

What do you think?

udryan10:
@nathanleclaire I like it! I'll update the PR.

udryan10:
@nathanleclaire the PR now has code that implements the more general `--engine-env` flag

nathanleclaire:
Very cool.  Any ideas how to test this in the integration suite?

udryan10:
@nathanleclaire I came up with a test that should work well -- spins up a machine and checks `/proc/<docker daemon pid>/environs` for the set Environment Variable. Let me know what you think.

ehazlett:
:+1: for this -- i know a lot will benefit :)

nathanleclaire:
Looking pretty good --- we are about to merge in some structural changes to our docs, I will need you to rebase after that and then we will be ready to rock.  I'll ping you when it's ready to rebase.

nathanleclaire:
OK I've merged that docs PR -- please rebase when it is convenient for you.  Thanks!

udryan10:
@nathanleclaire I have rebased my doc changes into the new doc structure

docs/reference/create.md
```diff
@@ -1,4 +1,4 @@
-<!--[metadata]>
+## Specifying Docker Swarm options for the created machine<!--[metadata]>
```
nathanleclaire: I'm not sure this is supposed to be here?

nathanleclaire:
cc @ehazlett for review of this PR

udryan10:
@nathanleclaire fixed up the mistake in the `docs/reference/create.md` you referenced - apologies for that oversight.

nathanleclaire:
Thanks @udryan10 and no problem.

Hey @tianon, could you please take a quick look at these changes to the various OS configurations and note if anything strikes you as incorrect?  For instance, I'm not sure if we should do `export VAR=blah` in `/var/lib/boot2docker/profile`, the rest of environment variables set there are not exported.

docs/reference/create.md
```diff
@@ -146,6 +147,17 @@ $ docker-machine create -d virtualbox \
     gdns
 ```
 
+Additionally, Docker Machine supports a flag, `--engine-env`, which can be used to
+specify arbitrary environment variables to be set within the engine with the syntax `--engine-env name=value`. For example, to specify that the engine should use `example.com` as the proxy server, you could run the following create command:
+
+```
+$ docker-machine create -d virtualbox \
+    --engine-env HTTP_PROXY=http://example.com:8080 \
+    --engine-env HTTPS_PROXY=http://example.com:8080 \
```
nathanleclaire: Heh... this is HTTPS_PROXY but the proto specified is `http` ;D

docs/reference/create.md
```diff
@@ -146,6 +147,17 @@ $ docker-machine create -d virtualbox \
...
+    --engine-env NO_PROXY=example2.com \
+    proxbox
```
nathanleclaire: Nice name :)

udryan10: :+1:

tianon:
It all looks reasonably sane, but what happens when I need quotes in my environment variable values?  ie: `--engine-env SOME_VAR='some "value" '` (which will come through into Go as `"SOME_VAR=some \"value\" "`)

tianon:
(ie, we probably ought to pass these through `%q` from `fmt` at the very least)

udryan10:
I've updated the code to run the env's through %q

ehazlett:
LGTM

nathanleclaire:
LGTM

chadswen:
It seems that some aspect of proxy variable parsing can be picky about the proxy URL schema. I was getting "Proxy Authentication" errors with a proxy that doesn't require credentials. Adding a trailing slash to my proxy variables fixed the issue, as in:
```
docker-machine create -d virtualbox \
    --engine-env HTTP_PROXY=http://example.com:8080/ \
    --engine-env HTTPS_PROXY=https://example.com:8080/ \
    default
```
This format is used in the [Docker Engine example](https://docs.docker.com/articles/systemd/#http-proxy), but it's not used in the [`docker-machine create` example](https://github.com/docker/machine/commit/ad6d8d49a8f281a167a689f000045f5bea469a60#diff-e22bac5e382dbfda7cf2840ffff1b3bfR154).

nathanleclaire:
@chadswen Could I get you to file a new issue about the documentation change required and cc @udryan10 and myself please?  Thanks!

PR #1499, opened 2015-07-10
[debian] Make enabling docker service work

### LINKED ISSUE #1495: docker service not enabled on debian
docker-machine creates an ``/etc/systemd/system/docker.service`` file that contains only ``[Service]`` section, this causes the following error later when it runs ``systemctl enable docker``:

```
About to run SSH command:
sudo systemctl enable docker
...
Synchronizing state of docker.service with SysV init with /lib/systemd/systemd-sysv-install...
Executing /lib/systemd/systemd-sysv-install enable docker
The unit files have no [Install] section. They are not meant to be enabled
using systemctl.
```

Which results in docker service not being enabled to start on boot.

I think overriding the entire service file shouldn't be needed, machine could re-use the service settings that the lxc-docker package installs, by using an override file instead of ``/etc/systemd/system/docker.service``.  E.g. tested with the following contents in ``/etc/systemd/system/docker.service.d/override.conf``:

```
[Unit]
After=network.target
Requires=

[Service]
ExecStart=
ExecStart=/usr/bin/docker -d -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --storage-driver btrfs --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=generic
```

(where ExecStart is the line generated by machine, also note ExecStart needs to be cleared first ([see here](https://bugzilla.redhat.com/show_bug.cgi?id=756787)))

This works fine, the service is started automatically after boot and inherits settings from docker.service provided by lxc-docker package.

chantra:
Fixes #1495
/etc/systemd/system/docker.service was missing the [Install] section
making the enable action fail with:

```
root@jessie:~# systemctl enable docker
Synchronizing state for docker.service with sysvinit using
update-rc.d...
Executing /usr/sbin/update-rc.d docker defaults
Executing /usr/sbin/update-rc.d docker enable
The unit files have no [Install] section. They are not meant to be
enabled
using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another
unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which
has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path,
timer,
   D-Bus, udev, scripted systemctl call, ...).
```

Signed-off-by: Emmanuel Bretelle <chantr4@gmail.com>

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -172,6 +172,9 @@ MountFlags=slave
 LimitNOFILE=1048576
 LimitNPROC=1048576
 LimitCORE=infinity
+
+[Install]
+WantedBy=multi-user.target
 `
 	t, err := template.New("engineConfig").Parse(engineConfigTmpl)
 	if err != nil {


ehazlett:
LGTM

/cc @nathanleclaire @tianon

chantra:
I reckon that should also be ported to other systemd based distro but did not get a chance to test it on those.

nathanleclaire:
LGTM

Issue #1266, opened 2015-05-29
Output from SSH command that fails should be printed


bfirsh:
```
$ docker-machine create ...
Creating SSH key...
Creating Digital Ocean droplet...
Error creating machine: exit status 255
You will want to check the provider to make sure the machine and associated resources were properly removed.
```

This isn't particularly helpful. It should probably be something like this:

```
$ docker-machine create ...
Creating SSH key...
Creating Digital Ocean droplet...
Error running command "sudo -E apt-get update" on the Machine. Here is the output:

---------------------------------
... error ...
---------------------------------

You will want to check the provider to make sure the machine and associated resources were properly removed.
```

prologic:
+1

nathanleclaire:
I'm all for storing / showing more information in the case of an error, although I think we need to think about the approach a bit.  Generally where I'm headed is towards having a `docker-machine logs` command which stores anything which gets logged by Machine in flight (including all debug output) to a log file on disk (just like `docker logs`), and that is something which can be queried later by the user, or by us if we want to selectively show things like above.

Bear in mind that there are many cases in Machine where things like SSH commands could fail and it is perfectly fine and normal (e.g. the SSH command timed out so we want to re-try it), so implementing this kind of thing without also spewing "undesirable"/scary error messages at the user is a little trickier than it may look at first glance.

nathanleclaire:
Update:  I may have been a bit hasty.  I think that the "good" SSH failures (testing if SSH is up / working) are largely exit code 255, so perhaps the proposed solution in my PR might work OK for now.

PR #1501, opened 2015-07-10
Bail on failed SSH command with information about what failed

nathanleclaire:
@ehazlett @bfirsh @prologic WDYT?

Closes https://github.com/docker/machine/issues/1266 

### DIFF
--- a/drivers/utils.go
+++ b/drivers/utils.go
@@ -2,10 +2,15 @@ package drivers
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/docker/machine/log"
 	"github.com/docker/machine/ssh"
 	"github.com/docker/machine/utils"
+)
+
+const (
+	ErrExitCode255 = "255"
 )
 
 func GetSSHClientFromDriver(d Driver) (ssh.Client, error) {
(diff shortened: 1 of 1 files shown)


prologic:
We can't just test for the exit code directly?

nathanleclaire:
@prologic There's no cross platform way to do so that I am aware of (see http://stackoverflow.com/questions/10385551/get-exit-code-go, for instance).

prologic:
Hmm I find this quite surpsingly really See [Popen.returncode](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.returncode) -- Unless the Python standard library has been lying to me this whole time!

prologic:
Go really has no platform independent way of returning the exit status of a subprocess?

nathanleclaire:
> Go really has no platform independent way of returning the exit status of a subprocess?

AFAICT not really.  There are some "hack-arounds", and it's fairly easy to tell whether a process failed or succeeded, but I think the actual numeric codes are fairly platform-specific.

prologic:
Weird but okay :)


James Mills / prologic

E: prologic@shortcircuit.net.au
W: prologic.shortcircuit.net.au

bfirsh:
Nice!

Might be worth explaining a little clearer what's going on to make it less scary: "Machine ran a command on a host using SSH and it failed:"

Does the upstream error also explain how you might recover from this? (e.g. remove the host and try creating it again)

nathanleclaire:
cc @ehazlett --- I'm not sure about this exact output format but I _really_ want to get this in for the next release as our failing SSH commands right now and not spitting out much that's useful.

ehazlett:
I think we can tweak the output later.  I agree it would be great to have this soon.

LGTM

nathanleclaire:
Pending Travis, LGTM.  Should work with native and external SSH both.

ehazlett:
LGTM

Issue #1410, opened 2015-06-23
Usability polish: per-command template text


rhunter:
Currently, all commands use the same usage guide template.

**`$ docker-machine help create`**
```
Usage: docker-machine create [OPTIONS] [arg...]
 ...
```

In the case of `create`, the reality is that exactly one final argument is required and that it refers to the machine name.

I don't know if there's a *formal* specification of Unix command usage notation, but there's certainly a convention that `[]` is for optional arguments and `...` is for repeating (one or more) occurrences. The convention for a mandatory argument seems to be `<>` (or underlines, or nothing at all).

So for `create`, a more helpful usage text would be:
```
Usage: docker-machine create [OPTIONS] <name>
```
The same would apply to `rm`, `ssh`, `config`, and several others. At a quick glance, it's not obvious how to provide the information to @codegangsta's "cli.go" but it looks like `SubcommandHelpTemplate` might be relevant.

(In the case of `create`, it would be even more helpful still to treat the per-driver mandatory arguments as their own use case, but that might be asking a bit much.)

nathanleclaire:
Sounds good to me.

Could be a good first PR to machine project to add these templates ;D

PR #1504, opened 2015-07-11
Suggest clean up of usage for #1410

ericsage:
Signed-off-by: Eric Sage <eric.david.sage@gmail.com>

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -239,14 +239,14 @@ var sharedCreateFlags = []cli.Flag{
 
 var Commands = []cli.Command{
 	{
-		Name:   "active",
-		Usage:  "Print which machine is active",
-		Action: cmdActive,
+		Name:        "active",
+		Description: "Print the active machine.",
+		Action:      cmdActive,
 	},
 	{
 		Name:        "config",
-		Usage:       "Print the connection config for machine",
-		Description: "Argument is a machine name.",
+		Usage:       "<machine-name>",
+		Description: "Print the specified machine's connection configuration.",
 		Action:      cmdConfig,
 		Flags: []cli.Flag{
 			cli.BoolFlag{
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Why not include some sample output?  It would save time in reviewing the PR.  Generally I'm kind of skittish about changing the help text.

ericsage:
Sure, I will post some sample output. Looking through the current descriptions, they seem to augment Usage, as the use of Usage doesn't really provide much in the way of useful information. By interpreting Usage as the unix style sample usage, and the description as describing what the command does, they have a better separation of concerns in my eyes. I think it would be even better to have an additional field in the template, but I wanted to go ahead and throw this out there. It seems to be difficult to add fields to the struct that gets applied to the templates without making some changes to codegansta/cli.

```bash
./machine

Commands:
  active		Print the active machine.
  config		Print the specified machine's connection configuration.
  create		Create a machine with the specified name.
  env			Display the commands to setup up the environment for the Docker client.
  inspect		Inspect information about a machine..
  ip			Get the ip(s) of one or more machines.
  kill			Kill one or more machines.
  ls			List machines.
  regenerate-certs	Regenerate TLS Certificates for a machine.
  restart		Restart one or more machines.
  rm			Remove one or more machines.
  ssh			Establish a secure connection to the machine, optionally supply a command to run on the target.
  scp			Copy files securely from host to the target machine, or vice-versa.
  start			Start one or more machines.
  stop			Stop one or more machines.
  upgrade		Upgrade one or more machines to the latest version of Docker.
  url			Get the url of a machine.
  help, h
```

```bash
./machine help ssh

Usage: docker-machine ssh <machine-name> [command]

Description:
   Establish a secure connection to the machine, optionally supply a command to run on the target.
```

nathanleclaire:
Cool, thanks.

@hairyhenderson Having done some work on the help text before, WDYT?

hairyhenderson:
Hmm... This is good _in principle_, but the whole `Usage` vs. `Description` thing is a little weird. The way the code stands right now, the auto-generated `help` command is missing its description, which looks a little odd.

Maybe if there's a way to get the help command's `Description` field populated, it would work...

It does seem that there's some use cases that aren't fully supported in `codegansta/cli`, so probably would be good to issue some PRs there too.

ehazlett:
@ericsage would you be willing to look at `codegangsta/cli` to see if we can address there instead of this way as @hairyhenderson suggested?

nathanleclaire:
I'm going to close this -- IMO it deviates too much from the prescribed `codegangsta/cli` style.  Please feel free to open an issue to discuss if you are adamant.

PR #1512, opened 2015-07-13
Add support for building Docker Machine binaries remotely (without bind mount)

imikushin:
Build on docker-machine provisioned docker

i.e. in the cloud

Signed-off-by: Ivan Mikushin <i.mikushin@gmail.com>

### DIFF
--- a/script/build-remote
+++ b/script/build-remote
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -e
+
+BUILD_IMAGE_NAME="docker-machine-build"
+GOOS=${GOOS:-"darwin linux windows"}
+GOARCH=${GOARCH:-"386 amd64 arm"}
+
+docker build -t ${BUILD_IMAGE_NAME} .
+
+BUILD_CONTAINER=$(docker run -d \
+    ${BUILD_IMAGE_NAME} \
+    gox \
+    -os "$GOOS" \
+    -arch "$GOARCH" \
+    -output="docker-machine_{{.OS}}-{{.Arch}}" \
+    -ldflags="-w -X github.com/docker/machine/version.GitCommit `git rev-parse --short HEAD`")
+cleanup_container() {
+    docker rm -v ${BUILD_CONTAINER}
+}
+trap cleanup_container EXIT
+
+docker logs -f ${BUILD_CONTAINER} &
+BUILD_STATUS=$(docker wait ${BUILD_CONTAINER})
+if [[ ${BUILD_STATUS} != 0 ]]; then exit ${BUILD_STATUS}; fi
+
+BUILT_IMAGE=$(docker commit ${BUILD_CONTAINER})
+cleanup_image() {
+    cleanup_container
+    docker rmi ${BUILT_IMAGE}
+}
+trap cleanup_image EXIT
+
+echo "Copying built binaries:"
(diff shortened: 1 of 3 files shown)


timfallmk:
:+1:

ehazlett:
Perhaps we could add this as another command instead of changing the current build?  I'm worried that this will cause slower build times by default if you are on a slow link.  Maybe a `script/build-remote` or something?  Otherwise I like the idea :)

nathanleclaire:
Yeah, I'm in favor of this, but as @ehazlett said, I'd rather have it as a separate script in `script/` than as a modification of the existing one.  The release script really shouldn't be changed, and there needs to be at least a small blurb in the documentation on contributing which mentions this remote build method.

ehazlett:
@imikushin would you mind making this an additional script and some docs about building remote?  thanks!

imikushin:
@ehazlett Are you sure you *absolutely* don't want to touch the build script? Reason I'm asking is people who are not on Linux (I guess, your target audience) won't be able to use the same build scripts as you guys, and so these scripts will have a good chance to divert in subtle ways.

I don't mind implementing your decision, whatever it is. Just wanted to make the build portable. 

PS: the speed difference will probably be barely noticeable for a local docker daemon. But if you can't natively run docker, you'll necessarily end up tweaking the build script to something like this.

ehazlett:
@imikushin actually they can.  the "preferred" way is to use b2d (or another docker host) and setup your environment for use with that docker host.  one way is to use machine.  when i was on a mac, i would use machine to create a "dev" host that i would then use to build machine.  since we have this working and is part of the release process, i would rather have an additional script to build remotely.  we can always merge them together if we see that it's desired.

imikushin:
@ehazlett NP. Pushing it shortly

nathanleclaire:
@tianon Could you please take a look at these scripts and let us know what you think?

script/build-remote
```diff
@@ -0,0 +1,28 @@
...
+
+BUILT_IMAGE=$(docker commit ${BUILD_CONTAINER})
+cleanup_image() {
+    cleanup_container
+    docker rmi ${BUILT_IMAGE}
+}
+trap cleanup_image EXIT
+
+echo "Copying built binaries to ./:"
+for f in `docker run --rm ${BUILT_IMAGE} ls | grep 'docker-machine_'`; do
```
tianon: I'm personally always scared when I see the output of `ls` scripted.  In this case, it's _probably_ safe, but the output from `ls` is usually supposed to be human-readable, so I'd trust a raw shell glob more personally, something like:
```bash
for f in $(docker run --rm ${BUILT_IMAGE} sh -c 'echo docker-machine_*'); do
    ...
```

tianon:
Seems sane, although I have to admit I'm not a big fan of `gox` (it does the job well enough though).

script/build-remote
imikushin: @tianon Thanks! I like your suggestion.

imikushin:
Updated according to @tianon's suggestion

nathanleclaire:
You'll want to update to conform to : https://github.com/docker/machine/pull/1686

Sorry about that.

imikushin:
Done

nathanleclaire:
LGTM

PR #1528, opened 2015-07-15
Allow the default user of an EC2 Instance to be set at creation time.

ekristen:
This adds the ability for you to change the default user for an EC2 instance via UserData. Allows a user to change the the user to which the default SSH key gets installed to on the machine. 

Use Case: Some users, like myself, use custom AMIs that do not have the default user accounts like ubuntu and ec2-user, this PR allows docker-machine to utilize customer user accounts as the default user by providing a cloud config via UserData.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -60,6 +60,7 @@ type Driver struct {
 	SpotPrice           string
 	PrivateIPOnly       bool
 	Monitoring          bool
+	DefaultUser         string
 }
 
 func init() {
@@ -164,6 +165,11 @@ func GetCreateFlags() []cli.Flag {
 			Name:  "amazonec2-monitoring",
 			Usage: "Set this flag to enable CloudWatch monitoring",
 		},
+		cli.StringFlag{
+			Name:  "amazonec2-default-user",
+			Usage: "Set the name of the default user for the AMI",
+			EnvVar: "AWS_DEFAULT_USER",
+		},
 	}
 }
 
@@ -209,6 +215,7 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.SSHPort = 22
 	d.PrivateIPOnly = flags.Bool("amazonec2-private-address-only")
 	d.Monitoring = flags.Bool("amazonec2-monitoring")
+	d.DefaultUser = flags.String("amazonec2-default-user")
 
 	if d.AccessKey == "" {
 		return fmt.Errorf("amazonec2 driver requires the --amazonec2-access-key option")
@@ -221,7 +228,7 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	if d.SubnetId == "" && d.VpcId == "" {
 		return fmt.Errorf("amazonec2 driver requires either the --amazonec2-subnet-id or --amazonec2-vpc-id option")
 	}
-
+	
 	if d.isSwarmMaster() {
 		u, err := url.Parse(d.SwarmHost)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


ehazlett:
Thanks for the submission!  I'm curious why you don't use the `--amazonec2-ssh-user`?  That is what is is for and we actually recommend to use it with other AMIs that are not ubuntu (debian, centos, rhel, etc)

ekristen:
To answer your question, I do use `--amazonec2-ssh-user` however the user provided with that option *must* be configured in the AMI as the "default" user in the cloud config. Until this morning I thought that had to be provided via `UserData` but I discovered editing the `/etc/cloud/cloud.cfg` has the same result of `UserData`.

Hopefully this makes sense. The nice thing about merging this in, is it will allow users to override the default user created on any AMI that supports cloud-init without having to build a custom AMI.

ehazlett:
I see the point and use case I'm just hesitant to introduce cloudinit data.  I'm curious the workflow for using a different user than the AMI but yet not using a custom AMI.  Is it just to have a different user specified for ssh?  For security?

ekristen:
It is really a security thing. Best practices is to never use the default anything. This option would allow someone to easily change from the defaults with docker-machine without having to build a custom AMI which I think to some would be a plus.

The EC2 driver is all about spinning build cloud servers, why the hesitation regarding cloudinit data? (just curious)

Perhaps there are others that are willing to comment?

hairyhenderson:
Thanks for this, @ekristen!

This sounds in general like a good idea. It's a bit scary having millions of VMs running in EC2 where the `ubuntu` and `ec2-user` account is present and not always protected by a good password. So, from that aspect, :+1:.

However, I'm not sure I understand why we can't just use `--amazonec2-ssh-user` to populate the default user field.

Also I'm curious about the default user - will cloud-init set up a `sudoers` entry for that user, or does that have to be part of the AMI (or something separate in the `cloud-config`)?

Having two separate flags allows this use:

```console
$ docker-machine create -d amazonec2 --amazonec2-default-user bob --amazonec2-ssh-user joe mymachine
...
```

This seems to me like it'd fail...

One caveat is that if `--amazonec2-ssh-user` is used to set the default user, we do need to make sure not to override it if using the default value (`ubuntu`).

Overall, I'm +1 on this, but I do want to make sure we don't add extra flags unnecessarily...

ekristen:
You could use the `--amazonec2-ssh-user` and probably makes sense to do so. If `docker-machine` is provisioning the instance, it has to be able to tell AWS which user to load the private key it creates against. 

I can go either way, we either need to make it so that `--amazonec2-ssh-user` sets the cloud-init data to indicate that is the default user to load the SSH key against, or we add the second option and then if you use a non-standard user you have to provide both options.

nrdlngr:
This seems like a cool idea, and I stumbled across it while trying to build my own way to add user data to EC2 with docker-machine.

My only hope is that any flag that injects user data for one function also allows for [injecting user data another way](https://github.com/docker/machine/pull/1637) without one overwriting the other.

Or at least the documentation could state that the incompatible flags cannot be used together.

Thoughts?

ekristen:
Probably easiest to make the two flags not compatible.

ehazlett:
I understand what you need here however, I think this goes a bit beyond the scope of Machine.  We want to enable the ability to use custom users but at the same time we don't want to be managing them when we can't do that uniformly across the other providers.  Thank you for the contribution but I'm going to close this for the that reason and similarly here: https://github.com/docker/machine/pull/1637#issuecomment-128555593

ekristen:
Is the scope of machine defined somewhere? 

Respectfully I disagree. If we want machine to be useful for more than a simple dev tool (which I do)

#1637 and this are totally separate issues in my opinion. While #1637 would be interesting to have, it is far to problematic. This PR is purely making it so that people can easily not use the default user if they want.

Without this option however, everyone is forced to **always** use the default user (ubuntu or ec2-user) **OR** go through the process of creating their own AMI.

I'd totally support having `--amazonec2-ssh-user` set the default user in the userdata if you'd rather that option.

Just my two cents. Thanks.

Issue #1511, opened 2015-07-13
Proposal: pass driver-specific flags to subcommands other than "docker-machine create"


jcmoore:
Certain subcommands (`rm`, `ls`, etc.) may need authentication credentials depending on what driver was used to create the relevant machines -- this is at least the case for the google drive and I suspect it is true for other cloud drivers.

Being able to pass useful driver configuration into subcommands other than `create` could greatly simplify their use.

EDIT:

I'm working towards a PR here -- https://github.com/jcmoore/machine/tree/1511-driverconfig
See the current diff -- https://github.com/docker/machine/compare/master...jcmoore:1511-driverconfig

nathanleclaire:
Well, it wouldn't really simplify their use, as today the auth credentials just get stored on disk alongside the other machine metadata, so `stop` and so on "just work" out of the box.  I agree that it's not really ideal from a security standpoint, but I don't think passing the same flags to each subcommand is how I'd like to solve it.

jcmoore:
After taking a closer look, most of my frustration may come from characteristics of the google driver (and is not an issue with other drivers) in that each machine on google requires it's own auth token when they could be shared (and since auth-tokens are limited per user, having too many machines invalidates existing tokens, throwing you into a bit of an authentication loop).

I'm going to take a closer look at the driver itself and see if I can't make more effective improvements there.  Should I close PR #1548 ?

ehazlett:
@jcmoore I agree that this is much more prevalent with the google driver.  I think perhaps a change in the driver would be more effective than #1548.

dgageot:
@jcmoore This should be fixed in 0.5.0 because google driver now uses Google's Default Authentication.

Feel free to reopen though.

PR #1548, opened 2015-07-19
Gather driver flags from env for non-create calls

jcmoore:
This permits passing driver flags (especially for authentication) required by the libmachine/host that may not otherwise be available because flags are strictly enforced at the subcommand level.

Addresses #1511


nathanleclaire:
Really appreciate the work, but I think we need to fix our broken credential storage model instead.  So, it will require a more fundamental change to the way Machine works.

Issue #1460, opened 2015-07-03
VMware Fusion: Host volume file corruption


adambiggs:
I've been testing Machine with VMware Fusion for local development now that #641 has been resolved, but have run into another issue with file corruption in host volumes.

When writing to an existing file on the host, in some cases the shared file in the VM is corrupted. Similar to #1275, but instead of the file having no contents, it contains the *old* data (before being written), but it's length is updated to match that of the host file...

- If characters are removed from host file, shared file is truncated to match same amount of characters
- If characters are added to host file, `^@` null characters are appended to shared file
- In both cases, shared file contents are not updated with actual changes from host file

Strangely, I only noticed this when compiling CoffeeScript files. Manually writing to a file with Vim works totally fine. Not sure why it works with Vim... but even a simple `$ echo "test 123" > test.js` on the host will trigger the problem.

I was able to reliably reproduce the issue with these steps:
- Create `test.coffee` with some dummy content and compile it on host
    - `test.js` looks good in VM

- Add or remove some lines from `test.coffee` and recompile
    - `test.js` is now corrupt in VM

- Open `test.js` with host Vim and save
    - `test.js` is fixed in VM

- `$ echo "test 123" > test.js` on host
    - `test.js` is corrupt again in VM

Very strange...

adambiggs:
I found this thread in the VMware forums which sounds very similar: https://communities.vmware.com/thread/462303

donnex:
I'm having the same issue. My python files got corrupt when I was testing docker-machine with Fusion. I had a shared folder from my Yosemite host to my VMWare Fusion Docker VM. Noticed some really strange file corruptions/files not updating when saved.

I was running Gunicorn inside the container and it picked up and reloaded the server when files were saved, but the content of the files did not update.

The only way I found to get the files to update was to restart the VM.

nathanleclaire:
cc @frapposelli

frapposelli:
@adambiggs @donnex I tried but cannot reproduce, can you get me a simple case I can test? also, which version of Docker/Boot2Docker are you using?

donnex:
## Versions
Docker version 1.7.0, build 0baf609

docker-machine version 0.3.0 (HEAD)

docker-compose version: 1.3.1
CPython version: 2.7.6
OpenSSL version: OpenSSL 0.9.8zf 19 Mar 2015

VMWare Fusion Version 7.1.2 (2779224)

Boot2Docker version 1.7.0, build master : 7960f90

## To reproduce
[...]

PR #1551, opened 2015-07-20
vmwarefusion: force aggressive time synchronization

steeve:
VMWare Fusion's shared folders rely heavily on the time to be kept in sync between the host and guest.
The drift can be significant enough so that shared folders become out of sync.

This ensure that vmware-tools keep the guest time in sync with the host every second.

### DIFF
--- a/drivers/vmwarefusion/vmx_darwin.go
+++ b/drivers/vmwarefusion/vmx_darwin.go
@@ -63,4 +63,6 @@ uuid.action = "create"
 numvcpus = "{{.CPU}}"
 hgfs.mapRootShare = "FALSE"
 hgfs.linkRootShare = "FALSE"
+tools.syncTime = "TRUE"
+tools.syncTime.period = 1
 `


frapposelli:
@steeve I'm not aware of any correlation between hgfs and time sync, where did you get this from?

steeve:
Indeed after more testing, it seems the issue is still here.
However, it's less frequent (used to be pretty much a paint to keep the folder in sync).

Could be an impression, though.

In any case, I guess this one doesn't hurt ?

steeve:
Guess I should open an issue instead?

frapposelli:
@steeve yes, take a look at #1460 first and then open an issue if that's not the problem you're facing.

I'm not ok with the PR as syncing time with tools is not the best way to go, and syncing with that every second  will just put unnecessary pressure on the host and guest, let's keep the time sync with ntp.

steeve:
Indeed I am having the same issue.

I do agree on the NTP front. Closing now, I'll follow up on #1460.

steeve:
Thanks guys!

PR #1552, opened 2015-07-20
VirtualBox: Update to support new B2D

ehazlett:
This PR is for machine support for the new version of Boot2Docker.  This does the following:

- Enable shared folders for all supported platforms (Mac, Win, Linux)
- Restrict default shared folder to user home directory
- Remove the SSH key loading via disk image
- Cleanup B2D provisioner

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -290,24 +290,7 @@ func (d *Driver) Create() error {
 		return err
 	}
 
-	var shareDir string
-
-	homePath := homedir.Get()
-
-	switch runtime.GOOS {
-	case "windows":
-		shareDir = homePath
-		mountDir, err := translateWindowsMount(homePath)
-		if err != nil {
-			return err
-		}
-		shareDir = mountDir
-	case "darwin":
-		shareDir = homePath
-	case "linux":
-		shareDir = homePath
-	}
-
+	shareDir := homedir.Get()
 	shareName := shareDir
 
 	log.Debugf("creating share: path=%s", shareDir)
@@ -321,12 +304,21 @@ func (d *Driver) Create() error {
 	}
 
 	if shareDir != "" {
+		log.Debugf("setting up shareDir")
 		if _, err := os.Stat(shareDir); err != nil && !os.IsNotExist(err) {
+			log.Debugf("setting up share failed: %s", err)
 			return err
 		} else if !os.IsNotExist(err) {
-			if shareName != "" {
-				// parts of the VBox internal code are buggy with share names that start with "/"
-				shareName = strings.TrimLeft(shareDir, "/")
+			// parts of the VBox internal code are buggy with share names that start with "/"
+			shareName = strings.TrimLeft(shareDir, "/")
+
+			// translate to msys git path
+			if runtime.GOOS == "windows" {
+				mountName, err := translateWindowsMount(shareDir)
+				if err != nil {
+					return err
+				}
+				shareName = mountName
 			}
 
 			log.Debugf("adding shared folder: name=%q dir=%q", shareName, shareDir)
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	isoFilename         = "boot2docker.iso"
+	isoFilename         = "boot2docker-virtualbox.iso"
 	defaultHostOnlyCIDR = "192.168.99.1/24"
 )
 
@@ -153,7 +153,7 @@ func (d *Driver) Create() error {
 		return err
 	}
 
-	b2dutils := utils.NewB2dUtils("", "")
+	b2dutils := utils.NewB2dUtils("", "", isoFilename)
 	if err := b2dutils.CopyIsoToMachineDir(d.Boot2DockerURL, d.MachineName); err != nil {
 		return err
 	}
@@ -274,7 +274,7 @@ func (d *Driver) Create() error {
 		"--port", "0",
 		"--device", "0",
 		"--type", "dvddrive",
-		"--medium", d.ResolveStorePath("boot2docker.iso")); err != nil {
+		"--medium", d.ResolveStorePath(isoFilename)); err != nil {
 		return err
 	}
 
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -1,16 +1,12 @@
 package virtualbox
 
 import (
-	"archive/tar"
-	"bytes"
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"math/rand"
 	"net"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
@@ -205,9 +201,10 @@ func (d *Driver) Create() error {
 		}
 
 		log.Debugf("Creating disk image...")
-		if err := d.generateDiskImage(d.DiskSize); err != nil {
+		if err := vbm("createhd", "--size", fmt.Sprintf("%d", d.DiskSize), "--format", "VMDK", "--filename", d.diskPath()); err != nil {
 			return err
 		}
+
 	}
 
 	if err := vbm("createvm",
@@ -572,30 +569,6 @@ func (d *Driver) diskPath() string {
 	return d.ResolveStorePath("disk.vmdk")
 }
 
-// Make a boot2docker VM disk image.
-func (d *Driver) generateDiskImage(size int) error {
-	log.Debugf("Creating %d MB hard disk image...", size)
-
-	magicString := "data"
-
-	buf := new(bytes.Buffer)
-	tw := tar.NewWriter(buf)
-
-	// magicString first so the automount script knows to format the disk
-	file := &tar.Header{Name: magicString, Size: int64(len(magicString))}
-	if err := tw.WriteHeader(file); err != nil {
-		return err
-	}
-
-	if _, err := tw.Write([]byte(magicString)); err != nil {
-		return err
-	}
-
-	raw := bytes.NewReader(buf.Bytes())
-
-	return createDiskImage(d.diskPath(), size, raw)
-}
-
 func (d *Driver) setupHostOnlyNetwork(machineName string) error {
 	hostOnlyCIDR := d.HostOnlyCIDR
 
@@ -643,69 +616,6 @@ func (d *Driver) setupHostOnlyNetwork(machineName string) error {
 		return err
 	}
 
-	return nil
-}
-
-// createDiskImage makes a disk image at dest with the given size in MB. If r is
-// not nil, it will be read as a raw disk image to convert from.
-func createDiskImage(dest string, size int, r io.Reader) error {
-	// Convert a raw image from stdin to the dest VMDK image.
-	sizeBytes := int64(size) << 20 // usually won't fit in 32-bit int (max 2GB)
-	// FIXME: why isn't this just using the vbm*() functions?
-	cmd := exec.Command(vboxManageCmd, "convertfromraw", "stdin", dest,
-		fmt.Sprintf("%d", sizeBytes), "--format", "VMDK")
-
-	if os.Getenv("DEBUG") != "" {
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-	}
-
-	stdin, err := cmd.StdinPipe()
-	if err != nil {
-		return err
-	}
-	if err := cmd.Start(); err != nil {
-		return err
-	}
-
-	n, err := io.Copy(stdin, r)
-	if err != nil {
-		return err
-	}
-
-	// The total number of bytes written to stdin must match sizeBytes, or
-	// VBoxManage.exe on Windows will fail. Fill remaining with zeros.
-	if left := sizeBytes - n; left > 0 {
-		if err := zeroFill(stdin, left); err != nil {
-			return err
-		}
-	}
-
-	// cmd won't exit until the stdin is closed.
-	if err := stdin.Close(); err != nil {
-		return err
-	}
-
-	return cmd.Wait()
-}
-
-// zeroFill writes n zero bytes into w.
-func zeroFill(w io.Writer, n int64) error {
-	const blocksize = 32 << 10
-	zeros := make([]byte, blocksize)
-	var k int
-	var err error
-	for n > 0 {
-		if n > blocksize {
-			k, err = w.Write(zeros)
-		} else {
-			k, err = w.Write(zeros[:n])
-		}
-		if err != nil {
-			return err
-		}
-		n -= int64(k)
-	}
 	return nil
 }
 
@@ -790,7 +700,7 @@ func (d *Driver) getLocalSSHClient() (ssh.Client, error) {
 }
 
 func translateWindowsMount(p string) (string, error) {
-	re := regexp.MustCompile(`(?P<drive>.+):\\(?P<path>.*)`)
+	re := regexp.MustCompile(`(?P<drive>[^:]+):\\(?P<path>.*)`)
 	m := re.FindStringSubmatch(p)
 
 	var drive, fullPath string
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -19,6 +20,7 @@ import (
 	"time"
 
 	"github.com/codegangsta/cli"
+	"github.com/docker/docker/pkg/homedir"
 	"github.com/docker/machine/drivers"
 	"github.com/docker/machine/log"
 	"github.com/docker/machine/ssh"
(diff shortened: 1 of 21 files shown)


drivers/virtualbox/virtualbox.go
```diff
@@ -288,26 +289,37 @@ func (d *Driver) Create() error {
...
+		username = os.Getenv("USERNAME")
+		shareDir = "c:\\Users\\" + username
+		shareMountDir = "/c/Users"
 	case "darwin":
-		shareName = "Users"
-		shareDir = "/Users"
-		// TODO "linux"
+		username = os.Getenv("USER")
+		shareDir = path.Join("/Users", username)
+		shareMountDir = "/Users"
```
hairyhenderson: Does this mean that `/Users/hairyhenderson` on my host will map to `/Users` in the VM? That kinda breaks things :)

Should this be `shareMountDir = path.Join("/Users/", username)` instead?

tianon: Do you think `HOME` would be better here instead? Maybe this username code as a fallback? 

In my experience, the `USER` environment variable is really unreliable. :cry:

drivers/virtualbox/virtualbox.go
```diff
@@ -336,13 +348,47 @@ func (d *Driver) Create() error {
...
+	if out, err := sshClient.Output(fmt.Sprintf(
+		"mkdir -p /home/%s/.ssh",
+		d.GetSSHUsername(),
+	)); err != nil {
+		log.Error(out)
+		return err
+	}
+
+	if out, err := sshClient.Output(fmt.Sprintf(
+		"printf '%%s' '%s' | sudo tee /home/%s/.ssh/authorized_keys",
```
tianon: Why `sudo` here?

drivers/virtualbox/virtualbox.go
tianon: If we need static compilation, we want something like https://github.com/docker/docker/blob/master/pkg/homedir/homedir.go.  If we don't need static compilation, then [`user.Current`](https://golang.org/pkg/os/user/#Current) is going to be the "proper" solution.

ehazlett: @hairyhenderson no it will map /Users/hairyhenderson -> /Users/hairyhenderson in the VM :)

@tianon lol i even use homdir in another project.  I will update :)  User.Current only works on Linux I believe.

drivers/virtualbox/virtualbox.go
ehazlett: Ah good point.  We should have removed that previously.  Will update.  Thx!

drivers/virtualbox/virtualbox.go
ehazlett: @tianon ah yes.  The reason I use the user env var is to translate the Windows home directory to the msys escaped format.  I'll see if I can fix that.

drivers/virtualbox/virtualbox.go
```diff
@@ -288,26 +289,37 @@ func (d *Driver) Create() error {
...
+		shareMountDir = "/home"
+		shareDir = path.Join("/home", username)
+	}
+
+	shareName := username
+
+	log.Debugf("creating share: path=%s", shareDir)
+
+	// let VBoxService do nice magic automounting (when it's used)
+	if err := vbm("guestproperty", "set", d.MachineName, "/VirtualBox/GuestAdd/SharedFolders/MountDir", shareMountDir); err != nil {
```
tianon: The point of setting this to `/` is that VirtualBox will then take a name like `home/tianon` and mount it at `/home/tianon` automatically.  Setting this to `/home` means that any automount shares added after the machine is created will live under `/home` (so a name of `mnt/files` would automount as `/home/mnt/files` instead of as `/mnt/files`).

drivers/virtualbox/virtualbox.go
tianon: There's a note about precisely that a little further down in this function: :+1:
```go
 				// TODO do some basic Windows -> MSYS path conversion
 				// ie, s!^([a-z]+):[/\\]+!\1/!; s!\\!/!g
```

tianon: Also, see https://golang.org/src/os/user/ (especially `lookup_windows.go`). :+1:

ehazlett: The replace seemed a bit hacky but I guess no different than what I currently have.  I do that.

`User.Current` does not work on Darwin which may be fine since we should be able to just use the path.

I think using the `homedir` and then replacing for msys sounds the best to me.  @tianon Your thoughts?

drivers/virtualbox/virtualbox.go
ehazlett: I think you are mixing that with `MountPrefix`.

drivers/virtualbox/virtualbox.go
tianon: Yeah, that sounds sane to me! :+1:

ehazlett: Cool thanks!

drivers/virtualbox/virtualbox.go
tianon: Perhaps, but don't they both get added together for the final result?  It still feels wrong to be setting this to `/home` or `/Users` since it's at the VM level not the share level, but I'm not sure what exactly the implications are since the docs for both of these are practically nonexistent. :disappointed:

tianon: Aha: http://en.helpdoc-online.com/virtualbox_4.1.2/source/ch04s03.html#sf_mount_auto

> The guest property `/VirtualBox/GuestAdd/SharedFolders/MountPrefix` determines the prefix that is used. Change that guest property to a value other than "sf" to change that prefix; see Section 4.6, “Guest properties” for details.

tianon: I guess https://www.virtualbox.org/manual/ch04.html#sf_mount_auto is the more official source of that content, but the words are the same. :smile:

ehazlett: Right so we can just leave it an empty string then and specify the full path?  I'll try that.

tianon: IIRC, I originally had to set these both to "/" for VirtualBox to not
ignore me, but if empty string actually works I'm all for it! :+1:

libmachine/provision/debian.go
```diff
@@ -44,7 +44,7 @@ func (provisioner *DebianProvisioner) Service(name string, action pkgaction.Serv
 		return err
 	}
 
-	command := fmt.Sprintf("sudo systemctl %s %s", action.String(), name)
+	command := fmt.Sprintf("sudo systemctl %s %s || true", action.String(), name)
```
nathanleclaire: Why this change?

ehazlett: Because systemd was causing issues and exiting at random even though the original issued command exited fine.  @tianon can you shed any more light?

tianon: I wasn't able to reproduce, but @ehazlett claimed `systemctl enable docker` was failing for him with `Failed to execute operation: File exists` because we're already enabled.  When I run that, I get a lot of harmful looking output, but it exits appropriately with a zero exit code. :confused:

tianon:
```console
docker@docker:~$ sudo systemctl enable docker
Synchronizing state for docker.service with sysvinit using update-rc.d...
Executing /usr/sbin/update-rc.d docker defaults
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.utf8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Executing /usr/sbin/update-rc.d docker enable
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.utf8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
docker@docker:~$ echo $?
0
```

tianon: I added a simple fix for the locale warnings, so here's the updated output (much easier to see what's going on):
```console
docker@docker:~$ sudo systemctl enable docker
Synchronizing state for docker.service with sysvinit using update-rc.d...
Executing /usr/sbin/update-rc.d docker defaults
Executing /usr/sbin/update-rc.d docker enable
docker@docker:~$ echo $?
0
```

ehazlett: I'll pull the latest and try to re-create.

ehazlett:
@tianon @nathanleclaire updated.  this now uses `homedir` instead of the env var.  Tested on Mac, Linux and Windows successfully.  PTAL

drivers/virtualbox/virtualbox.go
```diff
@@ -533,7 +576,7 @@ func (d *Driver) diskPath() string {
 func (d *Driver) generateDiskImage(size int) error {
 	log.Debugf("Creating %d MB hard disk image...", size)
 
-	magicString := "boot2docker, please format-me"
+	magicString := "data"
 
 	buf := new(bytes.Buffer)
 	tw := tar.NewWriter(buf)
```
tianon: Couldn't we skip the `tar` completely now and just initialize a blank disk instead?

drivers/virtualbox/virtualbox.go
```diff
@@ -755,3 +775,35 @@ func getRandomIPinSubnet(baseIP net.IP) (net.IP, error) {
...
+	sshClient, err := ssh.NewNativeClient(d.GetSSHUsername(), "127.0.0.1", d.SSHPort, sshAuth)
+	if err != nil {
+		return nil, err
+	}
+
+	return sshClient, nil
+}
+
+func translateWindowsMount(p string) (string, error) {
+	re := regexp.MustCompile(`(?P<drive>.+):\\(?P<path>.*)`)
```
tianon: Do you think the `drive` here should at least be `[^:]+` (instead of the raw `.+` you've got here) to make sure we don't accidentally match past the first `:` to a second `:`?

utils/b2d.go
```diff
@@ -89,8 +90,15 @@ func (b *B2dUtils) GetLatestBoot2DockerReleaseURL() (string, error) {
 		return "", fmt.Errorf("no releases found")
 	}
 
-	tag := t[0].TagName
-	isoUrl := fmt.Sprintf("%s/boot2docker/boot2docker/releases/download/%s/boot2docker.iso", b.githubBaseUrl, tag)
+	// find the latest "released" release (i.e. not pre-release)
+	isoUrl := ""
+	for _, r := range t {
+		if !r.PreRelease {
```
tianon: :heart: :purple_heart: :green_heart: :heart_eyes:

ehazlett: :+1: thx!

drivers/virtualbox/virtualbox.go
ehazlett: Yeah I think so.  I'm not too familiar with the disk creation.  I will give it a shot.  Thanks!

drivers/virtualbox/virtualbox.go
ehazlett: +1

drivers/virtualbox/virtualbox.go
```diff
@@ -543,36 +586,13 @@ func (d *Driver) generateDiskImage(size int) error {
...
-	if _, err := tw.Write([]byte(pubKey)); err != nil {
-		return err
-	}
-	if err := tw.Close(); err != nil {
-		return err
-	}
+
 	raw := bytes.NewReader(buf.Bytes())
+
 	return createDiskImage(d.diskPath(), size, raw)
```
tianon: I think the third parameter here can just be `nil` (https://github.com/docker/machine/blob/95a37e0f1b6a3a58fc97ee3276e8f841708e810f/drivers/virtualbox/virtualbox.go#L629-L631) and it'll work as intended. :+1:

tianon: Except reading that function, it seems like that comment might actually be lying...

tianon: Changing that function to be something more like this would probably be more correct:
```go
// createDiskImage makes a disk image at dest with the given size in MB. If r is
// not nil, it will be read as a raw disk image to convert from.
func createDiskImage(dest string, size int, r io.Reader) error {
	if r == nil {
		cmd := exec.Command(vboxManageCmd, "createhd", "--filename", dest, "--sizebyte", fmt.Sprintf("%d", sizeBytes))
		// ...
		return cmd.Wait()
	}
	// Convert a raw image from stdin to the dest VMDK image.
	sizeBytes := int64(size) << 20 // usually won't fit in 32-bit int (max 2GB)
	// FIXME: why isn't this just using the vbm*() functions?
	// ....
```

tianon: Although `given size in MB.` in that comment is still wrong because the argument to `convertfromraw` is bytes, lol.

tianon: For completeness:
```
  createhd                  --filename <filename>
                            [--size <megabytes>|--sizebyte <bytes>]
                            [--diffparent <uuid>|<filename>
                            [--format VDI|VMDK|VHD] (default: VDI)
                            [--variant Standard,Fixed,Split2G,Stream,ESX]

...

  convertfromraw            stdin <outputfile> <bytes>
                            [--format VDI|VMDK|VHD]
                            [--variant Standard,Fixed,Split2G,Stream,ESX]
                            [--uuid <uuid>]
```

So I guess `--format VMDK` would be important to add too for compatibility?

nathanleclaire:
The integration tests will have to be updated to not reference `overlay` if this is the kernel version we're going to be sticking with --- the kernel isn't old enough :(

ehazlett:
@nathanleclaire +1 -- I'll update

tianon:
If we manage to get a 4.x we will lose aufs and will need to specify
overlay as the default explicitly, so we have a trade-off either way.

nathanleclaire:
> If we manage to get a 4.x we will lose aufs and will need to specify
overlay as the default explicitly, so we have a trade-off either way.

:(

I'm inclined to `aufs` for this release, but at any rate we will still need to update the integration tests as the `--engine-storage-driver overlay` test will no longer work.

ehazlett:
@nathanleclaire already updated with `btrfs`.

nathanleclaire:
Ah, nice @ehazlett

ehazlett:
I've also updated the Fusion driver to use this.  /cc @frapposelli

nathanleclaire:
We should make sure that `hyper-v` doesn't break as well. cc @ahmetalpbalkan

ahmetalpbalkan:
@nathanleclaire I am not a Windows or Hyper-V user. :-) I haven't ever tried machine with hyper-v either.

tianon:
@nathanleclaire https://github.com/boot2docker/boot2docker/commit/3c22cf0924b6b531d68f9c568353115688d43474 just for you :heart:

(please don't hurt me @ehazlett; it's for the greater good! :see_no_evil:)

nathanleclaire:
> @nathanleclaire boot2docker/boot2docker@3c22cf0 just for you :heart:

Ah I thought this was going to be adding the ASCII art whale and I was going to post a very triumphant GIF

ehazlett:
@tianon :+1: :D

ehazlett:
I will attempt the hyper-v updates

tianon:
@nathanleclaire:

```Dockerfile
FROM dockercore/iso

RUN { \
		echo 'VARIANT="Nathan LeClaire Edition"'; \
		echo 'VARIANT_ID=nathanleclaire'; \
	} >> /etc/os-release

COPY whale.txt /etc/motd
RUN echo 'cat /etc/motd' >> /home/docker/.bashrc
```

(untested, but should totally work -- save it as `Dockerfile.nathanleclare` in your checkout of the `/tree/ng` branch and run `make nathanleclare.iso` and it should "just work")

tianon:
... and include https://github.com/boot2docker/boot2docker/blob/6247cc1e73eae6d03f5bfaeadad69b302cbea1eb/rootfs/rootfs/usr/local/etc/motd as `whale.txt` of course

nathanleclaire:
So are we just marking `upgrade` as a big "TODO: BEFORE RELEASE" in this?  If the user goes to `docker-machine upgrade vm` a VM created with this PR, it won't go get the ISO, it will actually try to run `apt-get upgrade` which is what the Debian provisioner wants.

nathanleclaire:
Something to note (we should probably make a separate issue as I am ready to merge this for now):

`upgrade` will be broken for local providers with this change as the provisioner will simply attempt a command like `apt-get upgrade lxc-docker` instead of actually upgrading the ISO.  It will need to be fixed before the release.

Other than that, LGTM

ehazlett:
This now has upgrade support.

ehazlett:
Now with Hyper-V support!

2015-07-24-102014_882x697_scrot

ehazlett:
@tianon just for you ;) ^^

ehazlett:
Fusion shared folders working again.  <3 to @tianon :D

tianon:
@frapposelli any idea why `vmtoolsd` doesn't actually do any amount of automounting?

ehazlett:
@nathanleclaire this should be good to go.  I believe we've closed the last few issues.

frapposelli:
@tianon it does automount but only when you enable/disable shared folders at the hypervisor level.

tianon:
@frapposelli oh, that's bizarre -- no auto-mount on boot for folders already set up then? :disappointed:

libmachine/provision/boot2docker.go
```diff
@@ -53,174 +45,62 @@ func (provisioner *Boot2DockerProvisioner) upgradeIso() error {
...
 	log.Infof("Upgrading machine %s...", machineName)
 
-	b2dutils := utils.NewB2dUtils("", "")
+	isoFilename := ""
+	switch provisioner.GetDriver().DriverName() {
+	case "virtualbox":
+		isoFilename = "boot2docker-virtualbox.iso"
+	case "vmwarefusion", "vmwarevsphere":
+		isoFilename = "boot2docker-vmware.iso"
+	case "hyperv":
```
nathanleclaire: It's actually `hyper-v` (with a hyphen).  We probably should have constants for the driver names for this reason, I think, but we can file a separate issue for that.

nathanleclaire:
LGTM

Issue #1560, opened 2015-07-21
Machine-readable output from driver


timfallmk:
This is tracking issue for an enhancement request. Work in Kinematic (and probably elsewhere) would greatly appreciate an unprocessed list of flags for drivers from `machine`. This is currently being done by scraping the help text, but it would be nice to add as a call to `machine` itself.

I've got a proposal PR for how to do this (as an added -q flag to `create`), but I'm open to suggestions.

It would look something like:
`docker-machine create -d <somedriver> -q`

I'll update with the tracking PR as I go.

dgageot:
Here the ugly code that can do what you want using libmachine in its current state:

```
package main

import (
	"encoding/json"
	"fmt"

	"log"

	"github.com/docker/machine/commands/mcndirs"
	"github.com/docker/machine/libmachine"
	"github.com/docker/machine/libmachine/drivers"
	"github.com/docker/machine/libmachine/mcnflag"
)

func main() {
	api := libmachine.NewClient(mcndirs.GetBaseDir(), mcndirs.GetMachineCertDir())
	defer api.Close()

	rawDriver, err := json.Marshal(&drivers.BaseDriver{})
	if err != nil {
		log.Fatalf("Error attempting to marshal bare driver data: %s", err)
	}

	// Hack to get a driver instance
	h, err := api.NewHost("virtualbox", rawDriver)
	if err != nil {
		log.Fatalf("Error getting new host: %s", err)
	}

	driver := h.Driver

	for _, flag := range driver.GetCreateFlags() {
		switch flag := flag.(type) {
		case *mcnflag.StringFlag:
			fmt.Printf("String %s: %s %s %s\n", flag.Name, flag.Usage, flag.EnvVar, flag.Value)
		case *mcnflag.IntFlag:
			fmt.Printf("Int %s: %s %s %s\n", flag.Name, flag.Usage, flag.EnvVar, flag.Value)
		case *mcnflag.BoolFlag:
			fmt.Printf("Bool %s: %s %s\n", flag.Name, flag.Usage, flag.EnvVar)
		case *mcnflag.StringSliceFlag:
			fmt.Printf("StringSlice %s: %s %s\n", flag.Name, flag.Usage, flag.EnvVar)
		}
	}
}
```

PR #1391, opened 2015-06-19
Give each driver a subcommand

cpuguy83:
Instead of sticking everything into the `create` command, create subcommands for each driver.
Also removes the driver flag's prefixes (which contained the driver name)

- [x] May be an issue with global flags being passed into subcommands codegangsta/cli#191 --- No issue
- [x] ~~Split swarm/engine opts from driver opts in the help output~~ not needed for this, can look at it separately
- [x] Fix help output command format - codegangsta/cli#246

```
# ./machine create --help
NAME:
   machine create - Create a machine

USAGE:
   machine create command [command options] [arguments...]

COMMANDS:
   amazonec2		Create a machine using the amazonec2 driver
   azure		Create a machine using the azure driver
   digitalocean		Create a machine using the digitalocean driver
   exoscale		Create a machine using the exoscale driver
   generic		Create a machine using the generic driver
   google		Create a machine using the google driver
   none			Create a machine using the none driver
   openstack		Create a machine using the openstack driver
   rackspace		Create a machine using the rackspace driver
   softlayer		Create a machine using the softlayer driver
   virtualbox		Create a machine using the virtualbox driver
   vmwarevcloudair	Create a machine using the vmwarevcloudair driver
   vmwarevsphere	Create a machine using the vmwarevsphere driver
   help, h		Shows a list of commands or help for one command

OPTIONS:
   --help, -h	show help
```

```
# ./machine create virtualbox -help
Usage: docker-machine create virtualbox [OPTIONS] [arg...]

Create a machine using the virtualbox driver

Options:

   --boot2docker-url 											The URL of the boot2docker image. Defaults to the latest available version [$VIRTUALBOX_BOOT2DOCKER_URL]
   --cpu-count "1"											number of CPUs for the machine (-1 to use the number of CPUs available) [$VIRTUALBOX_CPU_COUNT]
   --disk-size "20000"											Size of disk for host in MB [$VIRTUALBOX_DISK_SIZE]
   --hostonly-cidr "192.168.99.1/24"									Specify the Host Only CIDR [$VIRTUALBOX_HOSTONLY_CIDR]
   --import-boot2docker-vm 										The name of a Boot2Docker VM to import
   --memory "1024"											Size of memory for host in MB [$VIRTUALBOX_MEMORY_SIZE]
   --engine-install-url "https://get.docker.com"							Custom URL to use for engine installation [$MACHINE_DOCKER_INSTALL_URL]
[...]

PR #1561, opened 2015-07-22
Driver flags

timfallmk:
Working PR for issue #1560 .

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -235,6 +235,11 @@ var sharedCreateFlags = []cli.Flag{
 		Usage: "addr to advertise for Swarm (default: detect and use the machine IP)",
 		Value: "",
 	},
+	cli.StringFlag{
+                Name: "print-flags, p",
+                Usage: "Raw output of the flags available for a driver",
+                Value: "",
+        },
 }
 
 var Commands = []cli.Command{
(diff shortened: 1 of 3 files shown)


timfallmk:
Anyone have a neat trick for accessing nested values in the `[]cli.Flags` interface? I need to get `cli.StringFlag{Name}`.

cpuguy83:
Is this necessary with #1391 ?

timfallmk:
@cpuguy83 Hadn't seen you PR, would have saved me a lot of swearing at `create.go` for the way `[]cli.Flags` was used :hurtrealbad: .

Can I make a request that there be an option for non-pretty-print output? Something that's more API consumable, like `-q`?

If you'd like I'll blow this all away and rebase off you're PR to add it to that.

ehazlett:
Thanks for the contribution but I think we would go with #1391 in favor of this one.  Please feel free to comment on that one if you would like to see something in it.  Thanks!

timfallmk:
@ehazlett That's fine. I don't think they're necessarily conflicting, this was just to add an APIs consumable way to get flags for drivers. That would still be necessary given #1391 .

FYI, this is necessary for ongoing work with kitematic. Right now we have to distributed our own forked `machine` binary.

Issue #1537, opened 2015-07-16
machine 0.3.1-rc1 panic when running commands


mbentley:
Environment:
 - OS X
 - Previously running docker-machine 0.3.0; upgraded to `docker-machine version 0.3.1-rc1 (993f2db)`
 - Drivers in use: `amazonec2`, `softlayer`, `virtualbox`, and `none`

I am now getting panics anytime I run any command that reads my `amazonec2` or `none` based configs, such as `docker-machine ls` or `docker-machine ssh <instance-name>`.  I can do a `docker-machine ssh` successfully to instances utilizing the `softlayer` and `virtualbox` drivers but not `amazonec2` and `none` drivers.

My config files (with secrets removed)
https://gist.github.com/mbentley/40fba8ae44bdfe39327c

These configs have been around for quite some time and I know that @ehazlett and I did some hacking to migrate them around like 0.1.0 or so.

Output from `amazonec2` driver when attempting to `docker-machine ssh`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x1697e1]

goroutine 1 [running]:
github.com/docker/machine/libmachine.(*Host).LoadConfig(0x208c80320, 0x0, 0x0)
        /go/src/github.com/docker/machine/libmachine/host.go:287 +0x3d1
github.com/docker/machine/libmachine.Filestore.loadHost(0x208bf2f20, 0x1f, 0x208c404e0, 0x2c, 0x208c405a0, 0x30, 0x7fff5fbffb0b, 0xd, 0x8, 0x0, ...)
        /go/src/github.com/docker/machine/libmachine/filestore.go:34 +0x287
github.com/docker/machine/libmachine.Filestore.Get(0x208bf2f20, 0x1f, 0x208c404e0, 0x2c, 0x208c405a0, 0x30, 0x7fff5fbffb0b, 0xd, 0x667fc0, 0x0, ...)
        /go/src/github.com/docker/machine/libmachine/filestore.go:112 +0x65
github.com/docker/machine/libmachine.(*Filestore).Get(0x208c40750, 0x7fff5fbffb0b, 0xd, 0x208c8caf0, 0x0, 0x0)
        <autogenerated>:10 +0xcc
github.com/docker/machine/libmachine.(*Provider).Get(0x208c8caf0, 0x7fff5fbffb0b, 0xd, 0x0, 0x0, 0x0)
[...]

PR #1564, opened 2015-07-22
Revert configuration migration removal, include new version field, and introduce migration boilerplate

nathanleclaire:
cc @ehazlett @cpuguy83 

Closes https://github.com/docker/machine/issues/1537.
This PR reverts the removal of the "flat => nested" config migration and introduces a new field for versioning configs (it is independent of the Docker Machine version number).  I was going to skip it but I have received a few reports of panics using 0.3.1 now and I'd rather be safe than sorry, and longer-term move to a more smoothly architected config migration process.

__EDIT__: I also changed the "constants" in `version.go` to be `MixedCase` which is idiomatic for Go, even for constants.

### DIFF
--- a/drivers/openstack/client.go
+++ b/drivers/openstack/client.go
@@ -437,7 +437,7 @@ func (c *GenericClient) Authenticate(d *Driver) error {
 		return err
 	}
 
-	provider.UserAgent.Prepend(fmt.Sprintf("docker-machine/v%s", version.VERSION))
+	provider.UserAgent.Prepend(fmt.Sprintf("docker-machine/v%s", version.Version))
 
 	if d.Insecure {
 		// Configure custom TLS settings.
--- a/libmachine/filestore.go
+++ b/libmachine/filestore.go
@@ -19,7 +19,11 @@ type Filestore struct {
 }
 
 func NewFilestore(rootPath string, caCert string, privateKey string) *Filestore {
-	return &Filestore{path: rootPath, caCertPath: caCert, privateKeyPath: privateKey}
+	return &Filestore{
+		path:           rootPath,
+		caCertPath:     caCert,
+		privateKeyPath: privateKey,
+	}
 }
 
 func (s Filestore) loadHost(name string) (*Host, error) {
@@ -30,13 +34,15 @@ func (s Filestore) loadHost(name string) (*Host, error) {
 		}
 	}
 
-	host := &Host{Name: name, StorePath: hostPath}
+	host := &Host{
+		Name:      name,
+		StorePath: hostPath,
+	}
 	if err := host.LoadConfig(); err != nil {
 		return nil, err
 	}
 
-	h := FillNestedHost(host)
-	return h, nil
+	return host, nil
 }
 
 func (s Filestore) GetPath() string {
@@ -85,7 +91,7 @@ func (s Filestore) List() ([]*Host, error) {
 
 	for _, file := range dir {
 		// don't load hidden dirs; used for configs
-		if file.IsDir() && strings.Index(file.Name(), ".") != 0 {
+		if file.IsDir() && !strings.HasPrefix(file.Name(), ".") {
 			host, err := s.Get(file.Name())
 			if err != nil {
 				log.Errorf("error loading host %q: %s", file.Name(), err)
--- a/libmachine/filestore.go
+++ b/libmachine/filestore.go
@@ -35,7 +35,8 @@ func (s Filestore) loadHost(name string) (*Host, error) {
 		return nil, err
 	}
 
-	return host, nil
+	h := FillNestedHost(host)
+	return h, nil
 }
 
 func (s Filestore) GetPath() string {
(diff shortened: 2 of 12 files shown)


cpuguy83:
It's probably worthwhile to have versioned struct definitions and provide a migration between versions.

ehazlett:
@cpuguy83 +1 -- @nathanleclaire do you want to do that for this one?

nathanleclaire:
Yes, I'd much rather do it now than later on.

nathanleclaire:
OK, updated with versioned struct-based migration boilerplate to address @cpuguy83 concerns.

It's a little rough, but I think will pay off in the future.

@ehazlett @cpuguy83  Please take a look as soon as you get a chance.

Still to do:

- [ ] Write integration test for this functionality
- [ ] Write and fix unit tests around this functionality

drivers/openstack/client.go
```diff
@@ -437,7 +437,7 @@ func (c *GenericClient) Authenticate(d *Driver) error {
 		return err
 	}
 
-	provider.UserAgent.Prepend(fmt.Sprintf("docker-machine/v%s", version.VERSION))
+	provider.UserAgent.Prepend(fmt.Sprintf("docker-machine/v%s", version.Version))
```
nathanleclaire: I also changed the name of these versions variables because ALL_CAPS is not idiomatic to Go, even for constants.

libmachine/filestore.go
```diff
@@ -19,7 +19,11 @@ type Filestore struct {
 }
 
 func NewFilestore(rootPath string, caCert string, privateKey string) *Filestore {
-	return &Filestore{path: rootPath, caCertPath: caCert, privateKeyPath: privateKey}
+	return &Filestore{
```
nathanleclaire: Spacing pedantism

libmachine/filestore.go
```diff
@@ -30,7 +34,10 @@ func (s Filestore) loadHost(name string) (*Host, error) {
 		}
 	}
 
-	host := &Host{Name: name, StorePath: hostPath}
+	host := &Host{
```
nathanleclaire: More spacing pedantism

libmachine/filestore.go
```diff
@@ -84,7 +91,7 @@ func (s Filestore) List() ([]*Host, error) {
 
 	for _, file := range dir {
 		// don't load hidden dirs; used for configs
-		if file.IsDir() && strings.Index(file.Name(), ".") != 0 {
+		if file.IsDir() && !strings.HasPrefix(file.Name(), ".") {
```
nathanleclaire: This is simply a cleanup and not directly relevant to the migration code

libmachine/host.go
```diff
@@ -276,24 +283,28 @@ func (h *Host) LoadConfig() error {
...
 
-	driver, err := drivers.NewDriver(hostMetadata.DriverName, h.Name, h.StorePath, authOptions.CaCertPath, authOptions.PrivateKeyPath)
+	migratedHost, err := MigrateHost(h, data)
 	if err != nil {
-		return err
+		return fmt.Errorf("Error getting migrated host: %s", err)
 	}
 
-	h.Driver = driver
+	*h = *migratedHost
```
nathanleclaire: This seems a little ugly / dangerous to me, but I'm not sure if there's a better way unless `LoadConfig` was an external function and not a member of `Host`.

libmachine/migrate.go
```diff
@@ -0,0 +1,70 @@
...
+		authOptions.CaCertPath,
+		authOptions.PrivateKeyPath,
+	)
+	if err != nil {
+		return &Host{}, err
+	}
+
+	for h.ConfigVersion = migratedHostMetadata.ConfigVersion; h.ConfigVersion < version.ConfigVersion; h.ConfigVersion++ {
+		switch h.ConfigVersion {
+		case 0:
```
nathanleclaire: It may not be immediately obvious, but in the case of future migrations, we will simply move the current `Host` to, say, `HostV1`, add another `case` in this `switch` block here, and add the relevant migration functions.

libmachine/filestore.go
```diff
@@ -19,7 +19,11 @@ type Filestore struct {
 }
 
 func NewFilestore(rootPath string, caCert string, privateKey string) *Filestore {
-	return &Filestore{path: rootPath, caCertPath: caCert, privateKeyPath: privateKey}
+	return &Filestore{
+		path:           rootPath,
```
nathanleclaire: Side note:  Why do even have a constructor function if all it does it wrap a struct initialization directly?  ;D

ehazlett: Good question.  I think it was in case there was anything we wanted to change or initialize in the future if we moved to a different model.  I'm OK removing.

drivers/openstack/client.go
ehazlett: +1

ehazlett:
Looks good.  Do we want to have a follow up PR for docs/tests so we can get this into an RC2 for testing?

nathanleclaire:
> Looks good. Do we want to have a follow up PR for docs/tests so we can get this into an RC2 for testing?

Yeah, I'm happy to do so if it means we can get this into peoples' hands faster.

ehazlett:
LGTM

Issue #1334, opened 2015-06-08
Swarm: allow specifying listen port


ehazlett:
Allow the Swarm port to be configurable.

errordeveloper:
One of the usecase is for those who might wish to use [Weave proxy](http://docs.weave.works/weave/latest_release/proxy.html), which currently [works well](http://blog.weave.works/2015/05/27/multi-host-docker-deployment-with-swarm-and-compose-using-weave-0-11/), yet requires quite a bit of [custom handling](https://github.com/errordeveloper/weave-demos/blob/c6c6269d97de700c0954fc7b5df90c02e3f149ae/quartet/scripts/setup-cluster-simplified.sh).

nathanleclaire:
:+1: 

The longer we put this off, the more difficult it will get.

aperepel:
I took a quick look, but want to get clarity on the scope and expectations here:

* Swarm master is picking up the e.g. `--swarm-host tcp://0.0.0.0:3380` property correctly and stands it up at the specified port (latest dev builds), is there a problem with the way it works or am I looking at the wrong port assignment?
* The above command doesn't open up the new firewall rule correctly - I think this needs to propagate and update firewall rules. Today the code checks for the rule name only, and doesn't go deeper to check which ports are actually being open - a relatively straightforward way would be to wipe the firewall rule and re-create it (it adds both docker and swarm ports). If that's ok, we could add a note in the docs and warn the user about adding any custom ports to this rule (e.g. create your own rule, as this one will be wiped automatically)
* Finally, there's quite a drift in driver quality across providers, and I can't imagine implementing the logic for every possible backend. Any comment on this? I could keep hacking away around GCE and Azure otherwise.

errordeveloper:
The issue with `--swarm-host tcp://0.0.0.0:3380` is that you really should be specifying `--swarm-host tcp://_IP_OF_THIS_MACHINE_I_DONT_KNOW_YET_:3380`, which you cannot and really why should you have to, unless it's gonna be a DNS name, perhaps. So we just need to be able to set the port.

aperepel:
Is it a real issue? It's just that `0.0.0.0` instructs it to bind to all interfaces, external included (which we want, don't we?), and the actual in will be assigned once the vm is up. In that context we don't even need the ip part, as we always know which node is a swarm master. Is my understanding correct or is there anything else I might be missing?

errordeveloper:
[...]

PR #1574, opened 2015-07-23
add environment variable DOCKER_PORT to customize docker daemon port

TrumanLing:
add environment variable DOCKER_PORT to customize docker daemon port

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hairyhenderson:
Thanks for the PR, @TrumanLing!

There's been a bunch of talk in other issues and PRs about allowing customization of the engine port (though I can't find any specific issues right now).

While this would be useful, the change needs to be for more than just the `openstack` driver ;)

Also, we need to expose this as a command line flag - something like `--engine-port`. Another thing to keep in mind would be the Swarm port, and how to customize that. There's talk about that in #1334.

ehazlett:
Thanks @TrumanLing As @hairyhenderson said we want to do this for all of machine.  I'm going to add this to the 0.5 release.

ehazlett:
I'm going to close this in favor of #1334 which would cover this case.  Thanks for the contribution and please see that issue for progress.

Issue #1569, opened 2015-07-22
Non root ssh create sudo can't get password


pdxjohnny:
Sudo tty password
---

I saw this but i don't know if it relates
https://github.com/ehazlett/machine/commit/b43a72f91aad5f3067ba406e54534bdd4b834fb4
maybe I don't know how the allocate the tty

My issue is that when using the generic driver to create a machine I do:

```bash
docker-machine -D create --driver generic \
  --generic-ip-address=example_host.example.com --generic-ssh-user=john \
  --generic-ssh-key=/home/john/.ssh/id_rsa.pub example_host
```
machine is able to log into example_host.example.com as john with the correct
key file, however it then attempts to run

```bash
sudo hostname example_host && echo "example_host" | sudo tee /etc/hostname
```

To which sudo replies
```bash
sudo: no tty present and no askpass program specified
```

when sudo trys to ask for john's password

Here is the full debug:

```bash
Importing SSH key...
IP: example_host.example.com
Getting to WaitForSSH function...
Testing TCP connection to: example_host.example.com:22
Using SSH client type: external
About to run SSH command:
exit 0
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /home/john/.docker/machine/machines/example_host/id_rsa -p 22 john@example_host.example.com exit 0] []  <nil> <nil> <nil> [] <nil> <nil> <nil> ?reflect.Value? false [] [] [] [] <nil>}
SSH cmd err, output: <nil>:
Using SSH client type: external
About to run SSH command:
cat /etc/os-release
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /home/john/.docker/machine/machines/example_host/id_rsa -p 22 john@example_host.example.com cat /etc/os-release] []  <nil> <nil> <nil> [] <nil> <nil> <nil> ?reflect.Value? false [] [] [] [] <nil>}
SSH cmd err, output: <nil>: NAME="Ubuntu"
VERSION="14.04.1 LTS, Trusty Tahr"
[...]

PR #1584, opened 2015-07-25
SSH password issue

pdxjohnny:
I fixed an issue I had where there was no way for me to use docker machine as a non root user because I would get an error saying that sudo had no tty or askpass present https://github.com/docker/machine/issues/1569

I added an option for the generic driver of generic-ssh-pass which is the password of the generic-ssh-user. I modified all the ssh commands that involve sudo (I'm pretty sure I got them all) to pipe the password to `sudo -S`

I was able to fix my issue and docker compose now works with sudo because I can pass it my password

### DIFF
--- a/drivers/base.go
+++ b/drivers/base.go
@@ -1,6 +1,9 @@
 package drivers
 
-import "path/filepath"
+import (
+	"path/filepath"
+	"fmt"
+)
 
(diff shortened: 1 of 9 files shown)


pdxjohnny:
I had formatting errors I need to re run the coverage and such

PR #1586, opened 2015-07-25
Added generic-ssh-pass and modified sudo commands

pdxjohnny:
I found and fixed an issue I had where there was no way for me to use docker machine as a non root user because I would get an error saying that sudo had no tty or askpass present #1569

I added an option for the generic driver of generic-ssh-pass which is the password of the generic-ssh-user. I modified all the ssh commands that involve sudo (I'm pretty sure I got them all) to pipe the password to sudo -S

I was able to fix my issue and docker compose now works with sudo because I can pass it my password

If you see a bunch of pull requests from me its because I messed up a few times i forgo to run the formatter and stuff but this one should be good.

Signed-off-by: John Andersen <johnandersenpdx@gmail.com>

### DIFF
--- a/commands/scp_test.go
+++ b/commands/scp_test.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"os/exec"
 	"testing"
 
@@ -12,6 +13,7 @@ type MockHostInfo struct {
 	ip          string
 	sshUsername string
 	sshKeyPath  string
+	sshPass     string
 }
 
 func (h *MockHostInfo) GetMachineName() string {
@@ -24,6 +26,27 @@ func (h *MockHostInfo) GetIP() (string, error) {
 
 func (h *MockHostInfo) GetSSHUsername() string {
 	return h.sshUsername
+}
+
+func (h *MockHostInfo) GetSSHPassword() string {
+	return h.sshPass
+}
+
+func (h *MockHostInfo) SSHSudo(command string) string {
+	sudo := "sudo"
+	// If there is a password given pipe it to sudo to avoid no tty error
+	if h.sshPass != "" {
+		sudo = fmt.Sprintf(
+			"echo %q | sudo -SE",
+			h.sshPass,
+		)
+	}
+	command = fmt.Sprintf(
+		"%s %s",
+		sudo,
+		command,
+	)
+	return command
 }
 
 func (h *MockHostInfo) GetSSHKeyPath() string {
--- a/drivers/errdriver/error.go
+++ b/drivers/errdriver/error.go
@@ -70,6 +70,14 @@ func (d *Driver) GetSSHUsername() string {
 	return ""
 }
 
+func (d *Driver) GetSSHPassword() string {
+	return ""
+}
+
+func (d *Driver) SSHSudo(string) string {
+	return ""
+}
+
 func (d *Driver) GetState() (state.State, error) {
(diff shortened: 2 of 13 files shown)


osuka:
This is still happening to me, any chance of this getting merged? Otherwise the generic driver is pretty useless...

pdxjohnny:
I agree ^

britishbadger:
It would be nice if this could be merged in, as I still have the same issue with 0.5.0-rc3

pdxjohnny:
I will try to update this to master sometime soonish, within the next two weeks

pdxjohnny:
this is not finished I just had to push my changes up so I can work form home

pimpim:
This is definitely not working in the 1.9.0 release. Seems a major issue as this means that running docker-machine on local linux machines running docker is not possible.

pdxjohnny:
I have not tried to use this so If someone can please build and try to use the pass option let me know if it is working. I don't have time right now and the faster this can get merged the better. I'm not a fan of fixing merge conflicts so please help me test this!

jaycyb:
+1

dgageot:
Hi @pdxjohnny one base principle of docker-machine is that the ssh access should be root. It makes it possible to provision docker ourself and not rely on a docker installed from an unofficial source. This PR seems to go against this behaviour unless I got it wrong. @nathanleclaire @jeanlaurent wdyt?

nathanleclaire:
> Hi @pdxjohnny one base principle of docker-machine is that the ssh access should be root.

More or less -- so far it's been implicitly implied that the user we are SSHing into the machine can use `sudo` privileges without needing to enter a password (which root is a subset of).  For instance the default AMI user is `ubuntu`, not `root`, and can `sudo` without a password.  Likewise for `docker` user in boot2docker.

At any rate, we can't merge this in without at least an API version bump (painful).  It modifies methods on the RPC side which will break existing plugins.

pdxjohnny:
Understandable, Let me know if there is interest in the future otherwise
the fix for this is to use NOPASSWD for the user we are sshing with in the
suders file. But that should be documented, I don't think it is.

nathanleclaire:
I wonder if entering SSH password over and over might be supportable by something like this:

```console
$ yes my-ssh-pass | docker-machine create -d generic --generic-ssh-key /path/to/passwd/protectedkey foobar
```

nathanleclaire:
> the fix for this is to use NOPASSWD for the user we are sshing with in the
suders file. But that should be documented, I don't think it is.

Yes, I definitely think that would be a good addition to the documentation.

pdxjohnny:
I think NOPASSWD is the only way without changes to machine.
The `yes` pipe is essentially the same thing this patch does. It just pipes the password to sudo each time sudo is run.

So ya if there's interest in this as the fix I'll update it to master, otherwise NOPASSWD is the kind of fix.

marcellodesales:
@natemcmaster when will this be merged?

dgageot:
@marcellodesales @pdxjohnny I guess there is no plan to merge it soon since it breaks all existing plugins which is painful. @pdxjohnny are you ok if we only update the documentation for the generic driver?

pdxjohnny:
Ya that sounds good to me.

dgageot:
@pdxjohnny see https://github.com/docker/machine/pull/3085

pdxjohnny:
Closing because the update to documentation (#3085) is a sufficient fix to make the generic driver work.

PR #1597, opened 2015-07-27
Add Timeout state to ls command

nathanleclaire:
Today, if an `ls` command does not return or takes a really long time, it will hang the whole process waiting for the wonky machine.  This introduces a per-goroutine timeout to ensure that `ls` always returns (relatively) speedily, and indicates "Timeout" issues when they occur.

### DIFF
--- a/libmachine/host.go
+++ b/libmachine/host.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"time"
 
 	"github.com/docker/machine/drivers"
 	"github.com/docker/machine/libmachine/auth"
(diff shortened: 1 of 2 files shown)


hairyhenderson:
:+1:. I wonder if 1s is enough for all cases or if maybe it'd be a good idea to add a flag to change the timeout... Though it sort of seems that if a machine takes >1s something's very wrong ;)

ehazlett:
Yeah +1 to @hairyhenderson -- I've seen several timeouts greater than 1s just trying to reach slow providers.

nathanleclaire:
Yeah, 1 second is pretty aggressive, however I'd like to keep it pretty tight as one of the goals of this PR is to prevent non-responsive machines from freezing the UI up.  3 seconds?

hairyhenderson:
I'm good with 3s :+1:

ehazlett:
LGTM

Issue #1518, opened 2015-07-13
Machine state not preserved across reboots


arun-gupta:
Created a couple of machines as:

```
NAME            ACTIVE   DRIVER       STATE     URL                         SWARM
lab                      virtualbox   Running   tcp://192.168.99.101:2376   
summit2015               virtualbox   Running   tcp://192.168.99.100:2376   
```

Restarted the host machine, and now the status is shown as:

```
NAME            ACTIVE   DRIVER       STATE     URL   SWARM
lab                      virtualbox   Stopped         
summit2015               virtualbox   Stopped         
```

State of the machine is not preserved across host reboot.

cgmartin:
+1, same issue. Is there a workaround?

nathanleclaire:
Would you expect them to be running @arun-gupta ?  Generally I'd think it's pretty implicit in power-cycling your host machine that all of your VMs will be turned off as well.

arun-gupta:
If a host power cycles, then it cannot be relied upon to be part of the cluster in that case. That seems like a broken model. Or does that need to be configured somewhere?

nathanleclaire:
> If a host power cycles, then it cannot be relied upon to be part of the cluster in that case. That seems like a broken model. Or does that need to be configured somewhere?

I'm confused.  If you power-cycle a laptop which has VMs, there is no way for them to continue running, so this behavior seems normal to me.  If you power-cycle a laptop with references to cloud VMs, they will keep running just fine.  I would not really expect users to have their laptop as a node in a production cluster.

arun-gupta:
I guess the host in the cloud will "never" go through power recycle?

nathanleclaire:
> I guess the host in the cloud will "never" go through power recycle?

It is likely that the host computer will have to be rebooted at some point, but if all that it contains are references to cloud VMs, those will still keep running.  Machine does not do any magic related to stopping and starting of the hosts it manages on a host reboot.

arun-gupta:
I mean the host that really hosts the VM itself, not the host with just reference to those VMs.

buster:
Hi,

i am not sure if this is the correct ticket but it is referenced in https://github.com/docker/swarm/issues/668 .

As from the original ticket:
- I have docker-machine with virtualbox: 1 swarm master and 2 agents
- when i run docker-machine stop and start afterwards, i can see them running in "docker-machine ls" but they appear to not be part of the cluster anymore

It was mentioned that this is some TLS issue, but i don't know how to work around it.
[...]

PR #1598, opened 2015-07-27
Add docker-machine provision command

nathanleclaire:
This allows users to re-run provisioning on an existing Docker Machine instance.  If something blows up during the ordinary create or provisioning process, this allows users to potentially salvage the machine using `docker-machine provision`.

TODO:

- [ ] Add docs
- [ ] Add tests

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -331,6 +331,12 @@ var Commands = []cli.Command{
 		Action: cmdLs,
 	},
 	{
+		Name:        "provision",
+		Usage:       "Run standard Machine provisioning on one or more hosts",
+		Description: "Argument(s) are one or more machine names.",
+		Action:      cmdProvision,
+	},
+	{
 		Name:        "regenerate-certs",
 		Usage:       "Regenerate TLS Certificates for a machine",
 		Description: "Argument(s) are one or more machine names.",
@@ -413,8 +419,10 @@ var Commands = []cli.Command{
 // machineCommand maps the command name to the corresponding machine command.
 // We run commands concurrently and communicate back an error if there was one.
 func machineCommand(actionName string, host *libmachine.Host, errorChan chan<- error) {
+	// TODO: These actions should have their own type.
 	commands := map[string](func() error){
 		"configureAuth": host.ConfigureAuth,
+		"provision":     host.Provision,
 		"start":         host.Start,
 		"stop":          host.Stop,
 		"restart":       host.Restart,
--- a/commands/provision.go
+++ b/commands/provision.go
@@ -1,0 +1,13 @@
+package commands
+
+import (
+	"log"
+
+	"github.com/codegangsta/cli"
+)
+
+func cmdProvision(c *cli.Context) {
+	if err := runActionWithContext("provision", c); err != nil {
+		log.Fatal(err)
+	}
+}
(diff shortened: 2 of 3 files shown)


abronan:
Can this command help somehow with #1518 by re-provisioning Swarm agents and circumvent the certs issues on reboot? Although we lose all the containers running on the machine on the re-provisioning process right?

/cc @ehazlett

hairyhenderson:
:+1:. What's the expectation for how solid this will be? I'm guessing there will be a few weird cases where the previous provisioning died in an unsalvageable way and `provision` won't succeed. Is there a "normal" use case for this? i.e. any reason a user would want to use `provision` on a host where provisioning succeeded?

nathanleclaire:
> Can this command help somehow with #1518 by re-provisioning Swarm agents and circumvent the certs issues on reboot?

Ideally, yes.  One of my goals with this direction is to make re-provisioning of your Swarm system's state easier.  As you're aware, right now if you do something like reboot a machine where containers from `--swarm` etc. were running, they are gone and you are hosed (not to mention if you accidentally remove them manually).

> Although we lose all the containers running on the machine on the re-provisioning process right?

Yes, although in the future I'd like to consider making Machine capable of "elegant" restarts where it will remember which containers are running and restart them in the correct order on such an operation.

> What's the expectation for how solid this will be? I'm guessing there will be a few weird cases where the previous provisioning died in an unsalvageable way and provision won't succeed.

There probably are cases where such is the case, and this is why tools like Terraform have a notion of a "tainted" resource (it's impossible to establish whether or not it can be salvaged).  Longer-termish, I'd like to make provisioning as idempotent of an operation as possible (it's not right now), so that you could reasonably be assured that by calling the same transform on the machine over and over you are getting the same result.

> Is there a "normal" use case for this? i.e. any reason a user would want to use provision on a host where provisioning succeeded?

Oh, definitely -- it won't be super slick and usable right away, but medium-termish this type of command could be used to, say, update the `EngineOptions` and re-apply the changes to the instance.

hairyhenderson:
> I'd like to make provisioning as idempotent of an operation as possible

:+1: - would love to make this a less long-termish goal. Are there any issues logged about this yet? It'd be useful to flesh out all the different reasons why we don't have provisioning idempotence yet.

> this type of command could be used to, say, update the `EngineOptions` and re-apply the changes to the instance.

Sort of like `docker-machine apply`? This would be totally awesome.

abronan:
Agreed, SGTM! Makes me think of a Vagrant/Ansible workflow where you first do a `vagrant up` to spawn your VMs, running an Ansible playbook but you can still re-run your playbook manually after your VMs are provisioned. This might be useful in many cases.

nathanleclaire:
> :+1: - would love to make this a less long-termish goal. Are there any issues logged about this yet? It'd be useful to flesh out all the different reasons why we don't have provisioning idempotence yet.

Nope, but feel free to go ahead and do so!

> Sort of like docker-machine apply? This would be totally awesome.

Yeah, basically.

hairyhenderson:
> Nope, but feel free to go ahead and do so!

@nathanleclaire - done - #1605 ;)

efrecon:
Ah. This is what my proposed #1766 more or less is then. Cool that it's on the roadmap.

nathanleclaire:
This is a little outdated and I have a more up to date one in the works

PR #1601, opened 2015-07-28
Fix docs links

fixe:


ehazlett:
Thanks for the submission.  Would you mind signing your commits? (https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work)

/cc @moxiegirl

moxiegirl:
@fixe Thank you for contribution *but* your changes actually break the links in the generated static site.  When you make a change to the documentation, you need to test it with `make docs`  --- I see I need to add the readme explaining this.

fixe:
@moxiegirl so the reason I submitted the PR is accessing the docs through GitHub is redirecting to a 404 page. Can we implement something that will work on both sides?

moxiegirl:
@fixe You aren't the first person to make this attempt. Unfortunately, no we can't do this.  GitHub content even when its md aren't meant to be served as a web.   Moreover, our `docs` content is meant to be served from our branded docs.docker.com site which has to anticipate archiving docs sets and a lot more as Docker grows.

fixe:
You pre-process the documentation and have it work on both sides but that's your call.

moxiegirl:
There aren't a lot of large-scale open source projects doing that kind of set up.  Ruby, node.js, MongoDb --- they aren't doing that.  The staff, the development, and the on-going work to maintain that infrastructure typically isn't something companies are willing to spend on.  GitHub itself isn't doing it -- they have GitHub pages.

Issue #1596, opened 2015-07-27
/etc/os-release file warning on creating a machine OSX


moxiegirl:
```
$ sudo docker-machine create --driver virtualbox docker-vm
Password:
Creating CA: /Users/mary/.docker/machine/certs/ca.pem
Creating client certificate: /Users/mary/.docker/machine/certs/cert.pem
Creating VirtualBox VM...
Creating SSH key...
Starting VirtualBox VM...
Starting VM...
Error creating machine: Error parsing /etc/os-release file: Expected bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8) to split by '=' char into two strings, instead got 1 strings
You will want to check the provider to make sure the machine and associated resources were properly removed.
```

legal90:
Looks similar to GH-1481

Issue #1481, opened 2015-07-07
error creating digitalocean machine when on ko_KR.UTF-8 locale


difro:
I was trying to create digitalocean machine on a ubuntu machine which has ko_KR.UTF-8 locale.

Following error occurred.
```
Creating CA: /home/jihoonc/.docker/machine/certs/ca.pem
Creating client certificate: /home/jihoonc/.docker/machine/certs/cert.pem
Creating SSH key...
Creating Digital Ocean droplet...
Error creating machine: Error parsing /etc/os-release file: Expected bash: warning: setlocale: LC_ALL: cannot change locale (ko_KR.UTF-8) to split by '=' char into two strings, instead got 1 strings
You will want to check the provider to make sure the machine and associated resources were properly removed
```

difro:
same error happens on google-cloud driver.
```
...
Saving token in /home/jihoonc/.docker/machine/machines/gobox/gce_token
Creating host...
Generating SSH Key
Creating instance.
Creating firewall rule.
Waiting for Instance...
Uploading SSH Key
Waiting for SSH Key
Error creating machine: Error parsing /etc/os-release file: Expected bash: warning: setlocale: LC_ALL: cannot change locale (ko_KR.UTF-8) to split by '=' char into two strings, instead got 1 strings
```

difro:
prepending ```LC_ALL=C``` seems to fix the issue.
```LC_ALL=C docker-machine create --driver digitalocean ....```

I believe the culprit is this line in my host's ```/etc/ssh/ssh_config```.
```    SendEnv LANG LC_* ```

**Comment by pandrew** [comment]
yeah, got this issue on latest 0.4.0-rc1 on osx. Commented the line in /etc/ssh_config and the issue disappeared.

**Comment by legal90** [comment]
@tianon Is not it an issue of boot2docker NG? 
I got the similar warning after `docker-machine ssh <vm_name>` on OS X host:
```
$ docker-machine ssh b2d-prl-test

Docker version 1.7.1, build 786b29d

Last login: Wed Jul 29 16:13:32 2015 from 10.211.55.2
-bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
-bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
-bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)


docker@docker:~$ locale
locale: Cannot set LC_CTYPE to default locale: No such file or directory
[...]

PR #1449, opened 2015-07-02
Adding provisioner for Arch Linux

hairyhenderson:
Fixes #1293 

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

PR #1602, opened 2015-07-28
Skip and warn on wonky /etc/os-release lines

nathanleclaire:
Closes https://github.com/docker/machine/issues/1596
Closes https://github.com/docker/machine/issues/1481

If the lines don't split cleanly (occasionally STDERR gets mixed in, for
instance, due to our current SSH output setup), we should simply
log.Warn in the output instead of bailing completely.

Longer term we need to separate out STDERR and STDOUT in our SSH command output, but I think this should function as an OK hackaround for now.

cc @ehazlett 

### DIFF
--- a/libmachine/provision/os_release.go
+++ b/libmachine/provision/os_release.go
@@ -69,7 +69,8 @@ func (osr *OsRelease) ParseOsRelease(osReleaseContents []byte) error {
 	for scanner.Scan() {
 		key, val, err := parseLine(scanner.Text())
 		if err != nil {
-			return err
+			log.Warn("Warning: got an invalid line error parsing /etc/os-release: %s", err)
+			continue
 		}
 		if err := osr.setIfPossible(key, val); err != nil {
 			log.Debug(err)


libmachine/provision/os_release.go
```diff
@@ -69,7 +69,7 @@ func (osr *OsRelease) ParseOsRelease(osReleaseContents []byte) error {
 	for scanner.Scan() {
 		key, val, err := parseLine(scanner.Text())
 		if err != nil {
-			return err
+			log.Warn("Warning: got an invalid line error parsing /etc/os-release: %s", err)
```
hairyhenderson: Shouldn't we also skip this line since it's invalid? Probably just `continue` after the `log.Warn`

nathanleclaire: Ah, yeah, there should be a `continue` there.  Good catch.

hairyhenderson: Also - `continue`ing will also avoid an error I ran into recently in #1449 where `setIfPossible` barfed because of an extra blank line in `/etc/os-release`.

ehazlett:
LGTM

PR #1612, opened 2015-07-30
libmachine: Filestore#GetActive() should return an error if DOCKER_HOST is empty

skatsuta:
Currently `docker-machine active` returns a stopped machine name randomly if `DOCKER_HOST` is not set, because both `DOCKER_HOST` and `item.URL` are empty and `dockerHost == item.URL` is satisfied.

I think this behavior is very confusing so I suggest the command should return an error in that case.
(Please give me comments on the error message.)

### DIFF
--- a/libmachine/filestore.go
+++ b/libmachine/filestore.go
@@ -126,6 +126,10 @@ func (s Filestore) GetActive() (*Host, error) {
 	}
 
 	dockerHost := os.Getenv("DOCKER_HOST")
+	if dockerHost == "" {
+		return nil, errors.New("DOCKER_HOST not set")
+	}
+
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Seems reasonable to me.

I'd change the error message to `No active host found: DOCKER_HOST not set`.

nathanleclaire:
cc @ehazlett

skatsuta:
@nathanleclaire Thanks for your comment!
The message is better and more descriptive. I modified it.

ehazlett:
LGTM

nathanleclaire:
This would make a great test too ;D

PR #1613, opened 2015-07-30
Proposal: create --force

hairyhenderson:
I've hit a couple scenarios where I want to recreate a machine from scratch, and I don't really care that there's already a machine with the same name. Right now, I do:

```
$ docker-machine rm $NAME || true && docker-machine create -d ... $NAME
...
```

Not a terrible approach, but it feels kludgy. This PR allows the following:

```
$ docker-machine create --force -d ... $NAME
...
```

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -240,6 +240,10 @@ var sharedCreateFlags = []cli.Flag{
 		Usage: "addr to advertise for Swarm (default: detect and use the machine IP)",
 		Value: "",
 	},
+	cli.BoolFlag{
+		Name:  "force",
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Thanks for the contribution, but I'm -1 on this change.  I'd really prefer the remove operation to be a step which is explicitly taken.

hairyhenderson:
@nathanleclaire - thanks - mostly I was looking for a feeling of whether this is a good idea... I'm sort of +0 on the idea myself, so I could be swayed ;)

nathanleclaire:
Talked with Evan and he's also -1 -- closing :)

hairyhenderson:
Thanks! :)

PR #1620, opened 2015-07-31
libmachine: fix a test to check a host name is the same as expected

### LINKED ISSUE #1614: libmachine: what does host.Name != host.Name mean in filestore_test.go?
In https://github.com/docker/machine/blob/master/libmachine/filestore_test.go#L231-L233 what does the test below mean?

```go
if host.Name != host.Name {
    t.Fatalf("Active host is not 'test', got %s", host.Name)
}
```

I don't understand why the condition is this way.

skatsuta:
fixed #1614 

I fixed the test to check `host.Name` is the same as an expected host name.


ehazlett:
LGTM

Issue #1628, opened 2015-07-31
Move `goauth2/oauth` dependency to `golang.org/x/oauth2`


timfallmk:
https://code.google.com/p/goauth2/ is depreciated (along with all of code.google.com), and is being moved to read-only mode on August 25 in preparation for removal. We need to move this dependency to https://godoc.org/golang.org/x/oauth2, the suggested alternative.

dgageot:
This was done on the google driver.
Digital ocean driver is still using it though.

PR #1629, opened 2015-08-01
[WIP] Update `oauth2` interaction and library

timfallmk:
Reference to #1628 
Signed-off-by: Tim <timfall@gmail.com>

### DIFF
--- a/drivers/google/auth_util.go
+++ b/drivers/google/auth_util.go
@@ -11,7 +11,8 @@ import (
 	"strings"
 	"time"
 
-	"code.google.com/p/goauth2/oauth"
+	oauth "golang.org/x/oauth2" // This should use the specific /x/oauth2/google api set eventually
+	"golang.org/x/oauth2/google"
 	"github.com/docker/machine/log"
 	raw "google.golang.org/api/compute/v1"
 )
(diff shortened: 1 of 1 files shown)


timfallmk:
Anyone have more familiarity with the current `golang.org/x/oauth2/google` functions? I'm learning as I go.

mattjmcnaughton:
Thanks for doing this - great idea to get a jump on this ahead of Google code going read only.

I was just looking through the code and noticed the `digital_ocean` driver also uses `"code.google.com/p/goauth2/oauth"`, so it may make sense to switch it over to `golang.org/x/oauth2` as well.

mattjmcnaughton:
Also I don't have a ton of experience with the oauth, but it looks to me like `Transport` is not used as heavily in `golang.org/x/oauth2`. So instead of creating a `Transport` and then returning

timfallmk:
@mattjmcnaughton You're right on both counts. `Transport` is now a low level convention, used by `client` to construct an exchange. Changing this is what's taking most of my time, since I don't want to necessarily change the way the current code works.

As for the DO driver, I've submitted a patch here, which is now merged. https://github.com/digitalocean/godo/pull/62

I think the vendored version used here just needs to be updated.

mattjmcnaughton:
:+1: re `digitalocea/godo` update. Would this [line](https://github.com/docker/machine/blob/master/drivers/digitalocean/digitalocean.go#L8) in the Digital Ocean driver still need to be updated? Or is the `godo` patch you submitted taking care of that?

timfallmk:
@mattjmcnaughton This project uses `Godeps` to "vendorize" dependencies (freeze them at a specific version and include them with the code). When the maintainers update their included version of `godo` here, it will pull in the newest code. That would include the right import lines for *that* dependency as well.

dgageot:
Fixed by https://github.com/docker/machine/pull/2087

PR #1631, opened 2015-08-01
FIX #676 - Support Start/Stop GCE instance

### LINKED ISSUE #676: Starting a stopped google cloud machine does not work
This is the status of the machine:
```
$ gcloud compute instances list
NAME          ZONE           MACHINE_TYPE  INTERNAL_IP    EXTERNAL_IP   STATUS
gc-datalyse-1 europe-west1-b n1-standard-1 10.240.133.97                TERMINATED
```

```
$ docker-machine ls
NAME               ACTIVE   DRIVER       STATE     URL
gc-datalyse-1               google       Stopped   tcp://:2376
```

Issuing `docker-machine start gc-datalyse-1` returns:
```
INFO[0000] Creating instance.                           
INFO[0002] Waiting for Instance...                      
FATA[0003] Operation error: {RESOURCE_ALREADY_EXISTS  The resource 'projects/my-project/zones/europe-west1-b/instances/gc-datalyse-1' already exists} 
```

I guess it does not have to recreate it but instead just start it like 

```
$ gcloud compute instances start gc-datalyse-1
Updated [https://www.googleapis.com/compute/v1/projects/my-project/zones/europe-west1-b/instances/gc-datalyse-1].
```

```
$ gcloud compute instances list
NAME          ZONE           MACHINE_TYPE  INTERNAL_IP    EXTERNAL_IP   STATUS
gc-datalyse-1 europe-west1-b n1-standard-1 10.240.133.97  23.251.139.54 RUNNING
```

Which is then picked up by `docker-machine`

```
$ docker-machine ls
NAME               ACTIVE   DRIVER       STATE     URL
gc-datalyse-1               google       Running   tcp://23.251.139.54:2376
```

dgageot:


dgageot:
Hi @ehazlett can you explain what **design-review** tag means? Also why did you remove **driver-google** tag? Or maybe there's a document explaining those labels?

Can I help by improving the PR.

dgageot:
This PR fixes the start/stop of a GCE instance. Now that instances can be stopped property, we should really stop deleting and recreating instances.

drivers/google/google.go
```diff
@@ -216,28 +216,39 @@ func (d *Driver) GetState() (state.State, error) {
...
+// Start starts an existing GCE instance or create an instance with an existing disk.
 func (d *Driver) Start() error {
 	c, err := newComputeUtil(d)
 	if err != nil {
 		return err
 	}
-	if err = c.createInstance(d); err != nil {
-		return err
+
+	instance, _ := c.instance()
```
nathanleclaire: Why swallow the error?  Please don't do that.

drivers/google/google.go
```diff
@@ -260,20 +271,21 @@ func (d *Driver) Remove() error {
 	return c.deleteDisk()
 }
 
-// Restart deletes and recreates the GCE instance, keeping the disk.
 func (d *Driver) Restart() error {
+	return nil
+}
+
+// Kill deletes the GCE instance, but keeps the disk.
```
nathanleclaire: I know it's not your fault but, I don't think this is the intention of `Kill`.  It's meant to be a "hard stop".

drivers/google/google.go
dgageot: You are right. I pasted the code from GetState(). What really matter is if there's an instance or not. But this could swallow other kind of errors. I'll take a look.

drivers/google/google.go
dgageot: I'm not sure I understand what's the spirit of it.

Before the patch, Kill() would delegate to Stop(). And Stop() would delete the instance keeping the disk. This comes from a time where GCE instances couldn't be stopped.

I thought Stop() should keep the instance but keep it. And Kill() should delete the instance but keep the disk so that the instance could be recreated.

WDYT?

nathanleclaire: Passing off to `Stop` seems acceptable to me.  `Remove` is the method responsible for deleting and cleaning up the instance.

dgageot:
@nathanleclaire took your feedback into account.
Do you know why circleci fails?

dmp42:
@dgageot circleci integration testing is not yet ready to go (#1841).
Ignore circle entirely for now.

nathanleclaire:
@dgageot Code looks pretty reasonable. Please rebase (squash) all of your commits into one signed commit.

dgageot:
@nathanleclaire squashed. Thanks a lot

nathanleclaire:
LGTM

PR #1633, opened 2015-08-01
FIX #1297 - Support additional tags on GCE

### LINKED ISSUE #1297: Feature request - allow to pass in instance tags when creating machines
This is for GCE, but I see some basic code in AWS driver, too. We need to allow passing in instance tags for created machines. Internally, machine has a couple hard-coded ones it uses to set up basic firewall rules, but this needs to be opened up to allow more tags from a user.

Use case - I want to quickly open the swarm master machine port (3376), currently machine only sets up 2376 access. Ideally, this should be addressed via #948 , but I see how it will be useful beyond that. Having a tag makes firewall setup less brittle.

```
gcloud compute instances add-tags gce1 --tags docker-swarm-master --zone us-central1-f
gcloud compute firewall-rules create docker-swarm-master --allow tcp:3376 --target-tags docker-swarm-master
```

I noticed that machine terminates the instance in GCE when issued a `stop` command. On subsequent start the tag is lost, however. It is critical that machine saves these tags in instance metadata and re-applies them on `docker-machine start`, as cloud provider doesn't preserve them.

dgageot:


nathanleclaire:
You should perhaps use `cli.StringSlice` for this instead.

dgageot:
@nathanleclaire Thanks. It's much better this way

nathanleclaire:
Thanks.  Please rebase your commits into one signed commit and then I'll be ready to merge this.

dgageot:
@nathanleclaire squashed. Thanks a lot

nathanleclaire:
LGTM

PR #1637, opened 2015-08-03
Amazon EC2 - Adding option to inject EC2 user data at launch

nrdlngr:
Signed-off-by: Eric Nordlund <nrdlngr@gmail.com>

### DIFF
--- a/docs/drivers/aws.md
+++ b/docs/drivers/aws.md
@@ -33,6 +33,7 @@ Options:
  - `--amazonec2-spot-price`: Spot instance bid price (in dollars). Require the `--amazonec2-request-spot-instance` flag.
  - `--amazonec2-private-address-only`: Use the private IP address only.
  - `--amazonec2-monitoring`: Enable CloudWatch Monitoring.
+ - `--amazonec2-user-data`: Base64-encoded EC2 user data string to send.
 
 By default, the Amazon EC2 driver will use a daily image of Ubuntu 14.04 LTS.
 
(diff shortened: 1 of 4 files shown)


ehazlett:
@nrdlngr thanks for the contribution however we are really trying to have an opinionated control over the provisioning.  since not all providers support this we don't want to have different experiences across the drivers.  thanks!

nathanleclaire:
Thanks for your contribution! I agree with @ehazlett on closing this -- To elaborate a little further on my reasoning:

1. The EC2 driver already has a lot stuff in it and we'd still really like to do a rewrite using the Stripe API bindings (now official) instead of our homebrew solution.  It moved really quickly in the early days and we're still struggling to clean it up.
2. Before we go throwing things like custom user data into the mix, we need to get comfortable with the matrix we already have (OS x Provider) and provide the most uniform experience possible.  Before we allow flexibility in terms of provisioning, we have to decide what we are and are not willing to accept and define a real provisioning model, e.g. Do we support cloud-init only?  Do we allow users to run bash scripts at various points in provisioning?  Do we bootstrap Compose automatically?  There's a lot to consider.

It's not "no forever", but we need a definition of behavior which is relatively consistent across drivers before letting such a feature in.

bjaglin:
Any chance this decision could be revisited? https://github.com/docker/machine/pull/2141 was merged so this would actually bring consistency.

nathanleclaire:
I'd certainly review a pull request.  It will need to be different than this code probably since the EC2 driver has been re-written to use the official Go bindings for AWS.

PR #1642, opened 2015-08-04
openstack: New machine active timeout parameter

### LINKED ISSUE #1632: Creating openstack machine leads to maximum number of retries exceeded (it worked with 0.2)
Hi guys,

Using docker-machine-0.3.0 I can not create anymore openstack machines. It fails on some kind of timeout.
```
$ ./docker-machine_linux-amd64-0.3 create -d openstack --openstack-flavor-name="m1.small" --openstack-image-name="Ubuntu-14.04.2" --openstack-ssh-user=ubuntu cvl-iceda-everware01
Creating machine...
Error creating machine: Maximum number of retries (50) exceeded
```
But then the machine actually appears created in my openstack dashboard.

I still had the 0.2 binaries around and I tried with it, and it worked:
```
$ ./docker-machine_linux-amd64-0.2 create -d openstack --openstack-flavor-name="m1.small" --openstack-image-name="Ubuntu-14.04.2" --openstack-ssh-user=ubuntu cvl-iceda-everware
INFO[0009] Creating machine...                          
INFO[0747] "cvl-iceda-everware" has been created and is now the active machine. 
INFO[0747] To point your Docker client at it, run this in your shell: eval "$(docker-machine env cvl-iceda-everware)" 
```

And by the way, it took quite a lot of time (10-20 minutes, it is slow in our openstack installation because of network/DNS registration) to successfully create this machine with 0.2.
While with 0.3 I got the error rather quickly.

Seems to me the timeout is not letting enough time to the openstack driver to create the machine. And there does not seem to be a flag to actually increase the number of retries.

Cheers,
Axel

ddaze:
* Adds `--openstack-connection-retries` parameter to set the maximum
  number of retries. Closes #1632

Signed-off-by: David Zerulla <ddaze@outlook.de>

### DIFF
--- a/docs/drivers/openstack.md
+++ b/docs/drivers/openstack.md
@@ -36,6 +36,7 @@ Options:
    there is no IP address already allocated a new IP will be allocated and assigned to the machine.
  - `--openstack-ssh-user`: The username to use for SSH into the machine. If not provided `root` will be used.
  - `--openstack-ssh-port`: Customize the SSH port if the SSH server on the machine does not listen on the default port.
+ - `--openstack-active-timeout`: The timeout in seconds until the OpenStack instance must be active.
 
 Environment variables and default values:
 
@@ -62,3 +63,4 @@ Environment variables and default values:
 | `--openstack-floatingip-pool`    | -                      | -       |
 | `--openstack-ssh-user`           | -                      | `root`  |
 | `--openstack-ssh-port`           | -                      | `22`    |
+| `--openstack-active-timeout`     | -                      | `200`   |
--- a/drivers/openstack/client.go
+++ b/drivers/openstack/client.go
@@ -151,7 +151,7 @@ func (c *GenericClient) WaitForInstanceStatus(d *Driver, status string) error {
 		}
 
 		return false, nil
-	}, 50, 4*time.Second)
+	}, (d.ActiveTimeout / 4), 4*time.Second)
 }
 
 func (c *GenericClient) GetInstanceIpAddresses(d *Driver) ([]IpAddress, error) {
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -17,6 +17,7 @@ import (
 type Driver struct {
 	*drivers.BaseDriver
 	AuthUrl          string
+	ActiveTimeout    int
 	Insecure         bool
 	DomainID         string
 	DomainName       string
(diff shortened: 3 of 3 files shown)


ggiamarchi:
@ddaze I roughly understand your problem but i'm a bit perplex for two reasons :
* The timeout you want to increase is not about the connection to the machine. It concerns the time needed for the machine to become in the `ACTIVE` state. On the average OpenStack installation it takes a few seconds (appart in some specific circumstances) and the current timeout in machine is 200 seconds.
* The value of this timeout haven't changed between machine 0.2 and the current master (the code had been refactored but not the timeout duration)

In the whole machine creation process there are several waiting loops :
* To wait for the `ACTIVE` state of the instance (implemented by the OpenStack driver)
* To wait for the machine to be reachable using SSH (implemented in machine)
* To wait for Docker daemon to be started (implemented in machine)

Are you sure your need is about the `ACTIVE` state ? Are you able to explain why it is so long on your OpenStack ?

By the way, i'm not against the add of this option but we should find a better name because there's nothing about connection in this case.

ddaze:
@ggiamarchi Yes the timeout is because of the long delay for the machine to become `ACTIVE`.
The reasons for this are:
> - DNS entry using the server name specified so the name matches the IP address in LANDB
> - Active directory so that kerberos identity is established before booting
> - X.509 certificate for the host is available

> This leads to a delay of up to 15 minutes before the VM is created.
[CERN OpenStack](http://information-technology.web.cern.ch/book/cern-private-cloud-user-guide/command-line-tools/openstack-tools)

I don’t know if it was working with the version 0.2. But I had the same error message as in #1632.

And I agree the option name could be better, do you have a suggestion?

ggiamarchi:
@ddaze Ok, i understand.

I also forgot one point : i think it would be more user-friendly to give a flag to specify the timeout value in seconds instead of the number of retries. What do you think about `--openstack-active-timeout` ?

ddaze:
@ggiamarchi Sounds good!
And then calculate from the seconds the number of retries?
For example with `--openstack-active-timeout 900` seconds:
900/4 = 225 retries with 4 seconds each.

ggiamarchi:
@ddaze Exactly

ddaze:
@ggiamarchi done, anything else to improve?

ggiamarchi:
@ddaze The code looks good. Can you please describe the new flag in [the documentation](https://github.com/docker/machine/blob/master/docs/drivers/openstack.md) ?

ddaze:
@ggiamarchi also done :)

ggiamarchi:
@ddaze Thank you !

LGTM

PR #1646, opened 2015-08-05
Add Azure Swarm Master Port endpoint configuration when using docker machine

agup006:
When using docker-machine create with Azure driver, if using --swarm-master, Azure Ubuntu VM automatically opens port 3376. Resolving Issue 640


ahmetalpbalkan:
:metal: you rock

PR #1650, opened 2015-08-06
Add Virtualbox flag to clone existing Docker Machine VM

nathanleclaire:
This will be useful for Kitematic as they need to "rename" / migrate
some existing Docker Machine VMs, but is not intended to be used as a
primary feature.

cc @ehazlett @jeffdm @mchiang 


jeffdm:
Seems that it starts the VM but then exceeds the number of retries (60)

bonifaido:
What happened with this one?

Issue #1651, opened 2015-08-06
"active" and "ls" use different checks for active machine


ilkka:
Just noticed when commenting on brikis98/docker-osx-dev#77 that the "active" and "ls" commands use different methods for determining if a machine is active or not. Don't know if this is intentional, but for "active" using Filestore, only the value of DOCKER_HOST is checked against the host's URL. For "ls", `attemptGetHostState` in libmachine/host checks both the URL and also that the machine is not in state "Stopped". When DOCKER_HOST is set but e.g. the VM it points to is stopped, the commands now report slightly confusing results.

nathanleclaire:
Good catch.  They should definitely use the same, shared method.

mattjmcnaughton:
:+1: 

I'd love to call #dibs on this if that's cool (idk if machine uses #dibs like docker) - been looking to get involved with this project for a while and this seems like a good first issue.

nathanleclaire:
If you're interested, go for it @mattjmcnaughton !  I've been treating the "help-wanted" tag as kind of a "Hey, these issues are relatively non-controversial and might be worth looking into for someone wanting to contribute to Machine".

PR #1656, opened 2015-08-07
Use a single function for checking active machine

mattjmcnaughton:
This is my first time contributing to machine - I tried to follow the `CONTRIBUTING.md` as close as possible, but if I missed anything please let me know. Also, I wasn't sure about adding tests - the code I wrote seems to be covered by the existing tests and it didn't seem like each individual method had it's own test in `host_tests.go` so I ended up not writing any new ones - but I can if that would be good. Thanks all!

Fix https://github.com/docker/machine/issues/1651

As pointed out in the issue above, the `active` and `ls` commands used
different methods for determing the active machine. This commit defines
a single method on the `host` struct called `IsActive` which provides
a uniform check for machine activness. `IsActive` returns true only
if `DOCKER_HOST == url` and the state is not stopped - previously the
`active` command only checked the url.

* Add a single `host` method `IsActive` for determining if a machine is
  active.

Signed-off-by: Matt McNaughton <mattjmcnaughton@gmail.com>

### DIFF
--- a/libmachine/filestore.go
+++ b/libmachine/filestore.go
@@ -125,15 +125,10 @@ func (s Filestore) GetActive() (*Host, error) {
 		return nil, err
 	}
 
-	dockerHost := os.Getenv("DOCKER_HOST")
-	if dockerHost == "" {
-		return nil, errors.New("No active host found: DOCKER_HOST not set")
-	}
-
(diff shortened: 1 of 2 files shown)


nathanleclaire:
LGTM

cc @ehazlett

ehazlett:
LGTM

PR #1658, opened 2015-08-09
Teach `env` to run a child process

whilp:
Add support for running a child process in the `docker-machine` environment. This mirrors `ssh-agent`'s behavior when more than one argument is passed. For example:

```
$ env | grep SSH
$ ssh-agent /usr/bin/env | grep SSH
SSH_AUTH_SOCK=/tmp/ssh-ScOeqh91blLj/agent.70755
SSH_AGENT_PID=70757
$ env | grep SSH
```

Similarly, one can now invoke a child process within `docker-machine env`:

```
$ ./machine env dev /usr/bin/env | grep DOCKER
DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.99.101:2376
DOCKER_CERT_PATH=/Users/whilp/.docker/machine/machines/dev
DOCKER_MACHINE_NAME=dev
$ /usr/bin/env | grep DOCKER
```

This approach obviates the necessity for shell-specific export or unset syntax, so any shell (or, really, any program) is now supported. Previous behavior has been left in place.


whilp:
(Repushing after adding `Signed-off-by` headers.)

tianon:
Neat, seems like a reasonably sane idea IMO, especially since the pattern is already established by other tools like `ssh-agent`.  It'd be interesting to gather a list of popular tools users are likely to recognize that do the same thing so that when we get asked why this functionality exists we can point them at other tools.

This makes it possible to do things like this, at a simple level:
```console
$ alias docker-somemachine='docker-machine env somemachine docker'
$ docker-somemachine version
$ docker-somemachine ps
```

ehazlett:
@tianon thanks -- :+1:

ehazlett:
Would you mind squashing your commits?

whilp:
@ehazlett sure thing!

whilp:
@ehazlett squarshed and pushed.

ehazlett:
thanks!

When testing this locally, I'm getting:

```
ejh-x1:ehazlett> eval $(docker-machine env test-vbox)
OS type not recognized
```

whilp:
@ehazlett hm -- that manual test case works for me:

```shell
$ echo $DOCKER_MACHINE_NAME

$ docker-machine env -- dev /bin/sh -c 'echo $DOCKER_MACHINE_NAME'
dev
$ eval $(docker-machine env dev)
$ echo $DOCKER_MACHINE_NAME
dev
```

`dev` is a docker machine that was previously created. The error you pasted seems to be `ErrDetectionFailed`, which is returned by `DetectProvisioner()`. Is `test-vbox` a running machine? With this PR, I get the expected 'XXX is not running' or 'Host does not exist' errors with both the original shell and additional child behaviors.

whilp:
Hi all -- anything else I can do to improve this one?

emsi:
You probably should try to ping @ehazlett  once you make sure your patch works on all architectures :)

whilp:
@emsi happy to! I've read through the [CONTRIBUTING][contributing] doc and couldn't find notes about how to test against other architectures. With this PR, unit and manual tests pass on my OSX laptop and in Travis. Can you give me a pointer to test further? Thanks!

[contributing]: https://github.com/docker/machine/blob/c02b1c413e9f3d048a9be0cc588cff47827bc06f/CONTRIBUTING.md

dmp42:
cc @ehazlett anything left to do here? (save a rebase <- @whilp )

dgageot:
Hi @whilp can you rebase your commit please? Thanks a lot.

whilp:
@dmp42 @dgageot @ehazlett hi folks; sorry for the delay (had a baby). I just pushed a commit that syncs this up against master; I haven't actually rebased it yet because I had to adjust my approach a bit to incorporate the latest and would like your review. Please take a look and, if this seems legit to y'all, I'll rebase+sign. Thanks!

jeanlaurent:
I believe this is partially covered by `docker-machine config dev` although this does only apply to docker and not any other subcommand.

Two options here :
- Squashing + rebase ( code has been significanlty changed on this part ) + sign
- Do we really need this @docker/machine-maintainers ?  I fail to see much value since `config` does *almost* the same thing, but I'll be pleased to be proven wrong.

nathanleclaire:
It seems like a useful option although I am not 100% sure about the user interface.

Curious, @whilp why not use `config` + subshell instead? e.g.:

```console
$ docker $(docker-machine config default) ps
```

Is there something specific you had in mind where this comes in really handy?

gimoh:
@nathanleclaire it's longer to type ;-) and you may want to run other things than docker (e.g. [ptpython](https://pypi.python.org/pypi/ptpython) with docker-py or [wharfee](https://pypi.python.org/pypi/wharfee))

nathanleclaire:
@whilp OK cool I see now.  Docker Compose for instance has not accepted the same flags as Docker for connection settings and so `config` doesn't work.

I am in favor of the change, my main ask would be can you please add unit tests (and ideally an integration test as well) for this feature.  There should be quite a few existing env tests to draw from.

Thanks!

dweomer:
:+1: for the assist in machine-specific executions of `docker-compose` without polluting my shell

dgageot:
Hi @whilp could you make sure the PR is rebased? Thanks a lot!

nathanleclaire:
@whilp Any update? Would you like to rebase this?

PR #1664, opened 2015-08-10
Add DRIVER_OPTS variable in integration tests

ggiamarchi:
Some drivers need configuration that can be set only using specific command line flags. This is what the `DRIVER_OPTS` variable is for. Now, `core` tests can be run with any driver.

For instance, we can run the `core` tests on the OpenStack driver

```bash
$ export OS_USERNAME=user
$ export OS_PASSWORD=password
$ export OS_AUTH_URL=https://openstack/identity/v2.0
$ export OS_TENANT_ID=6d7fce9fee471194aa8b5b6e47267f03
$ export OS_REGION_NAME=region
$
$ export DRIVER=openstack
$ export DRIVER_OPTS='
    --openstack-image-name      ubuntu-14.04_x86_64
    --openstack-flavor-name     2_vCPU_RAM_4G_HD_10G
    --openstack-net-name        net
    --openstack-floatingip-pool PublicNetwork
    --openstack-ssh-user        ubuntu'
$
$ ./run-bats.sh core
```

### DIFF
--- a/test/integration/core/arbitrary-engine-envs.bats
+++ b/test/integration/core/arbitrary-engine-envs.bats
@@ -5,7 +5,7 @@ load ${BASE_TEST_DIR}/helpers.bash
 
 @test "$DRIVER: create with arbitrary engine envs" {
 
-run machine create -d $DRIVER \
+run machine create -d $DRIVER $DRIVER_OPTS \
   --engine-env=TEST=VALUE \
   $NAME
   [ $status -eq 0  ]
--- a/test/integration/core/arbitrary-engine-options.bats
+++ b/test/integration/core/arbitrary-engine-options.bats
@@ -3,7 +3,7 @@
 load ${BASE_TEST_DIR}/helpers.bash
 
 @test "$DRIVER: create with arbitrary engine option" {
-  run machine create -d $DRIVER \
+  run machine create -d $DRIVER $DRIVER_OPTS \
     --engine-opt log-driver=none \
     $NAME
   [ $status -eq 0 ]
--- a/test/integration/core/core-commands.bats
+++ b/test/integration/core/core-commands.bats
@@ -9,7 +9,7 @@ load ${BASE_TEST_DIR}/helpers.bash
 }
 
 @test "$DRIVER: create" {
-  run machine create -d $DRIVER $NAME
+  run machine create -d $DRIVER $DRIVER_OPTS $NAME
   [ "$status" -eq 0  ]
 }
 
--- a/test/integration/core/regenerate-certs.bats
+++ b/test/integration/core/regenerate-certs.bats
@@ -3,7 +3,7 @@
 load ${BASE_TEST_DIR}/helpers.bash
 
 @test "$DRIVER: create" {
-  run machine create -d $DRIVER $NAME
+  run machine create -d $DRIVER $DRIVER_OPTS $NAME
 }
 
 @test "$DRIVER: regenerate the certs" {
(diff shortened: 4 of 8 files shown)


nathanleclaire:
Generally I have been just setting the corresponding flags for the environment variables, since it is a uniform way to share them.  With this method it seems you have to remember to update the tests in every spot you might want to use them, and it's inconsistent in e.g. `cli` directory or the driver-specific ones.

Are there some flags you want to use which don't have environment variables already?

ggiamarchi:
Yes, actually none of the flags i use above get a corresponding environment variable.

Originally in the OpenStack driver there are environment variables only for things related to authentication, region, HTTP connection, ... ; everything not related to the machine actually. This is why there is currently no way to run integration tests on OpenStack.

I don't see why this solution would be inconsistent (?) Setting five environment variables or one containing five flags looks to be the same.

I'm not against the idea to add environment variables for all flags even if from a conceptual point of view it's better to use environment variables for things really related to the environment (i'm not sure to be clear :disappointed_relieved:)

I agree environment variable provide a uniform way to share them ; this make it easier. By the way, i'm aware to either solutions :
* Add environment variable for each flag
* Use the `DRIVER_OPTS` variable

__Note__ :
> In the other drivers most of the flags get a corresponding environment variable but not all. This is just an observation, not an argument.

WDYT ?

nathanleclaire:
> I don't see why this solution would be inconsistent (?) Setting five environment variables or one containing five flags looks to be the same.

Well, my concern is mostly about having to keep the `DRIVER_OPTS` variable updated in each test (every test which does a create should honor it).  

Outside of `core/`, there are also directories such as `cli/`, and the driver-specific ones, which are intended to run as part of the integration suite as well.  Forgetting `DRIVER_OPTS` in a create will introduce inconsistent behavior, but including it will force us to remember to manually update it whenever we add a new test or add to the existing ones.

Maybe a better solution would be to move to using some kind of `machine-create` alias (like how there is `machine`) which will add those flags?  And then change `machine create` usage in the tests to use `machine-create` instead?  What do you think about that idea?  Also cc @ehazlett for this discussion.

ggiamarchi:
Starting from your idea, we can basically do that in the existing `machine` function. if the `machine` call is a `create` action we insert driver specific flags before the machine name. This way, we don't need to change anything in the existing tests.

It would look something like

```bash
function machine() {
    if [[ "$@" =~ " -d $DRIVER " ]] ; then
        args=(${@:1:${#}-1} ${DRIVER_OPTS} ${@:${#}:${#}-1})
    else
        args=($@)
    fi
    "$MACHINE_ROOT"/"$MACHINE_BIN_NAME" "${args[@]}"
}
```

I did successfully test it with both `virtualbox` and `openstack` drivers.

Then, we could probably wrap the `-d $DRIVER` boilerplate in the function but the name is not always the same value because several tests create more than one machine.

ehazlett:
I agree with @nathanleclaire that using env vars provides a uniform way.  I'm also weary of trying to remember each flag in a `DRIVER_OPTS` var.  That being said, I think your suggestion could be used @ggiamarchi to override the machine test func.

I think my personal opinion would be to create environment variables for the flags as this is the convention and also keeps consistency with the other drivers.

KennethNagin:
I want to run docker tests on openstack, but the current build does not seem to allow it.
This fix might be the solution.  What is its status?

-Ken

nathanleclaire:
For the interim, I would prefer to either:

1. Add environment variables to the OpenStack driver so they can be specified directly (this would be easiest to get merged upstream quickly)
2. Add a `machine-create` alias for `machine create` which uses `$DRIVER`, `$NAME`, etc. by default

Or, both.

Thank you a lot for the contribution @ggiamarchi, but having to add `DRIVER_OPTS` to every single test which does a `create` is not something I am OK with.

PR #1667, opened 2015-08-11
digitalocean: enable specifying ssh user

ehazlett:
This enables specifying the ssh user in DigitalOcean.


nathanleclaire:
LGTM

This is starting to look like one of those things we should just have a "global" create flag for, though (it's duped in quite a few drivers).  Maybe we can file an issue for that.

PR #1685, opened 2015-08-12
Fix Docker daemon wait

nathanleclaire:
Also, a few various cleanups are bundled:

1. Only call GetDriver() once to get the object in provision/utils.go
2. SSH command wrapper will return the error and let the consumer decide
   what to do with it instead of bailing automatically on non-255

## Motivation

This PR is a first step towards fixing various issues that exist with regards to the way that Docker Machine (VirtualBox driver, mostly) handles itself in a networking environment using a proxy or VPN.  It will not fix the issue directly, but it should allow `create` to succeed without the dreaded "`ca.pem` missing" error message -- previously, provisioning would hang on attempting to dial the Docker daemon to establish whether it was running or stopped and this would prevent the rest of provisioning (including copying the certs) from happening.


nathanleclaire:
cc @ehazlett @tianon @jeffdm

libmachine/provision/utils.go
```diff
@@ -172,10 +174,17 @@ func ConfigureAuth(p Provisioner) error {
...
-	}
+	waitForDocker(p, dockerPort)
 
 	return nil
 }
+
+func waitForDocker(p Provisioner, dockerPort int) {
+	utils.WaitForSpecific(func() bool {
+		// HACK: Check netstat's output to see if anyone's listening on the Docker API port.
+		if _, err := p.SSHCommand(fmt.Sprintf("netstat -a | grep %d", dockerPort)); err != nil {
```
tianon: Do you think it'd be better to use `netstat -l` directly here to make sure we don't have false positives?

Also, how consistent is the output of `netstat`?  Can we narrow down that `grep` further to get rid of even more false positives?  (like a Docker port of `253` matching something listening on `1253` for an obvious example of a problem edge case)

tianon: Probably even want `netstat -ln` to make sure we don't get names from `/etc/services` mucking with our output.

nathanleclaire: Maybe something like `netstat -ln | grep ':::2376'`?  That starts to get into kind of specific output though, and I'm not sure how much the output of `netstat` deviates between systems.

tianon: Or maybe `grep '\b2376$'`, but likely doing the regex in Go instead of relying on `grep` inside the VM.

tianon: (`grep` is going to be much more wildly varying than `netstat`, in my experience, but it's worth testing a few different distros like `busybox`, `debian`, and something RPM-based like `fedora` to verify `netstat` output is reasonably sane)

ehazlett:
LGTM.  I've tested on Boot2Docker, Debian, Ubuntu and Fedora successfully.

nathanleclaire:
OK cool, @ehazlett I am going to update it to do the search for the relevant string using Go pattern matching like @tianon suggested, then I am g2g on this PR.

ehazlett:
@nathanleclaire :+1:

nathanleclaire:
This would also help situations such as: https://github.com/docker/machine/issues/1785#issuecomment-136883288 succeed in the create, but the firewall would still be blocking so I wonder if something could be done about that.  Perhaps a warning message when we can't reach <ip>:2376 indicating that a firewall or other network issue may be interfering with the communication.

drivers/virtualbox/virtualbox.go
```diff
@@ -136,7 +136,7 @@ func GetCreateFlags() []cli.Flag {
 }
 
 func (d *Driver) GetSSHHostname() (string, error) {
-	return "localhost", nil
+	return "127.0.0.1", nil
```
nathanleclaire: This is a tradeoff based on the fact that many users on Windows were getting `::1` (ipv6) for `localhost`.  We need to ensure IPv4.

libmachine/provision/boot2docker.go
```diff
@@ -182,6 +184,35 @@ func (provisioner *Boot2DockerProvisioner) GetOsReleaseInfo() (*OsRelease, error
 }
 
 func (provisioner *Boot2DockerProvisioner) Provision(swarmOptions swarm.SwarmOptions, authOptions auth.AuthOptions, engineOptions engine.EngineOptions) error {
+	const (
+		dockerPort = 2376
+	)
+
+	defer func() {
```
nathanleclaire: Not sure if a `defer` is appropriate here, or if it's simply something we should do at the end of the function on an otherwise happy path.

ehazlett:
LGTM

PR #1686, opened 2015-08-13
Prioritize using make in build, cleanup a few things in build script

nathanleclaire:
cc @ehazlett @tianon @hairyhenderson 

It's such a universal standard that `make` in your project root should build your project, I would really prefer for the build to be done through `make`.  This also reduces the complexity of the `script/build` script by just letting users specifying `GOOS` / `GOARCH` instead of our current `$1`/`$2` parsing weirdness.

This also replaces `SKIP_BUILD` with an automatic check to see if the build image is around already.

WDYT?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,17 +10,20 @@ Machine is a part of the [Docker](https://www.docker.com) project, and follows
 the same rules and principles. If you're already familiar with the way
 Docker does things, you'll feel right at home.
 
-Otherwise, go read
-[Docker's contributions guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
+Otherwise, please read [Docker's contributions
+guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
+
+# Building
 
 The requirements to build Machine are:
 
 1. A running instance of Docker
 2. The `bash` shell
+3. [Make](https://www.gnu.org/software/make/)
 
 To build, run:
 
-    $ script/build
+    $ make
 
 From the Machine repository's root.  Machine will run the build inside of a
 Docker container and the compiled binaries will appear in the project directory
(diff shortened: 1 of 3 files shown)


ehazlett:
I was hesitant in the beginning but I think it makes sense now.  LGTM

tianon:
I'm usually a +1 for `make`, since it can make things simpler :smile:

(and simpler for discoverability -- knowing about `script/build` requires reading the docs :smile:)

script/build
```diff
@@ -1,24 +1,20 @@
...
-    OS_PLATFORM_ARG=($1)
-fi
-
-if [ -z "$2" ]; then
-    OS_ARCH_ARG=(-arch="386 amd64 arm")
-else
-    OS_ARCH_ARG=($2)
-fi
+BUILD_IMAGE_NAME="docker-machine-build"
+OS_PLATFORM_ARG=${GOOS:-"darwin linux windows"}
```
nathanleclaire: Want to actually update this variable to `GOOS` instead of `OS_PLATFORM_ARG` for consistency / ease of mental burden's sake, then I'll be ready to merge this.

CONTRIBUTING.md
```diff
@@ -10,17 +10,19 @@ Machine is a part of the [Docker](https://www.docker.com) project, and follows
...
-[Docker's contributions guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
+Otherwise, please read [Docker's contributions
+guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
+
+# Building
 
 The requirements to build Machine are:
 
 1. A running instance of Docker
-2. The `bash` shell
```
hairyhenderson: Not sure this should be redacted - `bash` is still required (esp. important for systems where it isn't installed by default).

hairyhenderson:
:+1:  I like this.

CONTRIBUTING.md
nathanleclaire: That's true, the script does actually rely on bash.

Issue #1681, opened 2015-08-12
Upgrade completely broken on Ubuntu. lxc-docker no longer valid package name.  


nathanleclaire:
Due to the new package archives this: https://github.com/docker/machine/blob/master/libmachine/provision/ubuntu.go#L68 is wrong (therefore upgrade is broken).

We need to change this code to reference `engine-docker` and figure out how to address the situation with current users.

See: https://blog.docker.com/2015/07/new-apt-and-yum-repos/ for how to migrate from `lxc-docker` to `docker-engine`.

cc @ehazlett

tianon:
I actually see a few things in that function that worry me (like `apt-get upgrade some-package`, which doesn't actually upgrade just `some-package` and will really try to upgrade all packages on the system).

Here's what I'd propose:

`install` + `upgrade`:

1. always overwrite `/etc/sources.list.d/docker.list`
2. always `apt-get update`
3. always `apt-get install -y docker-engine` (which will both install or upgrade as appropriate)

`remove`:

1. same as today, although possibly slightly more like `apt-get remove -y docker-engine lxc-docker` to make sure we cover our bases

Installing `docker-engine` _should_ automatically remove `lxc-docker`, but that's worth some testing (ie, don't just take my word for it :innocent:).

mahnunchik:
+1

```bash
$ docker-machine upgrade test-machine
SSH cmd error!
command: DEBIAN_FRONTEND=noninteractive sudo -E apt-get upgrade -y  lxc-docker
err    : exit status 100
output : Reading package lists...
Building dependency tree...
Reading state information...
Package lxc-docker is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'lxc-docker' has no installation candidate
```

I've tried to upgrade machine from 1.8.0 to 1.8.1

nathanleclaire:
Sorry for the trouble @mahnunchik.  Patch submitted by the incorrigible @ehazlett here: https://github.com/docker/machine/pull/1687

ehazlett:
@tianon thx!  I don't know how i missed the `upgrade` thing :(  That's fixed in #1687.  I didn't implement the dual package logic in `remove` as we do not ever "remove" Docker and I want to try to keep the backward logic out.  If a user is on an old version of the install, they simply will just live with it.  We do not offer a way to remove it to hit that case.  If they upgrade, the upgrade will remove it for them and then it should be good.

[...]

PR #1687, opened 2015-08-13
fixes upgrade for debian/ubuntu for new package name

ehazlett:
This fixes upgrade issue with Debian / Ubuntu and Docker 1.8.

To test, you can create a host using an old install script.  Here is an example for Debian:

```
docker-machine create \
  -d digitalocean \
  --digitalocean-image debian-8-x64 \
  --engine-install-url http://docker-install.evanhazlett.com/docker-old \
  migration-test
```

Then you can upgrade with `docker-machine upgrade migration-test`.  After the upgrade you should end up with Docker 1.8.1.

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -92,7 +92,7 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
 		}
 
 		for _, cmd := range commands {
-			command := fmt.Sprintf("DEBIAN_FRONTEND=noninteractive sudo -E %s", cmd)
+			command := fmt.Sprintf("sudo DEBIAN_FRONTEND=noninteractive %s", cmd)
 			if _, err := provisioner.SSHCommand(command); err != nil {
 				return err
 			}
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -59,13 +59,11 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
 	updateMetadata := true
 
 	switch action {
-	case pkgaction.Install:
+	case pkgaction.Install, pkgaction.Upgrade:
 		packageAction = "install"
 	case pkgaction.Remove:
 		packageAction = "remove"
 		updateMetadata = false
-	case pkgaction.Upgrade:
-		packageAction = "upgrade"
 	}
 
 	switch name {
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -70,12 +70,34 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
 
 	switch name {
 	case "docker":
-		name = "lxc-docker"
+		name = "docker-engine"
 	}
 
 	if updateMetadata {
 		if _, err := provisioner.SSHCommand("sudo apt-get update"); err != nil {
 			return err
+		}
+	}
+
+	// handle the new docker-engine package; we can probably remove this
+	// after we have a few versions
+	if action == pkgaction.Upgrade && name == "docker-engine" {
+		// run the force remove on the existing lxc-docker package
+		// and remove the existing apt source list
+		// also re-run the get.docker.com script to properly setup
+		// the system again
+
+		commands := []string{
+			"rm /etc/apt/sources.list.d/docker.list || true",
+			"apt-get remove -y lxc-docker || true",
+			"curl -sSL https://get.docker.com | sh",
+		}
+
+		for _, cmd := range commands {
+			command := fmt.Sprintf("DEBIAN_FRONTEND=noninteractive sudo -E %s", cmd)
+			if _, err := provisioner.SSHCommand(command); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/libmachine/provision/ubuntu.go
+++ b/libmachine/provision/ubuntu.go
@@ -47,10 +47,9 @@ func (provisioner *UbuntuProvisioner) Service(name string, action pkgaction.Serv
 }
 
 func (provisioner *UbuntuProvisioner) Package(name string, action pkgaction.PackageAction) error {
-	var (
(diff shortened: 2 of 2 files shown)


nathanleclaire:
You're a legend my friend.

cc @tianon Can you please take a look?

nathanleclaire:
Also @ehazlett I'm assuming you saw @tianon comment here: https://github.com/docker/machine/issues/1681#issuecomment-130472503 ?

nathanleclaire:
Will `apt-get remove -y lxc-docker` nuke all images / containers and other data too, or is it safe?

ehazlett:
@nathanleclaire thx!  The data is safe.  I tested on two instances and all containers and images were left intact.  However, this does restart the daemon.

Updated to support proper upgrade functionality for packages (https://github.com/docker/machine/issues/1681#issuecomment-130472503)

libmachine/provision/debian.go
```diff
@@ -79,6 +77,28 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
...
+		// the system again
+
+		commands := []string{
+			"rm /etc/apt/sources.list.d/docker.list || true",
+			"apt-get remove -y lxc-docker || true",
+			"curl -sSL https://get.docker.com | sh",
+		}
+
+		for _, cmd := range commands {
+			command := fmt.Sprintf("DEBIAN_FRONTEND=noninteractive sudo -E %s", cmd)
```
tianon: What's the reason for the `sudo -E` here?

If it's just to make sure `DEBIAN_FRONTEND` is passed through, you should be able to swap them and accomplish the same, ie: `sudo DEBIAN_FRONTEND=noninteractive %s`.

tianon:
```console
$ sudo A=b env | grep ^A
A=b
```

libmachine/provision/ubuntu.go
```diff
@@ -47,36 +47,55 @@ func (provisioner *UbuntuProvisioner) Service(name string, action pkgaction.Serv
...
+	if action == pkgaction.Upgrade && name == "docker-engine" {
+		// run the force remove on the existing lxc-docker package
+		// and remove the existing apt source list
+		// also re-run the get.docker.com script to properly setup
+		// the system again
+
+		commands := []string{
+			"rm /etc/apt/sources.list.d/docker.list || true",
+			"apt-get remove -y lxc-docker || true",
+			"curl -sSL https://get.docker.com | sh",
```
tianon: I cringe erry-time :smile:

tianon:
Logic looks sane, just curious about `sudo -E` usage. :+1:

libmachine/provision/ubuntu.go
nathanleclaire: lol

nathanleclaire: Longer-term, I'm wondering if it might not be worthwhile to replace that script with Go code in machine, but for now it Gets The Job Done.

nathanleclaire:
> However, this does restart the daemon.

That's expected for an upgrade so OK with me.

I'll review this ASAP.

nathanleclaire:
I've gotten an error trying to upgrade one of my Ubuntu VMs (another went successfully).

This seems to be the relevant bit in the log:

```
About to run SSH command:
DEBIAN_FRONTEND=noninteractive sudo -E apt-get remove -y lxc-docker || true
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /Users/nathanleclaire/.docker/machine/machines/droplet/id_rsa -p 22 root@104.
236.189.32 DEBIAN_FRONTEND=noninteractive sudo -E apt-get remove -y lxc-docker || true] []  <nil> <nil> <nil> [] <nil> <nil> <nil> <nil> false [] [] [] [] <nil>}
SSH cmd err, output: <nil>: Reading package lists...
Building dependency tree...
Reading state information...
The following packages were automatically installed and are no longer required:
  aufs-tools cgroup-lite git git-man liberror-perl lxc-docker-1.7.0
Use 'apt-get autoremove' to remove them.
The following packages will be REMOVED:
  lxc-docker
/bin/sh: 1: /usr/sbin/dpkg-preconfigure: not found
0 upgraded, 0 newly installed, 1 to remove and 48 not upgraded.
After this operation, 0 B of additional disk space will be used.
(Reading database ... 93638 files and directories currently installed.)
Removing lxc-docker (1.7.0) ...
```

`/usr/sbin/dpkg-preconfigure` is not found, so the package is not removed.  Consequently, the new install of `docker` fails because it's already around on the system.

libmachine/provision/debian.go
ehazlett: :+1:

ehazlett:
Um that looks like some system issue.  `dpkg-preconfigure` shouldn't be removed for any reason.

libmachine/provision/ubuntu.go
ehazlett: :(

ehazlett:
thx @tianon -- i'll update the sudo usage.

libmachine/provision/ubuntu.go
tianon: I want that script and the functionality it provides to just die completely, and the long-long-term goal is to deprecate it in favor of people actually knowing what OS they're running and doing the few simple steps to set it up properly themselves. :smile:

(Platform detection is hairy, and wrought with failure.)

ehazlett:
@nathanleclaire updated and ready

nathanleclaire:
Alright hopefully the issues I was seeing were anomalous (one of my machines upgraded smoothly) so I'm ready to go on this PR and I'm just going to rebuild my b0rked host.  We can re-visit if we run into more issues in the future.

LGTM

Issue #1630, opened 2015-08-01
docker-machine create --driver virtualbox hangs on ssh


glyph:
`docker-machine create --driver virtualbox dev` hangs when I run it on my mac.

With debug output on, I see:

```
About to run SSH command:
exit 0
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /Users/glyph/.docker/machine/machines/dev/id_rsa -p 57660 docker@localhost exit 0] []  <nil> <nil> <nil> [] <nil> <nil> <nil> <nil> false [] [] [] [] <nil>}
```

If I copy and paste the SSH command into a different terminal while it's stuck, I see:

```
$ /usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /Users/glyph/.docker/machine/machines/dev/id_rsa -p 57660 docker@localhost
                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~~ ~ /  ===- ~~~
           \______ o           __/
             \    \         __/
              \____\_______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|
Boot2Docker version 1.7.1, build master : c202798 - Wed Jul 15 00:16:02 UTC 2015
Docker version 1.7.1, build 786b29d
docker@boot2docker:~$ 
```

glyph:
It seems this was at least partially an error in my own configuration.  I had this in my SSH configuration:

```
ControlMaster auto
ControlPath ~/.ssh/s/%u@%l-%r@%h:%p
ControlPersist 1m
```

[...]

PR #1690, opened 2015-08-13
Disable ssh multiplex explicitly

jonmort:
This PR fixes https://github.com/docker/machine/issues/1630 by disabling multiplexing


nathanleclaire:
Oh, good times.

Looks pretty good to me.  cc @ehazlett @tianon WDYT

nathanleclaire:
I'm merging this, we can revert if it turns out to be problematic.

Issue #1696, opened 2015-08-13
docker-machine ls timeout should be configurable.


nathanleclaire:
The current timeout for machines reporting state is 3 seconds.  This is intended to promote a relatively snappy UX.  However, this should be configurable in case the user has far away machines they want to include, e.g.:

```console
$ docker-machine ls --timeout=10
```

mattjmcnaughton:
I'd love to tackle this if no one else is - working on a pr now!

nathanleclaire:
Nice @mattjmcnaughton

ehazlett:
@mattjmcnaughton thanks!

mrumpf:
Under Windows 7 (VirtualBox) it seems that the timeout is too low. When I run the command ```docker-machine ls``` frequently I see that the virtual machine is running. When I wait 30 seconds and run the ls command again, I get a timeout for the same machine. Running the command again shows active again.

chigix:
Same situation encountered as @mrumpf  said in windows 10. And with `docker-machine --native-ssh ls`, every thing works very well.

Thanks, @mrumpf , your suggestion provided is of great helpful.

dgageot:
Fixed in https://github.com/docker/machine/commit/02c235b8b91e093fc185713e965fd17d7c462873

PR #1700, opened 2015-08-13
Make timeout for `ls` command configurable

mattjmcnaughton:
Fix https://github.com/docker/machine/issues/1696

A previous commit made it so machine would timeout after a set interval
of time per host. This commit makes the timeout time configurable
through a command line flag.

This required a slight architecture change. It also suggested that
perhaps `active` should have a similar timeout method because of the
amount of overlap in methods between `active` and `ls`.

* Add command-line configurable flags for timeout of `active` and `ls`
  command.

Signed-off-by: Matt McNaughton <mattjmcnaughton@gmail.com>

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/log"
@@ -27,7 +28,9 @@ func cmdActive(c *cli.Context) {
 		log.Fatal(err)
 	}
 
-	host, err := provider.GetActive()
+	timeout := time.Duration(c.Int("timeout")) * time.Second
+
+	host, err := provider.GetActive(timeout)
 	if err != nil {
 		log.Fatalf("Error getting active host: %s", err)
 	}
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -40,6 +40,7 @@ var (
 	ErrUnknownShell       = errors.New("Error: Unknown shell")
 	ErrNoMachineSpecified = errors.New("Error: Expected to get one or more machine names as arguments.")
 	ErrExpectedOneMachine = errors.New("Error: Expected one machine name as an argument.")
+	DefaultTimeout        = 3
 )
 
 type machineConfig struct {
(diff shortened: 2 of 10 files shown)


mattjmcnaughton:
@nathanleclaire or anyone else I'd love some feedback on this if you have the time! I think it works but idk how good of a design it is. The main thing I'm concerned about is that the timeout value is set as a command line flag, so its value is only accessible in `commands/ls.go`. But then I feel like I'm passing the timeout value to a lot of functions where it is simply passed along to another function - for example [here](https://github.com/docker/machine/pull/1700/files#diff-5fd65e2bf2a039f86e89c8b48f395db1R446).

Additionally, there's a lot of overlap between the methods used by `ls` and `active` so making `ls` use a configurable timeout kind of forced `active` having a configurable timeout (at least based on the design I used). So I don't know if that's good. This could perhaps be avoided by variadic function, but that seemed like added complexity.

Anyway, I'm definitely open to feedback on this pr - this is just my first idea!

nathanleclaire:
Ah yeah, that's kind of a really unfortunate antipattern.  I will have to take a look more closely when I get some time and see if I can think of a better solution.

nathanleclaire:
Thanks for the contribution btw!

mattjmcnaughton:
My pleasure - thanks for your help! 

One option I thought of was maybe creating a `config.go` file in libmachine and then there would be a singleton instance of `config` with just this kind of information included. So `cmdLs` would write the `timeout` value to `config` and then `libmachine/host#getHostState` could just read the value from the singleton config. This would prevent changes to all of the methods except `getHostState` and would also solve the problem of the change to `active`.

The reason I didn't do this in the first place is it seemed like a fairly large architectural change for a fairly small feature. But maybe it is the way to go.

Thoughts? If this sounds good I can revise my pr or keep thinking!

ehazlett:
@mattjmcnaughton there is some refactoring going on with libmachine (https://github.com/docker/machine/pull/1729) so you may want to hold off for a little bit until that is complete.

mattjmcnaughton:
@ehazlett Sounds great! I'm not really happy with the design in the pr I submitted, so I'll take another shot at it after the refactoring ends.

nathanleclaire:
Yes, the refactor is unfortunately quite large and affects a lot of the code.  However, after it is done hopefully this should be more elegant since I want to move all of the `getHostState` etc. code into `commands` module (since currently it's very directly tied to the `ls` codepath).

mattjmcnaughton:
@nathanleclaire Sounds great - I'll update this once the refactor is done!

nathanleclaire:
I'm going to close this one -- head on over to https://github.com/docker/machine/pull/1729/files#r39319122, where you can start taking a look at the new `libmachine` stuff and either make a PR on my repo to that open branch, or simply hold off until we merge that upstream (should be pretty soon).

PR #1701, opened 2015-08-14
Proposal: support to use GitHub API token to fetch the latest version of Boot2Docker

### LINKED ISSUE #1483: Feature Request: Add support for Github Token to avoid Github API Rate limits
I tried creating a machine today for the first time and I got the following output:

```
➜  ~  docker-machine create --driver virtualbox dev
Creating CA: /Users/smithm/.docker/machine/certs/ca.pem
Creating client certificate: /Users/smithm/.docker/machine/certs/cert.pem
Image cache does not exist, creating it at /Users/smithm/.docker/machine/cache...
No default boot2docker iso found locally, downloading the latest release...
Error creating machine: Error demarshaling the Github API response: json: cannot unmarshal object into Go value of type []struct { TagName string "json:\"tag_name\"" }
You may be getting rate limited by Github.
```

I'm presuming the offending code is here: https://github.com/docker/machine/blob/master/utils/b2d.go#L71-L95

I'm pretty sure I ran into this because I used `brew search docker-machine` which dinged my rate limit. Homebrew allowed an option for me to add an environment variable to define a token, which is what allow me to install docker-machine on my laptop. Which may be related to unseen issues introduced by #130 which discussed the addition of the homebrew formula for docker-machine.

Github does allow you to authenticate with an auth token which significantly increases the rate limit you can use. Has there been any consideration to adding configuration support to define such a token? Perhaps through an environment variable?

Without such support I have to wait up to 30 minutes to run this command. Not the end of the world, but slightly frustrating.

skatsuta:
This PR is a proposal that temporally resolves #1483 until we find a better way to fetch the latest version of Boot2Docker.
(I really want Machine to support this kind of feature because in my workplace GitHub API rate limit is frequently exceeded 😩)

I suggest that, like Homebrew's `HOMEBREW_GITHUB_API_TOKEN`, Machine supports to use an env var such as `MACHINE_GITHUB_API_TOKEN` when calling GitHub API to fetch the latest version of Boot2Docker.

issues:
- env var or flag? (I think an implicit env var is better because this feature is used only when a host is B2D)
- what env var name / flag name?
- test

### DIFF
--- a/utils/b2d.go
+++ b/utils/b2d.go
@@ -73,7 +73,15 @@ func NewB2dUtils(githubApiBaseUrl, githubBaseUrl string) *B2dUtils {
 func (b *B2dUtils) GetLatestBoot2DockerReleaseURL() (string, error) {
 	client := getClient()
 	apiUrl := fmt.Sprintf("%s/repos/boot2docker/boot2docker/releases", b.githubApiBaseUrl)
-	rsp, err := client.Get(apiUrl)
+	req, err := http.NewRequest("GET", apiUrl, nil)
+	if err != nil {
+		return "", fmt.Errorf("Error parsing a GitHub API URL: %s", err)
+	}
+	token := os.Getenv("MACHINE_GITHUB_API_TOKEN")
+	if token != "" {
+		req.Header.Set("Authorization", "token "+token)
+	}
+	rsp, err := client.Do(req)
 	if err != nil {
 		return "", err
 	}


nathanleclaire:
Cool, thanks for the contribution.  One solution we were considering for this was to have a file in the repository which we call out to instead.  I don't think raw files are rate limited like the API and that solution would stay transparent, although it would add an extra step to doing a boot2docker ISO release.  WDYT about that method?

ehazlett:
@nathanleclaire What about putting the b2d releases (and perhaps a metadata file) on get.docker.com?

skatsuta:
@nathanleclaire @ehazlett 
Ah, I think the method is better to avoid rate limit. Boot2Docker repo has https://github.com/boot2docker/boot2docker/blob/master/VERSION file, so how about using it? Putting the B2D releases on get.docker.com is also great, but it'd need to add an extra release step?

nathanleclaire:
Ah, cool, I didn't realize there's already a file for it.  Maybe we can switch over to using that.  @tianon have you given this matter any more thought?

tianon:
I'm personally -1 to using the `VERSION` file directly as-is because that has to be updated, committed, and pushed before the releases are so there'll be a window where it will outright fail to download the updated ISOs.  I think the discussion over in #1483 ought to continue with more testing of the `raw.githubusercontent.com` WRT rate limits (since that's still somewhat questionable, although looking positive).

tianon:
Although as far as allowing use of an API token, that sounds totally great IMO. :+1:

skatsuta:
Alright, we need further discussions in #1483. A pending PR is not good, so I close this for now.

PR #1703, opened 2015-08-14
`docker-machine status` should return non-0 for errors

alindeman:
Without this patch:

```
❯ docker-machine status foo
error getting state for host foo: machine does not exist
Error

❯ echo $?
0
```


alindeman:
Will re-open after I sort out the licensing details.

PR #1729, opened 2015-08-19
Make libmachine consumable by outside world

nathanleclaire:
Alright, I'm hesitant to turn this in incomplete (which it is so far-- probably about 60-70% done), but better early than never, and I'd prefer to iterate on something crude and incomplete rather than toil away and learn that others have some fundamental disagreements on my approach.

Sorry for the size of the patch, but you'll notice (in the current state at least) there are about 2x as many deletions as additions, and a chunk of it is now-or-never cleanup work to make the API more sane.

I'll annotate the code a little to guide since the patch is so large.

## Design Principles:

- [x] Never, ever write to disk inside the `host` module.  That's the `persist` module's job.  To that end, it would be good to get another `Store` implementation to prove viability outside of `Filestore`, and automatically test it.  To that extent, ideally `Store` needs some more re-working so it can keep track of artifacts (secrets such as certs and mundane artifacts like ISOs) instead of having everybody write them to the filesystem willy-nilly.
- [x] `Provider` in its current form is completely gone as its role conflicted heavily with `host.Host`.  Later, I want to bring back `Provider` as a notion of something which "provides" host creation and management and provides the things needed to bootstrap control of them, such as API tokens.

Something like the following code is my rough roadmap for where I'm working towards with the `libmachine` module.  I'm not really sure how to best handle the awkward `DriverOptions` bit.  `interface{}` seems wrong, but we don't have strongly typed `Provider`s to give us `Host`s (like alluded to above).

```Go
package main

// Sample Virtualbox create independent of Machine CLI.

import (
    "fmt"

    "github.com/docker/machine/libmachine"
    "github.com/docker/machine/libmachine/engine"
    "github.com/docker/machine/libmachine/host"
    "github.com/docker/machine/drivers/virtualbox"
)

func main() {
    // returns the familiar store at $HOME/.docker/machine
    store, _ := libmachine.GetDefaultStore()

    h := &host.Host{
        Name: "mycoolhost",
        HostOptions: &host.HostOptions{
                EngineOptions: &engine.EngineOptions{
                    StorageDriver: "overlay",
                },
        },
        Driver: &virtualbox.Driver{
            CPU: 2,
            Memory: 2048,
        },
    }
    
    // This method is a very high-level method which includes host creation,
    // saving the newly created host in the store, and provisioning.
    h, _ = libmachine.Create(store, h)

    out, _ := h.RunSSHCommand("df -h")

    fmt.Printf("Results of your disk space query:\n%s\n", out)

    fmt.Println("Powering down machine now...")
    _ := h.Stop()
}
```

Thoughts:

- I'd like to rename `host` module to something a bit less confusing, to avoid collision with the instantiated `Host` struct.  To that extent, I really am wondering if we should just move `Host` to be `Machine`, as it _is_ `libmachine` after all.
- You'll notice that `store` is now in its own module, as are most of the `libmachine`-related functions.  The core `libmachine` namespace will be reserved for very high-level functions meant for calling from the outside world, and/or `commands`.
- `libmachine` depends on other things such as `github.com/docker/machine/log`, do those go in `libmachine` as well?  How does that all work?  Logging etc. since I'm sure consumers of the API won't want us logging to terminal all over the place implicitly.  Guess we should pass in an `io.Writer` as a "config" property of... something?
- The API should probably be versioned, how do we do so?  Sounds like another case for a `libmachine` client struct that you should be instantiating before doing anything.

cc @ehazlett @ibuildthecloud @hairyhenderson @SvenDowideit @tianon @bfirsh  :musical_note:  MACHINE PARTY :tada: :balloon: :confetti_ball:


commands/active.go
```diff
@@ -12,22 +12,8 @@ func cmdActive(c *cli.Context) {
 		log.Fatal("Error: Too many arguments given.")
 	}
 
-	certInfo := getCertPathInfo(c)
```
nathanleclaire: This pattern is / was all over the place, and it's super repetitive, so it's been replaced by a call to `getStore` which is a helper function that performs this logic.

commands/commands.go
```diff
@@ -88,23 +90,38 @@ func confirmInput(msg string) bool {
...
+	certInfo := getCertPathInfo(c)
+	return &persist.Filestore{
+		Path:           c.GlobalString("storage-path"),
+		CaCertPath:     certInfo.CaCertPath,
+		PrivateKeyPath: certInfo.CaKeyPath,
+	}
 }
 
-func setupCertificates(caCertPath, caKeyPath, clientCertPath, clientKeyPath string) error {
+func getFirstArgHost(c *cli.Context) *host.Host {
```
nathanleclaire: There was a method `getHost` to do this before but this seems more descriptive.

commands/commands.go
```diff
@@ -88,23 +90,38 @@ func confirmInput(msg string) bool {
...
+		// I guess I feel OK with bailing here since if we can't get
+		// the host reliably we're definitely not going to be able to
+		// do anything else interesting, but also this premature exit
+		// feels wrong to me.  Let's revisit it later.
+		log.Fatalf("Error trying to get host %q: %s", hostName, err)
+	}
+	return h
+}
+
+func setupCertificates(certPathInfo cert.CertPathInfo) error {
```
nathanleclaire: Calls to this function involved separately passing all of the parameters of this object, so I went ahead and cleaned that up.

commands/commands_test.go
```diff
@@ -52,7 +53,11 @@ func getTestStore() (libmachine.Store, error) {
 
 	os.Setenv("MACHINE_STORAGE_PATH", tmpDir)
 
-	return libmachine.NewFilestore(tmpDir, hostTestCaCert, hostTestPrivateKey), nil
+	return persist.Filestore{
```
nathanleclaire: the `persist.Filestore` thing turns out a LOT cleaner than `store.Filestore` which is part of why I want to rename `host` module.

commands/config_test.go
```diff
@@ -1,104 +0,0 @@
-package commands
```
nathanleclaire: It is unfortunate to have to remove these tests (they were very hardcoded to the old `provider` system but I am replacing them with BATS-based integration tests (if you look closely, you'll notice these are basically integration tests anyway).

commands/create.go
```diff
@@ -42,29 +53,26 @@ func cmdCreate(c *cli.Context) {
...
+	store := &persist.Filestore{
+		Path:           c.GlobalString("storage-path"),
+		CaCertPath:     certInfo.CaCertPath,
+		PrivateKeyPath: certInfo.CaKeyPath,
 	}
 
-	provider, err := newProvider(defaultStore)
-	if err != nil {
-		log.Fatal(err)
+	switch driver {
```
nathanleclaire: This doesn't do anything for now but essentially is intended to be the block where the driver options (strongly typed) are added in later on.

commands/commands.go
```diff
@@ -13,24 +13,26 @@ import (
...
 
 	"github.com/docker/machine/drivers"
-	_ "github.com/docker/machine/drivers/amazonec2"
-	_ "github.com/docker/machine/drivers/azure"
-	_ "github.com/docker/machine/drivers/digitalocean"
-	_ "github.com/docker/machine/drivers/exoscale"
-	_ "github.com/docker/machine/drivers/generic"
-	_ "github.com/docker/machine/drivers/google"
-	_ "github.com/docker/machine/drivers/hyperv"
+	//_ "github.com/docker/machine/drivers/amazonec2"
```
nathanleclaire: These naturally will be added back in but are left out to speed up compiles and make the refactor simpler for now.

drivers/driver_options_mock.go
```diff
@@ -0,0 +1,21 @@
+package drivers
```
nathanleclaire: This was relied on by a couple modules and previously assumed to be in the same so it's been split out into its own exported file.

drivers/google/compute_util.go
```diff
@@ -97,7 +97,7 @@ func (c *ComputeUtil) staticAddress() (string, error) {
 		return "", err
 	}
 
-	if (!isName) {
+	if !isName {
```
nathanleclaire: Ah, these are accidental acts of `gofmt`.

ehazlett:
Huge +1 to this!  Great work!  I think this is a great start.  :+1: for the deletes!

This. Is. Awesome (to be said in a Spartan voice). :)

nathanleclaire:
So, the tests are hella broken, BUT!  I finally have a working POC standalone usage of `libmachine` which you can just `go run` and it Just Works.  And the generated store can be used interchangeably with the Docker Machine CLI (in this branch at least- I'm not 100% positive about `master`).

nathanleclaire:
My general thinking is, now that something relatively working-ish is put together, we can iterate on the design of the API and talk about next steps.  There's still a lot of work left to go, mostly converting existing drivers over to the new method, and discussing how to handle the change, but this script works out of the box now!

```console
~                                                                                  │➜  machine git:(git_r_done_libmachine) ✗ (default)
import (
        "fmt"
        "log"

        "github.com/docker/machine/libmachine"
        "github.com/docker/machine/libmachine/drivers/virtualbox"
)

func main() {
        libmachine.SetDebug(true)

        // returns the familiar store at $HOME/.docker/machine
        store := libmachine.GetDefaultStore()

        // over-ride this for now (don't want to muck with my default store)
        store.Path = "/tmp/automatic"

        hostName := "myfunhost"

        // Set some options on the provider...
        driver := virtualbox.NewDriver(hostName, "/tmp/automatic")
        driver.CPU = 2
        driver.Memory = 2048

        h, err := store.NewHost(driver)
        if err != nil {
                log.Fatal(err)
        }

        h.HostOptions.EngineOptions.StorageDriver = "overlay"

        if err := libmachine.Create(store, h); err != nil {
                log.Fatal(err)
        }

        out, err := h.RunSSHCommand("df -h")
        if err != nil {
                log.Fatal(err)
        }

        fmt.Printf("Results of your disk space query:\n%s\n", out)

        fmt.Println("Powering down machine now...")
        if err := h.Stop(); err != nil {
                log.Fatal(err)
        }
}
```

nathanleclaire:
My thinking around people wanting to create REST APIs and such with this is, you can unmarshal the `HostOptions` and `Driver` options onto the structs created via the factory methods before doing the create, and all the boilerplate set up for those structs will still be in place, but the user-defined options will over-ride all the other defaults.

ehazlett:
Looks good!  Could we get a list of remaining items so we could help knock those out?

nathanleclaire:
> Looks good! Could we get a list of remaining items so we could help knock those out?

Definitely!  I intend to squash down to a single commit as well soon (it will help with rebase hell down the line).

Here's my list so far:

## Must haves

- [ ] Fix unit tests (I intend to do this before handing off this PR and going on vacation)
- [ ] Fix integration tests (It's been a while since I've run them unfortunately and there may be things which I haven't foreseen that have broken.)
- [ ] Integrate all drivers to the new model of doing things.  This will be a decent chunk of rote work and I also intend to put together a guide, even if it is brief, on how to do it.  Mostly it is writing a new `NewDriver` method and adding them to the `switch` block in `drivers/drivermaker/maker.go`.
- [ ] Speaking of `NewDriver`, we need to introduce a uniform way to handle driver defaults.  That is to say, we have defaults in the flags today which we can no longer reliably expect the CLI to handle for machines created with `libmachine`.  So, we need to introduce into each driver, a set of per-module level default `const`s for flags that get re-used for the flag defaults _and_ struct defaults, i.e. `defaultCPU`, `defaultMemory` for VirtualBox.  I'll take a look at using that one as an example to show what I mean.
- [ ] There are places where we are reading from / setting environment variables in `libmachine` and they are changing behavior.  I've debugged what I can so far of this but I know there are still instances lurking around the corner waiting to bite us.  READING AND WRITING TO ENVIRONMENT VARIABLES SHOULD NEVER EVER HAPPEN INSIDE THE `libmachine` PACKAGE!!!!!  We have to fix every single instance of this or it will cause really sneaky bugs.
- [ ] There may be a bug persisting the `ArtifactPath` (it gets over-written as an empty string on some subsequent commands after `Create`), but I haven't debugged it yet.  So, that needs to be fixed.
- [ ] Documentation for the API, and preferably some more examples as well.  In general I think we should probably just have the documentation for the API on Github and this will allow us to move quickly on updating it.  When it becomes fairly stable, we can talk about how / where we want to put it on the official docs site.
- [ ] Create a real `libmachine.Client` struct.  Possibly, we can make it embed `store.Store`, so that you can just call `mcnClient.NewHost`, `mcnClient.Load`, etc. directly.
- [ ] I haven't figured out a good way to pass the `Host`-level `SwarmOptions` information to the drivers yet, so that's just been punted for now :(  Maybe they will need to get passed in to `NewDriver` methods, and `BaseDriver` will just embed `SwarmOptions` as well.

## Nice to haves

- I think that we need a per-driver `log` in addition to the "global" level `log`.  This will be a pretty good size chunk of search-replace work to integrate into all drivers though.
- Roundup easy / obvious PRs to the core and merge them before merging this in so contributors don't have to go through rebase hell.
- Add more examples of usage to `examples` module.
- There are some methods in `drivers.go` that seem kind of spurious to me (they should be moved to a different module), but maybe we can delay that to another time.
- Before subjecting other people to the API, I highly suggest that we rename `store.Get` to `store.Load`  ;P (`Save` and `Load` makes more sense to me than `Save` and `Get`)
- Again, maybe a bikesheddy point, but I _really_ want to rename `Host` to `Machine`.  It's just way more consistent - "I use `libmachine` to create 'hosts'" vs. "I use `libmachine` to create 'machines'".  Or at least rename `host` module to something like `hostmanage`.
- I know I sound like a broken record on this point, but the longer we put off a real "Provider" model the worse off we will be.  Others downstream will want to have a notion of a "Provider" that they can use to launch resources in and continuing down the path of managing API tokens etc. directly in the 'machine' drivers like we have so far just seems very wrong to me.
- Move swarm container provisioning to use `libcompose`?  I think it would allow us a lot more flexibility in terms of re-running provisioning (right now Swarm is the only thing that chokes when you re-run a provision on a finished machine, because you can't `docker run` the containers again if they're already there).
- I really don't like our current `ssh` stuff and want to clean it up to use variadic args (like `exec.Cmd`) and to return STDOUT and STDERR separately instead of mashing them together like we do today.
- Game plan for moving `libmachine` into its own repo, and the drivers we will be maintaining into _their_ own repos.  It would allow us to move faster on both projects and keep the API barrier nicely separated.  If Machine CLI and API stay in the same repo it is a huge risk that the boundary will bleed over between them again and totally ruin the hard work to get a clean separation between them.
- The "Artifact" model as it is now is a little hacky and needs some love / design discussion.  The basic dilemma is that drivers need to be able to write to the local filesystem somewhere to store, say, SSH keys BUT I wanted to ensure in `libmachine` that this was kept as separate from storing host metadata (`Store`) as possible.  So, it's a bit hacky today as each driver has a `LocalArtifactPath` (for per-machine artifacts like hard disks and SSH key files)  and `GlobalArtifactPath` (almost exclusively for ISOs in the current implementation) method.  There may well be a better way.
- Really would like to switch the "action" strings over to a proper type and expose the concurrent action stuff to the outside world, it's pretty great and I bet people will want to use it.

Side note: I've been thinking about how to integrate this with plugins and wondering, maybe we should just have one `plugin` driver which wraps the real-deal binaries?  Interesting food for thought at least.

nathanleclaire:
There seems to be a bug with `CaPrivateKeyPath` as well (I think reading from the environment when it shouldn't be).

nathanleclaire:
No one's said anything at : https://github.com/docker/machine/issues/1730 so I'm just going to go ahead and remove the useless coverage check for Travis here.

nathanleclaire:
Still a long way to go but all the unit tests are passing again :)

nathanleclaire:
I had to change some of the migration code to clean up some of the env-var reading bits so we should make sure to test that machines created with 0.1.0 still work (I would do it but my bandwidth here is bad and 0.1.0 doesn't play nice with Virtualbox 5).

nathanleclaire:
(All of the `getHostState`-type code in `host/host.go` needs to be moved to `commands` too).

libmachine/drivers/drivermaker/maker.go
```diff
@@ -0,0 +1,29 @@
...
+func NewDriver(driverName, hostName, artifactPath string) (drivers.Driver, error) {
+	var (
+		driver drivers.Driver
+	)
+
+	switch driverName {
+	case "virtualbox":
+		driver = virtualbox.NewDriver(hostName, artifactPath)
+	case "digitalocean":
+		driver = digitalocean.NewDriver(hostName, artifactPath)
```
nathanleclaire: Note the lack of an error return.  There's not that much which can go wrong here.  So, all of the other drivers need to be updated to not return errors as well in their "factory" methods.

nathanleclaire:
Mmm, also I am thinking it is time we should switch `AuthOptions` to embed https://github.com/docker/docker/blob/master/pkg/tlsconfig/config.go .  That would allow us to hand off to native Docker code / libcompose much more easily.  It would require a config migration, though.

nathanleclaire:
Also, should we rename `drivermaker` to `driverfactory`?  I felt conscious since Factory is so over-abused, but it definitely is one.

hairyhenderson:
> should we rename drivermaker to driverfactory?

+1 - if it's a factory, call it a factory ;)

hairyhenderson:
> Mmm, also I am thinking it is time we should switch `AuthOptions` to embed https://github.com/docker/docker/blob/master/pkg/tlsconfig/config.go . That would allow us to hand off to native Docker code / libcompose much more easily. It would require a config migration, though.

In theory yes, but this might be something best done in a follow-up PR :)

commands/config_test.go
hairyhenderson: Will new unit tests for the new code appear at some point? If so, it might be useful to keep this empty (mostly) file around as a reminder that tests need to be written...

libmachine/drivers/base.go
```diff
@@ -5,36 +5,19 @@ import "path/filepath"
...
 
 // GetSSHKeyPath -
 func (d *BaseDriver) GetSSHKeyPath() string {
-	return d.ResolveStorePath("id_rsa")
-}
-
-// ResolveStorePath -
-func (d *BaseDriver) ResolveStorePath(path string) string {
-	return filepath.Join(d.storePath, path)
+	return filepath.Join(d.ArtifactPath, "machines", d.MachineName, "id_rsa")
```
hairyhenderson: Why not keep it as
```go
return d.LocalArtifactPath("id_rsa")
```
like before?

hairyhenderson:
@nathanleclaire - I see that you've moved the `drivers` dir under `libmachine` - I'm missing something obviously... why was this necessary? Seems to me it'd be a much smaller delta if they were left where they were, especially in light driver plugins that will destroy that anyway.

commands/config_test.go
nathanleclaire: I felt that way before, but now I actually think the opposite is true.  It's very hard to tell just from looking at the filesystem which parts of the code have tests and which need them.  We haven't had any "Hey, noticed there was an empty test file so I submitted a patch" type contributions, so I think we just have to be stricter in code review.

libmachine/drivers/base.go
nathanleclaire: Ah yeah good point -- I'll make sure to change that when I get a chance.

nathanleclaire:
Good callouts @hairyhenderson -- I'll update to address them when I get back.

nathanleclaire:
If you want to see what this looks like in action (+ the provisioning work from https://github.com/docker/machine/pull/1598), I made https://github.com/nathanleclaire/moby to show off a little :)

libmachine/main.go
```diff
@@ -0,0 +1,89 @@
+package libmachine
```
ehazlett: This file should be named `libmachine.go` not `main.go`.

ehazlett:
I just started playing with a sample RPC implementation.  The first glaring thing is that this isn't going to be very friendly with our existing drivers (one of the big goals for plugins).  Almost every single import path has changed.  Is there a reason for that?

libmachine/main.go
nathanleclaire: Fair

nathanleclaire:
> The first glaring thing is that this isn't going to be very friendly with our existing drivers (one of the big goals for plugins)

Could you elaborate on what you mean by "this" here?  The whole PR, the specific changes to the drivers code, etc.?

> Almost every single import path has changed. Is there a reason for that?

A few reasons:

1. Some things which are used by `libmachine`, are outside of that module today (e.g. `github.com/docker/machine/utils`).  That breaks the modularity of that abstraction, so they have to be moved inside of `libmachine` to make it independently usable.
2. Some of the import path re-writing was more in the wheelhouse of "cleanup work", but mostly to facilitate the goal of making the `libmachine` more modular / independently usable.
3. In the case of things in `drivers` module, error on my part.  I think I got a bit carried away with that and it's probably best to keep them outside of `libmachine` as @hairyhenderson noted.

I know we want to be able to drop in existing drivers as much as possible, but I think we need to admit to ourselves as well that there are several things about the existing driver model which need to be improved because changing them after making driver plugins etc. seems like it could be astronomically more difficult than changing them today.  I'm curious to know where your specific concerns lie.  Import paths should be pretty easy to fix with tools like `goimports`.

hairyhenderson:
> but I think we need to admit to ourselves as well that there are several things about the existing driver model which need to be improved because changing them after making driver plugins etc. seems like it could be astronomically more difficult than changing them today

Could you call out these things? IMO the best course of action right now would be to get the driver plugin model going as soon as possible.

Also, we're still at 0.x - semantically, this means we can still break APIs :wink:. I think we'll have a much better idea of what exactly needs fixing about the driver model _after_ the plugin transition. Some of it might be hard to change later, but I thing getting plugins going faster is way more beneficial to our users.

ehazlett:
I agree with @hairyhenderson.   The original intent of libmachine refactor was to support the plugins, not to be a production grade external library.  The scope has increased quite a bit.  I'm not saying that's bad, I'm just saying that even running `goimports` on 14 drivers is going to be tough and when we stated that we want to be as drop-in replaceable as possible, it was to not force everyone to rewrite the drivers or deviate from the current driver design.  I agree libmachine needs a lot of work (it was never intended to be used externally from the beginning) but we also have to keep in mind the end goal.  Getting a workable plugin system is more important at this stage rather than a perfect libmachine.

ehazlett:
> The first glaring thing is that this isn't going to be very friendly with our existing drivers (one of the big goals for plugins)

What I meant by this was I tried to just take the existing VirtualBox driver and update.  Most were path conflicts (which hopefully would be alleviated by `goimports`) but clearly requires some decent level of intervention across all drivers.

It's very possible my thoughts were far too utopian, but my original hope was to have a lightweight RPC wrapper that would simply instantiate the driver and call the existing methods.  We are already going to have to refactor a bit for the logging.  I just didn't want to have to go too far beyond that.  As said, maybe too far idealistic.

/cc @ibuildthecloud @crunchywelch @miqui

crunchywelch:
Thanks for all the awesome work so far on this! Looking good to me so far.

+1 to this: `I think that we need a per-driver log in addition to the "global" level log. This will be a pretty good size chunk of search-replace work to integrate into all drivers though.` I'd love also to be able to plug in other logging providers, for example I have a small fork of machine that logs to logentries, which is awesome when you are using machine to automate things like tests (which we do).

FWIW, regarding the driver refactor and path changes, I took a look at the few that have been ported already, and it doesn't seem like that big of a deal to me. I know it's going to be a fair amount of work, but it's mostly just busy work and I think folks are expecting to have to do at least some refactor to get their drivers working with the new system. These changes don't seem unreasonable to me I guess, but I do appreciate the effort to make the impact as small as possible!

commands/commands.go
```diff
@@ -8,56 +8,55 @@ import (
...
+
+	"github.com/docker/machine/commands/mcndirs"
+	"github.com/docker/machine/libmachine/cert"
+	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/log"
+	"github.com/docker/machine/libmachine/persist"
 )
 
 var (
+	stateTimeoutDuration  = time.Second * 3
```
nathanleclaire: @mattjmcnaughton Here's the relevant bit!

nathanleclaire:
Update:  All drivers have been updated to use the new method, and integrate with the compiled binary!

commands/commands.go
mattjmcnaughton: Thanks @nathanleclaire - once this is merged I'll tackle the change I was working on. Thanks for all your work!

nathanleclaire: :+1:

ehazlett:
Tests are failing

nathanleclaire:
Yea, I've had some unfortunate rebase drama.  Working on it!

nathanleclaire:
Tests are fixed.

nathanleclaire:
I feel confident about merging this very soon -- currently I am running through the integration tests to check up, but all new development work on this branch is completed, the unit tests are passing, and I'm really keen to move this upstream ASAP and get started PRing the plugin work.

I do want to make a followup PR to add some documentation and more examples, though.  Additionally, there seems to be at least one issue with cert placement when using the standalone program.

nathanleclaire:
screen shot 2015-09-17 at 6 38 35 pm

These integration tests are failing with Virtualbox for this branch, going to try with DigitalOcean and see which ones are consistently failing, I'm not sure which failures might be inherited from master as well.

nathanleclaire:
The `env_shell` is one of my own doing, at least.  I had added that in to account for the unit test(s) for config and env I had taken out.  I'd say this PR is "ready to merge pending fixed integration tests" status.

nathanleclaire:
OK, everything looks G2G now on integration tests:

image

nathanleclaire:
So I'm ready to merge this whenever comfortable, but first there are a few things I want to merge upstream and rebase to.

nathanleclaire:
It should be noted, though, that while this is a backwards-compatible change, it is not necessarily forwards-compatible (due to a config migration) -- so, it is a little dangerous to use older versions of Machine after using this on hosts you have created.

hairyhenderson:
> it is a little dangerous to use older versions of Machine after using this on hosts you have created.

probably out of scope for this PR, but I wonder if we should prompt before migration so users are duly warned...

nathanleclaire:
> probably out of scope for this PR, but I wonder if we should prompt before migration so users are duly warned...

Or at least make a backup copy of their old config before attempting migration, etc.?  I'm more concerned about users (not neccesarily intentionally) switching back and forth between old versions of Machine (currently it will FUBAR your config to do so if it has been migrated) than issues with the migration itself.

nathanleclaire:
I've added a `config.json.bak`, as well as made the store save the JSON in "pretty" format for easier editing on disk just in case we need to instruct a small subset of users to edit the JSON manually.

nathanleclaire:
I know of at least one issue using this lib independently but this PR LGTM

zchee:
@nathanleclaire Awesome :tada:

nathanleclaire:
:+1:

hairyhenderson:
> I've added a `config.json.bak`, as well as made the store save the JSON in "pretty" format for easier editing on disk just in case we need to instruct a small subset of users to edit the JSON manually.

:+1: awesome!

also, yay for merge! :D

briceburg:
@nathanleclaire curios here -- does the merging of libmachine mean we'll be able to interface with machines from the `docker` command (without having to eval first)? Something like `docker --machine ocean-1 ps` ?

Thanks!

nathanleclaire:
@briceburg Not exactly -- `libmachine` is intended to be more like an SDK for Docker Machine, written in the Go language.  There's some rough edges in the current implementation, though.

dweomer:
+1 on the branch name

PR #1743, opened 2015-08-20
commands: show "-" in ACTIVE column for inactive machines

### LINKED ISSUE #1682: "Active" star in ls output breaks UNIX-ness of program
Notice that if you have an active machine something like this will return unreliable results:

```console
$ docker-machine ls | awk '{ print $3; }'
DRIVER
Running
Running
Stopped
Running
virtualbox
```

:frowning: 

`awk` delimits by whitespace so I think we should have a character to indicate non-active as well, or come up with another solution to make the output a bit more UNIX-ey.  Even though it's meant to be just for humans, I know of at least two or three examples of people parsing this output mechanically.

cc @ehazlett

skatsuta:
close #1682 

The output is like this:

```
$ docker-machine ls
NAME            ACTIVE   DRIVER       STATE     URL                         SWARM
dev             *        virtualbox   Running   tcp://192.168.99.100:2376
swarm-master    -        virtualbox   Stopped                               swarm-master (master)
swarm-node-00   -        virtualbox   Stopped                               swarm-master
```

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -66,7 +66,7 @@ func cmdLs(c *cli.Context) {
 	sortHostListItemsByName(items)
 
 	for _, item := range items {
-		activeString := ""
+		activeString := "-"
 		if item.Active {
 			activeString = "*"
 		}


thaJeztah:
thanks! LGTM

nathanleclaire:
Nice!

LGTM

cc @ehazlett

ehazlett:
LGTM

PR #1780, opened 2015-08-30
Remove redundent assignment to InsecureSkipVerify

paultag:
This second codepath is redundent, since tlsConfig.InsecureSkipVerify is
assigned to earlier in that function body. In addition, there's no need
to wrap the assignment in a check on that boolean value in this case.

Signed-off-by: Paul Tagliamonte <paultag@gmail.com>


ehazlett:
LGTM

paultag:
Thanks!

PR #1791, opened 2015-09-01
Remove flag parsing operator requirement for docker-machine ssh, and allow SSH command binary args to be passed

nathanleclaire:
This removes the requirement for the `--` operator when using SSH, e.g.:

```console
$ docker-machine ssh default df -h
```

will work again.

It also allows the use of additional SSH command arguments, such as `-A` for SSH agent forwarding:

```console
$ docker-machine ssh default -A
```

The syntax is a little weird since the SSH args come after the machine name but it's better to have the feature than nothing I think.  People really missed this about boot2docker.

cc @ehazlett @SvenDowideit @hairyhenderson @sthulb  WDYT

__NOTE__:  If there are existing scripts with `ssh --`, it's possible they may break using this method, so maybe we can account for that / leave out that arg if it exists?

TODO:

- [x] Docs
- [x] Integration tests


hairyhenderson:
Looking good! :+1:

ehazlett:
Ya the syntax is a bit odd but I think overall it's good.  I would rather have the ability to pass the options than not.  :+1:

nathanleclaire:
cc @moxiegirl for docs review

docs/reference/ssh.md
```diff
@@ -62,6 +60,17 @@ cgroup                  499.8M         0    499.8M   0% /sys/fs/cgroup
 /mnt/sda1/var/lib/docker/aufs
 ```
 
+If you are using the "external" SSH type as detailed in the next section, you
+can include additional arguments to pass through to the `ssh` binary in the
```
moxiegirl: This clause is ambigous: `the default arguments for the command generated by Docker Machine`  what do you mean by "the command generated by Docker machine`  are you referring to the "ssh" command provided in your example or something else.  

Also unclear, does  this limitation only apply to the Go native `ssh`?  

Also, this paragraph doesn't really belong in here with the forward reference. In general, this should be an edge case that a user is using the Go native driver. If we say to them, don't set the `ssh` implementation explicitly and they don't, the user shouldn't even be aware which implementation they are using. So, they won't be aware enough to use the feature you point out.

I'm going to restructure the content of this command a bit.  I'll put your new content in the correct location in the file.  If you are willing, please use the Gist I provide.  If you aren't willing, I'm happy to carry.

--> https://gist.github.com/moxiegirl/dd03c1fea513342042a9



# ssh

Log into or run a command on a machine using SSH. To login, use the 
`docker-machine ssh machinename` command, for example:

```
$ docker-machine ssh dev
                        ##        .
                  ## ## ##       ==
               ## ## ## ##      ===
           /""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~ ~ /  ===- ~~~
           \______ o          __/
             \    \        __/
              \____\______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|
Boot2Docker version 1.4.0, build master : 69cf398 - Fri Dec 12 01:39:42 UTC 2014
docker@default:~$ 
```
The system displays a splash screen indicating the machine's driver. Then, it
displays the prompt inside the running machine. You can interact with the
command line as in any shell:

```
docker@boot2docker:~$ ls /
Users/   dev/     home/    lib/     mnt/     proc/    run/     sys/     usr/
bin/     etc/     init     linuxrc  opt/     root/    sbin/    tmp      var/
```

To leave an SSH session, type  `exit` at the prompt.

```
docker@default:~$  exit
```

Similar to the regular `ssh` program, you can run commands inside a machine
remotely by passing a command together with the `docker-machine ssh` command line:

```
$ docker-machine ssh dev free
             total         used         free       shared      buffers
Mem:       1023556       183136       840420            0        30920
-/+ buffers:             152216       871340
Swap:      1212036            0      1212036
```

When you run a command, you don't enter the shell. Instead, the command runs in
the shell and exits returning you to your local prompt. You can also pass flags
to the commands you run:

```
$ docker-machine ssh dev df -h
Filesystem                Size      Used Available Use% Mounted on
rootfs                  899.6M     85.9M    813.7M  10% /
tmpfs                   899.6M     85.9M    813.7M  10% /
tmpfs                   499.8M         0    499.8M   0% /dev/shm
/dev/sda1                18.2G     58.2M     17.2G   0% /mnt/sda1
cgroup                  499.8M         0    499.8M   0% /sys/fs/cgroup
/dev/sda1                18.2G     58.2M     17.2G   0%
/mnt/sda1/var/lib/docker/aufs
```

## The SSH implementation

The first time `docker-machine` is invoked, it first checks to see if you have
the `ssh` binary locally. If you do, Machine attempts to use the local `ssh` for
the SSH commands it needs to run. It does this  whether the commands are a part
of an operation such as creation or were requested by the user directly. 

If it does not find a `ssh` binary locally, Machine uses the native Go
implementation from [crypto/ssh](https://godoc.org/golang.org/x/crypto/ssh).
This is useful in situations where you may not have access to traditional UNIX
tools, for example, if you are using Docker Machine on Windows without an
`msysgit` installation.

Typically, you should allow Machine to handle this implementation detail for
you. However, if you want to ensure Machine uses the Go native version, you can.
To do this, set a global command line flag / environment variable like so:

```
$ docker-machine --native-ssh ssh dev
```

If you are using the Go native binary, you can also include additional  `ssh`
arguments to pass through to the `ssh` binary.  You can do this as long as the
arguments do not conflict with the default arguments for the command generated
by Docker Machine. For instance, the following command forwards port `8080`
from the `default` machine to `localhost` on your host computer:

```
$ docker-machine ssh default -L 8080:localhost:8080
```

There are variations between the standard `ssh` behavior and the Go native
version, please report any issues or inconsistencies you come across to [Docker
Support](https://support.docker.com).

moxiegirl:
ping @theJetzah for any nits

nathanleclaire:
@moxiegirl, such a comprehensive re-write of the documentation in the `ssh` section seems pretty out of scope for what this PR is changing, how about we go ahead and merge as-is and you can follow up later with a PR if to iterate on it?

moxiegirl:
@nathanleclaire  That is what a carry is --- I'll make the changes in another PR

If you want me to do that, please answer the technical questions in the first part of my comments, those do apply to your specific paragraph addition and are still outstanding.

thaJeztah:
Erm, want me to review this, or wait for the carry for nitpicking? :smile:

nathanleclaire:
My understanding was that "carrying" a PR included finishing up work in an existing PR, but not introducing new work.  That's a new PR with a different scope.

> This clause is ambigous: the default arguments for the command generated by Docker Machine what do you mean by "the command generated by Docker machine` are you referring to the "ssh" command or the command passed to the shell or something else?

The `ssh` command generated by Docker Machine for execution by the shell.

> Also unclear, does this limitation only apply to the Go native ssh?

Which limitation?  The inability to pass SSH command line arguments to it?  If so, then yes.

> Also, this paragraph re Gp native doesn't really belong in here with the forward reference. In general, this should be an edge case that a user is using the Go native driver. If we say to them, don't set the ssh implementation explicitly and they don't, the user shouldn't even be aware which implementation they are using. So, they won't be aware enough to use the feature you point out.

If it's not explicitly called out, I find that users just assume every "default" feature is available with every option.  They should absolutely be able to figure out that they are using "native ssh" or "external ssh" if needed.  Docker Machine will default to native SSH if the `ssh` binary is not found, so it's a default behavior in some cases as well.

moxiegirl:
@nathanleclaire Thanks LGTM on your initial paragraph answers. I'll incorporate that perspective into the command rewrite PR.  So, I'm cool with merging and I'll immediately open a replacement if you are more comfortable with doing it that way.  This PR is ok to merge as is. LGTM

@thaJeztah Let's nit on the replacement rather than this as mine will be a rewrite of the command entire.

thaJeztah:
@moxiegirl gotcha, will save my time for that one :dancer:

nathanleclaire:
Thanks @ehazlett I will be ready to merge this once I add / update integration tests.

ehazlett:
Sounds good.  I know a lot of people will love this :)

nathanleclaire:
Boom, tests updated and added a new one for this feature.

I'm ready to merge this whenever you are comfortable @ehazlett .

ehazlett:
LGTM

nathanleclaire:
:+1:

kevinsimper:
@nathanleclaire This happens when i try to set up a tunnel, it justs gets stuck and I can't even Ctrl-c out of it.
```
kevinsimper$ docker-machine-dev --version
docker-machine-dev version 0.5.0-dev (ebd45f5)
kevinsimper$ docker-machine-dev ssh kubernetes -L 8080:localhost:8080

```

Running a ps aux gives this
```
kevinsimper      9498   0.0  0.0 573388680   6288 s002  S+    9:51PM   0:00.02 docker-machine-dev ssh kubernetes -L 8080:localhost:8080
```

VSZ is unusually high!

PR #1809, opened 2015-09-05
Filter openstack floating IPs by tenant ID

j2sol:
If the user in question happens to be an admin user, they may see
floating IPs that are from other tenants. Attempts to use such IPs will
result in an error, as the tenant of the IP doesn't match the tenant of
the instance. Filtering by ID limits the returned IPs to those that are
valid for the tenant.

Unfortunately the filtering has to be done by tenant ID, not name, so we
need to interact with the identity end point to map the tenant name to
ID if only the name was provided. Currently this only supports the v2
identity endpoint.

Signed-off-by: Jesse Keating <jkeating@j2solutions.net>

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -391,6 +391,7 @@ const (
 	errorUnknownFlavorName       string = "Unable to find flavor named %s"
 	errorUnknownImageName        string = "Unable to find image named %s"
 	errorUnknownNetworkName      string = "Unable to find network named %s"
+	errorUnknownTenantName       string = "Unable to find tenant named %s"
 )
 
 func (d *Driver) checkConfig() error {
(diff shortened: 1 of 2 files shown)


drivers/openstack/openstack.go
```diff
@@ -516,6 +517,27 @@ func (d *Driver) resolveIds() error {
 		}).Debug("Found floating IP pool id using its name")
 	}
 
+	if d.TenantName != "" {
```
ggiamarchi: If think it would be better to check `if d.TenantId == ""` rather than `if d.TenantName != ""` because it's common to have an OpenStack rc file containing both the tenant id and the tenant name. In such a situation it would be a pity to compute the id from the name.

dmp42:
@j2sol can you address @ggiamarchi comments?
@ggiamarchi save your one comment, is this going in the right direction?

Thank you both!

j2sol:
Yup, I'll address it.

dgageot:
Merged with https://github.com/docker/machine/pull/2589

PR #1816, opened 2015-09-07
Adding FUSE HGFS mount option

frapposelli:
This is to enable the new FUSE HGFS feature with `open-vm-tools` 10.x, FUSE mount is going to be the default for any Linux Kernel > 4.x.

This maintains backwards compatibility with previous `open-vm-tools` versions that used the `vmhgfs` kernel module.


nathanleclaire:
LGTM cc @ehazlett

PR #1522, opened 2015-07-14
Full rewrite of the vSphere driver

frapposelli:
This is a complete rewrite of the vSphere driver, this drops the `govc` requirement by using the `govmomi` library directly.

Also enhances auto detection of resources and general performance improvement.

Passes bats test:

```shell
machine · vspheredriver-with-govmomi ⟩ env DRIVER=vmwarevsphere \
                                       VSPHERE_DATASTORE="nfs_fas2020_15k_01" \
                                       VSPHERE_VCENTER="monster.ad.lab.gosddc.com" \
                                       VSPHERE_NETWORK="VM Network" \
                                       VSPHERE_USERNAME="root" \
                                       VSPHERE_PASSWORD="xxxxxxxxxxxx" \
                                       ./test/integration/run-bats.sh ./test/integration/core/core-commands.bats
=> ./test/integration/core/core-commands.bats
 ✓ vmwarevsphere: machine should not exist
 ✓ vmwarevsphere: create
 ✓ vmwarevsphere: ls
 ✓ vmwarevsphere: run busybox container
 ✓ vmwarevsphere: url
 ✓ vmwarevsphere: ip
 ✓ vmwarevsphere: ssh
 ✓ vmwarevsphere: docker commands with the socket should work
 ✓ vmwarevsphere: stop
 ✓ vmwarevsphere: machine should show stopped after stop
 ✓ vmwarevsphere: url should show an error when machine is stopped
 ✓ vmwarevsphere: env should show an error when machine is stopped
 ✓ vmwarevsphere: machine should not allow upgrade when stopped
 ✓ vmwarevsphere: start
 ✓ vmwarevsphere: machine should show running after start
 ✓ vmwarevsphere: kill
 ✓ vmwarevsphere: machine should show stopped after kill
 ✓ vmwarevsphere: restart
 ✓ vmwarevsphere: machine should show running after restart

19 tests, 0 failures
```

PR #1829, opened 2015-09-09
Credentials and server can be passed in GOVC_ environment variables.

emsi:
vsphere/ESXi server IP and credentials are not mandatory on command line as long as they are set in GOVC_ env variables.


ehazlett:
This probably needs docs updated as well.

frapposelli:
It's true, but we're going to get rid of the dependency on `govc` so still want to have them mandatory so transition will be smooth, see #1522

frapposelli:
Also, you can still set them through the `docker-machine` env vars.

emsi:
You are right. It can be done via machine VMWARE_ variables rather that GOVC.
I'm closing it.

PR #1830, opened 2015-09-09
Clarify Install Location

### LINKED ISSUE #1827: No download link to Docker Machine for Linux
I've been trying to look at https://docs.docker.com/machine/ https://docs.docker.com/machine/get-started/ and https://docs.docker.com/linux/started/ and can't find a link to Download Docker Machine for Linux.

Is Docker Machine only available for OS X and Linux?

ehazlett:
This adds a link to the release page and calls out the install section.

Closes #1827

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -26,6 +26,11 @@ managing them. Using these commands you can
  - upgrade the Docker client and daemon
  - configure a Docker client to talk to your host
 
+## Install
+You can install Docker Machine by installing the [Docker Toolbox](https://www.docker.com/toolbox
+) or to get the binary only, visit the [GitHub Releases](https://github.com/docker/machine/releases)
+for the latest release.
+
 ## Understand Docker Machine basic concepts
 
 Docker Machine allows you to provision Docker on virtual machines that reside either on your local system or on a cloud provider. Docker Machine creates a host on a VM and you use the Docker Engine client as needed to build images and create containers on the host.


docs/index.md
```diff
@@ -26,6 +26,11 @@ managing them. Using these commands you can
  - upgrade the Docker client and daemon
  - configure a Docker client to talk to your host
 
+## Install
```
moxiegirl: I would tend to leave the heading off all together as the sentence would come at the end of the bullet list in the intro section --- it gives enough emphasis.  If you don't like that, I"d tweak the heading level and the head itself so as not to imply these are instructions which the command 

`#### Need the installation instructions?`

nathanleclaire:
LGTM

nathanleclaire:
Thanks for doing this

dmp42:
Maybe a word about using homebrew? wdyt?

moxiegirl:
@ehazlett Are we advising them to use brew?  If we want them to use brew we should update the installation docs --- this is just a pointer about where to find them.

dmp42:
I would assume most reasonable people on OSX would pick a brew solution first.
If we are happy with brew formula quality, I would point it as the main solution - getting binaries manually as an alternative.

moxiegirl:
@dmp42 On OSX, we are actually expecting most people to start off with Toolbox which installs Machine and many other things including Kitematic.

On the Engine side, we seemed to be discouraging tools like `brew` in favor of our installation mechanisms.  This is because things like brew and other similar package managers weren't being kept up-to-date.

nathanleclaire:
> This is because things like brew and other similar package managers weren't being kept up-to-date.

This would be my concern with recommending `brew` as well.

dmp42:
Expecting people to go with Toolbox as a first choice is certainly fine, and is already the suggestion here. This discussion is about the second choice (binaries + manual finickying vs. brew)

About homebrew being up to date, they seem to be doing a good job at job: https://github.com/Homebrew/homebrew/commits/master/Library/Formula/docker-machine.rb

moxiegirl:
Is there a reason we just wouldn't recommend curl for OSX which is what the other Linux platforms use.  Omitting brew from our documentation doesn't prevent people from using it.  It **does** simplify our recommendations and keep us internally consistent with our other projects.

nathanleclaire:
Yeah, I agree with @moxiegirl -- I think we risk inconsistency if we recommend brew, and shouldn't do it unless we're willing to commit to officially supporting it (i.e. we maintain the `brew` recipe directly).

dmp42:
> Is there a reason we just wouldn't recommend curl for OSX which is what the other Linux platforms use. 

Overall: 

 * dangerous (no verification of any sort)
 * cumbersome (manual change of permissions)
 * anti-friendly as far as upgrading/maintenance is concerned

There is even a "wall of shame" for softwares recommending `curl | sh` (what we do here is basically the same thing): http://curlpipesh.tumblr.com/

I certainly agree there is some level of endorsement if we are going to recommend using brew - but at the very least we should stop detailing steps that are counter to best practices (curling without any verification, using `/usr/local/bin` on OSX for eg).

2 cents.

nathanleclaire:
I agree with all of your concerns @dmp42 , but we followed Compose's lead on this one: https://github.com/docker/compose/releases (although they do have an optional installation through a package manager (pip) as well).  I think that before either project changes the recommendation, we need:

1. Be agreed upon by both projects that this is now the recommended install, and present it in a uniform way.
2. An assigned team member who is clearly responsible for updating the brew recipes, otherwise there is a risk that it will not get done adequately.

nathanleclaire:
After discussion with @dmp42 we've decided to move forward with this and we can follow-up discussion and further changes in new PRs.  Thanks @ehazlett @moxiegirl

moxiegirl:
@nathanleclaire this actually didn't have my docs LGTM.  I'll carry the changes I asked for in another PR.

PR #1831, opened 2015-09-09
VMware Workstation driver

zaps:
First attempt at a VMware Workstation driver. Most of the code is direct duplicate of the Fusion driver with minor tweaks.

Current limitations:
* Only works on Windows version of Workstation
* Paths to vmrun, DHCP client table, and shared "Users" folder are hard-coded (with sensible values)
* Assumes 64-bit Windows


emsi:
+1
nice work. There should be a way to provide VMWare installation path on command line and hence use relative paths to it or fall back to default hard-coded values :)

zaps:
@emsi great idea! I will work on that.

zaps:
@emsi I've added you as a collaborator in my fork in case you would like to contribute. I'm sort of learning Go as I... "go" :smile: so if you have more experience it might be faster.

drivers/vmwareworkstation/vmrun.go
```diff
@@ -0,0 +1,66 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/docker/machine/log"
+)
+
+// TODO un-hardcode these locations
```
nathanleclaire: You could use an `exec.LookPath` for this which falls back to a `os.Stat` of the hardcoded paths to check if they're around.

nathanleclaire:
We have a CoreOS provisioner already IIRC, so that addition shouldn't be neccessary.

I'm going to close this for now (we're not accepting drivers into the core, we are moving to a pluggable model), but stayed tuned to make this into a driver plugin.  I have the initial versions of that (plugin) code done now too so contact me nathan@docker.com if you'd be willing to try it out :)

emsi:
@zaps: thanks. I'm new to golang as well (just did few obvious hacks) but I found your patch extremely useful.  :)

masaeedu:
@nathanleclaire Is this closed in favor of some other issue for the plugin version?

nathanleclaire:
@masaeedu Perhaps @zaps can comment on whether they have continued development in their own repo.  Feel free to PR and add it to the `AVAILABLE_DRIVER_PLUGINS.md` as well if so.

PR #1841, opened 2015-09-11
Integration tests & build system enhancements

dmp42:
This is work in progress.
See detailed commits for more.


ehazlett:
So far looks good!  Thanks!

nathanleclaire:
@dmp42 Is this ready to go?

.dockerignore
```diff
@@ -1,2 +1,4 @@
 docker-machine*
-.git
```
dmp42: Ignoring git would break `Version.Commit` initialization at compile time. Removing it so we get actual commit hash in machine versions.

nathanleclaire: I have to say I'm still torn on this.  It will blow up the size of our build context significantly.  Why not re-work the scripts to pass the git SHA to the compile container as an environment variable and use that instead?

dmp42:
@nathanleclaire @ehazlett this is ready to go

The only point I'm not confident with is the release script.
@ehazlett can we look into this later?

.dockerignore
dmp42: It's 11MB currently.
The total size of the folder is about 50MB.
So, it's a 20% increase on the size of the build context.

wdyt?

dmp42: Are you concerned about speed? If so, it should not impact perf in any significant way.

nathanleclaire: Well, when we move to plugins, it should reduce the size of our directory hopefully pretty significantly due to all the vendored code.  But, perhaps we should move forward with this as-is and deal with that if/when it is a problem.

.travis.yml
```diff
@@ -1,15 +1,13 @@
 language: go
-sudo: false
 go:
-- 1.4.2
 - 1.5
 install:
-- export GOPATH=${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH
```
nathanleclaire: Why will we not need this anymore?

dmp42: Because of vendoring capabilities in golang 1.5, that basically removes the need for any GOPATH monkeying, or relying on calls to godep itself during normal build/test operations.
Furthermore, leveraging the make infrastructure let us keep that complexity out of the specific CI systems definition files.

.dockerignore
dmp42: FWIW, on my laptop, the difference between ignoring .git and not ignoring it is 0.10 seconds on average.
My 2 cents is that it's not worth the effort. Speed freaks (like me) either way tend to build natively instead of inside a container.

nathanleclaire:
LGTM

PR #1847, opened 2015-09-13
Add sleep on adapter creation so 'list hostonlyifs' provides the correct info. 

### LINKED ISSUE #1843: Duplicate host only adapters created when creating a VM with virtualbox, VM creation hangs
##### Description of problem:
I'm running the latest Docker Toolbox (1.8.2) on Windows 10
When creating my first VM docker-machine creates 2 network adapters, the 2nd one gets assigned to the vm with garbage settings and create process hangs.

##### The culprit
It appears that there is significant latency from when a `hostonlyif` is created to being able to accurately query `hostonlyifs`. When docker-machine checks to see if there is an already created `hostonlyif` the netmask returned from `list hostonlyifs` doesn't match the expected netmask, instead we see a value of 15.0.0.0. This causes docker-machine to create a 2nd `hostonlyif`.

I'm not sure if this a bug in virtualbox or just a quirk we'll have to deal with.

##### Environment details (AWS, VirtualBox, physical, etc.):
Windows 10. Docker Toolbox 1.8.2

##### How reproducible:
Very reproducible

##### Steps to Reproduce:
1. Start with a fresh Docker Toolbox install. 
2. Remove all VM's and orphaned host only adpaters in VirtualBox
3. Attempt to create the default vm with docker quickstart,kitematic, or docker-machine

##### Actual Results:
VM creation hangs and two host only network adapters are created
here's the docker-machine create with debug
http://pastebin.com/raw.php?i=i868Tt7c

##### Expected Results:
VM creation to succeed with only one host only adapter

rwilliams:
Fixes #1843

### DIFF
--- a/drivers/virtualbox/network.go
+++ b/drivers/virtualbox/network.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 )
 
 var (
@@ -160,6 +161,9 @@ func getOrCreateHostOnlyNetwork(hostIP net.IP, netmask net.IPMask, dhcpIP net.IP
 	if err := addHostonlyDHCP(hostOnlyNet.Name, dhcp); err != nil {
 		return nil, err
 	}
+
+	//sleep for a few seconds to allow virtualbox's network adapter creation to catch up
+	time.Sleep(5 * time.Second)
 
 	return hostOnlyNet, nil
 }


thaJeztah:
Hm, may be just me, but never a real fan of arbitrary "waits"; is there a way to *detect* if the interface is up?

rwilliams:
I'm all for a more programmatic alternative to the sleep call.  It's a quirk in virtualbox and I'm not sure if we're going to find an event we can listen for to know if `list hostonlyifs` is gonna return good info or not.

rwilliams:
I've opened a ticket with virtualbox.  https://www.virtualbox.org/ticket/14582

nathanleclaire:
I think @thaJeztah means if there is some way we can actually call out to Virtualbox and find out the specific moment when the interface comes up.  Otherwise, this fix will not work for all cases.  I'm not a fan of putting arbitrary sleeps in either, if we can avoid it.

rwilliams:
Closing this. Lets hope VirtualBox gets this fixed soon. I know we can check for a bad netmask returned from VirtualBox and just say ok that's the right adapter, but i feel like it might lead to issues on more complex setups.

Issue #1845, opened 2015-09-13
Machine name should be case insensitive


dmp42:
Using FileStore, case sensitiveness depends on the underlying filesystem, leading to inconsistent behaviors.

Typically on Linux, both "a" and "A" can exist - on OSX (by default), that won't work.

Assuming we have alternate stores in the future, case sensitiveness might be an issue as well.

2 cents: normalize names so they are always lowercased for the Store.

wdyt @ehazlett @nathanleclaire ?

dmp42:
Under a case sensitive filesystem, creating both "a" and "A" ends-up with the following weird output:

```
deploy@loft:~$ ./docker-machine ls
NAME   ACTIVE   DRIVER   STATE   URL    SWARM
a               none             none
a               none             none
```

A first `rm A` will be successful. A second `rm A` will fail, but `rm a` will succeed.

dmp42:
cc @dgageot @jeanlaurent

nathanleclaire:
Current proposed solution is to transform name to all lowercase before doing anything with a directory on disk.  Therefore `A` and `a` would always appear with their same original casing to users but reference only `a` on disk (therefore two machines with the same name but different casing would not be allowed).

PR #1848, opened 2015-09-13
Fix #1846

### LINKED ISSUE #1846: Valid machine name like "." or ".." don't work
Current hostname validation allows for such names, but that will break on the FileStore (obviously they "exist").

Suggestion: restrict hostname validation to prevent such names.

cc @ehazlett @nathanleclaire

dmp42:


thaJeztah:
(You may want to add `fixes #1846` to the description as well, not sure if GitHub looks at the title of a PR)

Looking at the regex, this PR would allow `--something`This regex would solve that;

`^[a-zA-Z0-9][a-zA-Z0-9_.-]*$`

I haven't looked deeply in the way this regex is used; is this just the name of a *machine* or also a *hostname*? If the latter, the regex should be slightly modified to disallow, for example `a...b` as a name (consecutive sequences of `.` or `-`, which can only be used as a separator in host names [rfc952](https://tools.ietf.org/html/rfc952), [rfc1123](https://tools.ietf.org/html/rfc1123#section-2.1)), something like this;

`^[a-zA-Z0-9]([a-zA-Z0-9][.-]?)*$`

(The above would still allow trailing `.` or `-`, I can come up with one that fixes that if needed)

Should machine names be case-sensitive?

dmp42:
Thanks @thaJeztah 

Case-sensitiveness, look here: #1845

I'm wary about not breaking things for current users. This fix here should just prevent non-currently working cases (".", "..").

Agreed that "--foo" should be prevented, though. Will amend to this.

A wider brainstorm about what should be a "hostname", and how that would translate on the store is certainly warranted, possibly as a later PR?

thaJeztah:
> A wider brainstorm about what should be a "hostname", and how that would translate on the store is certainly warranted, possibly as a later PR?

SGTM, I just "stumbled" on this PR, adding some thoughts on possible pit-falls here :-)

(Thanks for the link to https://github.com/docker/machine/issues/1845, I missed that one)

hairyhenderson:
@dmp42 - thanks! It'd be really useful to update the `TestValidateHostnameInvalid` test (in `host_test.go`) to add more examples of strings that shouldn't be valid. Otherwise, :+1:.

ehazlett:
LGTM

ehazlett:
@dmp42 if you want to add more invalid examples in the tests as @hairyhenderson mentioned that would be great.  i don't think this is blocked on it though.

nathanleclaire:
So, are we unconcerned about what effect this might have on users who currently have hostnames which validate this scheme?  It seems unlikely to be a problem but you never know-  If that's a risk we're willing to accept, I'm OK to merge.

dmp42:
@nathanleclaire 
This is a risk for the following names:

 * `.`
 * `..`
 * `-foo`

The first two cases don't even work (because of the FileStore), so, if we are breaking something, it's for names in the third case (dash something).

I believe we should accept this risk - but obviously open to discussion.

ehazlett:
I'm also ok with this risk.  LGTM

nathanleclaire:
LGTM

dmp42:
I re-enabled circleci now. Should work ok against master.
Will now progressively add more integration tests to it.

nathanleclaire:
Cool!

PR #1850, opened 2015-09-14
Migrate 'amazonec2' driver to AWS SDK

### LINKED ISSUE #858: Replace AWS driver
With the release of an AWS (alpha) SDK for Go, https://github.com/awslabs/aws-go

Using the official SDK has advantages over our current implementation since we wouldn't have to maintain lower level code.

euank:
Closes #858

This replaces the old custom ec2 client with the AWS maintained SDK.

This does not have any intentional behavioural differences other than log / error output differences.

This, unfortunately, does not have any more tests than the the driver previously did.
I've attempted to manually test all features of it.


dgageot:
Hi @euank is it something you still you still like to work on?

euank:
@dgageot absolutely; some other stuff has been distracting me from it, but I expect to be back at it soon.

If you could answer my query about whether there's any objection to adding in `gomock` or if there's a preferred way to mock dependencies here, I'd appreciate that as well.

I'll rebase and do the last bits of code cleanup and testing as soon as I have time.

jeanlaurent:
@euank about `gomock` can you do without it for now ?

We would be happy to help in any way to move this forward.

euank:
I've updated the original comment and title to reflect that I'd like this to be reviewed / considered for merge.
I significantly decreased the scope (removed the layers of abstraction I previously added, removed gomock and related test additions, no longer added other extraneous  changes).

Although I think the other changes in the previous version of this PR were valuable, I think it would be easier all around if this were kept small and the other issues were addressed separately.

If there's any mocking library you guys are fans of, I'd be happy to look into it, but I'd rather not write a by-hand mock for the EC2 client. I think the most valuable testing to add now would require such a mock.

jeanlaurent:
Rebased on current master as-is in #2684. Thanks @euank for the good work.

dgageot:
Merged with https://github.com/docker/machine/pull/2684

PR #1864, opened 2015-09-16
Update generic.md

zijam:
Parameter for '--generic-ssh-key' should point to users public key


dmp42:
@zijam can you sign your commits? Thanks

zijam:
Done :wink:

dmp42:
@zijam Sorry, I'm confused - why do you think the doc should be changed here?

dmp42:
@zijam I'm closing this for now since I don't think it's correct, but let me know if I missed something and we can certainly reopen.

PR #1889, opened 2015-09-21
Remove beta warning message

nathanleclaire:
The amount of confusion over this message has been very surprising to me.  It was intended to warn against using Machine in production, but users have taken it to indicate that Machine is unreliable in lower environments as well.  Since I'm not sure how to phrase the message to explain this to someone who doesn't grok the difference between "development" and "production", perhaps it's best just to take out.

cc @moxiegirl @ehazlett 

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,9 +11,6 @@ parent="smn_workw_machine"
 
 # Docker Machine
 
-> **Note**: Machine is currently in beta, so things are likely to change. We
-> don't recommend you use it in production yet.
-
 Machine lets you create Docker hosts on your computer, on cloud providers, and
 inside your own data center. It automatically creates hosts, installs Docker on
 them, then configures the `docker` client to talk to them. A "machine" is the


moxiegirl:
If the warning still applies, we can clarify it rather than remove it.  Your choice.  It might present a company liability to remove it. You'd have to check with the legal team on that account.

nathanleclaire:
Any suggestions on how to clarify the message to people who don't understand what production is?  "Docker Machine is great for developing on your local computer, but ..."

moxiegirl:
*Machine is currently in beta so its features are likely to change frequently.   For this reason, you shouldn't rely on Machine for use in software processes or production workflows that your company depends on daily. You can use Docker Machine for your own evaluations, in sandboxes, in test, and other non-production environments.*

nathanleclaire:
To me, "you shouldn't rely on Machine for use in software processes or production workflows that your company depends on daily" implies that it's not suitable for development use (as development is a daily process), so the problem seems to remain the same with that phrasing.

moxiegirl:
*Machine is currently in beta so its features are likely to change frequently.   For this reason, you shouldn't rely on Machine for use in software processes or production workflows that your company depends on daily. You can use Docker Machine for your own evaluations, in sandboxes, in test, and other non-production environments.*

moxiegirl:
Let me play with that and see what I come up with.

moxiegirl:
*Machine is currently in beta so its features are likely to change frequently.   For this reason, you shouldn't rely on Machine for use in critical software processes or production workflows that your company depends on. You can use Docker Machine for your own evaluations, in development environments, in sandboxes, in test, and other non-production environments.*

nathanleclaire:
I feel like that just introduces a lot more uncertainty into the mix though. Our main problem today is that the current message doesn't clearly communicate that Machine is perfectly suitable for use in development, but we don't feel comfortable about recommending it in higher environments just yet.

moxiegirl:
@nathanleclaire Like my original comment said, these kind of warnings are standard in technical documentation.  My understanding has always been as a hedge against liability.  If you want to remove it, that is your choice.  

If the restriction is still the case, I would advise against removing it and would vote for rewriting it. I've offerred you several options that I think are appropriate. If you want to keep the warning, I suggest we meet to iron out the language to your satisfaction.  It would be more efficient.

nathanleclaire:
I'm going to take this risk and merge this

PR #1907, opened 2015-09-25
Do not quote machine name during hostname setup

### LINKED ISSUE #1906: Windows 10 - SSH Error When Setting Hostname (exit status 2: sh: syntax error: unterminated quoted string)
I wanted to try out the fix from PR #1882 so I am using a local build of master (c4cd238) on Windows 10 with Virtualbox. It looks like a bug may have been introduced in 2a15d98.

When I run `docker-machine -D create default --driver virtualbox` or `docker-machine regenerate-certs default`, it fails when setting the host name with the error **unterminated quoted string**. Here's a link to the [full output from both commands](https://gist.github.com/carolynvs/457f67bbcc1fffe3e78a), the important part is below.

```
About to run SSH command:
sudo /usr/bin/sethostname default && echo "default" | sudo tee /var/lib/boot2docker/etc/hostname
&{C:\Program Files\Git\usr\bin\ssh.exe [C:\Program Files\Git\usr\bin\ssh.exe -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=no docker@127.0.0.1 -i C:\Users\caro8994\.docker\machine\machines\default\id_rsa -p 49654 sudo /usr/bin/sethostname default && echo "default" | sudo tee /var/lib/boot2docker/etc/hostname] []  <nil> <nil> <nil> [] <nil> <nil> <nil> <nil> false [] [] [] [] <nil>}
SSH cmd err, output: exit status 2: sh: syntax error: unterminated quoted string
```

I edited `boot2docker.go` `SetHostname` to not quote the docker machine name in the echo command
which fixed the problem. I'd be happy to send a PR if this seems like an appropriate change (dropping the quotes).

```go
func (provisioner *Boot2DockerProvisioner) SetHostname(hostname string) error {
	if _, err := provisioner.SSHCommand(fmt.Sprintf(
		"sudo /usr/bin/sethostname %s && echo %s | sudo tee /var/lib/boot2docker/etc/hostname",
		hostname,
		hostname,
	)); err != nil {
		return err
	}

	return nil
}
```

carolynvs:
Fixes #1906

### DIFF
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -107,7 +107,7 @@ func (provisioner *Boot2DockerProvisioner) Hostname() (string, error) {
 
 func (provisioner *Boot2DockerProvisioner) SetHostname(hostname string) error {
 	if _, err := provisioner.SSHCommand(fmt.Sprintf(
-		"sudo /usr/bin/sethostname %s && echo %q | sudo tee /var/lib/boot2docker/etc/hostname",
+		"sudo /usr/bin/sethostname %s && echo %s | sudo tee /var/lib/boot2docker/etc/hostname",
 		hostname,
 		hostname,
 	)); err != nil {


carolynvs:
This fixed the immediate error message but it seems that something has changed recently in how ssh commands are sent to the host and there is more cleaning up to do. This PR fixes one instance of the problem, and the hostname is being set without error. 

However `docker-machine create default --driver virtualbox` still fails farther along, on another ssh command, when copying the certs to the host in `ConfigureAuth`...

nathanleclaire:
Hm, probably it's related to recent changes I've made to `ssh` code.  I'll take a look, thanks./

carolynvs:
@nathanleclaire Did you get a chance to look at this yet? I'm curious if you are seeing the same problem on master or if I should do more to provide steps to reproduce.

dmp42:
LGTM for this PR specifically.
@nathanleclaire do you think we can merge?

carolynvs:
I'd really like to hear back from someone else about all the SSH related problems I've been experiencing when using the latest from master before merging this. I fixed the first problem I encountered but it seems like there is a larger problem which this PR doesn't address.

nathanleclaire:
It's in the bucket of "Windows SSH related problems" and something I'll be devoting more focus on this week and next.  I'm hesitant to merge into master without a deeper understanding of the issue, although I have a general idea of what it probably is.

Sorry for the delay in processing @carolynvs, I assure you it's important to me to get this looked at and fixed before the release.

carolynvs:
@nathanleclaire Thanks, much appreciated! I just wanted to make sure this didn't get merged right now since it doesn't fix the real problem. :smile:

nathanleclaire:
I'm going to close this as it's been fixed through other areas thanks @carolynvs !

PR #1909, opened 2015-09-27
Change 'ls' to not report 'Saved' hosts as active when $DOCKER_HOST is not set

### LINKED ISSUE #1908: `docker-machine ls` reports saved hosts to be active when $DOCKER_HOST is not set
`docker-machine ls` reports saved hosts to be active when `$DOCKER_HOST` is not set.

For example:

```sh
bash-4.3$ echo $DOCKER_HOST

bash-4.3$ docker-machine ls
NAME      ACTIVE   DRIVER       STATE     URL   SWARM
contrib            virtualbox   Stopped
demo      *        virtualbox   Saved
dev       *        virtualbox   Saved
```

I think this is confusing.

skatsuta:
Fixes #1908.

Currently `isActive()` only checks whether the state of a host is not `state.Stopped`, but it should also check the state of a host is not `state.Saved`.
I also added a test for `GetHostListItems()` in the case where `$DOCKER_HOST` is not set.

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -342,10 +342,10 @@ func isActive(h *host.Host) (bool, error) {
 
 	dockerHost := os.Getenv("DOCKER_HOST")
 
-	notStopped := currentState != state.Stopped
+	running := currentState == state.Running
 	correctURL := url == dockerHost
 
-	isActive := notStopped && correctURL
+	isActive := running && correctURL
 
 	return isActive, nil
 }
--- a/commands/ls_test.go
+++ b/commands/ls_test.go
@@ -376,3 +376,92 @@ func TestGetHostListItems(t *testing.T) {
 		}
 	}
 }
+
+// issue #1908
+func TestGetHostListItemsEnvDockerHostUnset(t *testing.T) {
+	orgDockerHost := os.Getenv("DOCKER_HOST")
+	defer func() {
+		cleanup()
+
+		// revert DOCKER_HOST
+		os.Setenv("DOCKER_HOST", orgDockerHost)
+	}()
+
+	// unset DOCKER_HOST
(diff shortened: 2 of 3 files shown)


nathanleclaire:
Cool.  Code looks good, and thanks for adding a test.  Could you please rebase all of your commits to one and I will merge?

skatsuta:
@nathanleclaire Thanks!
Now I rebased, but CircleCI became to fail.
I don't know why...

dmp42:
@skatsuta did you add circleci on your fork of machine? Seems the build here points to your fork build instead of upstream.

skatsuta:
@dmp42 You're right! Now passed. Thanks a lot!

commands/ls.go
```diff
@@ -342,10 +342,10 @@ func isActive(h *host.Host) (bool, error) {
 
 	dockerHost := os.Getenv("DOCKER_HOST")
 
-	notStopped := currentState != state.Stopped
+	notStoppedSaved := currentState != state.Stopped && currentState != state.Saved
```
nathanleclaire: Sorry to think of this so late in the game, but perhaps we could just change this to something like:

```go
running := currentState == state.Running
```

The reason for this is that there may be other "invalid" states in the future, but `state.Running` is likely to stay the only valid one.

@dmp42 Any opinions?

skatsuta: @nathanleclaire That's reasonable. Will change if it's ok.

skatsuta:
@nathanleclaire Changed to check whether the current state is exactly running, as you commented. Easer to understand and I like it!

nathanleclaire:
LGTM

Thanks for being so persistent @skatsuta !

PR #1910, opened 2015-09-27
READY FOR REVIEW - Work-around erroneous SSL: CERTIFICATE_VERIFY_FAILED error with some buggy versions of OpenSSL

posita:
This was inspired by my investigation into #1880 (and sister issues such as docker/compose#890). This patch ensures that the `ca.pem` subject is different from the `cert.pem` subject for [*bootstrapped* certificates](/docker/machine/blob/master/libmachine/cert/bootstrap.go). This is a work-around for a bug in some versions of OpenSSL that would erroneously error out with `SSL: CERTIFICATE_VERIFY_FAILED` if the subjects are the same.

Note: for *provisioned* certificates (i.e., those generated with [`libmachine/provision/utils.go`](/docker/machine/blob/master/libmachine/provision/utils.go)), the subject of `ca.pem` is the username and the subject of `cert.pem` is the machine name. This generally does not implicate the OpenSSL bug unless the user tries to provision a machine with the same name as his username (e.g., `docker-machine create ... "$( id -u -n )"`). That situation is also addressed by this patch, which prefixes the machine name with the user name.

### DIFF
--- a/libmachine/cert/bootstrap.go
+++ b/libmachine/cert/bootstrap.go
@@ -20,7 +20,8 @@ func BootstrapCertificates(authOptions *auth.AuthOptions) error {
 	// TODO: I'm not super happy about this use of "org", the user should
 	// have to specify it explicitly instead of implicitly basing it on
 	// $USER.
-	org := mcnutils.GetUsername()
+	caOrg := mcnutils.GetUsername()
+	org := caOrg + ".<bootstrap>"
 
 	bits := 2048
 
@@ -42,7 +43,7 @@ func BootstrapCertificates(authOptions *auth.AuthOptions) error {
 			return errors.New("The CA key already exists.  Please remove it or specify a different key/cert.")
 		}
 
-		if err := GenerateCACertificate(caCertPath, caPrivateKeyPath, org, bits); err != nil {
+		if err := GenerateCACertificate(caCertPath, caPrivateKeyPath, caOrg, bits); err != nil {
 			return fmt.Errorf("Generating CA certificate failed: %s", err)
 		}
 	}
--- a/libmachine/provision/utils.go
+++ b/libmachine/provision/utils.go
@@ -63,7 +63,7 @@ func ConfigureAuth(p Provisioner) error {
 	driver := p.GetDriver()
 	machineName := driver.GetMachineName()
 	authOptions := p.GetAuthOptions()
-	org := machineName
+	org := mcnutils.GetUsername() + "." + machineName
 	bits := 2048
 
 	ip, err := driver.GetIP()


nathanleclaire:
Thanks for the patch @posita.

cc @ehazlett Please take a look at this.

dmp42:
@posita is it possible to add tests for this?

posita:
I wanted to include tests, but there were a few impediments (one specific to me personally):

1. Appropriate failure (and hence the effectiveness of the work-around) depends on using a certain version of OpenSSL, so unless there's a way to include a known bad build of OpenSSL in a harness somewhere (or at least identify whether the installed version has a problem, and then skip the tests if it doesn't comply), then I'm not sure tests would have much value.
1. Neither [`ConfigureAuth`](https://github.com/posita/machine/blob/posita/1880-ssl-cert-snafu/libmachine/provision/utils.go#L58) nor [`BootstrapCertificates`](https://github.com/posita/machine/blob/posita/1880-ssl-cert-snafu/libmachine/cert/bootstrap.go#L11) appear to be unit test friendly in their current state. Unless I've missed something, there don't appear to be any tests that already call either of them, that I can piggy back on.
1. I could make an attempt (perhaps refactoring out the certificate generation code from each into separate, testable functions), but keep in mind I'm a total Go newbie, so I'd probably have to spend more time than I currently have understanding Go's test harnesses, standard library, etc. Even then, I'd really only be testing whether certs generated from either `ConfigureAuth` or `BootstrapCertificates` matched anticipated characteristics. I'm not sure how valuable such tests would be (they'd mostly be testing whether `GenerateCACertificate` honored its arguments; although in fairness, it doesn't look like there's any existing test coverage for `GenerateCACertificate` either). That being said (based on my very rudimentary understanding), verification of certificates should be possible within [`crypo/x509`](https://golang.org/pkg/crypto/x509/) (i.e., without having to introduce another dependency).

Please don't mistake my comments on this issue as a general testing philosophy. The standard practice absolutely should be that PRs should either be covered by existing (failing) tests, or include augmented test code. But including them here was beyond my immediate personal limits. Given its simplicity, and in light of the extensive documentation of the problem (in issues like docker/compose#890), I hoped this PR would still have some value without them, perhaps being able to rely solely on a code review. If this was misguided, then please accept my apologies for any inconvenience.

dmp42:
@posita thanks for the comment. Tentatively LGTM.
cc @nathanleclaire @ehazlett

dgageot:
Hello @posita, can you rebase on master and fix the merge conflicts? That would be great!

posita:
> Hello @posita, can you rebase on master and fix the merge conflicts? ...

Done.

> ... That would be great!

[Heh.](https://youtu.be/JFRa7Ovym8s) :wink:

dmp42:
Thanks all.

PR #1920, opened 2015-09-30
Additional info message while waiting for IP address to be assigned to the Droplet.

yanzay:
It takes a lot of time to assign the IP to droplet, so informing user
about this hang looks like a good idea.


dmp42:
LGTM

cc @nathanleclaire

nathanleclaire:
I guess it's OK.  In general drivers aren't really supposed to be doing this sort of IP polling on their own though.

PR #1902, opened 2015-09-24
Add Docker Machine driver plugins

nathanleclaire:
# Docker Machine Driver Plugins

## Motivation

Many folks have expressed interest in writing or maintaining a driver for Machine but @ehazlett and I are incapable of reviewing and maintaining all of them ourselves.  Additionally, we are a bottleneck for when people want to add features and release new versions of existing drivers.  Therefore, we are moving to a plugin-based architecture.

## Usage

From the end user's perspective, the Machine CLI will be pretty similar to how it always has, with the additional caveat that binaries for the desired plugins must be installed as well as the core `docker-machine` binary.

For the plugin developer and/or distributor, there are some considerations to keep in mind.  All plugins must be Go binaries located in the end user's `PATH` with the name of `docker-machine-driver-name`.  Once present, they will be usable.  

To implement a plugin, the driver developer must make a new repository with the plugin code, which:

- includes a properly named module (e.g. `virtualbox`) which contains a `struct` that fulfills the `Driver` interface from `github.com/docker/machine/libmachine/drivers`
- includes a `main` module (I've chosen to make this in a directory called `bin` in the `virtualbox` example presented here for this) which calls `plugin.RegisterPlugin()` with an instantiated instance of the struct.

Example from VirtualBox:

```go
package main

import (
        "github.com/docker/machine/drivers/virtualbox"
        "github.com/docker/machine/libmachine/drivers/plugin"
)

func main() {
        plugin.RegisterDriver(virtualbox.NewDriver("", ""))
}
```

## Migrating an Existing Driver to a Plugin

This is for the time of writing and is likely to change, but I think through documenting things we can hopefully the most mundane details of this process through a script.

The most important / likely to trip up migrators details are:

- A lot of paths have changed, usually because the modules have been moved inside of `libmachine`.  Some examples:
    - `github.com/docker/machine/log` => `github.com/docker/machine/libmachine/log`
    - `github.com/docker/machine/utils` => `github.com/docker/machine/libmachine/mcnutils`
    - `github.com/docker/machine/drivers` => `github.com/docker/machine/libmachine/drivers`
- The usage of `cli.Flag` in the drivers has changed to a Machine-specific model which is more RPC-friendly.  Essentially:
    - `GetCreateFlags` is now a required method on the `Driver` itself, not a module-level function, which returns `[]mcnflag.Flag` (a custom type encompassing all types of flags).  All instances of `cli.Flag` should be replaced by the corresponding instance of `mcnflag.Flag`.
[...]

PR #1926, opened 2015-10-01
Add support for OpenNebula

jmelis:
This PR adds support for OpenNebula. It also includes the Godeps for the required libraries. It builds correctly (also with USE_CONTAINER option) and it has been tested in our environment successfully.

We are currently working with the boot2docker project in order to bring support to OpenNebula to b2d. While a new b2d image is released with OpenNebula support ([see this PR](https://github.com/boot2docker/boot2docker/pull/1067)), we have hosted a modified version of b2d (latest release) that includes OpenNebula support. The driver by omission fetches the image from that url, however it can be overridden, and we are planning on switching to the latest b2d release as soon as it has support.

Signed-off-by: Jaime Melis <jmelis@opennebula.org>

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -156,8 +156,21 @@
 			"Rev": "a09229c13c2f13bbdedf7b31b506cad4c83ef3bf"
 		},
 		{
+			"ImportPath": "github.com/OpenNebula/goca",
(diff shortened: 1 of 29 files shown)


nathanleclaire:
Hi @jmelis , we are currently not accepting drivers into the core, but working on driver plugins which will be distributable / usable in the next release!  Please see https://github.com/docker/machine/pull/1902 for details and leave feedback if you can.

Thanks!

- N

PR #1932, opened 2015-10-03
Updating CI builds to use Go 1.5.1

hairyhenderson:
Note: I recognize that the `FROM` in the Dockerfile is a no-op since `golang:1.5` == `golang:1.5.1`. I decided to change it anyway though, to make the next update simpler (i.e. we can just `grep -R 1.5.1 *`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.5
+- 1.5.1
 install:
 - go get github.com/mattn/goveralls
 script:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.5
+FROM golang:1.5.1
 
 RUN go get  github.com/golang/lint/golint \
             github.com/mattn/goveralls \
--- a/circle.yml
+++ b/circle.yml
@@ -8,7 +8,7 @@ machine:
     # - sudo apt-get install -y virtualbox
 
   post:
-    - gvm install go1.5 -B --name=stable
+    - gvm install go1.5.1 -B --name=stable
 
   environment:
   # Convenient shortcuts to "common" locations


dmp42:
Hi @hairyhenderson 

1.5 should always point to the latest released bugfix on the 1.5 branch (eg: 1.5.1 here), so, can you elaborate on what are the benefits in pinning it to a specific revision?

Thanks

hairyhenderson:
@dmp42 - my impression was otherwise for Travis (which uses `gimme`) and `gvm`... When I looked at them it seemed they pointed to the tags in Go's repo, and `1.5` was the 1.5.0 release. For the Dockerfile - yeah, that _should_ point to 1.5.1, but I wanted to be consistent as I noted above.

hairyhenderson:
Just as a followup, here's the output of `gimme 1.5` (which is what Travis does):

```console
$ eval "$(gimme 1.5)"
go version go1.5 linux/amd64
```

So, this means we certainly need to change the `.travis.yml` entry...

dmp42:
@hairyhenderson good point for gvm & travis

I would rather leave the Dockerfile alone, though (if only, because it's used for development only).

jeanlaurent:
@hairyhenderson would you mind rebasing on master and removing the Dockerfile changes ? This would then be ready to be merged then.

hairyhenderson:
@jeanlaurent - I've rebased. Are we _sure_ that removing the Dockerfile change is a good idea? IMO it should be kept in sync with CI.

When (if) Go 1.5.2 is released, new developers will effectively pull `golang:1.5.2`, while CI builds and other developers who have `golang:1.5` cached will still be using 1.5.1. IMO it's safer to keep them all in lock-step. Also, if the `FROM` is kept explicit, then developers will get their base image updated along with the CI changes, keeping the environment consistent for everyone.

What do you think?

jeanlaurent:
@hairyhenderson I think it make sense. We gain a bit of thoroughness at the expense of some longer build whenever we change the golang version. So let's keep it this way. 

Good to go.

cc @dmp42

hairyhenderson:
Thanks @jeanlaurent!

dmp42:
LGTM

PR #1941, opened 2015-10-06
Fix broken --storage-path flag

nathanleclaire:
Noticed that this was broken with the `libmachine` changes.

cc @ehazlett @dmp42 


nathanleclaire:
Fixed a couple instances in tests where we expect to be able to get and set the environment variable directly, I'm still not crazy about having a module-level variable to track this but it's better than reading and writing to the environment willy-nilly IMO.

dmp42:
LGTM
@dgageot do you want to have a look?

PR #1944, opened 2015-10-07
Fix #1943 - binaries location

### LINKED ISSUE #1943: Contributing Guide contains a couple of false information
The Contributing Guide says that docker-machine can be built with `make`. Actually, there is no default target to make so only `make build` will work. So either the doc is wrong or the Makefile needs to be fixed. I'll go for the doc.

Also, it is said that binaries are at the root of the repo. That used to be true. Not anymore.

dgageot:


dmp42:
LGTM
Kicking the build - not sure why it failed.

PR #1949, opened 2015-10-07
Prepare build for plugins PR

dmp42:
@nathanleclaire this contains part of the PR that we can merge right now instead of having sit in queue on top of your plugins PR (I'll then update my other one so it's smaller).

Key-point:

 * fix documentation confusion around integration testing inside/outside containers
 * overall enhance contribute documentation
 * fix a couple of build oddities
 * ready to build plugins as we drop them into cmd
 * no longer rely on gox
 * zip release support
 * small cleanups

Also cc @dgageot to the party :)


dmp42:
@nathanleclaire also, using `VERBOSE=true make FOO` you now have full output of the commands being run.

nathanleclaire:
LGTM

PR #1950, opened 2015-10-08
Display error message only when create was otherwise successful

nathanleclaire:
The new error messaging should be only be displayed in the case of a successful provision.

cc @dmp42 


dmp42:
cc @dgageot for additional eyes on this

PR #1952, opened 2015-10-08
Move VBox detection to Precreate and print version

nathanleclaire:
I'm not sure if we printed the VirtualBox version at some point and it just got lost (it's possible we never did), but it's absolutely critical for debugging to know which version of VirtualBox the user is on.


nathanleclaire:
cc @dmp42

PR #1953, opened 2015-10-08
Explain how to prepare a Go 1.5 development environment

gdm85:
Other minor documentation fixes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,29 +14,40 @@ Docker does things, you'll feel right at home.
 Otherwise, please read [Docker's contributions
 guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
 
-# Building using docker
+# Building
 
 The requirements to build Machine are:
 
-1. A running instance of Docker
+1. A running instance of Docker or a Golang 1.5 development environment
 2. The `bash` shell
 3. [Make](https://www.gnu.org/software/make/)
 
-Call `export USE_CONTAINER=true` to instruct the build system to use containers to build.
+## Build using Docker containers
 
-# Alternative: build using go only
+To build the `docker-machine` binary using containers, simply run:
 
-Alternatively, you can build without docker, using only golang.
+    $ export USE_CONTAINER=true
+    $ make build
 
-[Install and setup go](https://golang.org/doc/install), then clone the machine repository inside your gopath.
+## Local Go 1.5 development environment
 
-## Building
+Make sure the source code directory is under a correct directory structure to use Go 1.5 vendoring;
+example of cloning and preparing the correct environment `GOPATH`:
+```
+    mkdir -p projects/docker-machine/src/github.com/docker
+    export GOPATH="$PWD/projects/docker-machine"
+    cd projects/docker-machine/src/github.com/docker
+    git clone https://github.com/docker/machine
+    cd machine
+```
 
-To build the docker-machine binary, simply run:
+At this point, simply run:
 
     $ make build
 
-From the Machine repository's root. You will now have a `bin/docker-machine`.
+## Built binary
+
+After the build is complete a `bin/docker-machine` binary will be created.
 
 You may call:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,11 +34,11 @@ To build the `docker-machine` binary using containers, simply run:
 Make sure the source code directory is under a correct directory structure to use Go 1.5 vendoring;
 example of cloning and preparing the correct environment `GOPATH`:
 ```
-    mkdir -p projects/docker-machine/src/github.com/docker
-    export GOPATH="$PWD/projects/docker-machine"
-    cd projects/docker-machine/src/github.com/docker
-    git clone https://github.com/docker/machine
-    cd machine
+    mkdir docker-machine
+    cd docker-machine
+    export GOPATH="$PWD"
+    go get github.com/docker/machine
+    cd docker-machine/src/github.com/docker/machine
 ```
 
 At this point, simply run:


nathanleclaire:
Hm, @dmp42 has some changes in the pipe for this doc as well which I will probably merge first, but I'm interested, so stay tuned.

dmp42:
@gdm85 do you think you can rebase now the other PRs have been merged? Thanks a lot.

gdm85:
@dmp42 done.

Please note that I am omitting the link to Go 1.5 installation on purpose (I don't think it fits the scope of this documentation, but it's arguable) and then providing an example about how to work on sources directly under a ``GOPATH``. I am not sure if this is a best practice or not, but I couldn't figure out a better way to reuse the project-provided dependencies.

CONTRIBUTING.md
```diff
@@ -14,29 +14,41 @@ Docker does things, you'll feel right at home.
...
+    $ make build
 
-## Building
+## Local Go 1.5 development environment
+
+Make sure the source code directory is under a correct directory structure to use Go 1.5 vendoring;
+example of cloning and preparing the correct environment ``GOPATH``:
+````
+    mkdir -p projects/docker-machine/src/github.com/docker
+    export GOPATH="$PWD/projects/docker-machine"
```
dmp42: you probably want `GOPATH="$PWD/projects/docker-machine":$GOPATH` here

CONTRIBUTING.md
```diff
@@ -14,29 +14,41 @@ Docker does things, you'll feel right at home.
...
+    cd projects/docker-machine/src/github.com/docker
+    git clone https://github.com/docker/machine
+    cd machine
+    export GOPATH="$GOPATH:$PWD"
```
dmp42: I believe this one is useless.

CONTRIBUTING.md
```diff
@@ -14,29 +14,41 @@ Docker does things, you'll feel right at home.
...
-[Install and setup go](https://golang.org/doc/install), then clone the machine repository inside your gopath.
+    $ export USE_CONTAINER=true
+    $ make build
 
-## Building
+## Local Go 1.5 development environment
+
+Make sure the source code directory is under a correct directory structure to use Go 1.5 vendoring;
+example of cloning and preparing the correct environment ``GOPATH``:
+````
```
dmp42: Only three backticks here.

CONTRIBUTING.md
```diff
@@ -14,29 +14,41 @@ Docker does things, you'll feel right at home.
...
-1. A running instance of Docker
+1. A running instance of Docker or a Golang 1.5 development environment
 2. The `bash` shell
 3. [Make](https://www.gnu.org/software/make/)
 
-Call `export USE_CONTAINER=true` to instruct the build system to use containers to build.
+## Build using Docker containers
 
-# Alternative: build using go only
+To build the ``docker-machine`` binary using containers, simply run:
```
dmp42: You just need one backtick for these.

CONTRIBUTING.md
```diff
@@ -14,29 +14,41 @@ Docker does things, you'll feel right at home.
...
 
-To build the docker-machine binary, simply run:
+At this point, simply run:
 
     $ make build
 
-From the Machine repository's root. You will now have a `bin/docker-machine`.
+## Built binary
+
+After the build is complete a ``bin/docker-machine`` binary will be created.
```
dmp42: One backtick.

dmp42:
Thanks a lot @gdm85 !
I added a couple of remarks.
Also, we should still separate the native golang approach versus the docker only approach (people going with the latter *don't* need golang at all).

CONTRIBUTING.md
gdm85: Well, the imports from this same project (`docker/machine`) are not vendor'ed (obviously) and not in `GOPATH`, I remember trying without this line and getting errors.

gdm85:
@dmp42 I am still learning markdown, apparently :)

As per the approaches split, do you think it should be made visible in a different way?

gdm85:
I've removed also the extra GOPATH, tested, it works

jeanlaurent:
We could probably add the default simplest way of getting ready to code on machine which is @dgageot setup, this is what we want to achieve with this guide, make it easy to contribute.

And add an optional setup, where you don't share you `GOPATH` with other projects ( as @gdm85 suggested ). 

As my personal experience, I did them both, and there is no doubt the former is easier.

jeanlaurent:
@dmp42 IMHO this could be merged "as is", as it brings value right now. We should decide of the easiest way of getting up to speed on the project in another dedicated PR.

gdm85:
@jeanlaurent my personal opinion is that `GOPATH` is something that should be replaced by a command-line option (as it works in proven compilers like gcc/g++ with `-I`), but that is off-topic here :)

What I wanted to contribute here is a way for even an **occasional** contributor to do the right thing; I find myself contributing after weeks or a month and not updating the packages in `GOPATH` is dangerous in such cases.

With my latest commit 8ac92c2 I simplified the process even further by making use of `go get`; I didn't squash it to leave the option to you guys to either use it or drop it.

gdm85:
@dgageot ops, added the missing `go get` now :)

jeanlaurent:
thx @gdm85

@dmp42 ready to review & merge.

dmp42:
LGTM

@jeanlaurent you can hit that merge button at will.

nathanleclaire:
Should `GOPATH` really point to the Docker Machine repo's location on disk?  I'd generally be inclined towards recommending `export GOPATH="$(pwd)/vendor:$GOPATH"` or, better yet, `export GO15VENDOREXPERIMENT=1`.

dmp42:
`GO15VENDOREXPERIMENT=1` will not help (will just make golang resolve `vendor` but won't add the current directory to GOPATH)

Exporting `GOPATH="$(pwd)/vendor:$GOPATH"` is usually the right approach in a proper golang setup where you work on multiple different stuff - but it might also create difficulties for newcomers wrt package resolution, and is not needed per-se if you just want to hack on machine.

Either way, a follow-up PR can certainly enhance this if this is a problem.

gdm85:
@nathanleclaire now that you mention it, I notice that my wording is misleading as I wrote `to use Go 1.5 vendoring` although this project doesn't use it yet; sorry for that.

In this PR I saw the approach more from the perspective of allowing contributions without having to fetch sub-packages in a separate `GOPATH`, thus as easy and self-contained as possible.

nathanleclaire:
@gdm85 No problem, looks like a good start for now, and I'll revisit and PR any nitpicks I have.

PR #1438, opened 2015-06-29
Adding digitalocean-sshkey-id option so that we don't have to create …

usmanismail:
…a new ssh key every time we use machine.

Fixes Issue #1445

PR #1956, opened 2015-10-08
Fix for issue 1898,  use existing keypair when creating an EC2 instance.

### LINKED ISSUE #1898: Use existing keypair for new machines on EC2
Rather than create a new keypair for every machine it would be ideal to use an existing keypair.  We would like to use the same keypair for all of our dev machines.  Would make management of instances across the team easier.

jeffellin:
Fixes #1898. Add parameter to pass private key path to use when creating a machine.  It is
assumed that the corresponding public certificate will be the same file name + .pub


jeffellin:
@nathanleclaire having a little trouble understanding why this build failed.  Seeing the following

```
panic: interface conversion: interface is nil, not string [recovered]
	panic: interface conversion: interface is nil, not string
```

drivers/amazonec2/amazonec2.go
```diff
@@ -87,6 +88,11 @@ func GetCreateFlags() []cli.Flag {
 			EnvVar: "AWS_SECRET_ACCESS_KEY",
 		},
 		cli.StringFlag{
+			Name:   "amazonec2-ssh-keypath",
+			Usage:  "SSH Key for Instance",
+			EnvVar: "SSH_KEYPATH",
```
nathanleclaire: This needs to be prefixed with `AWS`, e.g. `AWS_SSH_KEYPATH`

test/integration/data/sshkey/id_rsa
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
nathanleclaire: I don't want to have keys checked into version control.  I understand that they're just for the test, but you should generate them and remove them in the test as a self-contained unit.

jeffellin: makes sense,  I can generate them to a tmp directory and remove them when complete.

nathanleclaire:
I'm a bit hesitant to merge in such a PR and continue the duplication of same type of flags across drivers (a lot of providers allow you to optionally specify an SSH key), but I'll think about it.

jeffellin:
Perhaps there needs to be some setting to specify keypairs beyond the driver level.  This would require refactoring a bunch of existing drivers for consistency.

Kosta-Github:
I am doing exactly the same `create-or-copy` operation in my driver, so maybe a base implementation would be useful, which could be reused by the concrete driver implementations?

jeanlaurent:
@nathanleclaire did you think about this ? It's probably easier to merge in this, and refactor away right after that.

@jeffellin The aws driver has been significantly changed by #2684, would you mind rebasing your work on this ?

jeffellin:
Only if I get a commitment from Nathan it will be merged.

jeanlaurent:
@nathanleclaire your call.

jeffellin:
@nathanleclaire any thoughts on this.  Seems to be a common theme, #1438

nathanleclaire:
Sure, having a keypair option for the driver SGTM

jeffellin:
@jeanlaurent Wonder if you can take a look at the bats test.  The assertions at the end aren't passing. However the output of $PUBLIC and $PRIVATE seem to be empty so it should be ok.

jeffellin:
@nathanleclaire @jeanlaurent not sure why the windows build is failing

jeanlaurent:
Sorry, I didn't saw that earlier. I'll give it a shot tomorrow.
Le mer. 13 janv. 2016 à 20:06, Jeffrey Ellin <notifications@github.com> a
écrit :

jeffellin:
> Sorry, I didn't saw that earlier. I'll give it a shot tomorrow.

@jeanlaurent No worries,  seems like its working now,  but It would still be great for you to take a look since I am by no means a bats expert.

jeffellin:
@nathanleclaire @jeanlaurent any thoughts on getting this in before I have to merge it again?

Anything else you need?

jeanlaurent:
Even if the bats file sounds like complicated, and due to the nature of the test, it can't really be that much simpler, It did run flawlessly 3-4 times in a row :+1: 

The only thing I would change is /tmp directory.
You should create your temp certificates in ``
That would limit file sprawling everywhere.
So change 
`export AWS_SSH_DIR=/tmp/mcnkeys`
to
`export AWS_SSH_DIR="$MACHINE_STORAGE_PATH/mcnkeys"`

once this change is done -> LGTM

dgageot:
LGTM

jeffellin:
> The only thing I would change is /tmp directory.
> You should create your temp certificates in $MACHINE_STORAGE_PATH

good idea. thanks for taking a look.

jeffellin:
@jeanlaurent should be g2g

jeanlaurent:
argh, I re-tested it erverything is ok, but the latest merge, made a conflict.

Can you please rebase ?

jeffellin:
@nathanleclaire @jeanlaurent is it possible to re-open this?  Not sure how I managed to close it.

nathanleclaire:
The "re-open" button is greyed out for me, unfortunately.  I'd recommend making a new branch and pull request, it will probably be easier than trying to fix Github's view of the state.

drivers/amazonec2/amazonec2.go
```diff
@@ -903,15 +889,13 @@ func (d *Driver) configureSecurityGroupPermissions(group *ec2.SecurityGroup) []*
 	hasDockerPort := false
 	hasSwarmPort := false
 	for _, p := range group.IpPermissions {
-		if p.FromPort != nil {
```
nathanleclaire: This `p.FromPort != nil` line should not be removed.  It was introduced in `f17cce330c5b4b7a0b04d3c931d3a96cdff53476` to prevent a panic from occuring.

jeffellin: yeah ,   I don’t think thats my latest commit,   I had some problems
rebasing and I think that resulted in some weird stuff,  Let me get a new
PR up so you can take a look at a clean diff.

nathanleclaire: Hm, looks like there's a few issues with rebase, so I'm guessing this wasn't intentional.

jeffellin: yeah,  was hoping to get the open issue fixed before pushing a fix.

artburkart:
```
Error with pre-create check: "There is already a keypair with the name <keyname>.  Please either remove that keypair or use a different machine name."
open : no such file or directory
notifying bugsnag: [Error with pre-create check: "There is already a keypair with the name <keyname>.  Please either remove that keypair or use a different machine name."]
```

I think this error message is related to this ticket. I'm using docker-machine 0.6 and my command is as follows:

```
dm --debug create --driver amazonec2 --amazonec2-vpc-id <vpc-id> --amazonec2-subnet-id <subnet-id> --amazonec2-zone c --amazonec2-security-group <sg-id> --amazonec2-iam-instance-profile <profile-name> --amazonec2-ssh-keypath /Users/arthurburkart/.ssh/<keyname> <keyname>
```

My question is, was it a design decision to not allow users to reuse a private key with more than one instance? If so, I'm fine with it. If not, I think this may be a bug.

Cheers

jeffellin:
We have kept with the model that the initial implementation used to create
a key pair using the name of the machine.  The name of the machine is
expected to be unique thus the key pair should be unique.   I can see a use
case where we don't create a new key pair and reuse one already uploaded to
aws but I was going with making a little changes as possible.

artburkart:
Ah, so it's legacy? :+1: That's fine. I'm using this for development anyway.

jeffellin:
I wouldn't exactly call it legacy it was just based on the existing implementation and not changed in the lack of requirements to do so.  Feel free to create a new issue if this is a blocker.

PR #1960, opened 2015-10-09
FIX #1914 Reject command lines with trailing flags

### LINKED ISSUE #1914: `create` silently ignores trailing options
Hi,

I spent most of my day debugging a problem caused by specifying trailing options on `docker-machine create --driver virtualbox default <more options>`, it'd be nice to have some sort of warning message when there are extra arguments specified to alert the user that it will not perform the desired behaviour.

dgageot:


dmp42:
@dgageot some integration tests are apparently relying on trailing flags - you will have to update them (see `test/integration`)

dmp42:
Also, github will magically link to an issue if fix #1914 is in the body of a comment - unfortunately, not in the title of the PR.

dgageot:
@nathanleclaire @dmp42 wdyt?

test/integration/cli/create-rm.bats
```diff
@@ -53,9 +53,9 @@ load ${BASE_TEST_DIR}/helpers.bash
 }
 
 @test "none: extraneous argument is ignored in name 'machine create -d none --url none a foo'" {
```
dmp42: Shall this test be renamed?

dmp42:
Looks good.
@nathanleclaire can you take a quick look?

dgageot:
I will change the name!
Le sam. 10 oct. 2015 à 21:07, Olivier Gambier <notifications@github.com> a
écrit :

dgageot:
@dmp42 test name fixed. If you or @nathanleclaire could take a look, that would be awsome

dmp42:
LGTM

ping @nathanleclaire can you take a look?

commands/create_test.go
```diff
@@ -20,3 +21,19 @@ func TestValidateSwarmDiscoveryAcceptsValidFormat(t *testing.T) {
 	err := validateSwarmDiscovery("token://deadbeefcafe")
 	assert.NoError(t, err)
 }
+
+func TestRejectCreateTrailingOptions(t *testing.T) {
```
nathanleclaire: This seems to be more of an integration test than a unit test (it tests the whole flow), so my preference would be to leave it out.

nathanleclaire:
Why change the `cmdCreate` method so much?  I have made some pretty serious changes in the whole `create` flow for the plugins so I'd prefer the changes to be minimal for now.

commands/create_test.go
dgageot: I strongly disagree @nathanleclaire. We should test most error cases without integration tests. This test might me a little complicated for now but it's not an integration test. It's just that the design has to be improved to make testing easier.

Until we do those improvements, that's one more test on a file that lacks tests

dgageot:
I changed `create` to make it testable. Having a dozen of `log.Fatal` is plain non-sense test-wise. It makes it impossible to unit-test.

test/integration/cli/create-rm.bats
```diff
@@ -52,10 +52,10 @@ load ${BASE_TEST_DIR}/helpers.bash
   [[ ${lines[0]} == "Error creating machine: Machine A already exists" ]]
 }
 
-@test "none: extraneous argument is ignored in name 'machine create -d none --url none a foo'" {
-  run machine create -d none a foo
+@test "none: fail with trailing options 'machine create -d none --url none a foo'" {
+  run machine create -d none --url none a foo
   [ "$status" -eq 1 ]
-  [[ ${lines[0]} == "Host already exists: \"a\"" ]]
+  [[ ${lines[0]} == "Invalid command line. All options should be either before or after the machine name. Found those trailing flags [foo]" ]]
```
nathanleclaire: This error message doesn't make much sense to me as `foo` is not a flag, it's an extra argument.

dgageot: Good point. I'll change the error message

dgageot:
@nathanleclaire I removed the modifications on `cmdCreate` since it's a pain for your PR. I still believe using `log.Fatal()` like this makes the code impossible to test with unit-tests so I'll create a PR later to address this issue.

I also fixed the wording.

Do you see any issue now?

nathanleclaire:
>  I removed the modifications on cmdCreate since it's a pain for your PR. I still believe using log.Fatal() like this makes the code impossible to test with unit-tests so I'll create a PR later to address this issue.

Yeah, I agree with this, and actually with the new plugins work anything we can do to reduce the possibility of a non-expected exit from the program is a good thing.

dgageot:
@nathanleclaire cool. So I guess it's ready to merge right? cc @nathanleclaire @dmp42

dmp42:
Thanks @nathanleclaire ! 
And thanks @dgageot 

LGTM

PR #1961, opened 2015-10-09
Fix building in docker

dmp42:
- now has default build target on `make`
- test depends on non-bogus target


dmp42:
cc @jeanlaurent @dgageot @nathanleclaire

nathanleclaire:
`go test` should already compile dependencies right?  Is `test` target doing other tests as well such as integration tests?  I thought those ensured that we compiled the binary first too.

dmp42:
`test` is also a folder inside the repo, and make is confused by this, unless the target depends on another target to run. This is just to workaround make here.

dmp42:
@nathanleclaire additionally the build now no longer tries to build _test files in cmd.

PR #1965, opened 2015-10-10
Fix filenames of loggers.

pragkent:
It seems that filenames of loggers are incorrect. Just exchanged names of standard.go and terminal.go

Signed-off-by: Kent Wang <pragkent@gmail.com>


nathanleclaire:
Hm, looks like it's a mistake I introduced with the `libmachine` code.  `terminal.go` should just be taken out completely and `standard.go` should be the one used.

pragkent:
Removed TerminalLogger and updated test.

nathanleclaire:
LGTM

Issue #1967, opened 2015-10-10
OpenStack instance starts up but docker-machine doesn't get its IP address


cmouttet:
The instance is started and I'm able to connect to it via ssh (with keypair). I found that the AddressType in client.go isn't set to any default. I guess that the address type is not replied by our OpenStack instance. It's a quite old installation (don't know the version).

```
Creating OpenStack instance...		 FlavorId=1 ImageId=e1310ae2-178b-4282-9be7-77249938ef0d
Creating machine...
Waiting for the OpenStack instance to be ACTIVE...		 MachineId=75015df8-5bff-43f1-848b-fde8a08ac1a4
Looking for the IP address...		 MachineId=75015df8-5bff-43f1-848b-fde8a08ac1a4
Error creating machine: No IP found for the machine
You will want to check the provider to make sure the machine and associated resources were properly removed.
```

After adding "AddressType: Fixed" as default, "docker-machine create" works perfectly.

```
Creating OpenStack instance...		 FlavorId=1 ImageId=e1310ae2-178b-4282-9be7-77249938ef0d
Creating machine...
Waiting for the OpenStack instance to be ACTIVE...		 MachineId=10b4db0c-2c61-4c6e-8e41-f421ea00a1a7
Looking for the IP address...		 MachineId=10b4db0c-2c61-4c6e-8e41-f421ea00a1a7
IP address found		 IP=172.19.9.44 MachineId=10b4db0c-2c61-4c6e-8e41-f421ea00a1a7
Waiting for machine to be running, this may take a few minutes...
Get status for OpenStack instance...		 MachineId=10b4db0c-2c61-4c6e-8e41-f421ea00a1a7
State for OpenStack instance		 MachineId=10b4db0c-2c61-4c6e-8e41-f421ea00a1a7 State=ACTIVE
```

cmouttet:
https://github.com/docker/machine/pull/1968

PR #1968, opened 2015-10-10
add default AddressType: Fixed

cmouttet:
Signed-off-by: Christian Mouttet <cmouttet@gmail.com>


nathanleclaire:
cc @ggiamarchi

ggiamarchi:
@cmouttet I don't have enough information to say why you get this behaviour from your OpenStack... By the way, this fix looks reasonable enough, it doesn't impact the "normal" OpenStack behaviour.

LGTM (Fix #1967)

Issue #1947, opened 2015-10-07
CoreOS Provisioner breaks ETCD


sylus:
I have been working with the new CoreOS provisioner which is scheduled for 0.5 or perhaps earlier. The new provisioner does indeed work and launches a CoreOS droplet (testing on digitalocean) successfully. The problem has to do with etcd not being configured as the discovery token was not sent via cloud-init. This means native functionality to CoreOS is broken and etcd is returning 404's. I asked the CoreOS folks about possible workarounds but the general opinion was that these parameters need to be entered at provision time.

As a workaround I though I could create CoreOS via cloud-config myself and then attach the docker-machine generic driver. While CoreOS was working and etcd was functional as soon as I attached docker-machine it broke all etcd functionality. Below are the exact steps I followed.

### Curl

```sh
curl --request POST "https://api.digitalocean.com/v2/droplets" \
     --header "Content-Type: application/json" \
     --header "Authorization: Bearer $DO_TOKEN" \
     --data '{
        "region":"'"${DO_REGION}"'",
        "image":"coreos-alpha",
        "size":"'"$DO_SIZE"'",
        "private_networking":true,
        "user_data": "'"$(cat conf/digitalocean/cloud-config.yaml)"'",
        "ssh_keys":["'"$DO_KEY_ID"'"],
        "name":"Master"
     }'
```

### Cloud Config

```sh
coreos:
  etcd:
    discovery: https://discovery.etcd.io/<token>
    # addr: $private_ipv4:4001
    addr: $public_ipv4:4001
    # peer-addr: $private_ipv4:7001
    peer-addr: $public_ipv4:7001
    peer-heartbeat-interval: 1000
    peer-election-timeout: 5000
  fleet:
    # public-ip: $private_ipv4   # used for fleetctl ssh command
    public-ip: $public_ipv4
  units:
    - name: etcd.service
      command: start
    - name: fleet.service
      command: start
```

After these steps CoreOS was launched and everything including etcd were configured properly. Now I attempted to link docker machine (recent dev version with coreos provisioner) to this existing droplet.

```sh
docker-machine create -d generic \
--generic-ssh-user core \
--generic-ssh-key ~/.ssh/xxx \
--generic-ip-address xxx.xxx.x.xx \
example
```

As soon as docker-machine was done its configuration and linked to the droplet etcd was disabled and its configuration lost. I'd welcome any ideas of how to solve this!

nathanleclaire:
[...]

PR #1479, opened 2015-07-07
[Enhancement] VMware Fusion driver

mschygulla:
This PR enhances the Docker Machine driver for VMware Fusion (OSX). It add's an additional cdrom as configuration drive datasource for the Cloud-init package. An example config drive which was tested with CoreOS can be found here: https://github.com/mschygulla/configdrive

The driver accepts some new flags:

**URL for cloud-init configdrive**
* EnvVar: "FUSION_CONFIGDRIVE_URL"
* Name:   "vmwarefusion-configdrive-url"

**SSH user (default is B2DUser)**
* EnvVar: "FUSION_SSH_USER"
* Name:   "vmwarefusion-ssh-user"

**SSH password (default is B2DPass)**
* EnvVar: "FUSION_SSH_PASSWORD"
* Name:   "vmwarefusion-ssh-password"

If the flag `--vmwarefusion-configdrive-url` is not used, the driver will behave as before.

If an URL is specified with the flag `--vmwarefusion-configdrive-url`, the driver will download the ISO and create an additional cdrom in the VMX file with the provided ISO. Since the VMware Tools are not always available, the generated public SSH key will be copied with SSH using the provided SSH User/Password.

This PR was tested with CoreOS and the CoreOS provisioner (https://github.com/docker/machine/pull/1463)

**Example Usage:**

```shell
$ docker-machine create -d vmwarefusion --vmwarefusion-boot2docker-url http://alpha.release.core-os.net/amd64-usr/723.0.0/coreos_production_iso_image.iso --vmwarefusion-configdrive-url https://github.com/mschygulla/configdrive/releases/download/v0.2/configdrive.iso --vmwarefusion-ssh-user core dev
```

or

```shell

$ export FUSION_BOOT2DOCKER_URL=http://alpha.release.core-os.net/amd64-usr/723.0.0/coreos_production_iso_image.iso
$ export FUSION_CONFIGDRIVE_URL=https://github.com/mschygulla/configdrive/releases/download/v0.2/configdrive.iso
$ export FUSION_SSH_USER=core

$ docker-machine create -d vmwarefusion dev1
$ docker-machine create -d vmwarefusion dev2
```

Feedback welcome :)

PR #1971, opened 2015-10-11
Issue #1947 CoreOS provisioner does not properly bootstrap etcd

sylus:
This referrence issue: #1947 

This is my first go at getting the new CoreOS provisioner to properly bootstrap ETCD. It works right now for the generation of the discovery token and using the Public IP. I'd welcome any and all feedback as is my first PR to docker.

Ideally we can also support the --digitalocean-private-networking and use the Private IP as well. Looking at the ec2 driver as a guide since I think it supports both. Right now the template enforces just the Public IP.

### DIFF
--- a/libmachine/provision/coreos.go
+++ b/libmachine/provision/coreos.go
@@ -21,6 +21,32 @@ const (
 hostname: %s
 EOF
 `
+
+	cloudConfigTmpl = `sudo tee /var/tmp/cloudinit.yml << EOF
+#cloud-config
+
+coreos:
+  etcd:
+    discovery: {{.Token}}
+    addr: {{.IP}}:4001
+    peer-addr: {{.IP}}:7001
+    # give etcd more time if it's under heavy load - prevent leader election thrashing
+    peer-election-timeout: 4000
+    # heartbeat interval should ideally be 1/4 or 1/5 of peer election timeout
+    peer-heartbeat-interval: 1000
+    # don't keep all day in memory
+    snapshot: true
+  fleet:
+    public-ip: {{.IP}}
+    # allow etcd to slow down at times
+    etcd-request-timeout: 15
+  units:
+    - name: etcd.service
+      command: start
+    - name: fleet.service
+      command: start
+EOF
+`
 )
 
 func init() {
@@ -74,6 +100,43 @@ func (provisioner *CoreOSProvisioner) SetHostname(hostname string) error {
 	}
 
 	if _, err := provisioner.SSHCommand("sudo systemctl start system-cloudinit@var-tmp-hostname.yml.service"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (provisioner *CoreOSProvisioner) SetCloudInit(hostname string) error {
+	log.Debugf("SetCloudInit: %s", hostname)
+
+	ip, err := provisioner.Driver.GetIP()
+	if err != nil {
+		log.Fatalf("Could not get IP address for created machine: %s", err)
+	}
+
+	token, err := provisioner.SSHCommand("curl -sSL https://discovery.etcd.io/new?size=3")
+	if err != nil {
+		return err
+	}
+
+	data := map[string]interface{}{
+		"IP":       ip,
+		"HostName": hostname,
+		"Token":    token,
+	}
+
+	t := template.Must(template.New("cloudinit").Parse(cloudConfigTmpl))
+	buf := &bytes.Buffer{}
+	if err := t.Execute(buf, data); err != nil {
+		panic(err)
+	}
+	cloudConfigTmpl := buf.String()
+
+	if _, err := provisioner.SSHCommand(cloudConfigTmpl); err != nil {
+		return err
+	}
+
+	if _, err := provisioner.SSHCommand("sudo systemctl start system-cloudinit@var-tmp-cloudinit.yml.service"); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


nathanleclaire:
cc @mschygulla please take a look and let us know your thoughts.

mschygulla:
I had a look and I have to say that I don't think this is a good way to configure the etcd service.

With this pull request the etcd configuration is hardcoded in the docker-machine executable. You cannot modify or adjust the etcd service settings to your personal needs. The hardcoded URL (https://discovery.etcd.io) to generate the discovery token might also not work for everyone (e.g. if you run your own etcd cluster in a corporate environment, or simply on your laptop inside virtualbox without internet connection).

I think the better way to configure the etcd service is to use cloud-config as described here:

https://coreos.com/os/docs/latest/booting-on-ec2.html

sylus:
I see your point about the etcd url being different and some of the hard coded values only able to be changed on the running droplet. Could you provide more explanation on how you see docker machine provisioning using cloud-config? For instance I don’t know how docker machine would provision the cloud-config on digitalocean which I don’t think exposes an API to set it ahead of time so running docker-machine create would work. I could just have docker-machine read from an external file with a new flag but if given more information will try to work on a more acceptable way. ^_^

mschygulla:
In my eyes the docker-machine driver needs to be extended to allow an option to provide a cloud-config file. For example, take a look at my [PR-1479](https://github.com/docker/machine/pull/1479) for the VMwareFusion driver. I can imaging something similar for other cloud providers, at least where it's technical possible.

so0k:
Hi, I've always been using Docker on CoreOS (and I'm on Hyper-V). I rely heavily on user-data to pull binaries (experimental docker releases, swarm, kubernetes, ...) at the moment I provision my docker hosts with powershell scripts and doctl.. but I'd like to use Machine.

I've never written go before, but...

for digital ocean you can provide string for cloud-config in createDropletRequest: https://github.com/digitalocean/godo/blob/master/droplets.go#L133
so, should just update driver to accept user data string and pass it on to godo here:
https://github.com/docker/machine/blob/master/drivers/digitalocean/digitalocean.go#L166

However, @mschygulla added support for specifying url to config drive, while local machine would take a userdata string, wouldn't it be more consistent to always accept a string and do something like: https://github.com/kelseyhightower/isod/blob/master/main.go to generate the iso? (or would this not cross compile?)

any suggestions, comments?

janeczku:
> However, @mschygulla added support for specifying url to config drive, while the above would take a userdata string for cloud providers..

And thats why the provisioner is the wrong place for implementing this stuff in my opinion. This should be done on the driver level.

nathanleclaire:
I think @janeczku has a pretty valid point about keeping this out of the provisioners in general, however if there's any way we can bootstrap the CoreOS node with a basic sane configuration and that isn't being done right now, I'm for it.  However, I don't want to make heavy modifications elsewhere to support it.  Maybe cloud config as an optional parameters is something we could tackle in the core drivers for the 0.6.0 release.

sylus:
Is it possible to get more information about the direction this should go? As the earlier comment mentioned I just want sane configuration. If the direction is this should be solved exclusively on the driver level then that is fine but just want to make sure we are all on the same page. It really is unfortunate we have to resort to custom scripts rather then leveraging docker-machine which is just phenomenal in its ease of use. I myself am not a go expert but can get my way around though just want to make sure that the direction we choose is appropriate.

sylus:
I wonder if the CoreOS folks have any insight on this (sorry for the ping) @crawford @philips

libmachine/provision/coreos.go
```diff
@@ -21,6 +21,32 @@ const (
...
+    peer-heartbeat-interval: 1000
+    # don't keep all day in memory
+    snapshot: true
+  fleet:
+    public-ip: {{.IP}}
+    # allow etcd to slow down at times
+    etcd-request-timeout: 15
+  units:
+    - name: etcd.service
+      command: start
```
philips: you need to use etcd2

libmachine/provision/coreos.go
```diff
@@ -21,6 +21,32 @@ const (
...
+    - name: fleet.service
+      command: start
```
philips: why are you starting fleet?

philips:
Does docker machine know the IP of the machines before they boot? (It seems it does based on the cloud-config). If so you can just use static bootstrap: https://coreos.com/etcd/docs/latest/clustering.html#static

Adding @gyuho to help assist further if needed.

dgageot:
@sylus Do you plan to improve the PR based on @philips's feedback? Or are you willing to close it?

sylus:
I plan to improve it I am going to try to take a look at it this weekend.

nathanleclaire:
@sylus Any update on this?

sylus:
Hey! Sorry been so slow getting back to this issue. I have made the corresponding fixes mentioned by the CoreOS folks but this still doesn't solve the earlier concerns that this was the wrong approach to do in the provisioner and should be in the drivers logic instead. I think the digital ocean driver might be simplest based on what @so0k mentioned in the earlier comments. However the virtualbox driver looks to be completely integrated with boot2docker so that would need some reoorg to get that working with CoreOS.

I'll close this P.R. though as is the wrong approach and this is ultimately going to need code fixes done on a particular driver level. I will try to reopen a new P.R. against digitalocean at some point. Sorry for keeping this issue open so long in the queue. Many thanks!

nathanleclaire:
Why not start by revising this PR to provide a basic sane configuration as is?  Maybe static bootstrap like Brandon suggested?

If we're going to implement cloudinit support in drivers, it should be done as across the board (for each driver) as much as possible, which is a massive undertaking.

so0k:
I'm interested to implemement cloud-init for digitalocean & hyper-v drivers
(the ones I use the most)

PR #1973, opened 2015-10-12
Add tests to Google Driver

### LINKED ISSUE #1972: Google Driver is broken in master
Google Driver was broken by PR #1633

```
docker-machine create --driver google --google-project code-story-blog toto
panic: interface conversion: flag.Value is *flag.stringValue, not *cli.StringSlice
```

dgageot:

### DIFF
--- a/drivers/google/google_test.go
+++ b/drivers/google/google_test.go
@@ -1,0 +1,153 @@
+package google
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/codegangsta/cli"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDriverName(t *testing.T) {
+	driverName := newDriver().DriverName()
+
+	assert.Equal(t, "google", driverName)
+}
+
+func TestFailWithoutProject(t *testing.T) {
+	_, err := create([]string{"test-name"})
+
+	assert.EqualError(t, err, "Please specify the Google Cloud Project name using the option --google-project.")
+}
+
+func TestParseFlags(t *testing.T) {
+	driver, err := create([]string{
+		"--google-project", "test-project",
+		"--google-zone", "test-zone",
+		"--google-machine-type", "test-machine-type",
+		"--google-disk-size", "10000",
+		"--google-disk-type", "test-disk-type",
+		"--google-address", "test-address",
+		"--google-preemptible",
+		"--google-auth-token", "test-auth-token-path",
+		"--google-scopes", "test-scope1,test-scope2",
+		"--google-tags", "test-tag1,test-tag2",
+		"--google-username", "test-ssh-user",
+		"test-name"})
+
+	assert.NoError(t, err)
+	assert.Equal(t, driver.Project, "test-project")
+	assert.Equal(t, driver.Zone, "test-zone")
+	assert.Equal(t, driver.MachineType, "test-machine-type")
+	assert.Equal(t, driver.DiskSize, 10000)
+	assert.Equal(t, driver.DiskType, "test-disk-type")
+	assert.Equal(t, driver.Address, "test-address")
+	assert.Equal(t, driver.Preemptible, true)
+	assert.Equal(t, driver.AuthTokenPath, "test-auth-token-path")
+	assert.Equal(t, driver.Scopes, "test-scope1,test-scope2")
+	assert.Equal(t, driver.Tags, "test-tag1,test-tag2")
+	assert.Equal(t, driver.SSHUser, "test-ssh-user")
+	assert.Equal(t, driver.SSHPort, 22)
+}
+
+func TestSSHHostname(t *testing.T) {
+	mockComputeUtilToReturn(&ComputeUtil{ipAddress: "host"})
+
+	hostname, err := newDriver().GetSSHHostname()
+
+	assert.Equal(t, "host", hostname)
+	assert.NoError(t, err)
+}
+
+func TestGetIp(t *testing.T) {
+	mockComputeUtilToReturn(&ComputeUtil{ipAddress: "host"})
+
+	ip, err := newDriver().GetIP()
+
+	assert.Equal(t, "host", ip)
+	assert.NoError(t, err)
+}
+
+func TestGetIpFailure(t *testing.T) {
+	mockComputeUtilToFail("Fail to create ComputeUtil")
+
+	ip, err := newDriver().GetIP()
+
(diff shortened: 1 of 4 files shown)


dmp42:
LGTM
@nathanleclaire wdyt?

nathanleclaire:
Looks OK, but what issue is it fixing?  Is there a bug filed somewhere?

dgageot:
#1972

Le lun. 12 oct. 2015 à 19:58, Nathan LeClaire <notifications@github.com> a
écrit :

nathanleclaire:
Ah, I see, it closes https://github.com/docker/machine/issues/1972 ?

dgageot:
Yes it does. Did I forget something?

Thanks a lot
Le lun. 12 oct. 2015 à 20:00, Nathan LeClaire <notifications@github.com> a
écrit :

drivers/google/google_test.go
```diff
@@ -0,0 +1,153 @@
...
+// fail with provided message.
+func mockComputeUtilToFail(msg string) {
+	newComputeUtil = func(driver *Driver) (*ComputeUtil, error) {
+		return nil, errors.New(msg)
+	}
+}
+
+// create parses the provided args and tries to configure the driver for
+// a `create` action.
+func create(args []string) (*Driver, error) {
```
nathanleclaire: This `create` function, and the tests which depends on it, are out of scope for a unit test in `google` module (it should not know anything about `cli` module) and are a bit too close to integration tests for my tastes. They should be removed.

drivers/google/google_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/codegangsta/cli"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDriverName(t *testing.T) {
```
nathanleclaire: This test really seems quite unnecessary to me.  Why test a function which simply returns a hardcoded string?

drivers/google/google_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestFailWithoutProject(t *testing.T) {
+	_, err := create([]string{"test-name"})
+
+	assert.EqualError(t, err, "Please specify the Google Cloud Project name using the option --google-project.")
+}
+
+func TestParseFlags(t *testing.T) {
+	driver, err := create([]string{
```
nathanleclaire: I think this would be a lot better suited by testing `SetConfigFromFlags` directly.  To do so, just pass in a custom version of the `DriverOptions` interface (use `DriverOptionsMock` from `libmachine/hosttest` if you can and if you run into circular import issues go ahead and move it to a separate module).

nathanleclaire:
I don't think setting the functions to be package-level variables instead of being defined directly will lead to a good long-term solution.

In this instance, instead of modifying the functions to be variables, we should have `ComputeUtil` be an `interface`, of which we return one implementation for the "real" use case and another we mock out directly for the tests.  See `fakedriver.Driver` and how it is used in tests in the codebase for an example of doing this type of operation.

drivers/google/google_test.go
dgageot: I won't argue for that one. It's just that I'm trying to cover as much code as possible.

drivers/google/google_test.go
dgageot: I beg to differ. The issue we are to fix comes from code that fails to parse the command line. If we unit test `SetConfigFromFlags` with a `DriverOptionsMock`, we don't spot the bug. There needs to be something that verifies that a given command line transforms into the appropriate action or error. An e2e test would cost too much, would rely on external services and couldn't cover all edge cases. By calling the `create` method with all the combinations of arguments, we can almost test 100% of the driver and this is cheap and doesn't rely on anything external.

drivers/google/google_test.go
dgageot: See my comment on `TestParseFlags` method. I think this is the best way to test the driver

dgageot:
@nathanleclaire I like the interface based approach. I tried it but didn't succeed. I'll retry tomorrow.
Regarding this PR, it looks like there will be more discussion on the tests than on the code itself. As the fix is really important, should we merge the code and not the tests or should we merge both and start a follow-up PR to improve the tests?

nathanleclaire:
>I like the interface based approach. I tried it but didn't succeed. I'll retry tomorrow.

:+1: 

> Regarding this PR, it looks like there will be more discussion on the tests than on the code itself. As the fix is really important, should we merge the code and not the tests or should we merge both and start a follow-up PR to improve the tests?

My preference would be to merge the code without the tests and do a followup PR for the tests.

dgageot:
@nathanleclaire If I remove the test commits from this PR, we'll loose our discussion right? Should I create a new PR only with the non-test code?

drivers/google/google_test.go
nathanleclaire: But this issue seems to be due to the fact that the tags flag was set as `String`, not as `StringSlice`: https://github.com/docker/machine/blob/master/drivers/google/google.go#L104

`codegangsta/cli` seems to be rightfully complaining that it received a `StringFlag` for that `google-tags` value and in `SetConfigFromFlags` you are trying to query it as a `StringSlice`.

nathanleclaire: i.e., The command line parsing is working fine.  It's the driver code that's broken.

nathanleclaire:
@dgageot Yeah, that seems reasonable to me, then we can change the title of this PR.  Make sure to note the comment inline above.  Unless I'm mistaken, you don't need to introduce any new code but rather could fix the issue simply by changing `flags.String("google-tags")` to `flags.StringSlice("google-tags")`

drivers/google/google_test.go
dgageot: You are right, the parsing is ok but unless we test at his level, we can't see the problem. The test will inject a `StringSlice` and the code will read a `StringSlice`, it will be green and it will fail at runtime because there's somewhere else a flag definition that is not aligned with `SetConfigFromFlags`. See what I mean?

dgageot:
Well, I dont think I want to do that. `StringSlice` is not a comma separated list of values, it's used with multiple usage of the same flag. Something like `create --google-flag flag1 --google-flag flag2`. It's not the approach used for scope. I'd rather parse the comma separated `String` value

dgageot:
I created a new PR just for the fix #1977

PR #1977, opened 2015-10-12
Fix broken google driver

dgageot:

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -177,8 +177,6 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 		}
 	}
 
-	tags := append(d.Tags, firewallTargetTag)
-
 	instance := &raw.Instance{
 		Name:        c.instanceName,
 		Description: "docker host vm",
@@ -200,7 +198,7 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 			},
 		},
 		Tags: &raw.Tags{
-			Items: tags,
+			Items: parseTags(d),
 		},
 		ServiceAccounts: []*raw.ServiceAccount{
 			{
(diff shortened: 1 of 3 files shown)


dgageot:
@dmp42 How is it possible to decrease coverage by adding tests?...

dmp42:
@dgageot because `go list` doesn't get to compute_utils  it seems.
I'll have a look at what's going on.

dgageot:
@dmp42 @nathanleclaire except for the coverage bug I think it's ready to merge. Wdyt?

dmp42:
@dgageot explanation: code with no test files will not appear in the coverage report.

Thus *adding* a test file providing coverage below the average for a file that wasn't in previously will bring the overall coverage down.

This is silly - and I would blame go cover for that (or maybe coverall).

Let me know if you want to get a crack at that but it should be low priority.
A simple workaround is to add empty test files for all go files - but we have been there and it's messy.
I would rather have proper tests :-)

Bottom-line: LGTM

@nathanleclaire wdyt?

nathanleclaire:
I don't mind the effect on coverage, but I do have a few nitpicks about code style.  I'll write inline

drivers/google/compute_util.go
```diff
@@ -277,6 +275,17 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 	return nil
 }
 
+// tags computes the tags for the instance.
+func tags(d *Driver) []string {
```
nathanleclaire: Why not make this a method directly on the driver and name it something slightly more descriptive (i.e. `func (d *Driver) parseTags() []string` ?

dgageot: @nathanleclaire ok for the better name. Making the method a function on the driver gives too much importance to it I think. I'd rather leave it as is.

dmp42:
Nitpicks are fixed. LGTM.

nathanleclaire:
Hm, I meant for it to be `d.parseTags` instead of `parseTags(d)`, but maybe that's splitting hairs.  It's OK.

PR #1979, opened 2015-10-13
FIX #1975 - Improve godoc output

### LINKED ISSUE #1975: machine godoc contains an import statement
Documentation at https://godoc.org/github.com/docker/machine contains

```
package machine
import "github.com/docker/machine"

Machine defines interfaces to manage a variety of docker instances deployed on different backends (VMs, baremetal). The goal is to allow users get from zero to docker as fast as possible.

Index
```

I believe the `import "github.com/docker/machine"` line shouldn't be here

dgageot:

### DIFF
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -1,4 +1,4 @@
-package main
+package machine
 
 import (
 	"fmt"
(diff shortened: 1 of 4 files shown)


dgageot:
OK, let's not do that since machine is moving towards being more of a library than an application

PR #1987, opened 2015-10-15
Fix GetUserName() and its test

dgageot:

### DIFF
--- a/libmachine/mcnutils/utils.go
+++ b/libmachine/mcnutils/utils.go
@@ -11,6 +11,9 @@ import (
 	"strconv"
 	"time"
 
+	"os/exec"
+	"strings"
+
 	"github.com/docker/machine/libmachine/log"
 )
 
@@ -23,22 +26,25 @@ func GetHomeDir() string {
 	return os.Getenv("HOME")
 }
 
+// GetUsername finds current user's login.
+// We can't rely on `user.Current()` in standard library because it doesn't cross-compile well.
 func GetUsername() string {
-	u := "unknown"
-	osUser := ""
-
+	username := ""
 	switch runtime.GOOS {
 	case "darwin", "linux":
-		osUser = os.Getenv("USER")
+		output, err := exec.Command("whoami").Output()
+		if err == nil {
+			username = strings.TrimSpace(string(output))
+		}
 	case "windows":
-		osUser = os.Getenv("USERNAME")
+		username = os.Getenv("USERNAME")
 	}
 
-	if osUser != "" {
-		u = osUser
+	if username != "" {
+		return username
 	}
 
-	return u
+	return "unknown"
 }
 
 func CopyFile(src, dst string) error {
--- a/libmachine/mcnutils/utils_test.go
+++ b/libmachine/mcnutils/utils_test.go
@@ -4,7 +4,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"runtime"
 	"testing"
 )
 
@@ -54,17 +53,10 @@ func TestCopyFile(t *testing.T) {
 }
 
 func TestGetUsername(t *testing.T) {
-	currentUser := "unknown"
-	switch runtime.GOOS {
-	case "darwin", "linux":
-		currentUser = os.Getenv("USER")
-	case "windows":
-		currentUser = os.Getenv("USERNAME")
-	}
-
 	username := GetUsername()
-	if username != currentUser {
-		t.Fatalf("expected username %s; received %s", currentUser, username)
+
+	if username == "unknown" {
+		t.Fatalf("Couldn't find username. Got %s", username)
 	}
 }
 


dgageot:
@nathanleclaire wdyt?

nathanleclaire:
It might work, but I'm still highly skeptical due to the cgo issues mentioned.  https://groups.google.com/forum/#!msg/Golang-nuts/LB71rodPxP0/xoAlpgOsPoQJ seems to indicate there won't be a problem until runtime.

dmp42:
@dgageot I can confirm these x-compiling darwin from linux generates a binary that returns <nil> from user.Current()

nathanleclaire:
So, I think we are unable to merge this.  Sorry @dgageot -- I would love to be able to use native Go stuff instead of the hacky way.

dmp42:
@dgageot alternatively we could consider cross-building with CGO.
That's more involved, but we might consider it.

Also, look at what libcontainer does:

https://github.com/docker/libcontainer/blob/83a102cc68a09d890cce3b6c2e5c14c49e6373a0/user/user.go

or dbus:

https://github.com/docker/libcontainer/tree/83a102cc68a09d890cce3b6c2e5c14c49e6373a0/vendor/src/github.com/godbus/dbus

I'm pretty there are some tricks with build flags / direct sys calls.

dgageot:
@dmp42 @nathanleclaire Don't you think that a pretty huge problem we are facing? Code that compiles and passes the tests can fail at runtime because of cross-compilation issue and we don't freak out?

dgageot:
Reopening the PR since I'd like to discuss about solutions and maybe change the code.

dgageot:
The code in this PR is not the real problem. There are ways to make it work:

+ We could release on each of our target platforms instead of relying on basic cross-compilation
+ We could cross-compile with CGO enabled. There's a docker image that does that and it seems to work for `user.Current()` https://github.com/karalabe/xgo
+ If we decide to not fix the release process, we should at least have tests that protect us against this dangerous standard libraries. Can we detect those calls? We could cross-compile on osx and test on linux.

dgageot:
In the mean time, I've pushed an implementation that relies on `whoami` to fix the issue. @nathanleclaire Can you take a look at it?

nathanleclaire:
For starters, what problem with the current code are we trying to solve?  This isn't clear to me from the discussion.  If it's just that tests fail when run in a container, adding `ENV USER` in the Dockerfile (which we had before) will solve that problem easily without needing to change our build toolchain.

@dmp42 I don't see how any of the linked code will help with the problem?  The libcontainer one seems pretty Linux-specific and I'm not sure which part of the dbus package you're referencing.

> Don't you think that a pretty huge problem we are facing? Code that compiles and passes the tests can fail at runtime because of cross-compilation issue and we don't freak out?

The issues with the standard library package that require cgo are well-known and the reason why libraries like https://github.com/mitchellh/homedir exist.  As long as we're using Go native build, there's nothing to be done about it except avoid the problematic packages.

> We could release on each of our target platforms instead of relying on basic cross-compilation

I'm not sure what this means.  Build with cgo for each platform?

> We could cross-compile with CGO enabled. There's a docker image that does that and it seems to work for user.Current() https://github.com/karalabe/xgo

It clocks in at about 2.3GB in image size (vs. ~500MB in `golang:1.5`) and is an additional upstream dependency to integrate into our build tooling, debug, and to maintain if the maintainer disappears.  Also, I'm not sure how it would integrate with our reliance on build tags, etc.

> If we decide to not fix the release process, we should at least have tests that protect us against this dangerous standard libraries. Can we detect those calls? We could cross-compile on osx and test on linux.

I'm in favor of static checking for these problematic packages where available, but I don't think it's super necessary.  The maintainers are generally pretty aware of these issues.

I'm really skeptical about cgo in general, the current cross-compilation method is trivial (especially with Go 1.5) and I think modifying our build toolchain heavily at this point just to get proper username lookups will not deliver value for us.  Granted, not having native DNS lookups and access to the system certificate store is annoying, but doesn't seem to be causing us a dramatic amount of pain so far.

As for this PR, this code has been humming along working fine for several releases now, so I'm inclined towards not changing it.

dgageot:
@nathanleclaire Thank you for your feedback. You are right that we have to focus on the root problem. 

You say that `GetUsername()` is fine in the real-world, so let's not change it. Do you mind if I add a user in the Dockerfile to fix the in-container build?

Are you ok if I investigate the cross-compilation subject? We can either:
+ Don't change anything
+ Compile our binaries on target platforms instead of relying on cross-compilation
+ Cross-compile with CGO enabled. For the record, xgo was just something I found, not the ideal solution.
I'd like to really understand what each solution has to bring and also would like to take a look at how the other projects solve this problem.

nathanleclaire:
> You say that GetUsername() is fine in the real-world, so let's not change it. Do you mind if I add a user in the Dockerfile to fix the in-container build?

Not at all - go for it.

As for investigating using cgo, go ahead and do some research, but my tentative forays have always seemed to indicate that it may be more hassle than its worth.  The utility of being able to generate binaries for each target platform and architecture with one command is really hard to beat.

PR #1989, opened 2015-10-15
CI should run in containers

dgageot:


.travis.yml
```diff
@@ -1,13 +1,5 @@
...
-# - make lint
-- make vet
-- make test-short
-- make test-long
-- make coverage-send
+sudo: required
+dist: trusty
+language: bash
+services: docker
+script: USE_CONTAINER=true make dco fmt vet test-short test-long
```
dmp42: coverage-send? not sure it will work out of the box though

dgageot:
Will close this PR. It's trying to achieve too many things.
One problem is that CI should run the in-container build,
The other is that the in-container build fails and does too many things
I'll create separate tickets and PRs

PR #1990, opened 2015-10-15
Host needs VT-x/AMD-v enabled for docker-machine to work on virtualbox

dgageot:


dgageot:
@nathanleclaire @dmp42 It's not finished yet but can you tell me what you think?

It catches the case where VT-X is disabled globally in the bios, on windows. But there's no way to find that on win7.
It also verifies the VM after it's started, on every OS, to make sure the logs don't say that VT-X is disabled.

This should catch most problems and give an actionable plan to the user.

drivers/virtualbox/virtualbox.go
```diff
@@ -182,23 +187,58 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
...
+		// TODO: Check on linux/OSX
+		return false
 	}
 
-	return nil
+	log.Debug("COMMAND: wmic cpu get VirtualizationFirmwareEnabled")
+	stdout, err := exec.Command("wmic", "cpu", "get", "VirtualizationFirmwareEnabled").Output()
+	if err != nil {
+		log.Debugf("Couldn't check that VT-X/AMD-v is enabled. Will check that the vm is properly created.")
+		return false
```
nathanleclaire: But if it's a version of Windows which doesn't have `wmic` for whatever reason, it will return `false` whether or not VT-X/AMD-v is enabled. So it's dangerous, because it could have false negatives.

dgageot:
@nathanleclaire @dmp42 Do you think it could be merged as is? (once squashed)

dmp42:
LGTM
@nathanleclaire wdyt?

nathanleclaire:
Just a few nitpicks but it looks generally OK

drivers/virtualbox/virtualbox.go
```diff
@@ -41,6 +45,7 @@ const (
 
 var (
 	ErrUnableToGenerateRandomIP = errors.New("unable to generate random IP")
+	ErrMustEnableVTX            = errors.New("This computer doesn't have VT-X/AMD-v enabled. Enabling it in the bios is mandatory.")
```
nathanleclaire: `bios` => `BIOS`

drivers/virtualbox/virtualbox.go
```diff
@@ -174,23 +179,71 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
...
 	}
 
-	return nil
+	stdout, err := runCommand("wmic", "cpu", "get", "VirtualizationFirmwareEnabled")
+	if err != nil {
+		log.Debugf("Couldn't check that VT-X/AMD-v is enabled. Will check that the vm is properly created.")
+		return false
+	}
+
+	disabled := strings.Contains(string(stdout), "FALSE")
```
nathanleclaire: `runCommand` already returns `string`, not `[]byte`, no?

drivers/virtualbox/virtualbox.go
```diff
@@ -174,23 +179,71 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
...
+	return disabled
 }
 
-func (d *Driver) Create() error {
-	if err := vbm("--version"); err != nil {
-		return fmt.Errorf("Checking VirtualBox version failed: %s", err)
+// runCommand runs a shell command.
+func runCommand(name string, args ...string) (string, error) {
```
nathanleclaire: Since it actually calls `Output` in addition to running a command, why not name it `cmdOutput`?

drivers/virtualbox/virtualbox.go
```diff
@@ -174,23 +179,71 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
...
+	}
+
+	log.Debugf("STDOUT:\n{\n%v}", string(stdout))
+
+	return string(stdout), nil
+}
+
+// IsVTXDisabledInTheVM checks if VT-X is disabled in the started vm.
+func (d *Driver) IsVTXDisabledInTheVM() (bool, error) {
+	file, err := os.Open(path.Join(mcndirs.GetBaseDir(), "machines", d.MachineName, d.MachineName, "Logs", "VBox.log"))
```
nathanleclaire: For finding the machine directory, please use `d.ResolveStorePath(".")` instead

drivers/virtualbox/virtualbox.go
```diff
@@ -417,6 +466,16 @@ func (d *Driver) Start() error {
 		log.Infof("VM not in restartable state")
 	}
 
+	// Verify that VT-X is not disabled in the started VM
+	disabled, err := d.IsVTXDisabledInTheVM()
+	if err != nil {
+		return err
```
nathanleclaire: Please wrap this in another error, e.g.:

```golang
return fmt.Errorf("Checking if hardware virtualization is enabled failed: %s", err)
```

It will be easier to track down an issue if something fails that way.

drivers/virtualbox/virtualbox.go
```diff
@@ -174,23 +179,71 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
...
 	// Check that VBoxManage exists and works
-	if err := vbm(); err != nil {
-		return err
+	return vbm()
+}
+
+// IsVTXDisabled checks if VT-X is disabled in the bios. If it is, the vm will fail to start.
+// If we can't be sure it is disabled, we carry on and will check the vm logs after it's started.
+func (d *Driver) IsVTXDisabled() bool {
+	// TODO: Check on linux/OSX
```
hairyhenderson: Here's an OS X implementation for you :gift: :

```go
import (
    "fmt"
    "strings"
    "syscall"
)

func inSlice(s string, slice []string) bool {
    for _, v := range slice {
        if v == s {
            return true
        }
    }
    return false
}

func IsVTXDisabled() {
    val, err := syscall.Sysctl("machdep.cpu.features")
    if err != nil {
        fmt.Errorf("%#v", err)
    }
    features := strings.Split(val, " ")
    return !inSlice("VMX", features)
}
```

nathanleclaire: Why not just `strings.Contains(features, " VMX ")`? :)

hairyhenderson: or I guess that works too ;)

drivers/virtualbox/virtualbox.go
dgageot: Nice catch!

drivers/virtualbox/virtualbox.go
dgageot: Thanks a lot @hairyhenderson!

dgageot: @hairyhenderson the code doesn't compile on linux. I'll put that on hold. Windows was really the priority.

hairyhenderson: @dgageot - right - it's darwin-only. There's a different method for checking on Linux, but IMO the virtualbox driver isn't going to be used that often on Linux ;)

dmp42:
Nitpicks are addressed. Merging.

drivers/virtualbox/virtualbox.go
dmp42: @hairyhenderson would you carry this part through as a follow-up PR?

hairyhenderson: @dmp42 - would be happy to. :)

PR #2013, opened 2015-10-20
Ensure log writes only to os.Stderr for env and config command

nathanleclaire:
This fixes some issue where if you run `eval $(docker-machine env default)` and it needs to regenerate the certs it was dumping output to STDOUT and messing up the subshell

cc @jeffdm FYI


nathanleclaire:
LGTM

PR #2014, opened 2015-10-20
Spit out at least some message when flag parsing etc. fails

nathanleclaire:
Prints error messages such as: `flag provided but not defined: -engine-laber` which were missing before (only got message "Incorrect Usage")

Inspired by conversation with @moxiegirl and others about our previous less useful flag error messages


nathanleclaire:
LGTM

Issue #1983, opened 2015-10-14
Host needs VT-x/AMD-v enabled for docker-machine to work on virtualbox


dgageot:
If VT-x or AMD-v are not enabled on the host machine, the vm won't boot boot2docker properly and the whole creation process will hang forever.

To reproduce:

```
docker-machine create -d virtualbox default
```

, on a host without VT-x/AMD-v

We need to:
+ Detect that VT-x/AMD-v is not enabled and fail as early as possible
+ Understand why the `start` command hangs for what it looks forever

dmp42:
Random googling :-)

http://stackoverflow.com/questions/21537769/programatically-detect-if-vt-x-or-amd-v-is-enabled-on-windows

https://www.virtualbox.org/ticket/10694

thaJeztah:
Also see https://github.com/docker/toolbox/issues/166

jeanlaurent:
Got this one which explain for windows and linux how to detect it.

http://stackoverflow.com/questions/840873/how-to-determine-if-hardware-virtualization-is-available-via-code-c-c

nathanleclaire:
I think my main concern in the immediate future is to detect this case for Windows users (the most likely to have it disabled), so maybe we can focus on that.  I don't see any reliable way to find the information in the links referenced without assuming something like Powershell, though.  Maybe I'm overlooking something.

At any rate, can `VBoxManage` itself detect whether virtualization is enabled through the command line?  I know that it complains if you attempt to run it in Xen, so there might be something there we can try and parse.

dgageot:
We should be able to detect that VT-x/AMD-v is enabled on windows with this command:

```
> wmic cpu get VirtualizationEnabled
VirtualizationEnabled
TRUE
```

If wmic is not available then we could create the vm anyway and then use `VBoxManage showvminfo` to see if it is enabled.

dgageot:
The only way I found to detect afterwards is to look at the vm's log in `~/.docker/machine/machines/default/default/Logs/VBox.log`

When is works, it says:

```
00:00:00.121781 HM: HMR3Init: VT-x w/ nested paging and unrestricted guest execution hw support
```

Otherwise it says:

```
00:00:00.121781 HM: HMR3Init: Falling back to raw-mode: VT-x is disabled in the BIOS for both all CPU modes
```

dgageot:
Closed by https://github.com/docker/machine/pull/1990

serjN:
[...]

PR #2031, opened 2015-10-20
Adding support for darwin to IsVTXDisabled

hairyhenderson:
Here's an OS X/darwin implementation for `IsVTXDisabled` - related to #1983, and a follow-on to #1990.

/cc @dmp42 

### DIFF
--- a/drivers/virtualbox/virtualbox_windows.go
+++ b/drivers/virtualbox/virtualbox_windows.go
@@ -1,0 +1,21 @@
+package virtualbox
+
+import (
+	"strings"
+
+	"github.com/docker/machine/libmachine/log"
+)
+
+// IsVTXDisabled checks if VT-X is disabled in the BIOS. If it is, the vm will fail to start.
+// If we can't be sure it is disabled, we carry on and will check the vm logs after it's started.
+func (d *Driver) IsVTXDisabled() bool {
+	errmsg := "Couldn't check that VT-X/AMD-v is enabled. Will check that the vm is properly created: %v"
+	output, err := cmdOutput("wmic", "cpu", "get", "VirtualizationFirmwareEnabled")
+	if err != nil {
+		log.Debugf(errmsg, err)
+		return false
+	}
+
+	disabled := strings.Contains(output, "FALSE")
+	return disabled
+}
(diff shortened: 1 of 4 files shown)


dmp42:
@hairyhenderson nice!

It doesn't build (yet) though.
Complaining about:

```
# github.com/docker/machine/drivers/virtualbox
drivers/virtualbox/virtualbox.go:205: undefined: syscall.Sysctl
make: *** [/home/travis/gopath/src/github.com/docker/machine/bin/docker-machine-driver-virtualbox] Error 2
```

hairyhenderson:
Ugh. I forgot about that - `syscall.Sysctl` isn't implemented on Linux. I'll see if I can hack around it.

drivers/virtualbox/virtualbox_windows.go
```diff
@@ -0,0 +1,21 @@
...
+	"strings"
+
+	"github.com/docker/machine/libmachine/log"
+)
+
+// IsVTXDisabled checks if VT-X is disabled in the BIOS. If it is, the vm will fail to start.
+// If we can't be sure it is disabled, we carry on and will check the vm logs after it's started.
+func (d *Driver) IsVTXDisabled() bool {
+	errmsg := "Couldn't check that VT-X/AMD-v is enabled. Will check that the vm is properly created: %v"
+	output, err := cmdOutput("wmic", "cpu", "get", "VirtualizationFirmwareEnabled")
```
hairyhenderson: FYI: An alternative to running the `wmic` command could be to use https://github.com/StackExchange/wmi, but this might be overkill!

hairyhenderson:
@dmp42 - ok, _now_ it builds... I sort of wonder if it might make sense to turn on multi-OS builds in Travis though so we can also compile on OS X...

dmp42:
@hairyhenderson we are already cross-compiling, but you have a point about the ability to *test* in different envs.

dgageot:
Thank you @hairyhenderson! LGTM
ping @nathanleclaire

dmp42:
LGTM

drivers/virtualbox/virtualbox_windows.go
nathanleclaire: Definitely overkill -- would prefer to just run the command directly

nathanleclaire:
I have to say that I find the complexity vs. benefit ratio suspect, but LGTM

drivers/virtualbox/virtualbox_windows.go
hairyhenderson: :+1:

PR #1313, opened 2015-06-05
Fixing Filestore.GetActive to handle the case where `--swarm` is used

hairyhenderson:
Fixes #1311 and fixes #1136

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

PR #2032, opened 2015-10-21
Returning active host when swarm is active

### LINKED ISSUE #1136: New active host tracking does not support swarm / master node
Because we query the URL directly, the new active host tracking does not match `<masterIp>:3376` correctly (the URL returned for that instance uses port `2376`).  It should support tracking the active master node if present and set as "active" with `env --swarm`, although I am not sure how to do so in a non-hacky way at the moment (obviously the inelegant solution is to replace `2376` with `3376` in the URL, and check if that matches).

### LINKED ISSUE #1311: `docker-machine active` can't find active when `env --swarm` was used
Business as usual:

```console
$ eval "$(machine env hairyhenderson)"
$ docker-machine active                      
hairyhenderson
```

But when I add `--swarm` to the mix:

```console
$ eval "$(docker-machine env --swarm hairyhenderson)"
$ docker-machine active
Error getting active host: Active host not found
```

hairyhenderson:
Fixes #1311 and fixes #1136.

Reimplements the changes from #1313.

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -1,9 +1,15 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/docker/machine/cli"
+	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/persist"
+	"github.com/docker/machine/libmachine/state"
 )
 
 func cmdActive(c *cli.Context) {
@@ -20,4 +26,35 @@ func cmdActive(c *cli.Context) {
 	if host != nil {
 		fmt.Println(host.Name)
 	}
+}
+
+func getActiveHost(store persist.Store) (*host.Host, error) {
+	hosts, err := listHosts(store)
+	if err != nil {
+		return nil, err
+	}
+
+	hostListItems := getHostListItems(hosts)
+
+	for _, item := range hostListItems {
+		if item.Active {
+			return loadHost(store, item.Name)
+		}
+	}
+
+	return nil, errors.New("Active host not found")
+}
+
+// IsActive provides a single function for determining if a host is active
+// based on both the url and if the host is stopped.
+func isActive(h *host.Host, currentState state.State, url string) (bool, error) {
+	dockerHost := os.Getenv("DOCKER_HOST")
+
+	// TODO: hard-coding the swarm port is a travesty...
+	deSwarmedHost := strings.Replace(dockerHost, ":3376", ":2376", 1)
+	if dockerHost == url || deSwarmedHost == url {
+		return currentState == state.Running, nil
+	}
+
+	return false, nil
 }
--- a/commands/active_test.go
+++ b/commands/active_test.go
@@ -1,0 +1,45 @@
+package commands
+
+import (
+	"os"
+	"testing"
+
+	"github.com/docker/machine/drivers/fakedriver"
+	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/state"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	h = host.Host{
+		Name:       "foo",
+		DriverName: "fakedriver",
+		Driver: &fakedriver.Driver{
+			MockState: state.Running,
+		},
+	}
+)
+
+func TestIsActive(t *testing.T) {
+	cases := []struct {
+		dockerHost string
+		state      state.State
+		expected   bool
+	}{
+		{"", state.Running, false},
+		{"tcp://5.6.7.8:2376", state.Running, false},
+		{"tcp://1.2.3.4:2376", state.Stopped, false},
+		{"tcp://1.2.3.4:2376", state.Running, true},
+		{"tcp://1.2.3.4:3376", state.Running, true},
+	}
+
+	for _, c := range cases {
+		os.Unsetenv("DOCKER_HOST")
+		if c.dockerHost != "" {
+			os.Setenv("DOCKER_HOST", c.dockerHost)
+		}
+		actual, err := isActive(&h, c.state, "tcp://1.2.3.4:2376")
+		assert.Equal(t, c.expected, actual, "IsActive(%s, \"%s\") should return %v, but didn't", c.state, c.dockerHost, c.expected)
+		assert.NoError(t, err)
+	}
+}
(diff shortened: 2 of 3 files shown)


commands/active.go
```diff
@@ -21,3 +27,38 @@ func cmdActive(c *cli.Context) {
 		fmt.Println(host.Name)
 	}
 }
+
+func getActiveHost(store persist.Store) (*host.Host, error) {
```
hairyhenderson: I moved this into `active.go` from `ls.go` because it wasn't referenced in `ls.go` at all...

commands/active.go
```diff
@@ -21,3 +27,38 @@ func cmdActive(c *cli.Context) {
...
+			return h, nil
+		}
+	}
+
+	return nil, errors.New("Active host not found")
+}
+
+// IsActive provides a single function for determining if a host is active
+// based on both the url and if the host is stopped.
+func isActive(h *host.Host, currentState state.State, url string) (bool, error) {
```
hairyhenderson: I moved this from `ls.go` into `active.go` ... because it seemed like it belonged here? But it's not directly referenced from here, so I might revert this. Part of the reason is I wanted the unit tests to go in `active_test.go` so it seemed right to have the function under test in the associated `active.go`...

So many interdependencies in the `commands` package!

dmp42:
Thanks @hairyhenderson 

Tests: +1 \o/

commands/active.go
nathanleclaire: Yeah, code in `ls.go` requires this function, although it looks like they should both use the same code if possible.  It may require some re-jiggering to get there, though

nathanleclaire:
@dgageot Is correct that the additional boilerplate in that little block calling `loadHost` seems unneeded.

hairyhenderson:
@nathanleclaire, @dgageot - right, I see that now. Since I've moved the function I'll clean it up.

commands/active_test.go
```diff
@@ -0,0 +1,51 @@
...
+	h = host.Host{
+		Name:       "foo",
+		DriverName: "fakedriver",
+		Driver: &fakedriver.Driver{
+			MockState: state.Running,
+		},
+	}
+)
+
+func TestIsActiveReturnsFalseGivenNoActiveHost(t *testing.T) {
```
nathanleclaire: Why not make this into just one test with a test table?  You may have seen me ~~pestering~~ commenting about it on some of the other issues. e.g.:

```go
	cases := []struct {
		dockerHost     string
		currentState   state.State
		url            string
		expectedResult bool
		expectedErr    error
	}{
		{"", state.Running, "tcp://1.2.3.4:2376", false, nil},
		{"tcp://1.2.3.4:2376", state.Stopped, "tcp://1.2.3.4:2376", false, nil},
	}

	for _, c := range cases {
		result, err := isActive(&h, c.currentState, c.url)
		assert.Equal(t, expectedResult.result)
		assert.Equal(t, expectedErr, err)
	}
```

Please make sure to check the error too.

hairyhenderson: I'm not really a fan of that style of unit test, since it introduces complexity into the test itself. Also single test per behavior means there's no ambiguity about why a test is failing. And a failing assertion doesn't prevent further tests from being run...

But good point about the error - I always seem to miss those :)

nathanleclaire: > I'm not really a fan of that style of unit test, since it introduces complexity into the test itself.

How do you mean?

>  single test per behavior means there's no ambiguity about why a test is failing. 

I don't understand.  Tests always fail when the expected outputs don't match the actual outputs.  What's ambiguous about that?

> And a failing assertion doesn't prevent further tests from being run...

A failing assertion doesn't prevent further tests from being run AFAICT.  `assert.Equal` will invoke `t.Errorf` if the case fails and doesn't prevent the rest of the test from running `t.Fatalf` would do that.

hairyhenderson: > How do you mean? I find it groups related tests in a really concise, readable fashion.

_Maybe_, but it adds complexity. As soon as you have a `for` loop or `if` block, the code is now twice (or many) times more complex. Unit tests are a different kind of code than actual code, especially since there's no unit tests to test the unit tests...

> I don't understand. Tests always fail when the expected outputs don't match the actual outputs. What's ambiguous about that?

In my `go test` output, if I get a failure that says something like `TestIsActive failed` I have no immediate idea _why_ it failed. If, however, the failure says `TestIsActiveReturnsFalseGivenNoActiveHost failed`, I have a better idea of _what exact behaviour_ is failing. If I have multiple tests in a single method, I then have to start putting more information in the assertion, which is information that should be encoded in the function's name.

> A failing assertion doesn't prevent further tests from being run AFAICT

Ah. That's news to me... Most unit testing frameworks in most other languages I've worked with will immediately abort the test on the first failure. On further inspection, it looks like `testing` _does_ behave like you say.

But, the fact remains, it's easier to tell _what exactly_ is wrong when there's a 1:1 relationship with the assertion and the test function. For example:

```go
package main

import (
    "github.com/stretchr/testify/assert"
    "testing"
)

func Foo(i int) int {
    return i
}

func TestFoo(t *testing.T) {
    assert.Equal(t, Foo(1), 1)
    assert.Equal(t, Foo(2), 1)
    assert.Equal(t, Foo(3), 3)
    assert.Equal(t, Foo(4), 1)
}
```

```console
$ go test .
--- FAIL: TestFoo (0.00s)
        Error Trace:    main_test.go:14
	Error:  	Not equal: 2 (expected)
			        != 1 (actual)
		
        Error Trace:    main_test.go:16
	Error:  	Not equal: 4 (expected)
			        != 1 (actual)
		
FAIL
FAIL	_/tmp	0.007s
```

This gives me precious little information about _why_ those two tests failed. I'd have to use something like `assert.Equal(t, Foo(4), 1, "Foo(4)")` at the very least to even know what that assertion was actually _for_.

Anyway, all this is pretty academic, and while I feel strongly about it, this is your code, not mine.

If you're unconvinced, let me know and I'll change the test :)

nathanleclaire: I'm still pretty unconvinced :)

hairyhenderson: Ok, changed.

hairyhenderson:
@nathanleclaire, @dgageot - addressed the issues - PTAL

dgageot:
LGTM

nathanleclaire:
Thanks @hairyhenderson !

LGTM

hairyhenderson:
Thanks! :)

PR #2033, opened 2015-10-21
Code cleanup

dmp42:
Individual commits are grouping similar changes.

Mainly:

 * trivial stuff (commenting, ordering)
 * code duplication removal
 * tests
 * hardening base and generic a bit

(this is smaller than it seems)

cc @hairyhenderson since this carries through part of #1220

### DIFF
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -55,6 +55,7 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 	}
 }
 
+// NewDriver creates and returns a new instance of the driver
 func NewDriver(hostName, storePath string) drivers.Driver {
 	return &Driver{
 		SSHKey: defaultSSHKey,
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -1,10 +1,12 @@
 package generic
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/docker/machine/libmachine/drivers"
@@ -16,21 +18,50 @@ import (
 
 type Driver struct {
 	*drivers.BaseDriver
-	SSHKey string
+	sourceSSHKey string
 }
 
 const (
-	defaultSSHUser = "root"
-	defaultSSHPort = 22
+	driverName     = "generic"
 	defaultTimeout = 1 * time.Second
 )
 
 var (
-	defaultSSHKey = filepath.Join(mcnutils.GetHomeDir(), ".ssh", "id_rsa")
+	defaultSourceSSHKey = filepath.Join(mcnutils.GetHomeDir(), ".ssh", "id_rsa")
 )
 
-// GetCreateFlags registers the flags this driver adds to
-// "docker hosts create"
+// NewDriver creates and returns a new instance of the driver
+func NewDriver(hostName, storePath string) drivers.Driver {
+	return &Driver{
+		BaseDriver: &drivers.BaseDriver{
+			MachineName: hostName,
+			StorePath:   storePath,
+		},
+		sourceSSHKey: defaultSourceSSHKey,
+	}
+}
+
+func (d *Driver) Create() error {
+	log.Info("Importing SSH key...")
+
+	// TODO: validate the key is a valid key
+	if err := mcnutils.CopyFile(d.sourceSSHKey, d.GetSSHKeyPath()); err != nil {
+		return fmt.Errorf("unable to copy ssh key: %s", err)
+	}
+
+	if err := os.Chmod(d.GetSSHKeyPath(), 0600); err != nil {
+		return fmt.Errorf("unable to set permissions on the ssh key: %s", err)
+	}
+
+	log.Debugf("IP: %s", d.IPAddress)
+
+	return nil
+}
+
+func (d *Driver) DriverName() string {
+	return driverName
+}
+
 func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 	return []mcnflag.Flag{
 		mcnflag.StringFlag{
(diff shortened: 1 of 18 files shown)


drivers/generic/generic.go
```diff
@@ -14,14 +15,13 @@ import (
 	"github.com/docker/machine/libmachine/state"
 )
 
-type Driver struct {
+type driver struct {
 	*drivers.BaseDriver
 	SSHKey string
```
hairyhenderson: This tripped me up - in #1220 I had renamed the references to this to `SSHKeyPath`, which was wrong, because what _this_ `SSHKey` is is actually the _original_ SSH key, which is always imported into `SSHKeyPath`.

Because of this confusion, I think it would be a good idea to rename this to `SourceSSHKey` or `OrigSSHKey` or something like that. It doesn't even need to be exposed, so maybe something like `sourceSSHKey` would work.

Obviously not technically necessary, but I think a rename would make the code a lot more clear :)

hairyhenderson:
:+1:. When this lands, I'll kill #1220 :)

drivers/generic/generic.go
dmp42: It tripped me too initially.
I'll update.

dgageot:
@dmp42 Appart from the couple of details, LGTM

dmp42:
@dgageot updated according to your comments, thanks.

nathanleclaire:
Changes mostly look fine but why do you rename `Driver` to `driver` in generic.go?  It should be exportable.

nathanleclaire:
Also, the diff is really hard to read, I guess because you seem to be alphabetizing / reordering methods at the same time as refactoring?  I'd appreciate having that done in separate PRs since I am really having a hard time figuring out what changed vs. just moved around in this one.

libmachine/drivers/base_test.go
```diff
@@ -0,0 +1,31 @@
+package drivers
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestIP(t *testing.T) {
+	ip, err := (&BaseDriver{}).GetIP()
```
nathanleclaire: Let's use a test table here, it'd be clearer and will show newcomers the right way how to do such a test, i.e.:

```go
	cases := []struct {
		baseDriver  Driver
		expectedIp  string
		expectedErr error
	}{
		{&BaseDriver{}, "", errors.New("IP address is not set")},
		{&BaseDriver{IPAddress: "2001:4860:0:2001::68"}, "2001:4860:0:2001::68", nil},
		{&BaseDriver{IPAddress: "192.168.0.1"}, "192.168.0.1", nil},
		{&BaseDriver{IPAddress: "192.168.0.1"}, "192.168.0.1", nil},
		{&BaseDriver{IPAddress: "::1"}, "::1", nil},
		{&BaseDriver{IPAddress: "whatever"}, "", errors.New("IP address is invalid: whatever")},
	}

	for _, c := range cases {
		ip, err := c.baseDriver.GetIP()
		assert.Equal(t, ip, c.expectedIp)
		assert.Equal(t, err, c.expectedErr)
	}
```

dgageot:
```
assert.Equal(t, ip, c.expectedIp)
assert.Equal(t, err, c.expectedErr)
```

should really be:

```
assert.Equal(t, c.expectedIp, ip)
assert.Equal(t, c.expectedErr, err)
```

Otherwise, in case of test failure, the error message will be confusing

libmachine/drivers/base_test.go
```diff
@@ -0,0 +1,31 @@
...
+	assert.Equal(t, err.Error(), "IP address is not set")
+
+	check := "192.168.0.1"
+	ip, err = (&BaseDriver{IPAddress: check}).GetIP()
+	assert.Equal(t, err, nil)
```
dgageot: This can be replaced by `assert.NoError(t, err)`

libmachine/drivers/base_test.go
nathanleclaire: Ah, nice catch @dgageot

dmp42:
Updated the tests with the suggestion, thanks :)

> Also, the diff is really hard to read, I guess because you seem to be alphabetizing / reordering methods at the same time as refactoring? I'd appreciate having that done in separate PRs since I am really having a hard time figuring out what changed vs. just moved around in this one.

I grouped them in different commits for that reason.
I'll split up in different PR if you want in the future.
Ordering now matches alphabetical order (and the interface order).

> why do you rename Driver to driver in generic.go? It should be exportable.

That's actually an open question: is there a reason to export from the driver implementation itself?
Wdyt?

nathanleclaire:
> That's actually an open question: is there a reason to export from the driver implementation itself?

Yes -- ideally (it is not quite there today), they would be independently usable as part of running `libmachine` directly.

nathanleclaire:
> I grouped them in different commits for that reason.
> I'll split up in different PR if you want in the future.
> Ordering now matches alphabetical order (and the interface order).

Yeah, would definitely appreciate it in the future to split up anything which is actually changing code vs. just moving it around.  Thanks!

dmp42:
> Yes -- ideally (it is not quite there today), they would be independently usable as part of running libmachine directly.

I don't understand. This is the driver implementation, that is used by the plugin only.

What am I missing?

nathanleclaire:
> What am I missing?

It's intended that the driver struct could be used by external modules -- I want to decouple the `libmachine` API from the specific use case in the Docker Machine CLI.  `Driver` structs should be usable from anywhere.  E.g. I should be able to create a `Host` and set the `Driver` by directly importing the relevant module without having to go through all the RPC song-and-dance.

Besides the theoretical concern, `Driver` is already being used in another package: the `main` module of the relevant plugin.  The only reason that it still compiles is because the plugin file uses `NewDriver` instead of directly instantiating it.  Since it's already "leaking" outside the module, I don't see any reason to try and confine it.

dmp42:
> Besides the theoretical concern, Driver is already being used in another package: the main module of the relevant plugin. The only reason that it still compiles is because the plugin file uses NewDriver instead of directly instantiating it. Since it's already "leaking" outside the module, I don't see any reason to try and confine it.

The reason why it works is because they implement libmachine Driver. The Driver interface and the individual driver implementations are different things. What "leaks", and what should (and is) used outside is the libmachine.drivers.Driver interface, and this is good. On the other hand, my question is: why would leaking individual drivers implementation be useful?

Sorry if this is confusing.

dmp42:
Put otherwise: we don't want people to call a specific Driver directly, we want them to use the NewDriver method. Thus we should not make the implementation exportable.

nathanleclaire:
> The reason why it works is because they implement libmachine Driver. The Driver interface and the individual driver implementations are different things. What "leaks", and what should (and is) used outside is the libmachine.drivers.Driver interface, and this is good

Ah, I mis-remembered the typical `NewDriver` method signature (and should it return `drivers.Driver` instead of the concrete type? Looks like it's handled inconsistently across the various drivers), but I still think the drivers need to be exported, in case you want to implement something like this: https://github.com/docker/machine/pull/1729#issuecomment-133333988 which invokes the `virtualbox.Driver` directly without going through all of the RPC / plugin code.  There are more intended users of the `libmachine` API than just the Machine CLI.

> Put otherwise: we don't want people to call a specific Driver directly, we want them to use the NewDriver method. Thus we should not make the implementation exportable.

No, I want them to be able to create bare-bones `Driver` structs if they really want to.

dmp42:
Yes, NewDriver should really return `drivers.Driver` (because that's what register driver expects). 

> No, I want them to be able to create bare-bones Driver structs if they really want to.

That makes us bound to maintaining existing implementations signatures - if people rely on them, and we can no longer evolve the implementations without breaking the contract.

Do such uses exist in the wild?

And what is the benefit for us in this case?

nathanleclaire:
> That makes us bound to maintaining existing implementations signatures - if people rely on them, and we can no longer evolve the implementations without breaking the contract.

I don't understand what the concern is here.  If you write your code towards a particular version of the library, sometimes things break between versions.  The whole point of a versioned 1.0 `libmachine` API would be to draw that line in the sand.  If we feel we need to move to a better model, we need to have a clean version break.

> Do such uses exist in the wild?

No, because it's not usable in such a way in its current state.  I'm trying to get it there.  Rancher is one such example of a company which is shelling out to Docker Machine CLI today and would be delighted to have native Go bindings.

> And what is the benefit for us in this case?

Building / prototyping new tools on top of and related to Docker Machine becomes much, much, easier.

dmp42:
We can discuss this later today if you want :)
We are probably talking about two different things.

dmp42:
> I don't understand what the concern is here.

Maintainability. Clarity. Code quality.
If something is exportable, it's public, and cannot change easily, has to be documented.

> If you write your code towards a particular version of the library, sometimes things break between versions.

Breaking the interface is something we are not going to do often.
This is why we have this generic interface.
The only benefit you get from exporting the concrete implementation is making the driver's specific methods accessible, and these do not implement a common interface. They are really driver specific, most of the time internal implementation details.

> No, because it's not usable in such a way in its current state.

Then why export it in this state?

>  Rancher is one such example of a company which is shelling out to Docker Machine CLI today and would be delighted to have native Go bindings.

And they could do that without having access to the concrete drivers implementation. Am I missing something?

> Building / prototyping new tools on top of and related to Docker Machine becomes much, much, easier.

But then again, it just gives access to specific implementation details of a specific driver, which the very purpose of machine is to abstract away, correct?

Hope that clarifies.

nathanleclaire:
Let's discuss in person -- I think we are communicating / understanding different things.

dmp42:
@nathanleclaire following our conversation, rolled-back and making the Driver implementation public again.

PTAL.

nathanleclaire:
LGTM

Issue #2011, opened 2015-10-19
0.5.0-rc1: amazonec2: Cannot create Machines due to aggressive SSH timeout


dave-tucker:
```
Running pre-create checks...
Creating machine...
Waiting for machine to be running, this may take a few minutes...
Machine is running, waiting for SSH to be available...
Error creating machine: Error waiting for SSH: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (6) exceeded
```

The machine wasn't yet ready for SSH :( A few mins later, it was ready but in an un-provisioned state

```
$ docker-machine ls
NAME        ACTIVE   DRIVER       STATE     URL                         SWARM
1.9rc1      -        virtualbox   Running
dev         -        virtualbox   Saved
devx        -        virtualbox   Running
mh-consul   -        amazonec2    Running   tcp://54.246.230.147:2376

```

Please increase the number of retries, or make it configurable!

PR #2048, opened 2015-10-21
Increase SSH timeout back to five minutes

nathanleclaire:
Closes https://github.com/docker/machine/issues/2011

### DIFF
--- a/libmachine/drivers/utils.go
+++ b/libmachine/drivers/utils.go
@@ -2,7 +2,6 @@ package drivers
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


dmp42:
LGTM
Do you remember why we moved to 6 x 5 secs earlier?

nathanleclaire:
It was kind of a mistake on my part actually.  I had set it a bit over-aggressively when trying to account for the possibility that the SSH dial might not work and needs to time out.  This is just changing it back to what it was before.

PR #2065, opened 2015-10-23
Adding localhost to the list of alt_names

dgageot:
When attempting to connect to the docker api from the machine itself,
the TLS verification of the certificate checked against the public
IP address of the primary interface.  This is undesirable on hosts
which have NAT rules that block access to that address by default.

Adding "localhost" to the list of alt_names allows the cert to be
verified and connections to localhost (either 127.0.0.1 or [::1]) to
the port to pass verification. Otherwise one would need to disable
verification just to connect to the local docker instance.


dgageot:
This PR replaces https://github.com/docker/machine/pull/1064

dmp42:
LGTM

PR #2053, opened 2015-10-22
Add heartbeat / automatic server cleanup code

nathanleclaire:
This replaces the previous method of attempting to clean up servers when
an unexpected exit occurs in the client (e.g. SIGINT or panic) by a
heartbeat protocol.  If the server does not hear from the connecting
client within a certain interval of time (500ms in this commit), it will
de-activate itself.  This prevents dangling Docker Machine server
processes from accumulating.

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

PR #2068, opened 2015-10-23
Make commands code easier to test

dgageot:
Made sure the commands return errors and don't just call `fatal`
It makes thing much easier to test.

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -12,20 +12,27 @@ import (
 	"github.com/docker/machine/libmachine/state"
 )
 
-func cmdActive(c *cli.Context) {
+var (
+	errTooManyArguments = errors.New("Error: Too many arguments given")
+)
+
+func cmdActive(c *cli.Context) error {
 	if len(c.Args()) > 0 {
-		fatal("Error: Too many arguments given.")
+		return errTooManyArguments
 	}
 
 	store := getStore(c)
+
 	host, err := getActiveHost(store)
 	if err != nil {
-		fatalf("Error getting active host: %s", err)
+		return fmt.Errorf("Error getting active host: %s", err)
 	}
 
 	if host != nil {
 		fmt.Println(host.Name)
 	}
+
+	return nil
 }
 
 func getActiveHost(store persist.Store) (*host.Host, error) {
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -19,8 +19,8 @@ import (
 
 var (
 	ErrUnknownShell       = errors.New("Error: Unknown shell")
-	ErrNoMachineSpecified = errors.New("Error: Expected to get one or more machine names as arguments.")
-	ErrExpectedOneMachine = errors.New("Error: Expected one machine name as an argument.")
+	ErrNoMachineSpecified = errors.New("Error: Expected to get one or more machine names as arguments")
+	ErrExpectedOneMachine = errors.New("Error: Expected one machine name as an argument")
 )
 
 func newPluginDriver(driverName string, rawContent []byte) (*rpcdriver.RpcClientDriver, error) {
@@ -32,29 +32,25 @@ func newPluginDriver(driverName string, rawContent []byte) (*rpcdriver.RpcClient
 	return d, nil
 }
 
-func fatal(args ...interface{}) {
-	log.Fatal(args...)
+func fatalOnError(command func(context *cli.Context) error) func(context *cli.Context) {
+	return func(context *cli.Context) {
+		if err := command(context); err != nil {
+			log.Fatal(err)
+		}
+	}
 }
 
-func fatalf(fmtString string, args ...interface{}) {
-	log.Fatalf(fmtString, args...)
-}
-
-func confirmInput(msg string) bool {
+func confirmInput(msg string) (bool, error) {
 	fmt.Printf("%s (y/n): ", msg)
+
 	var resp string
 	_, err := fmt.Scanln(&resp)
-
 	if err != nil {
-		fatal(err)
+		return false, err
 	}
 
-	if strings.Index(strings.ToLower(resp), "y") == 0 {
-		return true
-
-	}
-
-	return false
+	confirmed := strings.Index(strings.ToLower(resp), "y") == 0
+	return confirmed, nil
 }
 
 func getStore(c *cli.Context) persist.Store {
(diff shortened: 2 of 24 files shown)


dgageot:
@docker/machine-maintainers Can you please take a look? I'd like to add more tests and this is a pre-requisite.

nathanleclaire:
Nice yeah this is good, when the https://github.com/docker/machine/pull/2053 gets merged in (soon) we will be able to get rid of the `fatal`and `fatalf` functions completely.  They were an (admittedly bad) way to try and control the exit flow of the program so we could properly clean up plugins in a non-normal exit.

dmp42:
@dgageot can you rebase?

dmp42:
LGTM
@nathanleclaire PTAL

commands/commands.go
```diff
@@ -32,29 +32,29 @@ func newPluginDriver(driverName string, rawContent []byte) (*rpcdriver.RpcClient
 	return d, nil
 }
 
-func fatal(args ...interface{}) {
-	log.Fatal(args...)
+func fatalOnError(command func(context *cli.Context) error) func(context *cli.Context) {
+	return func(context *cli.Context) {
+		if err := command(context); err != nil {
+			fatal(err)
```
nathanleclaire: This can just call `log.Fatal` directly now.  Since the cleanup has been moved to server-side, we can eliminate the `fatal` and `fatalf` wrappers completely.

nathanleclaire:
Needs a rebase and squash, but looking good.  I'm excited about this.

dgageot:
@docker/machine-maintainers Fixed, rebased and squashed

dgageot:
@nathanleclaire To make the commands really easy to test, we should also:
+ Pass the commands an interface instead of a `*cli.Context`
+ Pass the commands an interface to manipulate the store. Right now the call functions defined in `commands.go`
Wdyt?

nathanleclaire:
> Pass the commands an interface instead of a *cli.Context

Yes, that'd be great!

> Pass the commands an interface to manipulate the store. Right now the call functions defined in commands.go Wdyt?

My preference would be to make it so the commands could get a `libmachine` Client  (type doesn't exist yet, but should) from the passed interface.  There's a variety of things which are implicit it module-level variables (e.g. SSH client type, output / error log writers, store path etc.) which would be far better off encapsulated in a `libmachine` client struct that can be configured, and more easily injected in for test.

dgageot:
@nathanleclaire Can you list one or two functions you expect on the libmachine struct?

commands/active.go
```diff
@@ -12,20 +12,23 @@ import (
 	"github.com/docker/machine/libmachine/state"
 )
 
-func cmdActive(c *cli.Context) {
+func cmdActive(c *cli.Context) error {
 	if len(c.Args()) > 0 {
-		fatal("Error: Too many arguments given.")
+		return errors.New("Error: Too many arguments given")
```
ehazlett: Typically I like to reserve `errors.New` for named errors and then use `fmt.Errorf` to return a one-off message.

i.e.
```
var ErrTooManyArguments = errors.New("too many arguments given")
...

return ErrTooManyArguments
```

commands/create.go
```diff
@@ -130,16 +132,16 @@ func cmdCreateInner(c *cli.Context) {
 
 	if name == "" {
 		cli.ShowCommandHelp(c, "create")
-		fatal("Error: No machine name specified.")
+		return errors.New("Error: No machine name specified")
```
ehazlett: same as above

commands/create.go
```diff
@@ -130,16 +132,16 @@ func cmdCreateInner(c *cli.Context) {
...
 	}
 
 	validName := host.ValidateHostName(name)
 	if !validName {
-		fatal("Error creating machine: ", mcnerror.ErrInvalidHostname)
+		return fmt.Errorf("Error creating machine: %s", mcnerror.ErrInvalidHostname)
```
ehazlett: same as above

nathanleclaire:
@dgageot Well, one example method is `Create`.  Right now the "wrapper" create method just gets invoked directly on the module -- `libmachine.Create`.  It should be in a struct to have some configurable parameters.  Likewise the `Store` should be embedded in the `libmachine` client and `Load` should go through that instead.

commands/scp.go
```diff
@@ -120,19 +120,17 @@ func runCmdWithStdIo(cmd exec.Cmd) error {
...
 
-func cmdScp(c *cli.Context) {
+func cmdScp(c *cli.Context) error {
 	hostLoader = &ScpHostLoader{}
 
 	args := c.Args()
 	if len(args) != 2 {
 		cli.ShowCommandHelp(c, "scp")
-		fatal("Improper number of arguments.")
+		return errors.New("Improper number of arguments")
```
ehazlett: same as above (`fmt.Errorf`)

ehazlett:
few minor nits, otherwise lgtm.   needs a rebase.  thanks!

dgageot:
@nathanleclaire @ehazlett Thanks for the feedback. Fixed and rebased!

dgageot:
@nathanleclaire Will you merge this one for 0.5.0?

nathanleclaire:
One thing to change before merge, sorry to be so late in asking for it, but I just now realized.

I think we should change the wrapper function name to something like `exitOnError` and simply `return` instead of calling `log.Fatal`.

The reasoning for this is that in `main.go` we call `app.Run` and the print the error if it returns one.  This error is important to return because sometimes it includes messages like `Flag -virtualbox-meomry not recognized`, which is a vital hint to the user that we were missing until recently.

Can you update it?  Once again, sorry for just now asking for this, but the rest of the code looks good and I feel OK about merging it for 0.5.0.

nathanleclaire:
Actually, I might be mis-judging behavior on this one.  Let me try and get back to you.

dgageot:
@nathanleclaire A simple `return` instead of a `log.Fatal` would swallow the error triggered by the command itself.
That'd be nice to merge this PR and then carry on with the improvements in another PR.

dgageot:
@nathanleclaire I've got wip on this subject here: https://github.com/dgageot/machine/tree/more-cleanup-for-more-tests

nathanleclaire:
@dgageot Yeah I think this is OK -- sorry for introducing some confusion into the mix.

nathanleclaire:
Yep it's G2G

PR #2069, opened 2015-10-23
Set default VirtualBox disk size to 200GB

tiangolo:
As commented in this pull request to Docker Toolbox: https://github.com/docker/toolbox/pull/203 the default disk size for a Docker Machine in VirtualBox is set to 20GB, it could be useful to set that default to 200GB.

I edited the code based on this comment: https://github.com/docker/toolbox/pull/203#issuecomment-144593368 by @SydOps, who nailed down where that default was set.

I submit this pull request as asked by @mchiang0610 and @nathanleclaire in other comments in that PR (https://github.com/docker/toolbox/pull/203#issuecomment-150402370).

If that other PR seems to work for Docker Toolbox users and don't rise any additional issues and if you see fit, you may consider merging this PR to update the default Docker-Machine VirtualBox disk size to 200GB.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -39,7 +39,7 @@ const (
 	defaultHostOnlyNictype     = "82540EM"
 	defaultHostOnlyPromiscMode = "deny"
 	defaultNoShare             = false
-	defaultDiskSize            = 20000
+	defaultDiskSize            = 204800
 )
 
 var (


nathanleclaire:
NOT LGTM until we see how setting the Toolbox default goes

mchiang0610:
^ yes @nathanleclaire let's see the results from separated Toolbox test first.

SydOps:
@tiangolo 

Seems there are more `defaultDiskSize` setting in this repository.

https://github.com/docker/machine/search?utf8=%E2%9C%93&q=defaultdisksize&type=Code

if we change it for VirtualBox, we should change the rest vm as well.

tiangolo:
OK, thanks @SydOps, I see there are several of those variables with the same default size, but some have very low values: "google", "exoscale" and "softlayer", I don't know the internals of any of those other drivers and which problems could arise.

Also, if any of those other drivers actually allocate the physical disk size from the beginning, that would probably be a problem. As VirtualBox has a dynamic disk size limited by the creation disk size but that allocates only about 200MB initially, I see how it could work and wouldn't be a problem, but for the others I'm not sure.

I would prefer to ask @nathanleclaire what we should do. I know this isn't even approved, but I would like to know if I should update this PR including any of those other drivers' default disk size and which of them.

nathanleclaire:
As I said before, NOT LGTM until we see the results from the ToolBox.  This is a really big setting to be playing around with, and if we make something dangerous the default in Machine, we're in a LOT of trouble.  Let's wait a few months and see.

PR #2070, opened 2015-10-23
FIX #1985 Build with Docker on TravisCI

### LINKED ISSUE #1985: CI should run in container
The project can be built in-container and that's the only way to make sure the developer host has no impact on the build/tests.

Currently, on test is broken in-container but passes on most developer workstations.
Nobody noticed because:

+ Developer tend to not use in-container build because it's slower. That's OK.
+ The CI doesn't run the in-container build. That's **not** OK.

We should update our TravisCI build to do everything in-container. That's ok since Travis supports Docker well http://docs.travis-ci.com/user/docker/

dgageot:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,5 @@
-language: go
-go:
-- 1.5.1
-install:
-- go get github.com/mattn/goveralls
-script:
-- make dco
-- make fmt
-# - make lint
-- make vet
-- make test-short
-- make test-long
-- make coverage-send
+sudo: required
+dist: trusty
+language: bash
+services: docker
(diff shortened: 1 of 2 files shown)


dgageot:
ping @dmp42

dmp42:
@dgageot this breaks s*** coveralls.
Can you keep the `coverage-send` target out of container? It doesn't matter at all.

dgageot:
@dmp42 There must be en env variable to set. I'll check that

dmp42:
@dgageot so, goveralls is supposed to detect a travis env somehow.
Check https://github.com/mattn/goveralls

dmp42:
LGTM

Issue #1887, opened 2015-09-20
Fix lint issues


dmp42:
After a first pass at this at #1886 we have:

`make lint ./... 2>&1 | grep -v "drivers/"  | grep -v "should have comment"  | wc -l`

47 non-trivial issues to fix (252 overall).

PR #2074, opened 2015-10-23
Linting 1

dmp42:
This is only cleanup:

 * dead code removal
 * linting variable names
 * harmonizing return values on NewDriver
 * harmonize a couple of inline code comments

Now, linter passes except:

 * code documentation inline on exported stuff
 * amazon driver, which is erf

### DIFF
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -13,7 +13,6 @@ import (
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/mcnflag"
-	"github.com/docker/machine/libmachine/mcnutils"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/docker/machine/libmachine/state"
 )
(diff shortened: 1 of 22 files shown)


dmp42:
This if a first step to fix #1887

These changes are innocuous, since they are really just renaming symbols.

Still it's important that we get this right before the drivers get out of the tree (since for example the dependency graph on libmachine is not clear).

dgageot:
LGTM

dgageot:
@dmp42 Can you rebase please?

nathanleclaire:
Mmm might want to hold off until after the release to rebase these, so that it only has to be done once.

nathanleclaire:
I think most of the changes probably will be OK, but we need to make sure not to change anything other than casing in the exported struct field names for Drivers (just a note for self and others) as it could cause issues with `json.Marshal`.

nathanleclaire:
Oh yes, also, don't change anything in `cli/` -- it's essentially a vendored directory and progress on getting the PR merged in looks good, so hopefully soon I can re-add it to Godeps.

dmp42:
Ok, thanks.
Since all this moved a lot, I'll close and resubmit soon.

PR #2075, opened 2015-10-23
Linting 2

dmp42:
Everything but libmachine & drivers (except inline documentation).

Is on top of the drivers cleanup.

PR #2076, opened 2015-10-24
Linting 3, final

dmp42:
Following #2074 & #2075 on top of which this is based.

This contains the final base linting effort to libmachine, that is specially important before we can move to the externalized plugins.

Why does go lint matters?

 * don't fight the language - go is opinionated (unlike, say, ruby) and provide sane defaults and great tooling, which we should leverage, instead of fighting against
 * we are the only docker project (AFAIK) that does not lint
 * doing it after the plugins are outsourced will be impossible, or very painful

This PR also activates partial linting of the codebase as part of the `test` target, and enable linting on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,11 @@ go:
 - 1.5.1
 install:
 - go get github.com/mattn/goveralls
+- go get github.com/golang/lint/golint
 script:
 - make dco
 - make fmt
-# - make lint
+- make lint
 - make vet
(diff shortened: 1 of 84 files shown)


nathanleclaire:
Will review these after the release.

dgageot:
ping @jeanlaurent

dgageot:
@dmp42 Can you rebase please?

PR #2088, opened 2015-10-26
Fix get hostname

dgageot:
These commits introduce a bug.

When a Driver overrides `GetIP()`, it expects `GetSSHHostname` to call his version of `GetIP()` not the version from `BaseDriver`


nathanleclaire:
Thanks and nice catch @dgageot 

LGTM

PR #2090, opened 2015-10-26
Stop heartbeat when there is an issue connecting to the server

nathanleclaire:
It's not a super common case, but I noticed this would indefinitely spam the log message instead of properly disconnecting if I closed my laptop case while SSHed into a machine.

cc @docker/machine-maintainers for review


dgageot:
LGTM

PR #2097, opened 2015-10-27
Print a better error when virtualbox fails

dgageot:
Instead of printing `exit status 1` we'll print the stderr output.


dgageot:
@docker/machine-maintainers Can you take a look please?

dmp42:
LGTM!

nathanleclaire:
LGTM

In the future, we should consider returning (and checking for) a custom error type, since we are starting to get more control over what our VirtualBox operations return, and it will be more robust than string checking.

PR #2099, opened 2015-10-27
Smooth rough edges around create help text

nathanleclaire:
Thanks for @tiborvass and others pointing it out, I noticed that `docker-machine create --help --driver virtualbox` would not include the driver-specific flags in the output, and also gives slightly misleading output when invoked via `docker-machine help create`.  This fixes these issues to a reasonable degree.

This adds a test to detect the issue but the test depends on https://github.com/codegangsta/cli/pull/283 getting merged to be valid (otherwise `SkipFlagParsing` does not honor its commitment and actually parses the flags, including `--help`).  The good news is that once that is merged, we can remove our terrible forked codegangsta/cli patch and go back to Good Old Fashioned Vendoring(tm).

Also includes a fix for a slightly silly behavior where the RPC client driver used to get the flags would have a name corresponding to `c.Args.First()` when really it's OK to simply hardcode it as something (I've done `flag-lookup` here).  Just in case help text is invoked with `--debug`, it makes things slightly clearer.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,10 +1,9 @@
+[![Coverage](http://gocover.io/_badge/github.com/codegangsta/cli?0)](http://gocover.io/github.com/codegangsta/cli)
 [![Build Status](https://travis-ci.org/codegangsta/cli.png?branch=master)](https://travis-ci.org/codegangsta/cli)
+[![GoDoc](https://godoc.org/github.com/codegangsta/cli?status.svg)](https://godoc.org/github.com/codegangsta/cli)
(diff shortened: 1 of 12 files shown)


nathanleclaire:

nathanleclaire:
If codegangsta/cli does not merge that patch before the release I am considering updating our "vendored" /cli fork to it and using that.

nathanleclaire:
Updated the `cli/` directory to use that branch from upstream `codegangsta/cli`.

PR #2100, opened 2015-10-27
New option for virtualbox driver to set custom shared folder

u-foka:
I'm an OSX user, who has his $HOME outside /Users. It was moved to a secondary hard drive.

With this setup, I had difficulties using docker volumes, so I've added a new parameter to docker-machine's virtualbox driver that I can use to specify the location of my alternative /Users directory.

The format is --virtualbox-share-folder /path/to/folder:sharename. Automatic mounting currently only works with sharename == Users, for custom sharenames some work is needed on the boot2docker side[1].

[1]: https://github.com/boot2docker/boot2docker/blob/master/rootfs/rootfs/etc/rc.d/vbox#L36

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -137,8 +137,8 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 		},
 		mcnflag.StringFlag{
 			EnvVar: "VIRTUALBOX_SHARE_FOLDER",
-			Name: "virtualbox-share-folder",
-			Usage: "Mount the specified directory instead of the default home location. Format: dir:name",
+			Name:   "virtualbox-share-folder",
+			Usage:  "Mount the specified directory instead of the default home location. Format: dir:name",
 		},
 	}
 }
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -60,6 +60,7 @@ type Driver struct {
 	HostOnlyNicType     string
 	HostOnlyPromiscMode string
 	NoShare             bool
+	ShareFolder         string
 }
 
 // NewDriver creates a new VirtualBox driver with default settings.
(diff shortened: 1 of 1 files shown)


frankcarey:
+1 For this - We're using https://github.com/adlogix/docker-machine-nfs to convert it to use nfs instead of the slow virtualbox share driver.  However, if you have other nfs mounted directories in your ~/ folder, then you get a conflict warning, as shown here: https://github.com/adlogix/docker-machine-nfs/issues/16

u-foka:
boot2docker share mounting: https://github.com/boot2docker/boot2docker/pull/1088

jeanlaurent:
As of now, I mean before https://github.com/boot2docker/boot2docker/pull/1088 is merged, adding support for this is not *that* useful since you have to manually mount the shared folder each time the vm starts.

Typing the below commands works but is a bit tedious, and I doubt automating them at the end of the process is a good idea either, since it does no sustain a vm restart
```
sudo mkdir -p /myshare
sudo mount -t vboxsf -o uid="$(id -u docker)",gid="$(id -g docker)" "myshare" "/myshare"
```

maximd:
+1 for this!

Here we do not use `/home` as the base folder for users but another folder to distinguish NFS user's homes from local homes #

netman92:
+1

nmgeek:
Wow.  I found 4 or more issues all related to the inability to easily auto mount folders when the host is running Linux.  Are any of the pull requests going to be accepted and pushed out to a release?  Or is there already a fix that I missed?

segator:
This pull request is necesary, is going to be integrated?

nathanleclaire:
@u-foka if you'd like to rebase and followup i'd be happy to take a look at this again

u-foka:
Hy,

sorry, somehow I ruined my branch during the rebase.. sourcetree tricked me :/
Anyway the changes were checked in to a new branch and here is the new pull request #3798

PR #2103, opened 2015-10-28
Pipe error output from git line to /dev/null

nathanleclaire:
This will simply leave the tag empty if compiled independent of a source
repository.

cc @dmp42 


dmp42:
LGTM!

Issue #1897, opened 2015-09-23
libmachine/examples/vbox_create.go does not copy certs correctly


nathanleclaire:
If you do `go run vbox_create.go`, you notice that there are some certs in the directly alongside the Go file, and also the server certs / keys seem to be copied to `$STORE/machines` instead of the Machine file.  This is wrong and should be addressed in the `libmachine` code.

PR #2104, opened 2015-10-28
Ensure copy of certs to do not end up in pwd

clintonskitson:
When using the libMachine as a package, the certificates would
be copied in the pwd on each run.  A simple if statement was
added to ensure that copies only occur if the StorePath
was specified when the AuthOptions were created.

Signed-off-by: Clinton Kitson <clintonskitson@gmail.com>

### DIFF
--- a/libmachine/provision/utils.go
+++ b/libmachine/provision/utils.go
@@ -71,18 +71,20 @@ func ConfigureAuth(p Provisioner) error {
 		return err
 	}
 
-	log.Info("Copying certs to the local machine directory...")
+	if authOptions.StorePath != "" {
+		log.Info("Copying certs to the local machine directory...")
 
-	if err := mcnutils.CopyFile(authOptions.CaCertPath, filepath.Join(authOptions.StorePath, "ca.pem")); err != nil {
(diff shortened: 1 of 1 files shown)


clintonskitson:
Addressing https://github.com/docker/machine/issues/1897

nathanleclaire:
Ooo, nice!  I think the real issue at play is that `StorePath` doesn't get populated at all, though.  It's why I'm badgering @dgageot on it in https://github.com/docker/machine/pull/2068#issuecomment-151266084 -- `libmachine.Client` should know about `StorePath` "properly" as part of its configuration.

dgageot:
LGTM

nathanleclaire:
To be clear: NOT LGTM

It needs to be fixed that `StorePath` is not populated to fix this issue (the certs _do_ need to be copied that that Machine's store directory), not workaround by skipping the copying block.

nathanleclaire:
@clintonskitson Thanks for the contribution, but it's not quite the right answer -- feel free to take a whack at fixing the `StorePath` population issue and ask any questions if you have them.

PR #2106, opened 2015-10-28
Fix panic when using openstack driver

flavio:
The openstack driver was unusable because some cli flags were redifined.

That caused a runtime error like:
```
  create flag redefined: openstack-ssh-user
  panic: create flag redefined: openstack-ssh-user
```
Signed-off-by: Flavio Castelli <fcastelli@suse.com>


dgageot:
LGTM
Thank you @flavio!

PR #2107, opened 2015-10-28
FIX #2093 warn in case we think VT-X is not enabled.

### LINKED ISSUE #2093: "This computer doesn't have VT-X/AMD-v enabled" (Bootcamp running Windows)
Whilst this is technically a Bootcamp issue, the recent change where Docker Machine checks if virtualization is enabled fails when running on a Macbook Pro using Windows. The virtualization flag is set to false when you boot straight into Windows for some unknown reason. A google suggests that there's no way around this other than first booting into OSX, then doing a soft boot into Windows each time :-/

Docker Machine was working fine for me until I updated to RC3 where it's now unusable, as it just throws the following error:

"This computer doesn't have VT-X/AMD-v enabled."

Is there any way this check can be bypassed? Perhaps with an explicit command line argument?

dgageot:

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -19,8 +19,6 @@ import (
 	"strings"
 	"time"
 
-	"path"
-
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/mcnflag"
@@ -44,7 +42,7 @@ const (
 
 var (
 	ErrUnableToGenerateRandomIP = errors.New("unable to generate random IP")
-	ErrMustEnableVTX            = errors.New("This computer doesn't have VT-X/AMD-v enabled. Enabling it in the BIOS is mandatory.")
+	ErrMustEnableVTX            = errors.New("This computer doesn't have VT-X/AMD-v enabled. Enabling it in the BIOS is mandatory")
 	ErrNetworkAddrCidr          = errors.New("host-only cidr must be specified with a host address, not a network address")
 )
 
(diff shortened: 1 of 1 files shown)


dmp42:
LGTM

drivers/virtualbox/virtualbox.go
```diff
@@ -205,7 +205,10 @@ func cmdOutput(name string, args ...string) (string, error) {
 
 // IsVTXDisabledInTheVM checks if VT-X is disabled in the started vm.
 func (d *Driver) IsVTXDisabledInTheVM() (bool, error) {
-	file, err := os.Open(path.Join(d.ResolveStorePath(d.MachineName), "Logs", "VBox.log"))
+	logPath := path.Join(d.ResolveStorePath(d.MachineName), "Logs", "VBox.log")
```
nathanleclaire: We almost ubiquitously use `filepath.Join` instead of `path.Join` for cross-platform compatibility reasons.  I'm not sure it strictly applies here, but can you please update to use that instead?

drivers/virtualbox/virtualbox.go
```diff
@@ -205,7 +205,10 @@ func cmdOutput(name string, args ...string) (string, error) {
...
+	log.Debugf("Checking vm logs: %s", logPath)
+
+	file, err := os.Open(logPath)
```
nathanleclaire: Looks like you may be missing a `defer file.Close()` after this `err` check block?

drivers/virtualbox/virtualbox.go
```diff
@@ -229,7 +232,11 @@ func (d *Driver) Create() error {
 	}
 
 	if d.IsVTXDisabled() {
-		return ErrMustEnableVTX
+		// Let's log a warning to warn the user. When the vm is started, logs
+		// will be checked for an error anyway.
+		// We could fail right here but the method to check didn't prove being
+		// bulletproof.
+		log.Warn("This computer doesn't have VT-X/AMD-v enabled. Enabling it in the BIOS is mandatory.")
```
nathanleclaire: I agree that this being hidden away in the plugin output (behind `--debug`) is not ideal -- in the future we can talk about how to improve the situation.  Let me know if you have any ideas.

nathanleclaire:
A few nitpicks but looks pretty good.

drivers/virtualbox/virtualbox.go
dgageot: @nathanleclaire It is here on line 214. Isn't is the right way to do it?

dgageot:
@nathanleclaire Thanks for your feedback. Ready to merge.

drivers/virtualbox/virtualbox.go
nathanleclaire: My mistake -- I misread the diff.

nathanleclaire:
LGTM

PR #2115, opened 2015-10-29
Add CommandLine interface for commands

dgageot:
Introduce an interface for the commands' command line.

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/docker/machine/cli"
 	"github.com/docker/machine/libmachine/host"
 	"github.com/docker/machine/libmachine/persist"
 	"github.com/docker/machine/libmachine/state"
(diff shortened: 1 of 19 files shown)


nathanleclaire:
Hm, how about let's start with just transitioning the existing `cli.Context` struct into an interface and getting some tests for those?  I don't think that injecting hosts directly into the commands is the correct solution, since you might want to parse some information from the context (e.g. arguments are machine names) in order to get hosts.  The libmachine interface I mentioned would allow you to query the store to do so instead of having each command create their own.

dgageot:
@nathanleclaire I think your are right. I wanted to try in this direction but was not happy with how it turned out.

nathanleclaire:
@dgageot Are you going to follow up in another PR or just force push in this one?

dgageot:
@nathanleclaire I'm going to force push tomorrow

nathanleclaire:
Excellent thx

dgageot:
@nathanleclaire PTAL

nathanleclaire:
Nice!  Will work on getting this in after the release.

dmp42:
@dgageot @nathanleclaire this needs a rebase. Otherwise, good to go?

nathanleclaire:
It LGTM after a rebase

PR #2117, opened 2015-10-29
Carry on commits from #2033

dmp42:
A couple of small cleanup and enhancements that were dropped after the revert.

### DIFF
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -1,10 +1,12 @@
 package generic
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/docker/machine/libmachine/drivers"
@@ -20,13 +22,11 @@ type Driver struct {
 }
 
 const (
-	defaultSSHUser = "root"
-	defaultSSHPort = 22
 	defaultTimeout = 1 * time.Second
 )
 
 var (
-	defaultSSHKey = filepath.Join(mcnutils.GetHomeDir(), ".ssh", "id_rsa")
+	defaultSourceSSHKey = filepath.Join(mcnutils.GetHomeDir(), ".ssh", "id_rsa")
 )
 
 // GetCreateFlags registers the flags this driver adds to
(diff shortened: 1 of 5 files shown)


dgageot:
@dmp42 It looks like GetIP() is called only by the test. Am I right?

dgageot:
@dmp42 Also, you could use that commit to fix `GetMachineName()`'s comment :-)

dmp42:
@dgageot fixed the comment and removed duplicated GetIP definitions.
PTAL

dgageot:
LGTM

hairyhenderson:
Nice. :+1:

drivers/generic/generic.go
```diff
@@ -16,17 +18,15 @@ import (
 
 type Driver struct {
 	*drivers.BaseDriver
-	SSHKey string
```
nathanleclaire: Changing this field name _will_ be problematic without adding a `json:"SSHKey"` tag to indicate that in marshalling / demarshaling it should look for that field name.  Existing `config.json` for `generic` machines will break trying to look for that field.  My feeling on the matter is largely that the benefit of changing such a field name does not outweight the cost.

drivers/generic/generic.go
```diff
@@ -83,33 +81,30 @@ func (d *Driver) GetSSHUsername() string {
 func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.IPAddress = flags.String("generic-ip-address")
 	d.SSHUser = flags.String("generic-ssh-user")
-	d.SSHKey = flags.String("generic-ssh-key")
+	d.sourceSSHKey = flags.String("generic-ssh-key")
 	d.SSHPort = flags.Int("generic-ssh-port")
 
 	if d.IPAddress == "" {
-		return fmt.Errorf("generic driver requires the --generic-ip-address option")
+		return errors.New("generic driver requires the --generic-ip-address option")
```
nathanleclaire: If you're going to be changing these to `errors.New` anyway, looks like a good candidate for a custom `ErrRequiredFlagMissing` type.

dmp42: Agreed. That can happen over all the codebase as a follow-up PR.

dmp42:
Reverted visibility change on the SSHKey in the driver interface.

dmp42:
PTAL

drivers/generic/generic.go
dgageot: Good catch @nathanleclaire 
Shouldn't we have a test to protect us from this kind of issue?

nathanleclaire:
LGTM

Issue #1896, opened 2015-09-23
~/.ssh/config interferes with SSH calls


bfirsh:
threescoops on IRC reported an issue where machines were taking ages to boot up. It turned out to be _something_ in his ~/.ssh/config causing it to be slow - moving the file and running again fixed the issue.

Perhaps Machine should ignore ~/.ssh/config, or have its own location for configuring SSH?

nathanleclaire:
We set some options manually in the `ssh` command with `-o`, but I actually hadn't thought of simply specifying an alternate config file- looks like the `-F` option for `ssh` will do it.  I'm for it, although I'm not sure what the best file to use would be.  `/dev/null` seems reasonable, I'm not sure how that works on Windows though (FWIW I think we're already doing it with `KnownHostsFile`, perhaps in error).

nathanleclaire:
OTOH, looks like using the config file is actually the desired behavior for some users, e.g. https://github.com/docker/machine/issues/1665

It'd be really helpful to know which SSH configuration settings in particular were problematic.

bfirsh:
Perhaps it could use `~/.docker/machine/ssh/config` or something like that.

bazbremner:
@bfirsh Thanks for opening this bug. Apologies for not being able to give an details of the specific issue, I had quite a substantial ~/.ssh/config, although I'd guess that a ProxyCommand statement was the likely culprit, but haven't had the time to experiment.

jitakirin:
Could it be SSH multiplexing (#1572)?

In any case I also think ignoring user's SSH config is undesired, it is usually there for a reason (reducing auth prompts, multi-hop connections, etc.).  I think it would be better to work out which option is causing issues and adjusting code to work with it.  E.g. for SSH multiplexing the most likely culprit is (as mentioned in #1572) that SSH has a bug where it leaves stderr open in the daemonized process, which could possibly be worked around.

bazbremner:
@jitakirin I suspect so - I did have multiplexing in my previous SSH config. It probably makes sense to close this issue in favour of #1572.

dgageot:
Closing in favour of https://github.com/docker/machine/issues/1572 that itself is fixed

PR #2118, opened 2015-10-29
Add ProxyCommand=none to SSH command arguments

jijojv:
People may use wildcards and localhost should not go through a ProxyCommand

### DIFF
--- a/libmachine/ssh/client.go
+++ b/libmachine/ssh/client.go
@@ -57,6 +57,7 @@ var (
 		"-o", "ConnectTimeout=10", // timeout after 10 seconds
 		"-o", "ControlMaster=no", // disable ssh multiplexing
 		"-o", "ControlPath=no",
+		"-o", "ProxyCommand=none", // Disable default ProxyCommand in ssh/config
 	}
 	defaultClientType SSHClientType = External
 )


jeanlaurent:
Thx @jijojv 

This works well, although I believe most people won't have such an aggressive `ProxyCommand` in their ssh/config, this could help having docker-machine works out of the box in more configurations.

Can you :

1. change the comment to something more informative to what the command actually do. Something like `// Disable  default ProxyCommand in ssh/config`or something along thoses lines.
1. Squash your commits.

And we should be good to go.

jeanlaurent:
Thx @jijojv 

 LGTM

dgageot:
LGTM

jijojv:
@dgageot ProxyCommand is used in office settings where you have to use a jump/bastion host e.g. http://backdrift.org/transparent-proxy-with-ssh . It is invalid for ssh localhost

dgageot:
@jijojv so if I understand correctly, I change to NOT LGTM. For Virtualbox, it makes a lot of sense. For remote hosts (ec2, do, gce...) users wouldn't be able to use a jump host. Wdyt?

jijojv:
@dgageot Wasn't aware/didn't think of remote docker.  However these 3 settings also presume docker machine is localhost and can be problematic and hard to debug with LogLevel=quiet
        "-o", "ConnectTimeout=10", // timeout after 10 seconds
        "-o", "ControlMaster=no", // disable ssh multiplexing
        "-o", "ControlPath=no",

nathanleclaire:
> so if I understand correctly, I change to NOT LGTM. For Virtualbox, it makes a lot of sense. For remote hosts (ec2, do, gce...) users wouldn't be able to use a jump host. Wdyt?

This kind of need to change SSH configuration granularity is why I think we should move to having a default SSH configuration file specifically for Machine SSH (https://github.com/docker/machine/issues/1896#issuecomment-142925456) and take the command line options out of the code.  Then we'd support `--ssh-config-file` option or equivalent. Then user could also use their default SSH configuration or a custom one for Machine if desired.

PR #2128, opened 2015-10-30
FIX #2127 Detect that HW virtualization is not supported

### LINKED ISSUE #2127: create hangs forever if linux host doesn't support HW virtualization
I tried to create a virtualbox machine on a google compute engine ubuntu vm. This hangs forever because it doesn't realise that boot2docker can't boot.

Looking at the vm's log, I see: `WARNING! 64-bit guest type selected but the host CPU does NOT support HW virtualization` and looking at a snapshot of the "running" vm, I see "This kernel requires an x86-64 CPU, but only detected an i686 CPU."

We should improve `IsVTXDisabledInTheVM()` so that is also detects this error.

dgageot:

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -218,6 +218,9 @@ func (d *Driver) IsVTXDisabledInTheVM() (bool, error) {
 		if strings.Contains(scanner.Text(), "VT-x is disabled") {
 			return true, nil
 		}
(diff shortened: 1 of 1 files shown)


dgageot:
Yet another HW virtualization check!

dmp42:
LGTM

nathanleclaire:
LGTM... might be better off moving these messages to be constants in case they end up getting re-used

PR #2135, opened 2015-10-31
Fail gracefully when plugin binary implements outdated API

janeczku:
Since drivers will (mostly) be distributed separately it is too be expected that the docker-machine and  plugin binaries on users systems might eventually get out-of-sync in regards to the API version they implement.
E.g. after i upgraded to v0.5.0-rc4 today, docker-machine stopped working. Eventually i found out that this was caused by a third-party driver plugin that was built against v0.5.0-rc3 and didn't satisfy API methods introduced in rc4.

This patch makes sure  that docker-machine fails gracefully in case the driver binary API version is outdated.

### DIFF
--- a/libmachine/drivers/plugin/register_driver.go
+++ b/libmachine/drivers/plugin/register_driver.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/drivers/plugin/localbinary"
 	"github.com/docker/machine/libmachine/drivers/rpc"
+	"github.com/docker/machine/libmachine/version"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


dmp42:
cc @nathanleclaire

libmachine/drivers/plugin/register_driver.go
```diff
@@ -20,6 +21,7 @@ var (
 
 func RegisterDriver(d drivers.Driver) {
 	if os.Getenv(localbinary.PluginEnvKey) != localbinary.PluginEnvVal {
+		fmt.Fprintln(os.Stdout, fmt.Sprintf("API version %d", version.ApiVersion))
```
nathanleclaire: Seems reasonable, but why not just change the below line to a `Fprintf` instead and including it in the multi-line message?  I'd prefer this to go to STDERR as well, since invoking a plugin binary directly is intended to be an error.

janeczku: > (...) since invoking a plugin binary directly is intended to be an error.

Good point

nathanleclaire:
LGTM

The ApiVersion is unlikely to change often FYI, sorry about the RC hiccup.

janeczku:
@nathanleclaire Clear. On another note, what do you think about version control for third-party drivers as they will not necessarily follow the release cycle of docker-machine? Maybe you could add a `DriverVersion()` method to  libmachine's driver interface and then print the driver's version along with the driver flags, when doing `docker-machine create -d third-party-driver`?

nathanleclaire:
@janeczku It certainly is something important that we need to think about.  It's too late to add anything additional to the API without a game plan, but safe(ish) versioning of drivers seems like a good thing to worry about to me.

janeczku:
@nathanleclaire :+1:

Issue #530, opened 2015-02-11
Odd console output in Windows


anweiss:
Noticing odd console output behavior in Windows.

![2015-02-11_12-52-36](https://cloud.githubusercontent.com/assets/2326106/6152681/15492146-b1ed-11e4-8873-0249c6d85ec7.png)

Looks as if I'm missing a specific font or something. The status lines also don't appear until I scroll down and then scroll back up in the console.

Running Windows Server Technical Preview (Version 6.4.9841)

ehazlett:
Thanks.  This is a duplicate of #501.

Issue #501, opened 2015-02-07
docker-machine output kind of funky on Cygwin on Windows


nathanleclaire:
I like Cygwin for Windows and haven't tried with Msysgit but the logging output looks pretty funky on Cygwin on Windows.

![windows_machine_logs](https://cloud.githubusercontent.com/assets/1476820/6093694/c33d746e-aebe-11e4-9cdd-933333865cf1.jpg)

Just marking it down for future reference; it's one reason why I'm in favor of not doing daemon-style logs on the client side.

ehazlett:
I agree.  This looks better than default msysgit (it shows escape codes).

ehazlett:
We also need to get an "official" supported way to run on all platforms.  OS X / Linux are pretty straight forward.  For Windows, we need to at least document out the recommended way and tools needed.  Right now that is msysgit but I'm probably the last one to recommend.  I chose msysgit because it seemed to be the lowest barrier to entry from the user side.

I think once we fix client logging this won't be as big of an issue.

SvenDowideit:
main thing with msysgit, is that we've used it with boot2docker for the last year. One thing we wanted to try, is to use a zero-install version of the tools and plonk them into the b2d/machine dir to avoid conflicts with other installed tools.

getting everything so it works in those unix shells and cmd.exe and powershell is a bit of a horror though.

ehazlett:
yeah i agree.  i'm all for using something else if we can get them to play
nice.

nathanleclaire:
This is done

znmeb:
FWIW I've had pretty good luck on Windows boxes with MSys from the MSysGit installer when I allow the installer to add the *nix utilities to the Windows PATH. The terminal windows have the usual goodies - bash, readline, tar, grep, etc., and IIRC an ancient Perl as well. And you still have the Windows command-window copy and paste with the mouse.

If at all possible I'd avoid CygWin - it does bizarre things to the Windows permission bits to get a Linux-like environment. If you must use CygWin, make a separate Windows partition for it rather than letting it touch your "C" drive.

nathanleclaire:
@znmeb Yep, our official install path (at least for now) on Windows is to use msysgit- this issue was more about the formatting of the log statements being hard to read.

PR #2144, opened 2015-11-02
[WIP] logrus + bugsnag integration

dmp42:
This is a work in progress proposal to:

1. get rid of the custom logging code, and leverage logrus
   * specifically, libmachine no longer depends on a specific logger, but on a (very) simple logger interface 
   * the user of libmachine (machine in that case) can use any logger that implements that interface (logrus, for example)
   * logger specifics are thus into the main code instead of being inside libmachine: specifically, a custom reporter, and proper redirection to stdout and stderr

The end result is a cleaner / less opinionated choice of logger for libmachine, and more flexibility for the embedder.

2. Furthermore, since logrus support hooks, we can now have bugsnag and slack.

To enable/test them, you can pass `MACHINE_SLACK` and `MACHINE_BUGSNAG` env variables (ask in PV for valid tokens).

3. for now, the choice was made to remove the `WithFields` method

 * this method was rarely used
 * doing so removes the need to depend on the logrus library inside libmachine, and let the embedder stick with the standardlogger if he wants so

Let me know what you think overall.


nathanleclaire:
I know I've been a terrible offender of this as well, but we've got to stop dropping huge patch bombs like this.

That aside, I'm not a fan of this PR at all.  I have not seen any users ask for Slack integration or Bugsnag integration.  The additional vendored dependencies slow down compile time and increase the (already enormous) scope of what Machine is expected to support.  Not to mention the problems with sending logs that contain private keys over the network, wherever they end up.

Additionally, logrus was problematic for us in the past.  It seems designed for a different use case than Machine has with its logging.  Two of the main issues we had with `logrus` when we used it in the past were:

1. No granularity between STDOUT and STDERR, consequently it's much harder to mix the two.  I like having `log.Error` go to STDERR and `log.Info` go to STDOUT.  In `config` and `env` we're working around this problem by just saying all logs go to STDERR and then writing very deliberately to STDOUT when needed, but I think it exists in other places as well.  E.g., you want to run integration tests with `--debug` (STDERR -- to view the logs later if something went wrong) on, but still check that STDOUT spit out what you expected.
2. Output on Windows, or any non-traditional terminal, or file etc. looks horrible and is 10x more difficult to read.  E.g.: https://github.com/docker/machine/issues/501 and https://github.com/docker/machine/issues/530

nathanleclaire:
BTW, the user of `libmachine` can already create and set custom interfaces for logging if they want to.  They just call `libmachine`'s `log.SetOutWriter` or `log.SetErrWriter`.

nathanleclaire:
I'm a fan of removing `WithFields` though.

dmp42:
Thanks @nathanleclaire 

> I know I've been a terrible offender of this as well, but we've got to stop dropping huge patch bombs like this.

This is a fairly small PR  - just because we vendor code doesn't make it a patch bomb.

> I have not seen any users ask for Slack integration or Bugsnag integration.

That's not a user feature.

This is about having proper debugging information, and easier ways to help users.

Slack here is purely for the demo, to show that it's possible to do, very simply, and that might be considered handy if we want to easily debug someone. I'm happy to remove it right now if you don't want to test it.

Having bugsnag support on the other hand is critical.
If people opt-in, we will get proper errors report and view into what goes wrong.
Currently, what we have is toolbox/kitematic pushing data into bugsnag on behalf of machine, which is silly. It thought we discussed it earlier.

> increase the (already enormous) scope of what Machine is expected to support.

There is nothing more to "support", since this is not a feature - it's a debug hook.

> No granularity between STDOUT and STDERR, consequently it's much harder to mix the two. I like having log.Error go to STDERR and log.Info go to STDOUT. In config and env we're working around this problem by just saying all logs go to STDERR and then writing very deliberately to STDOUT when needed, but I think it exists in other places as well. E.g., you want to run integration tests with --debug (STDERR -- to view the logs later if something went wrong) on, but still check that STDOUT spit out what you expected.

It's quite possible to mix STDOUT and STDERR with logrus, and there is granularity (this is what I did here, we have the same output logic as with the previous logging mechanism).

> Output on Windows, or any non-traditional terminal, or file etc. looks horrible and is 10x more difficult to read. E.g.: #501 and #530

I'm using the same formatter as you did in the previous logger, so, unless I missed something there is no reason to have specifically funky output 

> BTW, the user of libmachine can already create and set custom interfaces for logging if they want to. They just call libmachine's log.SetOutWriter or log.SetErrWriter.

Can I use logrus with the current libmachine?

nathanleclaire:
> This is a fairly small PR - just because we vendor code doesn't make it a patch bomb.

Sure it does - that's 14,000 additional lines of other peoples' code that we're now assuming responsibility for - and, while there certainly is a lot of vendored code, adding two new flags / environment variables, two new logging "drivers", re-adding logrus and replacing the current standard logger, removing `WithFields` and so on is definitely _not_ a fairly small PR.  It is a huge PR which affects almost every part of the Machine codebase, because almost every part calls `log`.

> That's not a user feature.
> This is about having proper debugging information, and easier ways to help users.

I'm confused by this bit, care to elaborate?  Sounds like a user feature is exactly what you're describing.

> Slack here is purely for the demo, to show that it's possible to do, very simply, and that might be considered handy if we want to easily debug someone. I'm happy to remove it right now if you don't want to test it.

Yeah, I do not ever want to merge in anything just for demo -- it _will_ get used and people _will_ expect us to maintain it.

> Having bugsnag support on the other hand is critical.
> If people opt-in, we will get proper errors report and view into what goes wrong.
> Currently, what we have is toolbox/kitematic pushing data into bugsnag on behalf > of machine, which is silly. It thought we discussed it earlier.

Yeah, I guessed that this was the motivation for this PR, and I'm comfortable with that -- but I'd prefer to get there in the smallest diff possible.  That might be through calling the BugSnag directly in batches on `Debug` with something like https://github.com/toggl/bugsnag, or through embedding logrus with BugSnag enabled in the `StandardLogger`.  +15,000 SLOC is really really heavy just to get logs sent to BugSnag.

> There is nothing more to "support", since this is not a feature - it's a debug hook.

Olivier, it's a feature.  It will need development effort and testing and time and debugging invested.  It will fail on users in the wild.  It expands what Machine is capable of, and in a pretty non-trivial way at that.

> It's quite possible to mix STDOUT and STDERR with logrus, and there is granularity (this is what I did here, we have the same output logic as with the previous logging mechanism).

It wasn't like that last time we tried it (or at least we didn't notice how to configure it as such), but I'll take a look.

> Can I use logrus with the current libmachine?

I think so -- it should implement the `interface` defined in the `log` package quite comfortably.  I think something like `log.SetErrWriter(logrus.New()); log.SetOutWriter(logrus.New())` will do the trick.

dmp42:
That was discussed offline.

Nevertheless, to address the specific point of the number of LOC.

 * `bootstrap.css` only is 5k lines.
 * other binary files account for several thousand LOC as well.

Actual code is in the range of *hundreds* of LOC.

nathanleclaire:
Yeah vendored deps really mess everything up in that regard.  To put the major points of our offline discussion here:

- Switching to logrus seems reasonable for a variety of reasons and I'm in favor as long as we test heavily on Windows shells to ensure the output looks the same
- Slack feature will be removed as it is not high priority
- Ensuring output is properly redirected to STDOUT/STDERR should also be tested heavily in this move

libmachine/drivers/plugin/localbinary/plugin.go
```diff
@@ -188,17 +189,20 @@ func (lbp *Plugin) execServer() error {
...
-		case err := <-stdErrCh:
-			log.Debug(fmt.Sprintf(pluginErrPrefix, lbp.MachineName), err)
+			data := out
+			if strings.Index(out, log.Prefix) == 0 {
+				cut := strings.Index(out, ":")
+				currentLogLevel = out[len(log.Prefix):cut]
+				data = out[cut+1:]
+			}
+			in := []reflect.Value{reflect.ValueOf(fmt.Sprintf(pluginOutPrefix, lbp.MachineName)), reflect.ValueOf(currentLogLevel), reflect.ValueOf(data)}
+			reflect.ValueOf(log.GetLogger()).MethodByName(currentLogLevel).Call(in)
```
nathanleclaire: What's going on here?

libmachine/drivers/plugin/register_driver.go
```diff
@@ -29,8 +28,6 @@ Please use this plugin through the main 'docker-machine' binary.
 		os.Exit(1)
 	}
 
-	libmachine.SetDebug(true)
```
nathanleclaire: Why remove this line?

libmachine/log/log.go
```diff
@@ -1,124 +1,76 @@
...
-
-func SetErrWriter(w io.Writer) {
-	l.ErrWriter = w
+	return logger
 }
 
+// Debug logs a message at level Debug
 func Debug(args ...interface{}) {
-	l.Debug(args...)
+	GetLogger().Debug(args...)
```
nathanleclaire: Why the whole `GetLogger` song and dance?  Why not just initialize `logger` in a `var` block in the top of the module? e.g.

```go
var (
    logger = newDefaultLogger()
)
```

log/hooks.go
```diff
@@ -0,0 +1,72 @@
...
+	bs.Configure(bs.Configuration{
+		APIKey: apiKey,
+		// XXX we need to abuse bugsnag metrics to get the OS/ARCH information as a usable filter
+		// Can do that with either "stage" or "hostname"
+		ReleaseStage: stage,
+		//		Hostname: stage,
+		ProjectPackages: []string{"github.com/docker/machine/[^v]*"},
+		AppVersion:      version.Version + " (" + version.GitCommit + ")",
+		Logger:          new(void),
+		Synchronous:     true,
```
nathanleclaire: I don't know much about the motivation for this, but maybe it should be async?  I certainly wouldn't like to block the main machine thread any time we log.

log/hooks.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// Add runtime informations
+	hook.Add("app", "compiler", fmt.Sprintf("%s (%s)", runtime.Compiler, runtime.Version()))
+
+	hook.Add("device", "os", runtime.GOOS)
+	hook.Add("device", "arch", runtime.GOARCH)
+
+	// Add environ
+	for _, v := range os.Environ() {
```
nathanleclaire: This could potentially leak secrets, e.g. `DIGITALOCEAN_ACCESS_TOKEN`.

log/hooks.go
```diff
@@ -0,0 +1,72 @@
...
+		idx := strings.Index(v, "=")
+		k := v[:idx]
+		v = v[idx+1:]
+		hook.Add("environ", k, v)
+	}
+
+	// Add network information
+	t, _ := net.Interfaces()
```
nathanleclaire: Maybe at least check the `err` to report that we weren't able to successfully run the call as a reason why the network information is unavailable?

log/log.go
```diff
@@ -0,0 +1,73 @@
+package log
+
+import (
+	"github.com/Sirupsen/logrus"
+	libmachinelog "github.com/docker/machine/libmachine/log"
```
nathanleclaire: I'm pretty confused by this import.  Maybe I'm missing something, but isn't the module this file is in `github.com/docker/machine/libmachine/log` already?

nathanleclaire:
Hey @jeanlaurent -- please take a look at this.  We won't be able to get it in time for the 0.5.2 release (which is fine), but it would be great to start having a member of the team rebasing, writing tests for, testing the actual Bugsnag bit and generally nurturing this PR actively.

log/hooks.go
dmp42: You need it to be synch if you want to send the report before the process exit.

libmachine/drivers/plugin/localbinary/plugin.go
dmp42: Reinjecting driver logs into the main process logs at the right level. Dynamic method calls need reflection, then.

libmachine/drivers/plugin/register_driver.go
dmp42: Drivers no longer need to set debug - all logs are sent back to the main process, and are forwarded (or not) depending on the overall log level.

libmachine/log/log.go
dmp42: I guess both work. The accessor has the benefit of lazy initialization (avoid the unnecessary creation of the default logger when it's explicitly set).
Either way is ok.

log/hooks.go
dmp42: If there is no network, I guess we will not be able to send anything?

log/log.go
dmp42: `libmachine/log` and `log` are two different packages.
In this proposal, libmachine doesn't preclude anything specific about logging (doesn't even depend on logrus), and drivers depend only on that.
On the other hand machine depends on logrus, and defines the logging strategy it uses (including bugsnag, etc).

dmp42:
@nathanleclaire I added a couple of comments.
Let me know if that clarifies.
Either way, it's a proposal in its current state, that can certainly be changed to reach maturity.

libmachine/drivers/plugin/localbinary/plugin.go
nathanleclaire: All (or at least most, there's a small bug at https://github.com/docker/machine/issues/2217) STDOUT and STDERR from the plugins gets passed to `log.Debug` in the main process though: https://github.com/docker/machine/blob/master/libmachine/drivers/plugin/localbinary/plugin.go#L195-L198 

Won't that cover logging from the drivers as long as the Bugsnag/logrus works in the main process?

libmachine/drivers/plugin/register_driver.go
nathanleclaire: Like noted above, if we simply set `log.IsDebug` to true, the main process is streaming the STDOUT and STDERR from the process and `log.Debug`ing them anyway.

libmachine/log/log.go
nathanleclaire: Cool.  My preference is definitely to initialize it at create time.  That's the way it's done in the standard library: https://golang.org/src/log/log.go#L71

log/hooks.go
nathanleclaire: Hm, OK. I'll have to look more into the mechanics of how the bugsnag hook looks.  If we can batch up the logs and only make a network call to send them during the program teardown procedure, that would be great, especially for users with slow network connections.

log/hooks.go
nathanleclaire: Yes, and the program should continue working as normal in that case -- but what I was referring to is if the `net.Interfaces` call itself failed.  It could return `err` that we shouldn't swallow.  e.g. a variety of `syscall` errors in the function here https://golang.org/src/net/interface_bsd.go#L18

log/log.go
nathanleclaire: Ah, OK.  I got confused by the pathing here.  I'll have to take a closer look at the arch.

nathanleclaire:
Yup @dmp42 NP just getting the wheels turning on this a bit.

jeanlaurent:
This is 3-4 things in one.
I'll split
* logrus
* bugsnag
* server/client log communication
* slack

dgageot:
@dmp42 OK to close this PR?

dmp42:
Sure :)

PR #2163, opened 2015-11-04
Move to vendor

dmp42:
Despite what it look like, this is not a patch bomb.
It just moves to vendor and away from Godeps/_workspace/src


dgageot:
LGTM

Issue #2138, opened 2015-11-02
proxy setting options for containers created by --swarm options.


kunalkushwaha:
[An environment behind proxy]

To setup docker swarm host, docker-machine provides options ``--swarm``.
While ``docker-machine create``,  using ``--engine-env HTTP_PROXY==<proxy>``, proxy for docker-machine can be configured.

But, if default token service from hub.docker.com is used as discovery service, It fails, since the ``proxy`` settings are not availble to containers, and it is not able to communicate to hub.docker.com.

```
kunal@kunal-PC MINGW64 ~
$ docker-machine  create -d virtualbox --engine-env HTTP_PROXY="http://example.com:8080" --engine-env HTTPS_PROXY="http://example.com:8080" --swarm --swarm-master --swarm-discovery token://d921ed483fe2d9f8116b48c24404af4f swarm-master
Running pre-create checks...
Creating machine...
Waiting for machine to be running, this may take a few minutes...
Machine is running, waiting for SSH to be available...
Detecting operating system of created instance...
Provisioning created instance...
Copying certs to the local machine directory...
Copying certs to the remote machine...
Setting Docker configuration on the remote daemon...
Configuring swarm...
To see how to connect Docker to this machine, run: C:\Program Files\Docker Toolbox\docker-machine.exe env swarm-master

kunal@kunal-PC MINGW64 ~
$ docker-machine ls
NAME             ACTIVE   DRIVER       STATE     URL                         SWARM

swarm-master     -        virtualbox   Running   tcp://192.168.99.105:2376   swarm-master (master)
test             -        virtualbox   Running   tcp://192.168.99.104:2376

kunal@kunal-PC MINGW64 ~
$ docker-machine ssh swarm-master
                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~~ ~ /  ===- ~~~
           \______ o           __/
             \    \         __/
              \____\_______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|

[...]

PR #2166, opened 2015-11-04
Environment variables passed to provision machine, will be passed to …

kunalkushwaha:
…containers for swarm also. Bug fix  https://github.com/docker/machine/issues/2138

### DIFF
--- a/libmachine/provision/arch.go
+++ b/libmachine/provision/arch.go
@@ -108,6 +108,7 @@ func (provisioner *ArchProvisioner) Provision(swarmOptions swarm.SwarmOptions, a
 	provisioner.SwarmOptions = swarmOptions
 	provisioner.AuthOptions = authOptions
 	provisioner.EngineOptions = engineOptions
+	swarmOptions.Env = engineOptions.Env
 
 	if provisioner.EngineOptions.StorageDriver == "" {
 		provisioner.EngineOptions.StorageDriver = "overlay"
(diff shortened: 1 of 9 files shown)


dgageot:
LGTM. Ping @docker/machine-maintainers

dmp42:
@nathanleclaire wdyt?

nathanleclaire:
Nice idea, have a few nitpicks in the implementation I want before merging though.  I'll post on the code.

libmachine/provision/arch.go
```diff
@@ -154,7 +154,7 @@ func (provisioner *ArchProvisioner) Provision(swarmOptions swarm.SwarmOptions, a
 	}
 
 	log.Debug("Configuring swarm")
-	if err := configureSwarm(provisioner, swarmOptions, provisioner.AuthOptions); err != nil {
+	if err := configureSwarm(provisioner, provisioner.EngineOptions.Env, swarmOptions, provisioner.AuthOptions); err != nil {
```
nathanleclaire: Instead of adding an additional argument to this function, please just add a `Env` value to the `SwarmOptions` struct and it will get passed through automatically.

libmachine/provision/configure_swarm.go
```diff
@@ -70,6 +71,7 @@ func configureSwarm(p Provisioner, swarmOptions swarm.SwarmOptions, authOptions
 
 	swarmCmdContext := SwarmCommandContext{
 		ContainerName: "",
+		Env:           Env,
```
nathanleclaire: As mentioned above, if you add the `Env` to `SwarmOptions` this can become `Env: swarmOptions.Env` instead and we don't have to add an additional argument to the function.

kunalkushwaha:
Made changes as per comments

nathanleclaire:
LGTM

PR #2168, opened 2015-11-04
FIX #2167 b2d download timeout during upgrade

### LINKED ISSUE #2167: b2d download timeout during upgrade
Having tested `docker-upgrade` process from a slow network, there are a couple of issues:

+ The download timeout is way too low. I couldn't upgrade at all because of that.
+ When the download is in progress, there is no visual feedback so the user might think it just hangs.
+ We should download b2d before we stop the machine otherwise it's really frustrating to restart the vm before retrying the upgrade.
+ Upgrade asks me to start the machine and the first message I see is that the machine is being stopped.. That adds to the frustration. There should be a log explaining what is done that need the machine to be running.

dgageot:
+ Download boot2docker before we stop the machine.
+ Use default(longer) timeout
+ Show b2d download progress
+ Fix typo
+ Add log before stop

### DIFF
--- a/libmachine/mcnutils/b2d.go
+++ b/libmachine/mcnutils/b2d.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"time"
 
 	"github.com/docker/machine/libmachine/log"
 )
@@ -20,12 +19,8 @@ var (
 	GithubAPIToken string
 )
 
-const (
-	timeout = time.Second * 5
-)
-
 func defaultTimeout(network, addr string) (net.Conn, error) {
-	return net.DialTimeout(network, addr, timeout)
+	return net.Dial(network, addr)
 }
 
 func getClient() *http.Client {
@@ -139,12 +134,14 @@ func removeFileIfExists(name string) error {
 // DownloadISO downloads boot2docker ISO image for the given tag and save it at dest.
 func (b *B2dUtils) DownloadISO(dir, file, isoURL string) error {
 	u, err := url.Parse(isoURL)
+
 	var src io.ReadCloser
 	if u.Scheme == "file" || u.Scheme == "" {
 		s, err := os.Open(u.Path)
 		if err != nil {
 			return err
 		}
+
 		src = s
 	} else {
 		client := getClient()
@@ -152,7 +149,12 @@ func (b *B2dUtils) DownloadISO(dir, file, isoURL string) error {
 		if err != nil {
 			return err
 		}
-		src = s.Body
+
+		src = &ReaderWithProgress{
+			ReadCloser:     s.Body,
+			out:            os.Stdout,
+			expectedLength: s.ContentLength,
+		}
 	}
 
 	defer src.Close()
@@ -170,7 +172,6 @@ func (b *B2dUtils) DownloadISO(dir, file, isoURL string) error {
 	}()
 
 	if _, err := io.Copy(f, src); err != nil {
-		// TODO: display download progress?
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


dgageot:

libmachine/mcnutils/b2d.go
```diff
@@ -20,12 +19,8 @@ var (
 	GithubApiToken string
 )
 
-const (
-	timeout = time.Second * 5
-)
-
 func defaultTimeout(network, addr string) (net.Conn, error) {
-	return net.DialTimeout(network, addr, timeout)
+	return net.Dial(network, addr)
```
nathanleclaire: Why eliminate the configured timeout?

libmachine/provision/boot2docker.go
```diff
@@ -90,6 +78,21 @@ func (provisioner *Boot2DockerProvisioner) upgradeIso() error {
...
+		return err
+	}
+
+	if err := mcnutils.WaitFor(drivers.MachineInState(provisioner.Driver, state.Stopped)); err != nil {
+		return err
+	}
+
+	machineName := provisioner.GetDriver().GetMachineName()
+
+	log.Infof("Upgrading machine %s...", machineName)
```
nathanleclaire: I usually like to use `%q` for this purpose or else the name tends to blend in with the rest of the text

nathanleclaire:
Cool.  Couple questions:

1. How about a unit test for `ReaderWithProgress`?  The embedded `ReadCloser` is already an interface and should be fairly straightforward to mock.
2. Have you tested this on every Windows shell we support? (git bash, powershell, cmd.exe)

libmachine/mcnutils/b2d.go
```diff
@@ -194,6 +191,39 @@ func (b *B2dUtils) DownloadISO(dir, file, isoUrl string) error {
 	return nil
 }
 
+type ReaderWithProgress struct {
+	io.ReadCloser
+	bytesTransferred  int64
+	expectedLength    int64
+	nexPercentToPrint int64
```
nathanleclaire: is it supposed to be `nextPercentToPrint`?

libmachine/mcnutils/b2d.go
dmp42: Out of curiosity - what was the reasoning for having the timeout initially? Do we have a fallback mechanism in case it wouldn't respond fast enough?

nathanleclaire: Introduced in `d1ee81453ca393acce39e16ece3474fd3fb234cc` by @ehazlett, I think to increase resilience in case of network failure?  The fallback is the image cache AFAIK.

libmachine/provision/boot2docker.go
dgageot: I only move existing code around. I'll fix that!

libmachine/mcnutils/b2d.go
dgageot: right...

libmachine/mcnutils/b2d.go
dgageot: We can't guess the duration of the download. I'm currently on a slow network and it takes more than 5 minutes to download.

dgageot:
Feedback taken, tests added. @docker/machine-maintainers PTAL

dmp42:
LGTM - this will solve problems in a bunch of cases.

@nathanleclaire wdyt?

libmachine/mcnutils/b2d.go
nathanleclaire: I think it's just a timeout for the `Dial` though, right?  i.e. the initial call to contact the server.

nathanleclaire:
Test looking good -- will review this and most likely merge today.

libmachine/mcnutils/b2d.go
dgageot: @nathanleclaire It should be yes. I don't understand why it fails here even with a >30s timeout.

dgageot:
@nathanleclaire If you fear, the timeout should just be increased. Please LGTM for the rest and I'll run tests tomorrow with different timeout values.

nathanleclaire:
Tested and this LGTM

Issue #2154, opened 2015-11-03
Error getting state for host using Docker Machine


arun-gupta:
Installed Docker Machine 0.5.0 and tried listing all the existing machines using `docker-machine ls`. The following error message is thrown:

```
~ > docker-machine ls
error getting state for host lab: /usr/local/bin/VBoxManage showvminfo lab --machinereadable failed:
VBoxManage: error: The object is not ready
VBoxManage: error: Details: code E_ACCESSDENIED (0x80070005), callee nsISupports
VBoxManage: error: Context: "COMGETTER(SharedFolders)(ComSafeArrayAsOutParam(folders))" at line 2186 of file VBoxManageInfo.cpp

NAME        ACTIVE   DRIVER       STATE     URL                         SWARM
cluster     -        virtualbox   Running   tcp://192.168.99.108:2376   
couchbase   -        virtualbox   Running   tcp://192.168.99.101:2376   
javaone     -        virtualbox   Running   tcp://192.168.99.109:2376   
lab         -        virtualbox   Error                                 
lab2        *        virtualbox   Running   tcp://192.168.99.112:2376   
```

nathanleclaire:
Weird. Does it work on subsequent invocations or same error message?

arun-gupta:
`docker-machine ls` gave that error consistently.

Removed and recreated the machine and now stuck with https://github.com/docker/machine/issues/2147#issuecomment-153519614

bfirsh:
I got the error when upgrading an existing install (a 1.9 RC).

```
$ docker-machine upgrade default                                                          !86
Stopping machine to do the upgrade...
/usr/local/bin/VBoxManage showvminfo default --machinereadable failed:
VBoxManage: error: The object is not ready
VBoxManage: error: Details: code E_ACCESSDENIED (0x80070005), component SessionMachine, interface IMachine, callee nsISupports
VBoxManage: error: Context: "COMGETTER(BandwidthControl)(bwCtrl.asOutParam())" at line 2243 of file VBoxManageInfo.cpp
```

Starting the machine and then upgrading again works fine - I can't reproduce now.

dgageot:
Hi @arun-gupta, thanks a lot for the feedback. I faced the same problem today and I know how to fix it. `docker-machine` ls contacts the hosts in separate go routines. But vbox isn't really happy about that. I'm going to shoot a PR.

PR #2172, opened 2015-11-04
FIX #2154 query virtualbox serially

dgageot:
Signed-off-by: David Gageot <david@gageot.net>

PR #2179, opened 2015-11-04
Add mutex and test for safe concurrent VBoxManage access

nathanleclaire:
Closes https://github.com/docker/machine/issues/2154

### DIFF
--- a/drivers/virtualbox/vbm.go
+++ b/drivers/virtualbox/vbm.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/docker/machine/libmachine/log"
 )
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Like I said I think it may allow us to also remove the serial bit in `runActionForeachMachine` as well.  We'll see.

nathanleclaire:

dmp42:
This is tripping a race condition apparently.

nathanleclaire:
Yeah, I noticed that.  I'm actually not sure how much good the test is doing in this case (since essentially it's elaborate test of mutex itself), so I might just take it out.

nathanleclaire:
Alright, how about this:  I've turned off running this test on `go test -race` using build tag exclusion.  I want to still have the test run in CI, however, so I also added a line to `test-long` target in Makefile to also run `go test` _without_ -race.  The overhead is not really substantial as far as I can tell (a few extra seconds).  This seems to be a reasonable compromise to me.

nathanleclaire:
To be clear, it's intentional that the test races.

dmp42:
> however, so I also added a line to test-long target in Makefile to also run go test without -race

Ah - I don't think you need this actually - `test-short` will already run your test (unless your test is specifically meant to run only on `long`?).

nathanleclaire:
Right, I see.  I'll take the Makefile change out then.  Should be good to go since you have to intentionally mark tests which will get skipped with short, which I haven't done here.

dgageot:
@nathanleclaire I've tested the code with #2154 and it does not solve the issue.

Also, adding a mutex is a way to not serialize actions in `runActionForeachMachine()`. Could you please simplify this function?

Regarding the test, it's well known that using `time.sleep` leads to unreliable and slow tests. Could we find another implementation for that test?

Also, FYI, reopening https://github.com/docker/machine/pull/2172 which actually fixes the #2154 issue. When the mutex is fully functional, it'll be a pleasure to simplify the code for `runActionForeachMachine()` and `ls.go`

dgageot:
@nathanleclaire shall we close this PR? Clearly, the mutex has no effect since each `virtualbox` driver is running in its own process.

nathanleclaire:
@dgageot Yep

PR #2181, opened 2015-11-05
Lint, second pass

dmp42:
@nathanleclaire this one is more involved than the first.

I would like to sit with you and review it four hands to make sure this is gold. 

Note that this also enable `make lint` by default on test and on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,4 @@ sudo: required
 dist: trusty
 language: bash
 services: docker
-script: USE_CONTAINER=true make dco fmt vet test-short test-long coverage-send
+script: USE_CONTAINER=true make dco fmt lint vet test-short test-long coverage-send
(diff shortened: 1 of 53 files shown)


dmp42:
Build is successful locally. Travis is perplexing.

dgageot:
@dmp42 Can you please rebase?

dmp42:
Yeah my bad.
And I added no-cache to try to force travis to refresh, but that shouldn't be required.

nathanleclaire:
It looks pretty good.  I see nothing which seems unreasonable.  LGTM

dmp42:
Thanks @nathanleclaire 

I believe doing this sooner than later is important - we may trigger a few bugs, but better now that when drivers are being outsourced.

nathanleclaire:
Yes my reasoning is that I gave it a sweep and everything seems safe, if we run into issues it will likely be before 0.5.1 and if any come up after it won't be too long to get a patch out to fix them.

PR #2185, opened 2015-11-05
FIX #2171 issue with amazonec2-private-address-only

### LINKED ISSUE #2171: EC2 issue with version .5 amazonec2-private-address-only
I am having an issue with the new version of docker machine.

Running the following command

docker-machine --debug create --driver amazonec2 --amazonec2-subnet-id subnet-24abec0f --amazonec2-zone b --amazonec2-private-address-only swarm-demo2

the result is 
Error: No machine name specified

If I remove amazonec2-private-address-only option, it works, but I can't connect to it since the machine isn't accessible using the public ip.

Running in debug shows this

(swarm-demo2) Calling RpcServerDriver.GetSSHPort
(swarm-demo2) Calling RpcServerDriver.GetSSHKeyPath
(swarm-demo2) Calling RpcServerDriver.GetSSHUsername
About to run SSH command:
exit 0
SSH cmd err, output: exit status 255: 
Error getting ssh command 'exit 0' : Something went wrong running an SSH command!
command : exit 0
err     : exit status 255
output  :

dgageot:

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -150,7 +150,7 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 			Usage: "AWS spot instance bid price (in dollar)",
 			Value: defaultSpotPrice,
 		},
-		mcnflag.StringFlag{
+		mcnflag.BoolFlag{
 			Name:  "amazonec2-private-address-only",
 			Usage: "Only use a private IP address",
 		},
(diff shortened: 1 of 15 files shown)


dgageot:

nathanleclaire:
There's some tests and stuff for other drivers in here.  Why not split them out to a separate PR?

dgageot:
@nathan I wrote a test to show the pb, then fixed, then applied the test for all the drivers to make sure we didn't have other bugs. So that's really a whole.

nathanleclaire:
OK -- in this case I don't think it's particularly problematic but I'm trying to move towards smaller PRs in general, so splitting this into 2 would have been nice.

Could you please change the commit message to something like `Fix faulty flag and add tests for GetConfigFromFlags` since it's more general than the original message?  After that, LGTM

nathanleclaire:
Cool thx

dmp42:
LGTM

PR #2187, opened 2015-11-05
[WIP] Inject the Store into commands

dgageot:
Testing the commands is still complicated. So I carried on with refactoring the code:

+ Inject the Store into command -> easy to mock.
+ Cleanup createInner and createOuter -> it's a mess to read and test.
+ Replace most helper methods on store be calls to store directly -> Easier to mock
+ Improve error messages by removing some long error chains that say the same thing 3x

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -15,12 +15,10 @@ var (
 	errTooManyArguments = errors.New("Error: Too many arguments given")
 )
 
-func cmdActive(c CommandLine) error {
+func cmdActive(c CommandLine, store persist.Store) error {
 	if len(c.Args()) > 0 {
 		return errTooManyArguments
 	}
-
-	store := getStore(c)
 
 	host, err := getActiveHost(store)
 	if err != nil {
@@ -35,7 +33,7 @@ func cmdActive(c CommandLine) error {
 }
 
 func getActiveHost(store persist.Store) (*host.Host, error) {
-	hosts, err := listHosts(store)
+	hosts, err := store.List()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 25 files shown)


dgageot:
Hi @docker/machine-maintainers Here's some more refactoring to make code easier to test and also easier to read. It's still WIP because a couple of things are broken.
But I could really use your feedback.

dmp42:
I like it. Expect to rebase this a couple of times though.

@nathanleclaire wdyt?

nathanleclaire:
Let's change just one piece at a time (i.e. let's not go modifying the `cmdOuter` and `cmdInner` stuff at the same time as working on injecting a libmachine client).

The `cmdOuter` and `cmdInner` stuff, while I know it _looks_ ugly and horrible, is that way for very specific reasons.  There has been a lot of effort put into making codegangsta/cli work with the types of dynamic lookups we're doing with pluggable drivers and I doubt we can take out large chunks safely overnight.  

It's split up into two functions for a very specific reason, for instance:  If `SkipFlagParsing` is not set to `true` for the outermost `create` function, codegangsta/cli will complain about flags it doesn't recognize (because it can't have flags provided by drivers compiled directly in, it doesn't know them ahead of time).  But the flag parsing actually _does_ need to be done for the second time around, in order to validate flags passed in and to parse them out into a form queryable by e.g. `c.Bool("blah")`.

Granted, I'm not averse to cleaning it up if possible, but we need to be really careful about not breaking the fragile logic around `create` in its current form.  And it should most likely be done in a few separate PRs.  This one purports to only deal with injecting store / `libmachine` client into the commands.

dgageot:
@nathanleclaire Sure. This PR is more of a spike to discover what can be done. That said, `cmdOuter` and `cmdInner` is clearly a mess. There has to be a better solution.

I'll keep you posted with a PR dedicated on `Store` injection.

PR #2191, opened 2015-11-05
Fix the local development environment instructions

jviide:
Omit the docker-machine/ path prefix from the last cd command. Having followed the instructions the user should already be in the docker-machine directory.


dmp42:
Agreed. LGTM

Issue #2196, opened 2015-11-05
Add a vmwarefusion driver flag to disable the mounting of the /Users directory


jviide:
It would be great to be able to disable the automatic mounting of the /Users directory inside the VMWare Fusion VM.

Maybe this could be done along the lines of pull request #1622 that adds the flag --virtualbox-no-share. The Parallels driver also has a similar --parallels-no-share (see https://github.com/Parallels/docker-machine-parallels/pull/1). Therefore it would make sense to name the flag --vmwarefusion-no-share.

jviide:
Opened pull request #2197 to have a go at implementing this.

dgageot:
Closed with https://github.com/docker/machine/pull/2197

PR #2197, opened 2015-11-06
Add a --vmwarefusion-no-share flag to disable the mounting of the /Users directory

jviide:
Implement the functionality mentioned in issue #2196: Add a vmwarefusion driver flag to disable the mounting of the /Users directory. The code is derived from the corresponding --virtualbox-no-share implementation.

I do have a couple of lingering questions though:

 * Should the defaultNoShare constant be defined although it is not used anywhere? The virtualbox driver implementation also has such constant defined.

 * Should shared folders get enabled at all when the sharing is disabled? Currently the ```vmrun("-gu", B2DUser, "-gp", B2DPass, "enableSharedFolders", d.vmxPath())``` line gets called regardless of whether /Users mounting is enabled or disabled.

### DIFF
--- a/docs/drivers/vm-fusion.md
+++ b/docs/drivers/vm-fusion.md
@@ -17,6 +17,7 @@ Options:
  - `--vmwarefusion-cpu-count`: Number of CPUs for the machine (-1 to use the number of CPUs available)
  - `--vmwarefusion-disk-size`: Size of disk for host VM (in MB).
  - `--vmwarefusion-memory-size`: Size of memory for host VM (in MB).
+ - `--vmwarefusion-no-share`: Disable the mount of your home directory.
 
 The VMware Fusion driver uses the latest boot2docker image.
 See [frapposelli/boot2docker](https://github.com/frapposelli/boot2docker/tree/vmware-64bit)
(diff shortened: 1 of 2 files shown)


dgageot:
Except this one detail, it looks good

jviide:
I'll remove ```defaultNoShare```. One thing to note is that the virtualbox driver (which this code is based on) seems to contain a similar unused constant.

dgageot:
@jviide Thanks for spotting that. You are right. I'll fix that.
You should also define a `EnvVar` that binds to the `vmwarefusion-no-share` flag. It is also missing in the vb driver. I'll also fix that :-)

jviide:
@dgageot Sure. I'll add an ```EnvVar``` called ```FUSION_NO_SHARE```. Can I assume (for documentation) that its default value is ```0```, as it's defined as ```mcnflag.BoolFlag```?

dgageot:
Yes. `false` might be even clearer. wdyt?

jviide:
Ah, indeed ```false``` is clearer. I'll go with that.

jviide:
@dgageot I updated the pull request based on your comments.

dgageot:
Thanks! Can you squash all the commits? Then we are ready to merge.

jviide:
Squashed.

dgageot:
LGTM

jeanlaurent:
LGTM

dgageot:
Thanks a lot @jviide!

jviide:
Thank you!

PR #2202, opened 2015-11-06
BugFix: IP in template need as per Structure

kunalkushwaha:
Lint changes in SwarmCommandOptions chnaged `Ip` to `IP`.
It required change in template code also.

### DIFF
--- a/libmachine/provision/configure_swarm.go
+++ b/libmachine/provision/configure_swarm.go
@@ -107,7 +107,7 @@ manage \
 {{range .Env}} -e {{.}}{{end}} \
 --name swarm-agent \
 {{.SwarmImage}} \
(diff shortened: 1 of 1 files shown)


dgageot:
Good catch @kunalkushwaha LGTM. ping @docker/machine-maintainers

dmp42:
Nice catch.
Templates are evil.
@dgageot any similar pattern?

dgageot:
I'll check

nathanleclaire:
Templates are not evil (:smiling_imp:), but they _are_ the wrong fit for this usecase.  I would really like to move starting the swarm containers over to use https://github.com/samalba/dockerclient instead.

nathanleclaire:
That said, this PR LGTM and thanks for catching that @kunalkushwaha

Issue #2098, opened 2015-10-27
0.5.0rc3 upgrade doesn't use most logical/recent b2d ISO


pvnovarese:
I had docker toolbox 1.9 rc1 installed on my machine, and I used the included docker-machine to create some hosts.  This went as expected and the hosts were created using the b2d 1.9 rc1 ISO.

I then upgraded toolbox to 1.9 rc2 (which upgraded b2d to 1.9 rc2 and machine to 0.5.0 rc3).  When I try upgrading the same host, docker-machine doesn't use the 1.9 rc2 ISO that ships with toolbox, and instead goes and downloads 1.8.3 ISO and "upgrades" my machine to that.

pvn@dewgong /Users/pvn> file .docker/machine/cache/boot2docker.iso
.docker/machine/cache/boot2docker.iso: ISO 9660 CD-ROM filesystem data 'Boot2Docker-v1.9.0-rc2         ' (bootable)

pvn@dewgong /Users/pvn> eval "$(docker-machine env test3)"
pvn@dewgong /Users/pvn> docker version
Client:
 Version:      1.9.0-rc2
 API version:  1.21
 Go version:   go1.4.3
 Git commit:   60d36f7
 Built:        Fri Oct 23 00:20:59 UTC 2015
 OS/Arch:      darwin/amd64

Server:
 Version:      1.9.0-rc1
 API version:  1.21
 Go version:   go1.4.3
 Git commit:   9291a0e
 Built:        Wed Oct 14 03:29:07 UTC 2015
 OS/Arch:      linux/amd64

pvn@dewgong /Users/pvn> docker-machine upgrade test3
Stopping machine to do the upgrade...
Upgrading machine test3...
Downloading https://github.com/boot2docker/boot2docker/releases/download/v1.8.3/boot2docker.iso to /Users/pvn/.docker/machine/cache/boot2docker.iso...
Starting machine back up...

pvn@dewgong /Users/pvn> docker version
Client:
 Version:      1.9.0-rc2
 API version:  1.21
 Go version:   go1.4.3
 Git commit:   60d36f7
 Built:        Fri Oct 23 00:20:59 UTC 2015
 OS/Arch:      darwin/amd64
Error response from daemon: client is newer than server (client API version: 1.21, server API version: 1.20)

pvn@dewgong /Users/pvn> docker-machine -v
docker-machine version 0.5.0-rc3 (a1e610b)

pvn@dewgong /Users/pvn> file .docker/machine/cache/boot2docker.iso
[...]

PR #2203, opened 2015-11-06
Add a "release-candidate" flag to machine creation

dave-tucker:
This flag will allow a machine created with this flag to be `upgrade`ed to the latest RC

- Add filtering logic to b2d check latest release
- Only use b2d RC's if `--release-candidate` is set
- Set EngineOpts.InstallURL to test.docker.com with `--release-candidate`
  - This ensures that most provisioners will work with this option too

Updates #2098 

Signed-off-by: Dave Tucker <dt@docker.com>

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -32,6 +32,10 @@ var (
 	errNoMachineName = errors.New("Error: No machine name specified")
 )
 
+const (
+	preReleaseInstallURL = "https://test.docker.com"
+)
+
 var (
 	sharedCreateFlags = []cli.Flag{
 		cli.StringFlag{
@@ -40,6 +44,10 @@ var (
 				"Driver to create machine with.",
 			),
 			Value: "none",
+		},
+		cli.BoolFlag{
+			Name:  "release-candidate",
+			Usage: "Use the latest release candidate versions of Docker",
 		},
 		cli.StringFlag{
 			Name:   "engine-install-url",
@@ -168,6 +176,11 @@ func cmdCreateInner(c CommandLine) error {
 		return fmt.Errorf("Error getting new host: %s", err)
 	}
 
+	installURL := c.String("engine-install-url")
+	if c.Bool("release-candidate") {
+		installURL = preReleaseInstallURL
+	}
+
 	h.HostOptions = &host.Options{
 		AuthOptions: &auth.Options{
 			CertDir:          mcndirs.GetMachineCertDir(),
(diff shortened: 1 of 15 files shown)


nathanleclaire:
Sorry to sit on this for so long @dave-tucker.  I think my preference would be to go for a `--allow-rc` flag in `docker-machine upgrade` instead.  Also, we can't make plugin API changes like are in included this PR.  cc @docker/machine-maintainers for comment and possible close.

nathanleclaire:
Oops, didn't mean to close right away.  Wanted to get the other maintainers' opinions first.

dgageot:
+1 for a flag to `update`

nathanleclaire:
What do you think about the flag for upgrade @dave-tucker ?  That way it would be `docker-machine create -d virtualbox rc; docker-machine upgrade --release-candidate rc` to create a VM with the RC and more flexibly update ones which are already created.

dave-tucker:
@nathanleclaire that was what I had hoped to do when starting out with this PR but IIRC, there didn't appear to be an easy way of changing the `InstallURL` once the machine was created... 

The flag in this PR means "when upgraded, this machine will always have the latest release, including pre-releases" - it's stored in the machine's config

Correct me if I'm wrong, but I think the flag you describe means "when checking the latest release in this upgrade operation, include pre-releases" - it's not stored in the machine's config

That has a very weird behaviour in certain cases which was what eventually convinced me that this made sense to be a creation time flag (changing the machine config to allow RC's after it's created could be solved at a later date)

Let's assume for a moment that 1.10-rc1 is out:

```
docker-machine create -d virtualbox rc                     # Installs Docker 1.9
docker-machine upgrade --release-candidate rc      # Upgrades to Docker 1.10-rc1
```

#### Weird behaviour #1 - downgrading

When rc2 is released:
```
docker-machine upgrade rc                                     # Downgrades to Docker 1.9
docker-machine upgrade --release-candidate rc     # Upgrades to Docker 1.10-rc2
```

The "upgrade" flag now performs a downgrade from the RC to the latest GA release if I forget to add the correct flag

#### Weird behaviour #2 - rc flag has no effect

Let's assume now that I've successfully upgrade my machine to the final RC, and Docker 1.10 is released

Using the `--release-candidate` flag I get the same result as without it!

```
docker-machine upgrade rc                                      # Upgrades to Docker 1.10
docker-machine upgrade --release-candidate rc      # Upgrades to Docker 1.10
```

Re: this PR, I see two options

1) Close it in favour of a flag on upgrade
2) Fix it

Re 2), I can do that.
I'll need some help from someone with some familiarity with the code...
We might be able to remove the Driver API changes... The only thing we need is to know if we are trying to provision a "RC" or not, if we can get that `bool` without having to call the driver that would be most awesome!

PR #2207, opened 2015-11-06
Fix #1737 - Detect non default VirtualBox Installation Path

### LINKED ISSUE #1737: create docker failure due to different oracle virtualbox installation path.
I can't install docker-vm successfully due to the different oracle virtualbox installation path in windows

The original path is `C:\Program Files\Oracle VM VirtualBox`, so when I do virtualbox upgrade, it will use this without change. `boot2docker` can handle this.

If it is the new installation, it will be `C:\\Program Files\\Oracle\\VirtualBox`

Now it is hard-coded in https://github.com/docker/machine/blob/master/drivers/virtualbox/vbm.go for new installation path, it looks someone already have the comment to use registry `HKEY_LOCAL_MACHINE\SOFTWARE\Oracle\VirtualBox\InstallDir`

It impacts the new installation of docker-vm and migration from boot2docker both.

jeanlaurent:
Signed-off-by: Jean-Laurent de Morlhon <jeanlaurent@morlhon.net>

This aims to fix #1737

### DIFF
--- a/drivers/virtualbox/vbm.go
+++ b/drivers/virtualbox/vbm.go
@@ -4,11 +4,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"os"
 	"os/exec"
-	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 
 	"github.com/docker/machine/libmachine/log"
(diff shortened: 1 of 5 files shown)


dmp42:
I like that.

@nathanleclaire @dgageot wdyt?

nathanleclaire:
Looks pretty OK. I would prefer very strongly to have this code with some accompanying unit tests.  Are you verifying all of these cases on an actual Windows machine @jeanlaurent ?

jeanlaurent:
@nathanleclaire Most of them, by hand for now. :sob:

I'll try to mock the `golang.org/x/sys/windows/registry`, even if my recent attempt in mocking external code in go, was a bit frustrating.

The other option is to write per-OS unit testing. but it sounds like completely overkill for this.

dgageot:
LGTM

jeanlaurent:
I wrote #2237 to discuss this per-os unit test.

I suggest we move forward with this PR now.

dgageot:
LGTM

Issue #2195, opened 2015-11-05
Upgrade doesn't work for machines where `--virtualbox-boot2docker-url` was used in create


dave-tucker:
To recreate:

```
$ docker-machine create -d virtualbox --virtualbox-boot2docker-url=https://github.com/boot2docker/boot2docker/releases/download/v1.8.3/boot2docker.iso 183

$ docker-machine upgrade 183
Stopping machine to do the upgrade...
Upgrading machine 183...
Downloading boot2docker.iso from https://github.com/boot2docker/boot2docker/releases/download/v1.8.3/boot2docker.iso...
Starting machine back up...

$ docker-machine ssh 183 docker -v
Docker version 1.8.3, build f4bf5c7
```

I'd expect that machine to be upgraded to the latest version of Docker

dmp42:
cc @nathanleclaire

nathanleclaire:
It's difficult to say what the right behavior should be, since there are certainly users who would expect a custom-used ISO to never be overwritten.

dave-tucker:
I would recommend two distinct behaviours

`docker-machine update`
- Gets the latest version of the ISO at the provided URL
- This is useful for folks using custom ISOs

`docker-machine upgrade`
- Get the latest boot2docker ISO and upgrades the machine
- Perhaps we allow for a version to be specified, if present, upgrade to that one
- Also, we could allow pre-release versions, and if enabled, upgrade to the latest release candidate

tiborvass:
I don't like having both `update` and `upgrade`.

Tibor Vass

dave-tucker:
I guess I was trying to say, instead of having one command doing 2 things for 2 different users, it might make sense to split the command in two.
`update` could be renamed `recreate` or `create --recreate` or similar if that makes it more palletable

Without splitting the behaviours, the other approach I can think of would be as follows:
- add some logic that looks at the b2d iso and see's if it's from the b2d repo
- if yes, then get the latest release and upgrade
- if no, just fetch the latest version of the ISO

I get bad vibes from just thinking about baking assumptions like that (only upgrade ISOs from b2d repo) in to code.

Especially as the "upgrade" logic could be useful if users were able to change the GitHub repo via a flag. For example, If I'm developing a boot2docker style ISO (e.g https://github.com/docker/boot2docker), it would be useful to me to have users specify that they want this at create time and to re-use the upgrade checking logic.

```
[...]

PR #2210, opened 2015-11-06
Fix integration tests for bash versions < 4.1

### LINKED ISSUE #2200: Some Integration Tests Always Pass on OSX
I was struggling to figure out why https://github.com/docker/machine/blob/master/test/integration/drivers/virtualbox/upgrade.bats was passing when I had reported #2195 

Turns out it was because of https://github.com/sstephenson/bats/issues/49

On OSX (or any Bash < 4.1)
``` 
$ bash
bash-3.2$ set -e
bash-3.2$ [[ "1.4.1" != "1.4.1" ]]
# Shell doesn't exit :(
```

Reading that ticket I see two possible fixes:
- use `[` in place of `[[` where possible
- append ` || false` to those lines

dave-tucker:
Fixes #2200

Note:

- "virtualbox: check created engine option (log driver)" fails due to
  docker/docker#17761
- "virtualbox: upgrade is correct version" fails per #2195

Signed-off-by: Dave Tucker <dt@docker.com>

### DIFF
--- a/test/integration/cli/help.bats
+++ b/test/integration/cli/help.bats
@@ -5,126 +5,126 @@ load ${BASE_TEST_DIR}/helpers.bash
 @test "cli: show info" {
   run machine
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "Usage:"  ]]
-  [[ ${lines[1]} =~ "Create and manage machines running Docker"  ]]
+  [[ ${lines[0]} =~ "Usage:"  ]] || false
+  [[ ${lines[1]} =~ "Create and manage machines running Docker"  ]] || false
 }
 
 @test "cli: show active help" {
   run machine active -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine active"  ]]
+  [[ ${lines[0]} =~ "machine active"  ]] || false
 }
 
 @test "cli: show config help" {
   run machine config -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine config"  ]]
+  [[ ${lines[0]} =~ "machine config"  ]] || false
 }
 
 @test "cli: show create help" {
   run machine create -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine create"  ]]
+  [[ ${lines[0]} =~ "machine create"  ]] || false
 }
 
 @test "cli: show env help" {
   run machine env -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine env"  ]]
+  [[ ${lines[0]} =~ "machine env"  ]] || false
 }
 
 @test "cli: show inspect help" {
   run machine inspect -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine inspect"  ]]
+  [[ ${lines[0]} =~ "machine inspect"  ]] || false
 }
 
 @test "cli: show ip help" {
   run machine ip -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine ip"  ]]
+  [[ ${lines[0]} =~ "machine ip"  ]] || false
 }
 
 @test "cli: show kill help" {
   run machine kill -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine kill"  ]]
+  [[ ${lines[0]} =~ "machine kill"  ]] || false
 }
 
 @test "cli: show ls help" {
   run machine ls -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine ls"  ]]
+  [[ ${lines[0]} =~ "machine ls"  ]] || false
 }
 
 @test "cli: show regenerate-certs help" {
   run machine regenerate-certs -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine regenerate-certs"  ]]
+  [[ ${lines[0]} =~ "machine regenerate-certs"  ]] || false
 }
 
 @test "cli: show restart help" {
   run machine restart -h
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "machine restart"  ]]
+  [[ ${lines[0]} =~ "machine restart"  ]] || false
 }
 
 @test "cli: show rm help" {
   run machine rm -h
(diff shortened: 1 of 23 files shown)


dmp42:
Landing on @jeanlaurent :)

dave-tucker:
Got my `gt` and `lt` in a muddle. This is for bash versions **less than** 4.1, like OSX's measly 3.2. Updating commit message too

dave-tucker:
will take care of any CI failures over the weekend

nathanleclaire:
Thanks for taking the torch up on this @dave-tucker.

nathanleclaire:
We should consider adding a check for the bash version in the `run-bats.sh` script, if we are writing towards one version and it is possible to get different behavior with other versions.

nathanleclaire:
@dave-tucker Is it ready for review?

nathanleclaire:
I do wonder how we can prevent such mistakes from happening again in the future (in an automated way vs. just us knowing not to use or merge in `[[`).  I don't think it's possible to alias `[[` to `[` or blow up if a user attempts to use `[[`.

dave-tucker:
@nathanleclaire yep ready for review.
tbh the best thing would be to write a short doc on style for integration tests
we could add another validate script for the makefile that looks for lines that contain `[[` and don't end in `|| false`

nathanleclaire:
> tbh the best thing would be to write a short doc on style for integration tests
we could add another validate script for the makefile that looks for lines that contain [[ and don't end in || false

Yeah, I think we should do something of that nature.  Also, would a helper function like this work?

```bash
check_cond () {
    [[ "$@" ]] || false
}
``` 

(Not sure if it would work, but it's an idea that would at least make the `[[` uses a little more readable and something we could suggest people use)

dave-tucker:
not sure i like the helper function unless it's used for everything e.g

```
assert() {
  [ "$@" ]
}

assert_ext() {
  [[ "$@ ]] || false
}
```

still, maintainers will need to remember not to merge tests that don't use the helpers for assertions

nathanleclaire:
Yeah, I'm still for programatic detection of faulty test code when possible, just thought it might be a little more easy on the eyes / easier to spot faulty `[[` uses with a ubiquitous helper.

nathanleclaire:
I'd say it looks pretty good to me, take a crack at automatic detection of invalid tests if you're feeling brave.  Maybe we can follow up with UI polish if desired.

dave-tucker:
@nathanleclaire created a linter to check for bad tests - it found one I missed ;)

test/integration/cli/help.bats
```diff
@@ -5,126 +5,126 @@ load ${BASE_TEST_DIR}/helpers.bash
 @test "cli: show info" {
   run machine
   [ "$status" -eq 0  ]
-  [[ ${lines[0]} =~ "Usage:"  ]]
-  [[ ${lines[1]} =~ "Create and manage machines running Docker"  ]]
+  [[ ${lines[0]} =~ "Usage:"  ]] || false
```
tianon: I'm not sure what problem this is solving, but I'm fairly positive it'll break the test; AFAIU, `bats` relies on `set -e`, for which the Bash manual has this to say (https://www.gnu.org/software/bash/manual/bashref.html#index-set):
> Exit immediately if a pipeline (see Pipelines), which may consist of a single simple command (see Simple Commands), a list (see Lists), or a compound command (see Compound Commands) returns a non-zero status. The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !. If a compound command other than a subshell returns a non-zero status because a command failed while -e was being ignored, the shell does not exit. A trap on ERR, if set, is executed before the shell exits.

What this means in this context is that the `|| false` makes this expression explicitly always "true" as far as `set -e` is concerned (it won't stop the script, or force the test to fail).

tianon: If this does actually work, that'd make me really happy, but from the Bash manual's text, it's not supposed to.

tianon: I guess maybe `except the command following the final && or ||` gives us our out-of-jail-free card?

dave-tucker: @tianon for context, the original issue in bats that I'm trying to work around is https://github.com/sstephenson/bats/issues/49

```
bash-3.2$ set -e
set -e
bash-3.2$ [[ 1 == 2 ]]
[[ 1 == 2 ]] 
bash-3.2$ [[ 1= = 2 ]] || false
[[ 1 == 2 ]] || false
# bash exits
```

dave-tucker: also

```
bash-3.2$ set -e
set -e
bash-3.2$ [[ 1 == 1 ]] || false
[[ 1 == 1 ]] || false
# bash doesn't exit
```

dave-tucker: so I *think* these test should be ok - at least, they pass for me

jeanlaurent:
Ran the test with an obvious error on :
* bash version 3.2.57
* bash version 4.3.42 
and I got the exact same behavior.

Could you craft a test which shows that this is really something that helps ?

As a side note, we are not that keen of introducing python deps to the build and we discussed the idea of writing bash syntax  > 4 by putting some sort of guard to the start of the script.

dgageot:
Closing. @dave-tucker please feel free to reopen if you think we missed something.

dave-tucker:
@dgageot @jeanlaurent #2200 is still an issue, and AFAICT this is the root cause.

My test case, https://github.com/docker/machine/blob/master/test/integration/drivers/virtualbox/upgrade.bats consistently passes before this test (although it shouldn't), and fails afterwards

dave-tucker:
@jeanlaurent here's a snippet you can try that illustrates this issue:

```
#!/usr/bin/env bats

load ${BASE_TEST_DIR}/helpers.bash

@test "1 is 1, 2 is 2 etc..." {
   [[ 1 == 1 ]]
   # This should case the test to fail
   [[ 2 == 5 ]]
   # Back to normal now
   [[ 3 == 3 ]]
}
```

dgageot:
Thanks a lot @dave-tucker, I would really prefer a guard on the integration tests that checks the bash version and fail if it is < 4.1. It is straightforward and doesn't impact the way we write tests. On OSX, bash 4.3 is one `brew install bash` away...

PR #2239, opened 2015-11-10
Faster CI

dgageot:
+ In-container build should start the docker container only once
+ No need to run short tests since test-long runs all tests


jeanlaurent:
LGTM

Issue #2244, opened 2015-11-10
make fails if you clone using https instead of ssh(i.e. git)


miqui:
make fails if you clone using https instead of ssh(i.e. git)
Looking at Makefile.inc, 

# Github infos
GITHUB_INFO := $(shell git config --get remote.origin.url | sed -e 's/.*:\(.*\).git/\1/')

assumes cloning with ssh.. is worth fixing/changing for both https and git?

jeanlaurent:
Thx @miqui to point this out.

Current code link is : https://github.com/docker/machine/blob/master/Makefile.inc#L4

As a side note, the build can't work offline.

dmp42:
That part is only used for release management.
Let's just `2>&1 > /dev/null` it and be happy.

dmp42:
The same thing has been done here: https://github.com/docker/machine/blob/34f818527debef02f5a3102f198e44fb911433ae/mk/main.mk#L2

nathanleclaire:
Why are we even doing that parsing at all?  As noted here, it seems to be brittle.  Why not just hardcode `docker` and `machine` for the repo information?

dmp42:
@nathanleclaire portability, and ability to test on a fork more easily come to mind.
Doesn't look like the fix is specially complex (and it does only matter for publication).

nathanleclaire:
@dmp42 What kind of portability are you referring to?  As for ability to test on a fork, editing two variables by hand seems a pretty reasonable tradeoff to me to get our Makefile simpler.

dmp42:
Portability of the makefile.

About "simplification", it's doubtful that replacing these by consts is going to make the makefile significantly simpler (what would make it simpler is extracting the in-container build, but that would come at a price too).

Nevertheless, we certainly have more important stuff to fix right now :)

jeanlaurent:
This is super tiny, it does not change the world, this is only affecting release management BUT it prevent some user from easily building docker-machine and thus probably affect the # of contribution we could get.

Note that, I did this with @dgageot in pairing and that we couldn't fully run the `make release` since it does a `git tag` + `git push`on master and we couldn't spot a "dryRun" flag or something alike.

I suggest moving forward with this, for the sole reason of helping our contributor a bit.

dmp42:
Agreed about fixing this :)

nathanleclaire:
[...]

PR #2245, opened 2015-11-10
Fix for #2244 - make fails with https remote url

jeanlaurent:


dmp42:
New version doesn't `cut` like it should.
You might want to bring back the initial sed.

dmp42:
Still not working :)

jeanlaurent:
Well if was working for ssh remote and broken for https remote. 
The 1st fix broke things, the other way around ssh ko  & http ok

Should be fixed for both now.

dgageot:
@dmp42 PTAL

dmp42:
LGTM. Maybe we want to still `2> /dev/null` if not in a git repo? (though admittedly we shouldn't allow publication at all in that case).

nathanleclaire:
Why not just set these variables to default to `docker` and `machine`?  It's trivially easy to override them from the command line if you want to change them, so why get clever?  e.g.:

```console
$ cat Makefile
FOO := bar

default:
        echo ${FOO}
$ make
echo bar
bar
$ make FOO=quux
echo quux
quux
$
```

dgageot:
+1

nathanleclaire:
I've submitted https://github.com/docker/machine/pull/2278 instead

dmp42:
@nathanleclaire I believe we disagree on this, so we should discuss this more.
Let's sit together when you can.

dgageot:
@nathanleclaire, @dmp42 let's move on with that one. Let's look at the facts:

+ Both https://github.com/docker/machine/pull/2245 and https://github.com/docker/machine/pull/2278 do the job of fixing #2244
+ https://github.com/docker/machine/pull/2245 keeps the current property of working on a clone or a renamed repo. So it's more user friendly.
+ We have bigger fish to fry

I'm going to LGTM that one and close https://github.com/docker/machine/pull/2278

dgageot:
LGTM

PR #2265, opened 2015-11-12
Build on all platforms

dgageot:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,14 @@ sudo: required
 dist: trusty
 language: bash
 services: docker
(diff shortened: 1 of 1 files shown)


jeanlaurent:
LGTM

dmp42:
Mmmm... if you don't call `make cross`, things are not going to be built.
Also, I think you can use `TARGET_OS=darwin,windows,linux` and `make cross` will build all three in one row.

dgageot:
Yes @dmp42, I realised that afterwards. I fixed it a couple of hours later in https://github.com/docker/machine/commit/98abeddf3be980483a6da71a382245f54f2ef4e1

PR #2272, opened 2015-11-12
Speed up Integration Tests

jeanlaurent:
This speedup integration tests in a number of ways :
* Copy boot2docker locally rather than download it each time for test using it.
* Allow to use an already running docker machine rather than create a new each time by specifying a `USECONTAINER=myowntestcontainer` instead of `DRIVER`
* cleanup properly the cache folder in case a test went :boom: kaboom :boom: before

Signed-off-by: Jean-Laurent de Morlhon <jeanlaurent@morlhon.net>

### DIFF
--- a/test/integration/run-bats.sh
+++ b/test/integration/run-bats.sh
@@ -4,6 +4,9 @@ set -e
 
 # Wrapper script to run bats tests for various drivers.
 # Usage: DRIVER=[driver] ./run-bats.sh [subtest]
+# or
+# If you just want to quickly run the test on an existing container
+# Usage: RECYCLE_MACHINE=[container name] ./run-bats.sh [subtest]
 
 function quiet_run () {
     if [[ "$VERBOSE" == "1" ]]; then
@@ -14,8 +17,12 @@ function quiet_run () {
 }
 
 function cleanup_machines() {
-    if [[ $(machine ls -q | wc -l) -ne 0 ]]; then
-        quiet_run machine rm -f $(machine ls -q)
+    if [[ -z "$RECYCLE_MACHINE" ]]; then
+        if [[ $(machine ls -q | wc -l) -ne 0 ]]; then
+            quiet_run machine rm -f $(machine ls -q)
+        fi
+    else
+        echo "Explicity using Container ${RECYCLE_MACHINE}, so not killing it, on your request..."
     fi
 }
 
@@ -44,7 +51,7 @@ function run_bats() {
 EXIT_STATUS=0
 export BATS_FILE="$1"
 
-if [[ -z "$DRIVER" ]]; then
+if [[ -z "$RECYCLE_MACHINE" ]] && [[ -z "$DRIVER" ]]; then
     echo "You must specify the DRIVER environment variable."
     exit 1
 fi
@@ -62,10 +69,18 @@ fi
 # TODO: Should the script bail out if these are set already?
 export BASE_TEST_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 export MACHINE_ROOT="$BASE_TEST_DIR/../.."
-export NAME="bats-$DRIVER-test"
-export MACHINE_STORAGE_PATH="/tmp/machine-bats-test-$DRIVER"
+
+if [[ -z "$RECYCLE_MACHINE" ]]; then
+    export MACHINE_STORAGE_PATH="/tmp/machine-bats-test-$DRIVER"
+    export NAME="bats-$DRIVER-test"
+else
+    export DRIVER=$(docker-machine inspect "$RECYCLE_MACHINE" | grep DriverName | cut -d\" -f 4)
+    export NAME=$RECYCLE_MACHINE
+fi
+
 export MACHINE_BIN_NAME=docker-machine
 export BATS_LOG="$MACHINE_ROOT/bats.log"
+B2D_LOCATION=~/.docker/machine/cache/boot2docker.iso
 
 # This function gets used in the integration tests, so export it.
 export -f machine
@@ -73,10 +88,26 @@ export -f machine
 touch "$BATS_LOG"
 rm "$BATS_LOG"
 
+if [[ -z "$RECYCLE_MACHINE" ]]; then
+
+    if [[ -d "$MACHINE_STORAGE_PATH" ]]; then
+        rm -r "$MACHINE_STORAGE_PATH"
+    fi
+
+    if [[ "$b2dcache" == "1" ]] && [[ -f $B2D_LOCATION ]]; then
+        mkdir -p "${MACHINE_STORAGE_PATH}/cache"
+        cp $B2D_LOCATION "${MACHINE_STORAGE_PATH}/cache/boot2docker.iso"
+    fi
+
+fi
+
 run_bats "$BATS_FILE"
 
-if [[ -d "$MACHINE_STORAGE_PATH" ]]; then
-    rm -r "$MACHINE_STORAGE_PATH"
+if [[ -z "$RECYCLE_MACHINE" ]]; then
+
+    if [[ -d "$MACHINE_STORAGE_PATH" ]]; then
+        rm -r "$MACHINE_STORAGE_PATH"
+    fi
 fi
 
 set +e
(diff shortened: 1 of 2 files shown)


jeanlaurent:
This a little impact as of now, since most integration test are not ran since the `none` driver has regression, and we have trouble to run virtuabox on Travis/CirclCI as of now.

So the impact is only *for now* on our workstation.

test/integration/run-bats.sh
```diff
@@ -62,10 +71,18 @@ fi
...
+    export MACHINE_STORAGE_PATH="/tmp/machine-bats-test-$DRIVER"
+    export NAME="bats-$DRIVER-test"
+else
+    export DRIVER=$(docker-machine inspect "$USECONTAINER" | grep DriverName | cut -d\" -f 4)
+    export NAME=$USECONTAINER
+fi
+
 export MACHINE_BIN_NAME=docker-machine
 export BATS_LOG="$MACHINE_ROOT/bats.log"
+BD2_LOCATION=~/.docker/machine/cache/boot2docker.iso
```
dmp42: `B2D`? (super minor nit)

dmp42:
Haven't tested it manually but LGTM

nathanleclaire:
__NOT LGTM__

We shouldn't re-use machines (or stores) between integration tests.  The integration tests deliberately set up a new machine and store each time for isolation purposes.  If they are allowed to interfere with each other, suddenly you have a situation where one integration test might be breaking, but the breakage could actually be caused by another integration test.  Not fun.

It's already trivial to run the integration tests for `virtualbox` without needing to download the boot2docker ISO every time.  Just set `VIRTUALBOX_BOOT2DOCKER_URL` to `file:///my/local/boot2docker.iso` if that's the issue and I think it should run without going off and downloading the new ISO.

dmp42:
I don't think this changes the way the stores are (not) shared (at all), does it?
Good point about `VIRTUALBOX_BOOT2DOCKER_URL` already existing though.

Nevertheless, please do not close PRs like that - we discuss it, reach an agreement, and if need be, the author will close the PR himself.

dmp42:
And by the way (unless this was changed recently) our current model is broken as-is: concurrent tests of the same driver will use the same store (this bite me in the past).

nathanleclaire:
> I don't think this changes the way the stores are (not) shared (at all), does it?
Good point about VIRTUALBOX_BOOT2DOCKER_URL already existing though.

It does change the way that things are handled in that now we have a custom shell script which actually does changing around of store properties across test runs.  Making the store directory ahead of time and hacking it to be in states that we want dilutes the power of integration tests that otherwise run from scratch.

> And by the way (unless this was changed recently) our current model is broken as-is: concurrent tests of the same driver will use the same store (this bite me in the past).

The integrations tests were not designed to be run concurrently and safely, so I'd hardly consider that "broken".  If we want to add that as a feature, we definitely could, but they're just not meant to be run at the same time right now.

dgageot:
When using integration tests, I'm happy to have two options: the "safe path" and the "fast path".

+ The safe path is what runs on the CI. It is running the integration tests from scratch with no side-effect. It gives maximum confidence.
+ The fast path is what a developer CAN run locally to minimize the feedback loop. It runs the integration tests against a prepared environment for the price of reducing confidence.

So this PR looks good to me. We can improve the code and keep the idea.

jeanlaurent:
The title may suggest otherwise but the speedup increase is meant to be in some situation where we don't change the host (`env` & `env -u` are one of those case), or during development to speed the process. You are right that we need most of the time a clean fresh host created just for this purpose, but that's not the only way of doing integration tests.

Facts :
* Running `test/integration/core/env_shell.bats` takes 5 minutes. (we have very poor bandwidth atm)
* With the automatic `cp` of b2d iso we fall down to 1 minutes & 10 secs.
* With the added, custom way of running a test within an already running container, this falls down to 5secs.

I'm glad to have saved 4min & 5 secs on the hundred launches of integration tests I have done yesterday, I thought that we could benefit from it.

I'll update the code to use the `VIRTUALBOX_BOOT2DOCKER_URL` rather than the custom `cp` it is doing right now.

nathanleclaire:
@jeanlaurent I'm still _really_ confused by the naming `USECONTAINER` here.  We do something like `USE_CONTAINER=true` to do the build in a Docker container in our Makefile, but here it seems to be referencing something more like using an existing store.  Can you clarify when you get a chance please?

jeanlaurent:
100% Agree, I'll rename `USECONTAINER` to something less confusing.

jeanlaurent:
renamed the confusing `USECONTAINER` to `RECYCLE_CONTAINER` which conveys more sense and is not confusing with the `USE_CONTAINER` variable.

If you want to try this out 
```
machine [speedup-integration|✔] > time DRIVER=virtualbox test/integration/run-bats.sh test/integration/core/env_shell.bats 
=> test/integration/core/env_shell.bats
 ✓ virtualbox: create
 ✓ virtualbox: test basic bash / zsh notation
 ✓ virtualbox: test powershell notation
 ✓ virtualbox: test bash / zsh notation with no-proxy
 ✓ virtualbox: test cmd.exe notation
 ✓ virtualbox: test fish notation
 ✓ virtualbox: test no proxy with NO_PROXY already set
 ✓ virtualbox: test unset with an args should fail
 ✓ virtualbox: test bash/zsh unset
 ✓ virtualbox: test unset killing no proxy
 ✓ virtualbox: unset powershell
 ✓ virtualbox: unset with fish shell
 ✓ virtualbox: unset with cmd shell

13 tests, 0 failures


real	1m43.918s
user	0m1.549s
sys	0m0.829s
```
If you have an already running container named testbox
```
machine [speedup-integration|✔] > time RECYCLE_CONTAINER=testbox test/integration/run-bats.sh test/integration/core/env_shell.bats 
=> test/integration/core/env_shell.bats
 ✓ virtualbox: create
 ✓ virtualbox: test basic bash / zsh notation
 ✓ virtualbox: test powershell notation
 ✓ virtualbox: test bash / zsh notation with no-proxy
 ✓ virtualbox: test cmd.exe notation
 ✓ virtualbox: test fish notation
 ✓ virtualbox: test no proxy with NO_PROXY already set
 ✓ virtualbox: test unset with an args should fail
 ✓ virtualbox: test bash/zsh unset
 ✓ virtualbox: test unset killing no proxy
 ✓ virtualbox: unset powershell
 ✓ virtualbox: unset with fish shell
 ✓ virtualbox: unset with cmd shell

13 tests, 0 failures

Explicity using Container testbox, so not killing it, on your request...

real	0m3.960s
user	0m0.629s
sys	0m0.610s
```

103s vs 4s...

nathanleclaire:
I meant that "Container" almost universally means "Linux container / process" in Dockerland, and it seems to mean "Machine" or "Store" here (not really sure which).  So, `RECYCLE_CONTAINER` is equally confusing wording to me.

I'm still really against this because:

1. Integration tests are meant to be run with a clean slate, and allowing to run them from a "dirty" working machine/store will cause issues with non-reproducibility of integration tests ("the integration tests work on my machine").  Integration tests are meant to test how each component of the program works together in a real use case. Short-circuiting bits of the program to make them faster defeats the purpose.
2. This PR in particular only updates one instance of `machine create` so it's usable with only one of our many integration tests, yet touches code that all of them depend on.  This is a risk.
3. I don't see how it's possible to update all tests to use this option since it's not always the case that the tests create just one machine with default options.  Some tests (e.g. of Swarm) create multiple machines in their setup.  Some create engines with different options than the default ones.  Some go mucking with the store to put a different `boot2docker.iso` in its cache.  So, it's impossible to make this option work for all of our integration tests.

If we want more tests that run quickly, we should write more unit tests.  Those are actually meant for testing components in isolation.  It's perfectly valid for the integration tests to _always_ take a long time.

jeanlaurent:
Rewrote it to `RECYCLE_MACHINE` because yes it's more a machine than a container.

I believe we should move step by step, I think this bashes lines are a pegleg in the transition from the poor state of testing on machine with unit and integration tests towards something better in the near future. Getting a 50x increase on local workstation testing time on some test sounds like a bargain.

jeanlaurent:
This PR contains 3 distincts parts, one part started an infinite-loop-time-consuming debate, so closing it, to free uptime for everybody.

To try to move forward with the two other relevant pieces I have open #2331 and #2332.

Let's try to merge them quicker than this one.

PR #2274, opened 2015-11-12
[WIP] Code quality

dgageot:
Trying to improve the codebase, step by step...


nathanleclaire:
I don't understand. What's the motivation for moving this code around like so?  It doesn't seem to introduce any new tests or cover any new code paths.  What concrete metric does this improve?

nathanleclaire:
I guess my concern is partially that this is about 5 different refactoring changes lumped into one PR, and there will be way more back-and-forth getting it merged in than having it as separate PRs.  That way, if there's one small change I'm unhappy about, we could debate on that one and merge the other ones in.

dgageot:
@nathanleclaire do there will be hundreds of PR because the code is a mess to test and read. All those commits are doing is moving code where it belongs and removing dead code. If we seriously want to improve the test coverage, we have MUCH more things to cleanup

nathanleclaire:
It shouldn't take hundreds of PRs.  Just group them more logically.  All of the dead code removal in one PR, a small refactor to other code and addition of a test in another.  The impetus is on the person who is changing the code to provide a reason why it is changing.  It's not clear what changing `map[string]func()error` to a `switch` instead will win us.  

If changes are for increased testability, add tests.  Otherwise things are just changing for the entirely subjective reason that "this code is messy and it sucks".  If it's being distributed and working today, changing it is inherently a risk, no matter how simple it seems.  Stuff breaks when you change it even when you didn't expect it.

jeanlaurent:
The project is as of now poorly tested. Whenever we try to add tests in significant areas, we came up with lots of refactoring, making giants changes, leading to huge pr which are difficult to merge. One way of obtaining this testability while avoiding giant PR, is to proceed with small steps, some of those steps are so insignificant (https://github.com/dgageot/machine/commit/403b9bcd584a9ac2acf21b6714f3c29522f7cf36) that they don't need an extra test. If we don't do this, we can't move towards increased testability as this is a vicious circle. No test -> Can't Change Code -> No Test -> Can't Change code...

So LGTM

dgageot:
@nathanleclaire Let's submit 7 PRs. I think it's crazy but let's give it a try!

nathanleclaire:
@dgageot To be clear, I did not mean 7 PRs which are depend on the last one.  I meant several PRs which are mergeable independent of one another.  Otherwise, it's not much different than one giant PR.

dgageot:
Some of them are independant. Please LGTM those you agree on and we can merge them independently later.

nathanleclaire:
@dgageot That's just not true. _None_ of them are independent.  They all rely on commits from the previous one.  Therefore I can't LGTM any of the ones further along the chain, without LGTMing the ones that came first.  My point was that ideally it should be possible to review and/or reject any of them without having to do so with the other ones.

screen shot 2015-11-13 at 12 06 34 pm
screen shot 2015-11-13 at 12 06 30 pm
screen shot 2015-11-13 at 12 06 27 pm
screen shot 2015-11-13 at 12 06 22 pm
screen shot 2015-11-13 at 12 06 18 pm

dgageot:
What I meant is some commits are Indépendant even if they are right now part of a combined PR. Could you at least review the commits?

nathanleclaire:
Sure, maybe there's just been a miscommunication about process -- I'm happy to review the individual commits, it just makes it more difficult to track if they're mixed in with other commits in the same PR.  In addition, I like to merge things upstream quickly if I find they're good changes, so that's why I may seem difficult / adamant about having each separate change in its own PR.  If you're not comfortable using `rebase` to split them out, I can give you some pointers as well.

That said, for these existing ones, some changes look really good and some look like they'll conflict quite a bit with the work I've been doing.  I'll comment on each one and also get my own changes up soon so we can discuss.

dgageot:
Good. Waiting for your feedback. I kept the previous commits in each pr to provide some context. I'll change that for my next refactoring commits. And don't worry, I know how to rebase :-)

PR #2277, opened 2015-11-12
Run integration tests on TravisCI

dgageot:
Replace Circle CI by TravisCI where we can run the integration tests inside a container


jeanlaurent:
Last stand of CircleCI as it seems...

dmp42:
Integration testing for drivers inside the container is going to be a PITA.
Drivers use a huge bunch of environment variables that you will have to pass one by one to the container.
I'm not so sure we want to live with that (maintenance burden).

Nevertheless, for the existing tests, that LGTM

dgageot:
I understand.We could be inside the container and run the tests outside the container. 
I'll check tomorrow how hard it is to add some of the disabled tests.

nathanleclaire:
What's the reasoning for wanting to run the integration tests in a container?  I am usually all for dogfooding wherever we can, however I don't know that it gains us much in this scenario.

dgageot:
@nathanleclaire As for the other make targets, running inside a container would give us reproducibility on any developer machine.

nathanleclaire:
So this PR also removes CircleCI entirely as far as I understand it?  I'm in favor of that (one CI system is much simpler to manage than two), just want to make sure we're all committing to that as the gameplan.

dmp42:
If travis provides a way to store and pass env variables at build time (for cloud integration tests), I'm all in favor - otherwise, we should keep circle for the time being while we await a better solution.

dgageot:
You know what, let's close it and we come back later when we have more reasons to move to Travis.

PR #2281, opened 2015-11-13
More accurate way to calculate drivers' urls with IPv6 support

### LINKED ISSUE #2219: Error "tcp: too many colons in address <IPv6 addr of created machine>:2376" when executing env command
I've successfully created a machine using `openstack` driver, but when I execute `docker-machine env <my-machine>` I get an error:

~~~
  › docker-machine -D env os01f      
                                                                                                                                              
Docker Machine Version: 0.5.0 (04cfa58)
Found binary path at /usr/local/bin/docker-machine-driver-openstack
Launching plugin server for driver openstack
Plugin server listening at address 127.0.0.1:61158
Using API Version 1
(os01f) Calling RpcServerDriver.GetState
(os01f) DBG | Get status for OpenStack instance...		 MachineId=<machine-id>
(os01f) DBG | Authenticating...		 AuthUrl=<openstack-auth-url> DomainID=default DomainName= Insecure=false TenantID= TenantName=<tenant-name> Username=<username>
(os01f) DBG | State for OpenStack instance		 MachineId=<machine-id> State=ACTIVE
(os01f) Calling RpcServerDriver.GetURL
Reading CA certificate from /Users/varankinv/.docker/machine/certs/ca.pem
Reading server certificate from /Users/varankinv/.docker/machine/machines/os01f/server.pem
Reading server key from /Users/varankinv/.docker/machine/machines/os01f/server-key.pem
Error running connection boilerplate: Error checking and/or regenerating the certs: There was an error validating certificates for host "<my machines ipv6 address>:2376": dial tcp: too many colons in address <my machines ipv6 address>:2376
You can attempt to regenerate them using 'docker-machine regenerate-certs name'.
Be advised that this will trigger a Docker daemon restart which will stop running containers.
~~~

My machine is running inside IPv6-only network and looks like the only way to manage it with docker-machine is to specify `$DOCKER_HOST` env variable manualy using machine's fqdn.

At the same time I can ssh to machine with `docker-machine ssh <my-machine>` as well as run other commands.

narqo:
Fix `drivers.GetUrl()` to build proper IPv6 addr.

Use [`net.JoinHostPort()`](https://godoc.org/net#JoinHostPort) as it has support for IPv6 literals, which should be enclosed in `[]`.

fix #2219 but depends on docker/docker#15557 which adds support for valid IPv6 addresses in `$DOCKER_HOST`. Unfortunately it hasn't been landed to any docker's release yet.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/url"
 	"strconv"
 	"strings"
@@ -430,7 +431,7 @@ func (d *Driver) GetURL() (string, error) {
 	if ip == "" {
 		return "", nil
 	}
-	return fmt.Sprintf("tcp://%s:%d", ip, dockerPort), nil
+	return fmt.Sprintf("tcp://%s", net.JoinHostPort(ip, strconv.Itoa(dockerPort))), nil
 }
 
 func (d *Driver) GetIP() (string, error) {
--- a/drivers/digitalocean/digitalocean.go
+++ b/drivers/digitalocean/digitalocean.go
@@ -3,6 +3,7 @@ package digitalocean
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"time"
 
 	"github.com/digitalocean/godo"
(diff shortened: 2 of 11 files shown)


jeanlaurent:
Thanks @narqo for taking the time to check each & every driver.

I wish we vould do some simple unit testing on those cases, but unfortunately the way things are put together makes this quite hard.

Sounds like docker/docker#15557 is merged, so here's my 

LGTM

dgageot:
LGTM

PR #2288, opened 2015-11-13
Quality01 - Remove unused code

dgageot:

Let's remove some unused code!
+ `store.Exists` is already checked in `loadHost()`
+ Let's get rid of an old comment I left after I removed the calls to `log.fatal()`
+ Some test helpers that are never used


nathanleclaire:
LGTM

PR #2289, opened 2015-11-13
Quality 02 move code

dgageot:
Not a big deal, but all this code is in the wrong place. It makes it difficult to understand.

To be merged after https://github.com/docker/machine/pull/2288


jeanlaurent:
Sounds like dead simple, just moving code around for better readability

LGTM

PR #2290, opened 2015-11-13
Quality 03 simpler test

dgageot:
Let's remove all the duplication in this test!

To be merged after https://github.com/docker/machine/pull/2289


dmp42:
@nathanleclaire no maintainer should merge its own code himself.
*LGTM are mandatory*.
 It should not be necessary to point that out.

PR #2291, opened 2015-11-13
Quality 04 inject store

dgageot:
Instead of creating a store in each action, let's create it in only one place and pass
it to the actions. It's not enough to make the actions easy to test but it's mandatory.

To be merged after https://github.com/docker/machine/pull/2290

PR #2292, opened 2015-11-13
Quality 05 use functions

dgageot:
Why use an indirection when we can just call the function that matches the action.
Simpler code

To be merged after https://github.com/docker/machine/pull/2291

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -12,12 +12,10 @@ var (
 	errTooManyArguments = errors.New("Error: Too many arguments given")
 )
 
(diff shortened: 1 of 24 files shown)


dgageot:
@nathanleclaire Go can pass functions as parameters. Why use an indirection with a lookup table? This commit makes it easier to understand which function each command really calls on the host.

dgageot:
Closing. Let's try to improve https://github.com/docker/machine/pull/2295

PR #2293, opened 2015-11-13
Quality 06 rename cli

dgageot:
A little renaming because who understands what `c` means?

To be merged after https://github.com/docker/machine/pull/2292

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -12,12 +12,10 @@ var (
 	errTooManyArguments = errors.New("Error: Too many arguments given")
 )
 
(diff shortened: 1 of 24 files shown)


nathanleclaire:
In the end, we may want to call it `client` since in the https://github.com/docker/machine/pull/2295 it can perform a few different functions.

dgageot:
@nathanleclaire This is a simple rename to make code easier to read. It has nothing to do with what #2295 could lead to

dgageot:
Closing. Let's try to improve https://github.com/docker/machine/pull/2295

PR #2294, opened 2015-11-13
Quality 07 rpc store

dgageot:
The *piece de resistance*!

Let's put remove those TODOs and put the rpc wrapping/unwrapping and the rawDriver hack in one place, ready to be refactored.

To be merged after https://github.com/docker/machine/pull/2293

PR #2295, opened 2015-11-14
Move to using more testable interfaces

nathanleclaire:
It is a combination of @dgageot commits / refactoring work and my own.  Demonstration of increased testability and proper rebase coming soon.  This contains all the steps needed to get us to extremely testable code at the subcommand level.

### DIFF
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -10,8 +10,6 @@ import (
 	"github.com/docker/machine/commands"
 	"github.com/docker/machine/commands/mcndirs"
 	"github.com/docker/machine/libmachine/log"
-	"github.com/docker/machine/libmachine/mcnutils"
-	"github.com/docker/machine/libmachine/ssh"
 	"github.com/docker/machine/version"
 )
 
@@ -74,16 +72,6 @@ func main() {
 	app.Name = path.Base(os.Args[0])
 	app.Author = "Docker Machine Contributors"
 	app.Email = "https://github.com/docker/machine"
-	app.Before = func(c *cli.Context) error {
-		// TODO: Need better handling of config, everything is too
-		// complected together right now.
-		if c.GlobalBool("native-ssh") {
-			ssh.SetDefaultClient(ssh.Native)
-		}
-		mcnutils.GithubAPIToken = c.GlobalString("github-api-token")
-		mcndirs.BaseDir = c.GlobalString("storage-path")
-		return nil
-	}
 
 	app.Commands = commands.Commands
 	app.CommandNotFound = cmdNotFound
--- a/commands/active.go
+++ b/commands/active.go
@@ -4,46 +4,33 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/persist"
 )
 
 var (
 	errTooManyArguments = errors.New("Error: Too many arguments given")
+	errNoActiveHost     = errors.New("No active host found")
 )
 
-func cmdActive(c CommandLine) error {
+func cmdActive(c CommandLine, api libmachine.API) error {
 	if len(c.Args()) > 0 {
 		return errTooManyArguments
 	}
 
-	store := getStore(c)
-
-	host, err := getActiveHost(store)
+	hosts, err := persist.LoadAllHosts(api)
 	if err != nil {
 		return fmt.Errorf("Error getting active host: %s", err)
 	}
 
-	if host != nil {
-		fmt.Println(host.Name)
-	}
+	items := getHostListItems(hosts)
 
-	return nil
-}
-
-func getActiveHost(store persist.Store) (*host.Host, error) {
-	hosts, err := listHosts(store)
-	if err != nil {
-		return nil, err
-	}
-
-	hostListItems := getHostListItems(hosts)
-
-	for _, item := range hostListItems {
+	for _, item := range items {
 		if item.Active {
-			return loadHost(store, item.Name)
+			fmt.Println(item.Name)
+			return nil
 		}
 	}
 
-	return nil, errors.New("Active host not found")
+	return errNoActiveHost
 }
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -9,14 +9,13 @@ import (
 
 	"github.com/codegangsta/cli"
 	"github.com/docker/machine/commands/mcndirs"
-	"github.com/docker/machine/drivers/errdriver"
+	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/cert"
-	"github.com/docker/machine/libmachine/drivers"
-	"github.com/docker/machine/libmachine/drivers/plugin/localbinary"
-	"github.com/docker/machine/libmachine/drivers/rpc"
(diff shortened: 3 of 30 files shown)


nathanleclaire:
You can see the new test in action here: https://github.com/nathanleclaire/machine/pull/5

Where I've added a test for `env` as well.

nathanleclaire:

nathanleclaire:
Naturally we should wait until after the 0.5.1 release to merge this, so no rush.

nathanleclaire:
OK good times, I have this passing in the CI now.

nathanleclaire:
Summary of changes in case it's confusing to figure out just from the diff:

- All "helper methods" that coordinate a combination of CLI-related things and Machine-related things (e.g., `RunActionWithContext` which runs actions on hosts based on what's in the `cli.Context`) have been moved into the `MachineCLIHelper` interface which is embedded inside of `MachineCLIClient`, the interface which _actually_ gets injected into the command functions.  This is kind of cool, since you can now call things like `c.GetCertPathInfoFromContext` and not have to pass `c` around as an argument to every helper function like we were doing before.
- The interface for wrapping `codegangsta/cli` operations has been renamed to `CLIer` and is an embedded property of the interface injected into `cmd*` functions instead of the thing directly passed in.
- Logic related to loading plugin drivers has been moved directly into the `Filestore`, which may not be an amazing choice long term, but frees up a whole bag of issues with trying to separate the two (e.g. what if some consumer of `libmachine`, including the Machine CLI, tries to `Load` a `Host`/`Driver` without going through the plugin binary).
- I know it seems like `c` / `contextCommandLine` might be at risk of becoming some kind of "God object" with many methods, but invocations like `c.List()` seem pretty reasonable to me given the scope of what the methods they're used are expected to be working with (a machine context and a command line context). That said, we can consider invoking `c.Store.List()` directly if we feel like that's more clear, `c.List()` is just a convenient shorthand which resolves to that anyway. 

These changes actually allow us to control `libmachine` operations and `cli`-related operations with quite a lot of granularity, so they'll be quite valuable for testing.  I'm not so concerned with getting naming etc. perfect on the first go-round as I am with just getting to a place where we can start writing real tests.

dgageot:
@nathanleclaire Thanks for the work, here's my feedback on a couple of points:

+ Lot's commands use `c commandLine` to extract the args which are either a machine name or multiple machine names. This should be obvious and not hidden behind `GetFirstArgHost`, `GetHostsFromContext` and `RunActionWithContext` functions. What we really need is a `GetHost(machineName string)` and `RunAction(machineName []string)

+ `MachineCLIHelper` is mixing things that are related to command line and the store.  Please give it a single responsibility.

+ Do you think `CLIer` is a better name than `CommandLine`? In go, it's the norm to call an interface `XXer` when it has a single function called `XX`. This is not the case here. I'm confused.

+ I don't think the `Filestore` should take care of loading plugin drivers. Let's group this code outside of `Filestore` to enforce a single responsibility to each struct and to make things easier to test.

+ Also with this kind of refactoring, please consider splitting the thing into multiple commits to make it easier to review. For example, I heard you once ask for commits dedicated to moving functions around. I think it's a great idea that would have helped understand this PR.

nathanleclaire:
> Lot's commands use c commandLine to extract the args which are either a machine name or multiple machine names. This should be obvious and not hidden behind GetFirstArgHost, GetHostsFromContext and RunActionWithContext functions. What we really need is a GetHost(machineName string) and `RunAction(machineName []string)

> MachineCLIHelper is mixing things that are related to command line and the store. Please give it a single responsibility.

Hm.. that _is_ the intent of the `MachineCLIHelper`: to contain the little helper functions we are using today that rely on both the `Store` / `libmachine` and the CLI (so that they can be mocked out).  `GetFirstArgHost` probably could go away and be replaced by something like `c.Load(c.First())` and the others could likely stand to be reduced down to a less complex form as well.  I'll see where I can take it in that regard.

> Do you think CLIer is a better name than CommandLine? In go, it's the norm to call an interface XXer when it has a single function called XX. This is not the case here. I'm confused.

Either is OK with me, `CLIer` was just shorter and is slightly incidental.

> I don't think the Filestore should take care of loading plugin drivers. Let's group this code outside of Filestore to enforce a single responsibility to each struct and to make things easier to test.

How about this:  `Filestore` implements `type interface Store` that we have today and there is an additional type `PluginStore` which also implements `Store`, embedding the `Filestore` and doing the additional plugin operations on top.  That would allow us to test `PluginStore` and the operations which depend on it both quite robustly.

> Also with this kind of refactoring, please consider splitting the thing into multiple commits to make it easier to review. For example, I heard you once ask for commits dedicated to moving functions around. I think it's a great idea that would have helped understand this PR.

Yeah, I agree it's a monster, and I was far too over-optimistic about how big of a diff it would take to get to what I was envisioning.  I'll be more conscientious in the future.

dgageot:
> > Lot's commands use c commandLine to extract the args which are either a machine name or multiple machine names. This should be obvious and not hidden behind GetFirstArgHost, GetHostsFromContext and RunActionWithContext functions. What we really need is a GetHost(machineName string) and `RunAction(machineName []string)

> > MachineCLIHelper is mixing things that are related to command line and the store. Please give it a single responsibility.

> Hm.. that _is_ the intent of the `MachineCLIHelper`: to contain the little helper functions we are using today that rely on both the `Store` / `libmachine` and the CLI (so that they can be mocked out).  `GetFirstArgHost` probably could go away and be replaced by something like `c.Load(c.First())` and the others could likely stand to be reduced down to a less complex form as well.  I'll see where I can take it in that regard.

+1 to replace `GetFirstArgHost` with  `c.Load(c.First())` and try do the same everywhere it makes sense. That would clearly mark the frontier between the cli
and the store.

> > Do you think CLIer is a better name than CommandLine? In go, it's the norm to call an interface XXer when it has a single function called XX. This is not the case here. I'm confused.

> Either is OK with me, `CLIer` was just shorter and is slightly incidental.  I'm happy to change it back.

I prefer to keep the old name. It seems easier to understand :smile:

> > I don't think the Filestore should take care of loading plugin drivers. Let's group this code outside of Filestore to enforce a single responsibility to each struct and to make things easier to test.

> How about this:  `Filestore` implements `type interface Store` that we have today and there is an additional type `PluginStore` which also implements `Store`, embedding the `Filestore` and doing the additional plugin operations on top.  That would allow us to test `PluginStore` and the operations which depend on it both quite robustly.

That would be good. Exactly what I did in https://github.com/docker/machine/pull/2294. I let you do that.

nathanleclaire:
OK @dgageot I've implemented the requested feedback.  PTAL and then I will rebase and submit my `env` test as a followup.

dgageot:
Thanks @nathanleclaire. I can see it goes in the right direction. I still don't understand how having one argument/one interface passed to each command is better than passing both a `CommandLine` and a `PluginStore`.

nathanleclaire:
I think your point is valid for understanding of the code.  We should inject the whole `libmachine.API` though.  I'll update to inject both objects.

nathanleclaire:
Updated; PTAL

dgageot:
LGTM

dgageot:
:dancers:

nathanleclaire:
:+1: :+1: :+1:

PR #2297, opened 2015-11-14
Change godep command call to use the GODEP variable

yamamoto-febc:
If not exists godep in $PATH , following error occurs.
```bash
make: godep: No such file or directory
```
Because it is calling directly godep command without a GODEP variable.
This PR fix it.

Signed-off-by: Kazumichi Yamamoto <yamamoto.febc@gmail.com>


dmp42:
If godep doesn't exist in the path, this should fail earlier.
LGTM either way, as hardening.

PR #2327, opened 2015-11-18
Add -n to get port number for netstat

nathanleclaire:
On Fedora the new "detect daemon is listening" method did not work because netstat needed the `-n` flag.

cc @tianon You might be curious to know that the `netstat` output actually _was_ different between distros.  Ah, the joys of OSS.


dgageot:
LGTM

PR #2331, opened 2015-11-18
IT: cleanup properly before starting

jeanlaurent:
This ensure that, we cleanup properly the work directory AND running containers whenever a test fails or crash bats


dgageot:
LGTM

When integration tests are stopped in the middle, this is very useful to start from a clean state.

PR #2332, opened 2015-11-18
Bootstrap test quicker by caching boot2docker iso if present

jeanlaurent:
by adding a b2dcache env variable before launching bats, we will copy the boot2docker iso before starting the test run in the test directory

This help by NOT downloading the b2d iso when you already have it locally.

The current implementation copy the iso by hand, but we could you use the ` VIRTUALBOX_BOOT2DOCKER_URL` but this would apply only to vbox.


dmp42:
Needs a rebase

jeanlaurent:
rebased

dgageot:

PR #2337, opened 2015-11-18
Fix its

dgageot:
With this PR, the virtualbox Integration tests will at last be all green!!

+ It makes sure everything is built before running the its
+ It fixes two tests that are broken because of a change in docker-machine and a change in docker
+ It removes the `pkill` thing at the end of the its because we don't need it anymore and it makes the its always fail
+ It gives a unique name to each machine created by the its. This has a huge effect on test robustness since a failing test won't break another test anymore.

### DIFF
--- a/mk/test.mk
+++ b/mk/test.mk
@@ -6,7 +6,7 @@ test-short:
 test-long:
 	$(GO) test $(VERBOSE_GO) -race -tags "$(BUILDTAGS)" $(PKGS)
 
-test-integration:
+test-integration: build
 	$(eval TESTSUITE=$(filter-out $@,$(MAKECMDGOALS)))
 	test/integration/run-bats.sh $(TESTSUITE)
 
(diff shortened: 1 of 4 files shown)


dgageot:

dgageot:
FYI, I also tested this PR with `vmwarefusion` driver and it's working great.

jeanlaurent:
re-tested on another workstation with virtualbox, runs fine !

LGTM

nathanleclaire:
Mostly looks good, but that log driver one worries me.  I'm somewhat surprised to find that the daemon is not allowing `--log-driver=none` by default.  I'll investigate.

nathanleclaire:
I think that error message / return code change is just Docker client exhibiting erratic behaviors from version to version.  How about this (hopefully less likely to be broken in version changes) check instead:

```
[ docker $(machine config $NAME) inspect -f '{{.HostConfig.LogConfig.Type}}' $(docker $(machine config $NAME) ps -lq) == "none" ]
```

I'll go ahead and merge this and submit that as a followup.

LGTM

PR #2338, opened 2015-11-18
Revert "Remove systemd-related driver duplication"

dgageot:
This reverts commit fda08f5e15627facb3e2cfe8ff7b0a891ba2c876 because it breaks at least google driver if not all drivers based on the provisioner.


dgageot:

PR #2340, opened 2015-11-18
Global set for fish env

rcoedo:
Using `set -x` only exports the variable in the local scope.
This is a limitation when an user needs to invoke `docker-machine` in a fish script.

Using `set -gx` instead exports the variable globally.

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -98,7 +98,7 @@ func set(c CommandLine) error {
 
 	switch userShell {
 	case "fish":
-		shellCfg.Prefix = "set -x "
+		shellCfg.Prefix = "set -gx "
 		shellCfg.Suffix = "\";\n"
 		shellCfg.Delimiter = " \""
 	case "powershell":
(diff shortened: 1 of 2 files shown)


jeanlaurent:
@rcoedo This probably break [this integration test](https://github.com/docker/machine/blob/master/test/integration/core/env_shell.bats#L45-L52)

Those integration tests are not yet part of the build system, we're working on it, that's why you don't have any failed check here.

As a side note, as I'm not a fish expert, does the globally exported variable as any impact in the way we [unset](https://github.com/docker/machine/blob/master/commands/env.go#L140-L143) them ?

rcoedo:
@jeanlaurent I added some changes to fix the integration test too.

You can still unset global variables with `set -e`. The only thing that changes, as far as I know, is the scope of the variable. One is local to the begin/end and the other one isn't.

You can find more information about `set` [here](http://fishshell.com/docs/current/commands.html#set)

rcoedo:
As a side note, the use case that motivated me to request this change is having a fish function to select my docker machine, export it, and add docker.local with the machine's IP to /etc/hosts.

jeanlaurent:
LGTM

dgageot:
LGTM

PR #2345, opened 2015-11-19
Change log driver check to be more future-safe

nathanleclaire:

I'm wondering if this method might be more future-safe since the exit code and output of the `docker run` changed between versions.

@jfrazelle @dcalavera @icecrime Would be good to get some input from you all about what you think the most robust (future-compatible) way to verify the creation with `--log-driver` on the daemon happened successfully is.  Reasoning for changing this is that exit code and message from `docker run` changed on us from 1.8.3 => 1.9.0 (which is fine, just wondering which might be least likely to change).


dgageot:
LGTM

PR #2350, opened 2015-11-19
Add a version command

jeanlaurent:
Sounds like every other docker project has a `version` command, and we don't. I couldn't stand it anymore :smiley: , so here it is.


dgageot:
LGTM

PR #2359, opened 2015-11-19
Reject VirtualBox version<=3

dgageot:
Should we test the version of virtualbox and refuse to work with something different from `5.X`?

### DIFF
--- a/docs/drivers/virtualbox.md
+++ b/docs/drivers/virtualbox.md
@@ -10,7 +10,7 @@ parent="smn_machine_drivers"
 
 # Oracle VirtualBox
 Create machines locally using [VirtualBox](https://www.virtualbox.org/).
-This driver requires VirtualBox to be installed on your host.
+This driver requires VirtualBox 4+ to be installed on your host.
 
     $ docker-machine create --driver=virtualbox vbox-test
 
--- a/drivers/virtualbox/vbm.go
+++ b/drivers/virtualbox/vbm.go
@@ -75,3 +75,11 @@ func (v *VBoxCmdManager) vbmOutErr(args ...string) (string, string, error) {
 
 	return stdout.String(), stderrStr, err
 }
+
+func checkVBoxManageVersion(version string) error {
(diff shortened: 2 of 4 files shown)


dgageot:

nathanleclaire:
I'm in favor of supporting only >5

We should make sure to include a note about it in the documentation somewhere.

dgageot:
Thanks @nathanleclaire and @jeanlaurent, I have updated the documentation and improved the error message

jeanlaurent:
Are you sure aa5dc59 && especially 6468cb5 belongs to this PR ?

dgageot:
@jeanlaurent Sorry, push error

dgageot:
@docker/machine-maintainers As pointed out by @jeanlaurent, `virtualbox 5.0` was released in `july 2015`. Do we want to force every user with a `4.3` to upgrade. This might be difficult for some corporate users.

@jeanlaurent said we could reject `<4.3` and warn is `<5`

jeanlaurent:
There is two options:

1) is what @dgageot propose: reject `<4.3`  and warn `<5`
2) is to reject `<5` and allow to bypass it, if at the own risk of the user with some sort of dissuasive flag like `--BYPASS_OUTDATED_VBOX_VERSION_AT_MY_OWN_RISK=1` or something alike.

We need not to frustrate user who will not be in a position of upgrading properly their virtualbox version at the same time we should'nt invest too much time on legacy vbox version and better spent it on making it work flawlessly on the latest, but we must ensure that they use the latest one for the best usage of `docker-machine`

dgageot:
This issue https://github.com/docker/machine/issues/2371 shows that `VB5.0` can be ore buggy than `VB4.3`...

dgageot:
Updated the PR to accept only VirtualBox 4+

nathanleclaire:
LGTM

Maybe we can talk about ending support for older versions as 5 grows more mature.  But, I think it is still in the wild quite heavily, and we should hold off until we have more info.

PR #2364, opened 2015-11-20
Make user supplied input more obvious

elliot:
The command example given does not make it obvious that `name` is to be supplied by the user. Add square bracket notation around it to indicate the need to replace it.


dgageot:
Hi @elliot, you should sign your commits for the PR to be merged

dgageot:
Closing. Fixed in https://github.com/docker/machine/pull/2389

PR #2369, opened 2015-11-20
Detect multiple host only

dgageot:
Fail if duplicated host only interfaces are detected

### DIFF
--- a/drivers/virtualbox/network.go
+++ b/drivers/virtualbox/network.go
@@ -15,7 +15,8 @@ const (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


jeanlaurent:
LGTM

nathanleclaire:
Looking pretty good.  Ideally we would be able to report back to the user the names of the problematic `hostonlyif` and/or if any VMs are currently using them.  Maybe we can follow up on this with that.

nathanleclaire:
LGTM

PR #2374, opened 2015-11-20
Add more tests to the VirtualBox driver

dgageot:
+ Add more tests to the virtualbox driver
+ Make the host only interfaces output parsing more robust


jeanlaurent:
LGTM

PR #2375, opened 2015-11-20
Move vbox integration test to core

jeanlaurent:
This move all the vbox specific test into core so we can more easely run them.

At least as of now all the tests in core runs for virtualbox & fusion


dgageot:
@jeanlaurent you forgot to remove the virtualbox specific test in `drivers`

jeanlaurent:
thx, repushed

nathanleclaire:
Don't you think the `core/` directory will get quite messy adding these tests in?  What about tests for other drivers?  Ideally all of the core drivers should have integration tests, don't you think that sticking them all in one directory is likely to cause headaches?

dgageot:
Maybe create a subfolder per driver in core. The idea is to be able to run all its with one command without thinking too much about it

dgageot:
Also using a unique test command for every driver is way simpler. Less error prone

nathanleclaire:
OK.  Maybe we could change `force_env` to skip a test instead of failing it if it's not the desired driver, then always invoke the tests on the `test/integration` directory instead of `test/integration/core`?  If it's possible in BATS to do that, that seems like a good solution we could implement without needing to change the directory layout.

nathanleclaire:
Something like this:

screen shot 2015-11-20 at 2 03 07 pm

nathanleclaire:
I'll submit a patch for that.  It'll allow us to keep our current structure with only a ~2 lines change to the code.

jeanlaurent:
This is of course temporary. The core directory contains with this PR all the tests that run for virtualbox and fusion. I suggest we commit those in, and build slowly something once we get more tests that passes. rather than design something for which we don't have much data now

nathanleclaire:
@jeanlaurent If we could get the same result by changing a few lines in `force_env` and not having to move any files, why change the directory structure?

nathanleclaire:
FWIW the only test that failed for me invoking them all with that method was the RancherOS one and that is only because the Docker client-daemon are mismatched.

jeanlaurent:
I have put again the vbox specific test in their own directory, using a flat structure right now.

I really think the rancheros.bats & redaht.bats file should belong in the makefile in a way that we should run the all suite against all drivers. But this need to be discussed right now, so to keep this pr small, I just removed a folder to follow the flat structure.

dgageot:
LGTM. I'd like to go even further and cleanup the rancheros.bats and redhat.bats exceptions.
I'll follow with a PR.

nathanleclaire:
NOT LGTM

We could have just updated `force_env` and let everything else stay the same, so why the extra effort?  I'll follow up with a PR to show what I mean.

PR #2377, opened 2015-11-21
Revendor codegangsta cli

nathanleclaire:
For 0.5.0 we had to use a forked version of `codegangsta/cli` due to a `SkipFlagParsing` bug (so we didn't want to move it to a totally different repo and use `Godeps` properly), but the fix just got merged upstream so this sets everything back to normal with the master `codegangsta/cli`.


nathanleclaire:
The coveralls failure is normal since when we first introduced the semi-vendored change it bumped code coverage by mistakenly assuming `cli/` tests were ours.

dgageot:
LGTM

nathanleclaire:
:+1:

PR #2379, opened 2015-11-21
Add reference for HPE OneView plugin

wenlock:
HPE OneView plugin allows you to manage docker on
composable infrastructure using HPE OneView.
Creating a bare metal cloud is now super simple.

Signed-off-by: Edward Raigosa <edward.raigosa@hp.com>


dgageot:
LGTM. I didn't test the driver though. We might come up with a **tested** list in the future

wenlock:
@dgageot thanks!

PR #2382, opened 2015-11-21
Add VULTR to available driver plugins list

janeczku:


dgageot:
LGTM. I didn't test the driver though. We might come up with a **tested** list in the future

janeczku:
@dgageot much obliged! :smile:

PR #2388, opened 2015-11-23
Fix for issue 2370, show machine state after kill is executed #2386

### LINKED ISSUE #2370: kill does not return any message to the user
```
> docker-machine ls
NAME         ACTIVE   DRIVER       STATE     URL                         SWARM
default      *        virtualbox   Running   tcp://192.168.99.100:2376      
tobekilled   -        virtualbox   Running   tcp://192.168.99.128:2376   
> docker-machine kill tobekilled
> docker-machine ls
NAME         ACTIVE   DRIVER       STATE     URL                         SWARM
default      *        virtualbox   Running   tcp://192.168.99.100:2376   
tobekilled   -        virtualbox   Stopped                               
```
the user is forced to `docker-machine ls` to know how things are going for the killed machine

ksasi:
Hi,

This fixes #2370 "kill does not return any message to the user #2370". docker-machine kill displays machine(s) state after kill is executed.

Thanks...

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -73,7 +73,7 @@ func runAction(actionName string, c CommandLine, api libmachine.API) error {
 		return ErrNoMachineSpecified
 	}
 
-	if errs := runActionForeachMachine(actionName, hosts); len(errs) > 0 {
+	if errs := runActionForeachMachine(actionName, hosts); len(errs) > 0 && len(errs) == len(hosts) {
 		return consolidateErrs(errs)
 	}
 
@@ -81,6 +81,10 @@ func runAction(actionName string, c CommandLine, api libmachine.API) error {
 		if err := api.Save(h); err != nil {
 			return fmt.Errorf("Error saving host to store: %s", err)
 		}
+	}
+
+	for _, h := range hosts {
+		log.Infof("Successfully ran action %q on machine %q", actionName, h.Name)
 	}
 
 	return nil


nathanleclaire:
Thanks for the contribution @ksasi.  Unfortunately I have to say, this method of implementing it doesn't look like how I'd like to solve the issue.

Instead of adding extra code to this command, why don't we just echo out the name of the machine after an action has been performed on it?  It would then be similar to how `docker rm`, `docker start`, etc. work and we'd only have to update the code in one place.

dgageot:
Fully agree with @nathanleclaire. What we expect is feedback on what was done more than the list of remaining machines.

ksasi:
Thanks for the feedback @nathanleclaire and @dgageot . I will explore on how we can  just echo out the name of the machine after an action has been performed on it.

jeanlaurent:
@ksasi Something like `Machine 'target' has successfully been killed` is sufficient

nathanleclaire:
It should be sufficient to do something like:

```go
log.Infof("Successfully ran action %q on machine %q", action, host.Name)
```

in `runAction` before returning `nil`.

ksasi:
@nathanleclaire adding log.Info before returning nil in runAction helped. log.Info doesn't seem get executed in case of multiple arguments i.e. docker-kill default tobekilled where default in "Running" and tobekilled already "stopped" unless len(errs) == len(hosts) is also included as below 

if errs := runActionForeachMachine(actionName, hosts); len(errs) > 0 && len(errs) == len(hosts) {
		return consolidateErrs(errs)
	}

dgageot:
Hi @ksasi the PR contains two commits that are not related to your work. You should try to rebate your commit on top of master branch

ksasi:
Hi @dgageot ,

Not returning an error if len(errs) != len(hosts)

$ ./docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL                         SWARM                ERRORS
default       -        virtualbox   Running   tcp://192.168.99.101:2376
tobekilled    -        virtualbox   Stopped

$ ./docker-machine kill default tobekilled
Machine "tobekilled" is already stopped.

Sasikanths-MacBook-Pro-2:bin sasikanthkotti$ ./docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL   SWARM                ERRORS
default       -        virtualbox   Stopped
tobekilled    -        virtualbox   Stopped

As shown above,it appears when we are using docker kill for multiple machines with different statuses, returning an error is causing to
display details of few machines.

Returning an error only if len(errs) == len(hosts)

$ ./docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL                         SWARM                ERRORS
default       -        virtualbox   Running   tcp://192.168.99.101:2376                              
tobekilled    -        virtualbox   Stopped

$ ./docker-machine kill default tobekilled
Successfully ran action "kill" on machine "default"
Successfully ran action "kill" on machine "tobekilled"

$ ./docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL   SWARM                ERRORS
default       -        virtualbox   Stopped
tobekilled    -        virtualbox   Stopped


Maybe ideal solution would be identify the machine(s) returning error and exclude them from the loop printing the status
and returning error after the loop.

Please advise.

dgageot:
Hey @ksasi, I think the original issue was not clear. I'm going to fix it here: https://github.com/docker/machine/pull/2645
I might have to close this PR afterwards. Sorry and thanks for the contribution

dgageot:
Fixed with https://github.com/docker/machine/pull/2645

PR #2391, opened 2015-11-23
Update Google Driver dependencies

dgageot:


dgageot:

nathanleclaire:
Probably minor issue, but does `metaDataValue` need to pass an address instead of just the string value?

Other than that, LGTM

dgageot:
Yes it does!

nathanleclaire:
Seems kind on add API but OK.  LGTM

dgageot:
Yes, their API is odd but de can't do anything about it. Don't you think so?

nathanleclaire:
Yeah there's nothing we can do.  I'm sure there's a reason it's that way.  Doesn't bother me too much was just curious.

PR #2393, opened 2015-11-23
Clarifying  description of hyperv-disk-size

afscrome:
Making clear the value is a maximum for a dynamic disk not the actual size of a fixed size disk.


dgageot:
LGTM

PR #2406, opened 2015-11-24
Add interfaces for command testing, and unit test for `docker-machine env`

nathanleclaire:
Contains interfaces for mocking `CommandLine` and `libmachine.API` as well as a test for most code in `docker-machine env`.

### DIFF
--- a/libmachine/persist/filestore_test.go
+++ b/libmachine/persist/filestore_test.go
@@ -3,15 +3,28 @@ package persist
 import (
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/docker/machine/commands/mcndirs"
-	_ "github.com/docker/machine/drivers/none"
+	"github.com/docker/machine/drivers/none"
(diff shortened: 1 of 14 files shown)


nathanleclaire:
I'm not 100% crazy about the way `ConnChecker`, `UsageHintGenerator`, and interaction with environment variables is handled, but it seems a solid start until we can build proper structs  from the component pieces.

libmachine/persist/filestore_test.go
```diff
@@ -3,16 +3,29 @@ package persist
...
 
 	"github.com/docker/machine/commands/mcndirs"
-	_ "github.com/docker/machine/drivers/none"
+	"github.com/docker/machine/drivers/none"
+	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/hosttest"
+	"github.com/stretchr/testify/assert"
 )
 
+const (
```
nathanleclaire: Hm, I think these updates might not be needed anymore.  I'll check.

dgageot:
LGTM. Let's improve it step by step

Issue #2155, opened 2015-11-04
Eliminate RawDriver from Host struct / config.json


nathanleclaire:
Unfortunately, I realized a bit too late that there are a couple of issues with this method. Originally it was intended to be the "preserved" state of the Driver between Machine runs, since the client side (which handles persistence to disk) doesn't know which interface is being used on the server side.

But now I have two main concerns with it having used and distributed it a while:

1. Users may not realize that the `[]byte` is only a very thin layer of obfuscation and still contains trivially decodable secrets, so I worry that they'll post `docker-machine inspect `output that contains secrets without realizing it.
2. Some users (including myself) are accustomed to being able to change the values in the `Driver` JSON object, and have them affect the result of the run program.  This is not possible with 0.5.0 today and makes things like manually rotating AWS keys in the JSON (which, while not advised, is being done) difficult to impossible.

cc @docker/machine-maintainers

charlesg3:
Out of curiosity -- are the AWS credentials being stored in this blob?

nathanleclaire:
@charlesg3 Yes -- it's the reason why editing the config manually to update the credentials no longer works.

charlesg3:
Ah, bummer. Can you describe any workflow that would allow updating the creds?

nathanleclaire:
Hm, I know this is a terrible hack (and really we should fix this since I think manually editing configs is a valid use case), but until we roll out a solution you might be able to paste your desired driver JSON into the JSON literal in this Go program (probably as a Go file running locally): http://play.golang.org/p/JTzFal8zfJ and run the code to see the what you should put in the `RawDriver` field.  That might help until we can get a proper fix.  Does it make sense?

(There's probably a more elegant way to do this in Go, but this closely mimics what's actually going on in the actual Machine code).

nathanleclaire:
Really sorry about this, and the issue is slated to be worked on for 0.5.1, which is scheduled to be released in about two weeks' time.

charlesg3:
I'm willing to give that a shot for now. Happy to help try to work around these issues as things progress. There's huge potential for docker-machine to help our workflow but I think we have a particular usage pattern that may not be the usual.

nathanleclaire:
Seems not too unusual to me so I'm happy to help.  In fact in the future I'd like to make the type of secret management you're doing by hand today a little more easy to do automatically.

charlesg3:
[...]

PR #2419, opened 2015-11-26
Remove RawDriver from persistence on disk

nathanleclaire:
Closes https://github.com/docker/machine/issues/2155

I noticed that with a few simple changes, we can remove `RawDriver` from being persisted on disk entirely.  As a bonus, we can get rid of some code too.

Please give this one a thorough spin, and if it seems sound it would be great to get in for 0.5.2.  If it seems problematic, no need to rush it.

I've tried to test it as thoroughly as possible given that changing migration-related code can be a little dangerous.

### DIFF
--- a/libmachine/host/host.go
+++ b/libmachine/host/host.go
@@ -30,7 +30,7 @@ type Host struct {
 	DriverName    string
 	HostOptions   *Options
 	Name          string
-	RawDriver     []byte
+	RawDriver     []byte `json:"-"`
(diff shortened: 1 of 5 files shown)


nathanleclaire:

nathanleclaire:
There's some known wonkiness with docs CI, I've pinged @SvenDowideit to look into it.

dgageot:
LGTM

PR #2421, opened 2015-11-26
Update DRIVER checking to universal function

nathanleclaire:
It can be used for all environment variables and can be used at the whole test file as well as the individual test level.

### DIFF
--- a/test/integration/core/inspect_format.bats
+++ b/test/integration/core/inspect_format.bats
@@ -23,6 +23,7 @@ load ${BASE_TEST_DIR}/helpers.bash
 }
 
 @test "$DRIVER: check .Driver output is not flawed" {
+  only_if_env DRIVER virtualbox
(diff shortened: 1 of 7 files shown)


test/integration/core/inspect_format.bats
```diff
@@ -23,7 +23,8 @@ load ${BASE_TEST_DIR}/helpers.bash
 }
 
 @test "$DRIVER: check .Driver output is not flawed" {
+  only_if_env DRIVER virtualbox
```
nathanleclaire: Note the motivation for this change -- `SSHUser` is not consistent across all drivers. if our driver is `amazonec2` the default `SSHUser` will be `ubuntu`, for instance.

nathanleclaire:

dgageot:
LGTM

jeanlaurent:
LGTM

PR #2424, opened 2015-11-26
Move doubtful provision tests out of the integration folder

jeanlaurent:
This moves out of the way the doubtful so called provision tests out of the integration folder.

This enables easier integration testing by launching : `DRIVER=virtualbox make test-integration test/integration`

I really think they need to be removed, but just in case someone disagree on it.


dgageot:
LGTM

PR #2425, opened 2015-11-26
B2D_CACHE can be used to speedup integration tests

dgageot:


jeanlaurent:
A Pro Tip is often wise.

LGTM I say.

PR #2427, opened 2015-11-26
add debug message in dockerDaemonResponding function due to helping f…

newpcraft:
…or troubleshooting of docker daemon state

It might be difficult to know why it was failed and what command was executed, instead, we only can see the following message:

```
WARNING >>> Error getting SSH command to check if the daemon is up: exit status 1
```

It shows an output message of the SSH command like the following if the user use ```--debug``` option of the ```docker-machine```.

```
WARNING >>> Error getting SSH command to check if the daemon is up:
Client:
 Version:      1.9.1
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   a34a1d5
 Built:        Fri Nov 20 13:25:01 UTC 2015
 OS/Arch:      linux/amd64
Cannot connect to the Docker daemon. Is the docker daemon running on this host?
 : exit status 1
```

The message may be helped for troubleshooting as I was.

### DIFF
--- a/libmachine/provision/arch.go
+++ b/libmachine/provision/arch.go
@@ -72,8 +72,11 @@ func (provisioner *ArchProvisioner) Package(name string, action pkgaction.Packag
 }
 
 func (provisioner *ArchProvisioner) dockerDaemonResponding() bool {
-	if _, err := provisioner.SSHCommand("sudo docker version"); err != nil {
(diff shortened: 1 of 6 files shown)


dgageot:
This code could really be extracted to a helper function.
Still, LGTM

jeanlaurent:
LGTM

PR #2434, opened 2015-11-27
Check machine after create

dgageot:
Check that docker is running just after Create. This is to avoid the case where a machine is created
and `docker-machine env` fails just after that.

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -2,38 +2,12 @@ package commands
 
 import (
 	"fmt"
-	"net/url"
 	"os"
-	"strings"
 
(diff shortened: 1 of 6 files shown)


jeanlaurent:
LGTM

nathanleclaire:
Looks pretty good.  I think I would prefer `conncheck` as the module name since there is _a lot_ of different meanings that `check` could have.

Medium-term, let's work on defining the different things that could potentially go wrong with a check.  E.g., tell the user if it was impossible to dial the host IP.

LGTM

PR #2435, opened 2015-11-27
Lint all Markdown documents

dgageot:


jeanlaurent:
This needs to be done now, if we don't do this now all subsequent markdown modifications will fail.

LGTM

Issue #2355, opened 2015-11-19
Driver 'none' should work when a non-TLS URL is provided


gordontyler:
I have a Docker host configured with insecure access on port 2375. I can access it using `docker -H torneosth1:2375 ps` with no other DOCKER_* env vars set. I would like to register this host in docker-machine to make it easier to switch between various hosts.

However, after registering the host using the 'none' driver (`docker-machine create -d none --url tcp://torneosth1:2375 torneosth1`), when I try to make it the active machine, it fails with the following:

```
Error running connection boilerplate: Error checking and/or regenerating the certs: There was an error validating certificates for host "torneosth1:2375": open C:\Users\GTyler\.docker\machine\machines\torneosth1\server.pem: The system cannot find the file specified.
You can attempt to regenerate them using 'docker-machine regenerate-certs name'.
Be advised that this will trigger a Docker daemon restart which will stop running containers.
```

It would appear that docker-machine assumes that all hosts are secured with TLS. This host is on an internal network and will never be Internet-accessible -- it doesn't need TLS. The 'none' driver should accept an option to indicate whether the host should be accessed using TLS.

gordontyler:
As a corollary, it would be useful if the 'none' driver also accepted options providing paths to the certificates needed for a TLS connection, so that docker-machine could install the files in the correct location.

dgageot:
Duplicate of https://github.com/docker/machine/issues/1532

Issue #1532, opened 2015-07-15
Cannot specify --tlsverify=false


mikedougherty:
I have some hosts that are swarmed together, and I'm trying to add the master to machine with `-d none`. Everything has worked as expected so far, but the last bit I'm stuck on is that there's no way to specify the equivalent of `--tlsverify=false` or `DOCKER_TLS_VERIFY=`. This is my current workaround:

```bash
$ eval $( docker-machine env --swarm aws-stage | sed 's/"1"//g' )
```

Being able to specify this option during creation would be great; just having a field in the machine's config.json that I can manually edit is fine too.

* * * 
rationale:
The server cert is generated by our infra team and is for a particular domain, but the hosts do not have DNS names assigned. So, I get the following outcome when TLS verification is enabled:
```bash
$ docker ps
FATA[0000] An error occurred trying to connect: Get https://10.128.47.21:3376/v1.18/containers/json: x509: cannot validate certificate for 10.128.47.21 because it doesn't contain any IP SANs
```

ehazlett:
Could you try `--engine-tlsverify=false`?

dobesv:
$ docker-machine --version
    docker-machine version 0.4.0 (9d0dc7a)
    $ docker-machine create localhost -d none --url unix:///var/run/docker.sock --engine-tlsverify=false
    .. prints usage, does nothing ...

dobesv:
See also https://github.com/docker/machine/issues/1435

dgageot:
Hi @dobesv, the problem you are having is that before 0.5.0, trailing options (after the machine name) would be ignored without a warning. Please try `docker-machine create -d none --url unix:///var/run/docker.sock --engine-tlsverify=false` localhost`. Even better, upgrade to 0.5.0

dgageot:
@mikedougherty any update on this issue? Did you try 0.5.0?

mikedougherty:
Hi all, sorry, I've been very busy lately. I will give this a try ASAP. Thanks for the feedback so far

aaronfeng:
There's no option called `--engine-tlsverify`.  I also tried `--engine-opt tlsverify=false` and `--engine-env tlsverify=false`  `DOCKER_TLS_VERIFY` is always set to `1`.

nathanleclaire:
`TlsVerify` on the created daemons today is always set to true.

mikedougherty:
[...]

Issue #1435, opened 2015-06-28
Managing local unix-socket based docker via docker-machine


jleskovar:
Hi there,

For the sake of consistency, I'd like to have my local docker be managed via docker-machine. When I try to create and use my local docker-machine using the 'none' driver:

```
docker-machine create -d none --url=unix:///var/run/docker.sock localhost
eval "$(docker-machine env)"
```

I end up with: ```FATA[0000] Couldn't read ca cert /home/james/.docker/machine/machines/localhost/ca.pem: open /home/james/.docker/machine/machines/localhost/ca.pem: no such file or directory ```

To get around this, I have to manually unset DOCKER_TLS_VERIFY just for my localhost docker, which is kind of annoying. Are there plans to add the ability to disable TLS when creating a docker-machine using -d none?

TeckniX:
What about just adding the host without specifying the driver?
`docker-machine create --url=unix:///var/run/docker.sock localbox`

jleskovar:
Adding the host without specifying the driver is the same as -d none, AFAIK...
Just to clarify, I get CA cert error when I attempt to run any docker commands after running the eval -- e.g. docker ps.

The problem is that since DOCKER_TLS_VERIFY=1 gets set as part of the eval, the client will attempt to use TLS to talk to the local daemon via the unix socket, which fails as I haven't setup any certs for my local daemon.

nathanleclaire:
I'm a fan of detecting any hosts listening at `/var/run/docker.sock` and displaying them in `docker-machine ls` by default, so maybe we can make this work.

dkirrane:
Any workaround for this? I cannot use docker virtualbox machines on windows at all due to `certificate signed by unknown authority`
```
docker run hello-world
Unable to find image 'hello-world:latest' locally
Pulling repository docker.io/library/hello-world
Error while pulling image: Get https://index.docker.io/v1/repositories/library/hello-world/images: x509: certificate signed by unknown authority
```

dobesv:
Normally if you are managing local stuff you can just not use docker-machine.  I wanted to use PyCharm with docker, and PyCharm only manages docker via docker-machine.  In this case I could work around it, to some degree, by writing a fake docker-machine shell script that emulated docker-machine for local docker.

martynsmith:
@dobesv - could I grab a copy of that script, I have exactly the same use case :-)

dobesv:
@martynsmith 

[...]

PR #2437, opened 2015-11-27
Rename "none" driver to "test"

### LINKED ISSUE #2270: `none` driver is broken
## Synopsis

I have several manually configured docker servers that I added to my `docker-machine` inventory using the `none` driver:

```
# This is using v0.4.1
$ docker-machine create --driver=none --url=tcp://myhost.domain.tld:2376 myhost
# Then copy certificates and keys to various places under ~/.docker/machine/machines
```

I can then switch between a local VM image and my staging/production servers, as expected

```
$ eval "$(docker-machine env myhost)"
$ docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL                                         SWARM
myhost      *        none                   tcp://myhost.domain.tld:2376   
```
 
After upgrading to docker-machine 0.5.0, when I run `env` I get this:

```
$ docker-machine -D env myhost
Docker Machine Version: 0.5.0 (HEAD)
Found binary path at /usr/local/bin/docker-machine-driver-none
Launching plugin server for driver none
Plugin server listening at address 127.0.0.1:64831
Using API Version 1
(devops01) Calling RpcServerDriver.GetState
Error running connection boilerplate: myhost is not running. Please start it in order to use the connection settings
```

If I try recreating the entry I still get the same error.

Would love some ideas on how to further debug this aside from rolling back to 0.4.1.

## Client
```
$ uname -a
Darwin mymac 13.4.0 Darwin Kernel Version 13.4.0: Wed Mar 18 16:20:14 PDT 2015; root:xnu-2422.115.14~1/RELEASE_X86_64 x86_64 
```

```
$ docker version
Client:
 Version:      1.9.0
 API version:  1.21
 Go version:   go1.5.1
 Git commit:   76d6bc9
 Built:        Tue Nov  3 21:17:48 UTC 2015
 OS/Arch:      darwin/amd64
```

## Server

```
$ uname -a
Linux myserver 3.10.0-229.14.1.el7.x86_64 #1 SMP Tue Sep 15 15:05:51 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ docker version
Client:
 Version:      1.9.0
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   76d6bc9
 Built:        Tue Nov  3 18:00:05 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.9.0
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   76d6bc9
 Built:        Tue Nov  3 18:00:05 UTC 2015
 OS/Arch:      linux/amd64
```

jeanlaurent:
Following @ehazlett advice [here](https://github.com/docker/machine/issues/643#issuecomment-76769741) this PR rename the none driver to test, and remove it from the release package, set the `generic` driver as default driver and remove the documentation related to it. We keep using it to run the various integration test, where it is of great help in allowing a faster execution time.

The driver `none` in it's current state is not usable. The hardcoded ssh user is a blank string the ssh port is 0, and those can not be overriden.

The main purpose of this driver,or at least how it was partially documented to, is to register an existing running docker instance, which is what the `generic` driver do.

This will close #2270, #2355, #1532, #1435 and #643

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -16,7 +16,7 @@
 		"github.com/docker/machine/drivers/generic",
 		"github.com/docker/machine/drivers/google",
 		"github.com/docker/machine/drivers/hyperv",
-		"github.com/docker/machine/drivers/none",
+		"github.com/docker/machine/drivers/test",
 		"github.com/docker/machine/drivers/openstack",
 		"github.com/docker/machine/drivers/rackspace",
 		"github.com/docker/machine/drivers/softlayer",
--- a/circle.yml
+++ b/circle.yml
@@ -33,7 +33,7 @@ test:
         pwd: $BASE_STABLE
 
   override:
-    - DRIVER=none test/integration/run-bats.sh test/integration/cli:
+    - DRIVER=test test/integration/run-bats.sh test/integration/cli:
         pwd: $BASE_STABLE
         timeout: 600
     # - DRIVER=virtualbox test/integration/run-bats.sh test/integration/core:
--- a/cmd/machine-driver-test.go
+++ b/cmd/machine-driver-test.go
@@ -1,10 +1,10 @@
 package main
 
 import (
-	"github.com/docker/machine/drivers/none"
+	"github.com/docker/machine/drivers/test"
 	"github.com/docker/machine/libmachine/drivers/plugin"
 )
 
 func main() {
-	plugin.RegisterDriver(none.NewDriver("", ""))
+	plugin.RegisterDriver(test.NewDriver("", ""))
 }
--- a/commands/create.go
+++ b/commands/create.go
@@ -38,7 +38,7 @@ var (
 			Usage: fmt.Sprintf(
 				"Driver to create machine with.",
 			),
-			Value: "none",
+			Value: "generic",
 		},
 		cli.StringFlag{
 			Name:   "engine-install-url",
--- a/docs/get-started-cloud.md
+++ b/docs/get-started-cloud.md
@@ -85,16 +85,19 @@ To remove a host and all of its containers and images, use `docker-machine rm`:
     $ docker-machine ls
     NAME      ACTIVE   DRIVER       STATE     URL
 
-## Adding a host without a driver
 
-You can add a host to Docker which only has a URL and no driver. Therefore it
-can be used an alias for an existing host so you don’t have to type out the URL
-every time you run a Docker command.
+## Adding an existing host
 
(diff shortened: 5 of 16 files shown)


dgageot:
I would add that everything that targets an existing host is done with the `generic` driver. If this driver is not good enough (I'm thinking documentation, supporting non TLS, not restarting the docker daemon...), then we should create issues for the `generic` driver.

jeanlaurent:
This is ready for review.

dgageot:
LGTM

jeanlaurent:
We don't want to push this without you taking a glance @nathanleclaire.But this would help for the release since end the 'none' driver misunderstanding.

dnephin:
Isn't the `none` driver the only way to support a daemon without SSH ? This is really useful for anyone that has local socket, or has already provisioned some instances and just wants a way to add them to machine.

It would be good to add this support to some other driver before removing what little support exists.

cc @mikedougherty

nathanleclaire:
Yes, `none` is and always has been targeted at exactly the use case @dnephin mentions: Adding an existing instance of the Docker daemon to be managed in Machine's store with necessarily needing SSH access, provider credentials, etc.

@jeanlaurent Can we please open a thread and discuss what's to be done about the `none` driver before making any sudden moves like this?

My feeling is that the communication with the community and existing users of the `none` driver has not been thorough.  I'd prefer to chat more with folks and get everyone on a similar page before dropping support for the feature completely, and/or redirecting it to a different expected workflow.

nathanleclaire:
Note that Docker's existing [feature deprecation policy](https://docs.docker.com/engine/misc/#feature-deprecation-policy) indicates that such an action (deprecating `none` in favor of `generic`) is usually expected to take a timespan of 2 releases.

dgageot:
@nathanleclaire I'm wondering when `none` driver has ever worked. Do you have an idea? I understand that people complain if we remove something that works. But something that never really worked. cc @ehazlett

mikedougherty:
@dgageot My understanding is that it worked prior to 0.5.0. #2270 has some more context.

Even if it never worked, I would personally rather see the feature fixed than removed, per @dnephin's comment:
> This is really useful for anyone that has local socket, or has already provisioned some instances and just wants a way to add them to machine.

metasim:
@dgageot `none` worked in 0.4.1. Have had to roll back to regain access to my infrastructure. :-(

jeanlaurent:
Closing will, re-open if needed by #2667

PR #2441, opened 2015-11-27
Display driver in error without crashing

jeanlaurent:
This will prevent `docker-machine ls` to crash with no explanation whenever you have a host in your configuration that is broken.

This is super useful, if you had a driver in your path and it has been removed.

step to reproduce.

1. Have an old release ( 0.5.0 for instance ) of docker machine installed in your path through toolbox
2. Setup a machine with the `none` driver for instance.
3. build and remove the none driver `make clean build && rm bin/docker-machine-driver-none`
4. setup your path to pickup your locally build binaries `export PATH=$(pwd)/bin:$PATH`
5. run `docker-machine ls`

docker-machine will pick-up the none driver from your path, and it will crash when run with the new client.

Final result ( check foo2 ) :
```
> docker-machine ls
NAME          ACTIVE   DRIVER       STATE     URL                         SWARM   ERRORS
default       *        virtualbox   Running   tcp://192.168.99.100:2376           
foo           -        generic      Running   tcp://178.32.221.192:2376           
foo2          -        not found    Error                                         rpc: can't find service RPCServerDriver.GetVersion
foo3          -        generic      Running   tcp://178.32.222.191:2376           
fromhome      -        virtualbox   Stopped                                       
tobekilled    -        virtualbox   Running   tcp://192.168.99.164:2376           
tobekilled3   -        virtualbox   Running   tcp://192.168.99.190:2376       
```

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -18,12 +18,12 @@ func cmdActive(c CommandLine, api libmachine.API) error {
 		return errTooManyArguments
 	}
 
-	hosts, err := persist.LoadAllHosts(api)
+	hosts, hostsInError, err := persist.LoadAllHosts(api)
 	if err != nil {
 		return fmt.Errorf("Error getting active host: %s", err)
 	}
 
-	items := getHostListItems(hosts)
+	items := getHostListItems(hosts, hostsInError)
 
 	for _, item := range items {
 		if item.Active {
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -64,9 +64,14 @@ func (c *contextCommandLine) Application() *cli.App {
 }
 
 func runAction(actionName string, c CommandLine, api libmachine.API) error {
-	hosts, err := persist.LoadHosts(api, c.Args())
-	if err != nil {
-		return err
+	hosts, hostsInError := persist.LoadHosts(api, c.Args())
+
+	if len(hostsInError) > 0 {
+		errs := []error{}
+		for _, err := range hostsInError {
+			errs = append(errs, err)
+		}
+		return consolidateErrs(errs)
 	}
 
 	if len(hosts) == 0 {
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -49,7 +49,7 @@ func cmdLs(c CommandLine, api libmachine.API) error {
 		return err
 	}
 
-	hostList, err := persist.LoadAllHosts(api)
+	hostList, hostInError, err := persist.LoadAllHosts(api)
 	if err != nil {
 		return err
 	}
@@ -81,7 +81,7 @@ func cmdLs(c CommandLine, api libmachine.API) error {
 		}
 	}
 
-	items := getHostListItems(hostList)
+	items := getHostListItems(hostList, hostInError)
 
 	for _, item := range items {
 		activeString := "-"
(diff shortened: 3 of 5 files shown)


thaJeztah:
Nice!

dgageot:
LGTM

jeanlaurent:
Ping @nathanleclaire.

This would be a great addition to the 0.5.2 release.

nathanleclaire:
I don't understand the problem that this PR solves so I've moved it to the 0.5.3 release milestone.

I followed the instructions stated above and I'm not sure what they're intended to be illustrating.  If I have a version of Docker Machine installed, create a machine with `none` driver, then build and `rm` `bin/docker-machine-driver-none`, nothing has changed.  What am I missing here?

nathanleclaire:
Side note: If we're shipping a client that is known to be broken with older versions of the RPC API that's quite problematic.

nathanleclaire:
As far as implementation goes: Why not just add an `Error` field to `Host` struct that we access later instead?  It will allow us to keep the method signatures the same way they are now and will avoid creating additional garbage for the GC to pick up.

jeanlaurent:
To ease your understanding, check the second line of the description : " This is super useful, if you had a driver in your path and it has been removed."
Meaning you have probably an old driver in your path still called rather than the new one, since it has been deleted.

I updated the steps to be more precise so that you can reproduce it locally.

dgageot:
@nathanleclaire Adding an `Error` field to `Host` seems very odd. The rest of the code wouldn't know how to deal with this field. And the GC... I don't see how such a small thing can be a problem for the GC

nathanleclaire:
@jeanlaurent If a driver is removed today everything is fine already, so what additional gain does this win us?

```console
$ docker-machine create -d none --url foo.bar foobar
Running pre-create checks...
Creating machine...
To see how to connect Docker to this machine, run: docker-machine env foobar
$ rm /usr/local/bin/docker-machine-driver-none
$ docker-machine ls
NAME      ACTIVE   DRIVER         STATE     URL                          SWARM   ERRORS
default   *        virtualbox     Running   tcp://192.168.99.100:2376
droplet   -        digitalocean   Running   tcp://159.203.249.243:2376
foobar    -        not-found      Error                                          Driver "none" not found. Do you have the plugin binary accessible in your PATH?
```

> Adding an Error field to Host seems very odd. The rest of the code wouldn't know how to deal with this field

@dgageot What do you mean?  Getting and setting struct fields is no problem.  And it could be passed through to the `HostListItem` directly by including `host.Err` as one of the possible pass-throughs here: https://github.com/docker/machine/blob/master/commands/ls.go#L259

> Adding an Error field to Host seems very odd. The rest of the code wouldn't know how to deal with this field. And the GC... I don't see how such a small thing can be a problem for the GC

Yes, it's less of a concern for me than adding more moving parts to the mix.

jeanlaurent:
@nathanleclaire yes I know that the **complete** driver removal is already handled.

This is covering the case were the driver **was** installed and it **no longer** distributed thus while youre using the latest `docker-machine` binary with and **old** driver still pickup from your path. As driver are pickup from the path and you can't really handle that, this PR allow to detect this tiny little corner case and avoid crashing badly.

dgageot:
Let's try to make some progress and look and the facts!

The feature and the code LGTM:
- The feature LGTM because it stops `docker-machine ls` from failing really badly when it encounters any unexpected error (bad communication with a driver, buggy driver, regression in the driver protocol...).
- The code LGTM because it doesn't pollute `Host` that is used everywhere. It just pollutes the calls to `LoadHosts` because it's where the errors take place.

The feature is useful, yet minor. It doesn't have a major impact on the design, so when it's rebased, I'm going to merge it right away.

As a side note, the issue that @jeanlaurent was facing is https://github.com/docker/machine/issues/2474. This PR doesn't fix the issue but makes it (and any other future similar case) bearable.

PR #2442, opened 2015-11-27
Add version to machine ls

dgageot:
ping @docker/machine-maintainers wdyt?


libmachine/host/version.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+
+	"github.com/docker/machine/libmachine/cert"
+	"github.com/samalba/dockerclient"
+)
+
+var DefaultDockerVersionner DockerVersionner = &BasicDockerVersionner{}
+
+type DockerVersionner interface {
```
nathanleclaire: Nitpick: `Versioner` would seem to be the correct way to spell this.

libmachine/host/host.go
```diff
@@ -149,6 +149,10 @@ func (h *Host) GetURL() (string, error) {
 	return h.Driver.GetURL()
 }
 
+func (h *Host) GetDockerVersion() (string, error) {
```
nathanleclaire: Nitpick: The idiomatic Golang convention would be to call this function simply `DockerVersion`.

libmachine/host/version.go
```diff
@@ -0,0 +1,44 @@
+package host
```
nathanleclaire: Why not move the code to get such a `DockerClient` struct to its own module, e.g. `dockerclient`?  There are other parts of the code such as the provisioners which are likely to want to use import this as well.

libmachine/host/version.go
```diff
@@ -0,0 +1,44 @@
...
+
+	version, err := client.Version()
+	if err != nil {
+		return "", fmt.Errorf("Unable to query docker version: %s", err)
+	}
+
+	return version.Version, nil
+}
+
+func dockerClient(host *Host) (*dockerclient.DockerClient, error) {
```
nathanleclaire: Why not move the code to get such a DockerClient struct to its own module, e.g. `mcndockerclient`? There are other parts of the code such as the provisioners which are likely to want to use import this as well.

commands/ls.go
```diff
@@ -232,11 +233,16 @@ func matchesName(host *host.Host, names []string) bool {
...
 	url := ""
 	hostError := ""
+	dockerVersion := "?"
 
 	currentState, err := h.Driver.GetState()
 	if err == nil {
 		url, err = h.GetURL()
 	}
+	if err == nil {
+		dockerVersion, err = h.GetDockerVersion()
```
nathanleclaire: If the Docker daemon is not running, won't this bit hang?  Why not wrap this code in its own timeout to avoid the catch-all timeout falsely reporting state as `state.Timeout` if it's actually `state.Running`?

Likewise, what will happen here if `state` is `state.Stopped`?  Will it hang?

nathanleclaire:
Some general comments:

- Isn't it a bit worrying to have yet another column in `ls`?  What's our upper limit on number of columns?
- Do you think it would be helpful to report that the daemon is up as well as which version it is?  After all, if the daemon isn't up, we won't be able to make the API call to determine the version at all.

libmachine/host/host.go
dgageot: Like `GetURL`? ;-)

libmachine/host/version.go
dgageot: Good idea. This could be done when we need it then.

libmachine/host/version.go
dgageot: Thanks. I fixed it

commands/ls.go
dgageot: This would print something like that:

```
NAME      ACTIVE   DRIVER       STATE     URL                         SWARM   VERSION   ERRORS
default   -        virtualbox   Running   tcp://192.168.99.100:2376                     Unable to query docker version: Get https://192.168.99.100:2376/v1.15/version: dial tcp 192.168.99.100:2376: getsockopt: connection refused
```

So it doesn't hang. Also, we call this piece of code only when the host is running and has a url.

dgageot:
Thank you @nathanleclaire for the feedback. I improved the code.

> Some general comments:

> - Isn't it a bit worrying to have yet another column in `ls`?  What's our upper limit on number of columns?

Yes, I don't know what the upper limit should be... `docker ps` shows 7 columns. We show 8 narrower columns. I think it's ok. We could hide columns such as `ERROR` and `SWARM` when they are empty. Is it worth the effort?

> - Do you think it would be helpful to report that the daemon is up as well as which version it is?  After all, if the daemon isn't up, we won't be able to make the API call to determine the version at all.

Sorry I don't understand...

libmachine/host/version.go
nathanleclaire: I would normally agree but I want to move the provisioning code which runs Swarm containers over to using `samalba/dockerclient` instead of shell ASAP.  It doesn't add any additional code, just moving into another module, so why not?

nathanleclaire:
> Yes, I don't know what the upper limit should be... docker ps shows 7 columns. We show 8 narrower columns. I think it's ok. We could hide columns such as ERROR and SWARM when they are empty. Is it worth the effort?

Yes I think this one is OK and arguably pretty important.  Don't think I'd like to hide columns (too inconsistent).

> Sorry I don't understand...

I mean, why not report if the daemon is down? Something like:

```console
$ docker-machine ls
NAME         ACTIVE   DRIVER         STATE     URL                          SWARM   DAEMON          ERRORS
b2dcompile   -        digitalocean   Running   tcp://104.236.195.179:2376           Stopped
dcompile     *        virtualbox     Running   tcp://192.168.99.166:2376            v1.10.0-dev
default      -        virtualbox     Running   tcp://192.168.99.100:2376            v1.9.1
droplet      -        digitalocean   Running   tcp://159.203.247.243:2376           v1.9.1
```

Also, as noted above, my preference would be for the column heading to be simply `DAEMON` (or maybe `DOCKER`?).  `VERSION` could mean a lot of things -- many things have versions.  What do you think?

libmachine/host/version.go
nathanleclaire: Also, can you please rename this file to something like `docker_versioner.go` instead?  We already have a couple of `version.go` files in the repository and they're usually just holders for static constants.

nathanleclaire:
Code looking good -- I have a few more points for feedback, largely aesthetic things.  Please address the ones that you agree with and then feel free to merge at your discretion.

jeanlaurent:
This is a great addition to the upcoming 0.5.3 release. 

LGTM

PR #2445, opened 2015-11-28
Put all core drivers in docker-machine binary

dgageot:
Let's package all core-drivers into the main `docker-machine` binary.
This has multiple advantages:

- Much faster build (8s vs 47s on my laptop)
- Much smaller distribution (15Mb vs 125Mb)
- One step towards being go gettable


dgageot:

dgageot:
Pushed in the wrong branch. Will recreate it

PR #2446, opened 2015-11-28
FIX 2444 Command scp being called twice

dgageot:


dgageot:
Pushed to the wrong branch. Will recreate it

PR #2447, opened 2015-11-28
Reference OVH Cloud driver

yadutaf:
As discussed at the DockerCon, here is an OVH Cloud driver for Docker Machine, based on the OpenStack driver.

It has been tested and confirmed to work using a fresh build from docker-machine master. It does not work with the latest released version. 

In the future, I may rework it to use directly OVH APIs instead of OpenStack.

And the good point: It's go get-able ;)


dgageot:
LGTM

jeanlaurent:
LGTM

PR #2448, opened 2015-11-28
Add the --virtualbox-host-dns-resolver flag

grampajoe:
This adds a new flag, `--virtualbox-host-dns-resolver`, that tells VirtualBox to start the VM with `--natdnshostresolver1 on` instead of `off`.

A common problem I've run into is getting DNS to work correctly in VirtualBox VMs when I'm connected to a VPN that does special magic to the host's DNS resolver. This makes it easier to set up a Docker host with working DNS without having to shut it down and run the `VBoxManage` command manually.

The integration test for this isn't great, because I couldn't find a way to inspect the `natdnshostresolverX` property. I'd love to hear any suggestions for how to properly test this! :rainbow: 

Signed-off-by: Joe Friedl <joe@joefriedl.net>

### DIFF
--- a/docs/drivers/virtualbox.md
+++ b/docs/drivers/virtualbox.md
@@ -26,6 +26,7 @@ Options:
 -   `--virtualbox-memory`: Size of memory for the host in MB.
 -   `--virtualbox-cpu-count`: Number of CPUs to use to create the VM. Defaults to single CPU.
 -   `--virtualbox-disk-size`: Size of disk for the host in MB.
+-   `--virtualbox-host-dns-resolver`: Use the host DNS resolver. (Boolean value, defaults to false)
 -   `--virtualbox-boot2docker-url`: The URL of the boot2docker image. Defaults to the latest available version.
 -   `--virtualbox-import-boot2docker-vm`: The name of a Boot2Docker VM to import.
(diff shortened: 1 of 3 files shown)


jeanlaurent:
This is working.

I believe the integration test does not bring anything new to the table.
I just spent a good deal of time trying to figure out how we can check that this flag is actually correctly set in the box, but the only place where I found it is in the log file.

For a machine named `grampajoe-resolver`
you can
`cat ~/.docker/machine/machines/grampajoe-resolver/grampajoe-resolver/Logs/VBox.log | grep UseHostResolver | grep -q '(1)'`
and check the output code with `echo $?`

I suggest you remove the not-that-useful integration test. If you really feel like it, you can use the code above to hack your way into a new integration test.

Either way I tested this works, and I'm ready to move forward with a merge once you took a decision on integration tests.

nathanleclaire:
My feeling is also that the integration test doesn't really add anything and we'll just have to wait until better unit testing is available for VirtualBox driver to test this option.

So, when the integration test is removed, I'm game to merge this.

grampajoe:
Thanks for the review! I removed the integration test and added a note about the boolean flag in the docs.

nathanleclaire:
LGTM

Issue #2130, opened 2015-10-30
Major install size regression 


ljrittle:
0.3.1 was about a 12MB install on OS X
0.4.1 was about the same
0.5.0-rc4 is about 138MB

Is it safe to delete drivers (or extract only those required during the install step) that will never be used for a given install site?

dmp42:
@ljrittle it is safe to remove the drivers you have never used and don't plan on using.

PR #2453, opened 2015-11-30
Put all core drivers in docker-machine binary

dgageot:
Let's package all core-drivers into the main docker-machine binary.
This has multiple advantages:

- Much faster build (8s vs 47s on my laptop)
- Much smaller distribution (15Mb vs 125Mb)
- One step towards being go gettable


jeanlaurent:
Sounds really good. Re-exec FTW.

LGTM

jeanlaurent:
I honestly don't know if we can put this in the 0.5.2 release. 
Moving it to 0.5.2 to sparkle discussion.

nathanleclaire:
Given that documentation (https://github.com/docker/machine/blob/master/docs/install-machine.md), release notes, etc. will need to be updated, and this is such a big change from the previous method, I'm going to go ahead and push the milestone on this out and really want to discuss and design it more thoroughly before doing it.

Some thoughts:

- Won't this create a difficult situation when we eventually take the non-core drivers such as `rackspace` out and they're managed by others?  Users won't be able to just download the new binaries in place of the old ones, they will have to upgrade their `docker-machine` as well.  Seems like lots of potential for timing problems there.
- Isn't this change really risky from the standpoint of possibly breaking non-core driver plugins?  If the core drivers have this special, privileged way of being accessed that other drivers don't share, aren't we a lot more likely to break the "second-class" method?
- On that same note, shouldn't any PR we make in this regard also introduce testing of one or several non-core drivers to make sure that we don't have a regression there?
- Though it will speed up time to compile everything, won't it slow down compiles when we just want to build just the main Machine binary?  `make machine` on my laptop builds in ~2 seconds -- `make build` on this PR builds in ~8 seconds.  I really like being able to compile different pieces very quickly.  It's possible to build just the plugin binaries too and they're also only 2-3s each.
- Can we just update the Makefile to only re-compile driver plugins when needed?  e.g. if and only if files in `rpcdriver` / `virtualbox` modules are changed, re-compile `docker-machine-driver-virtualbox`.  That will help with compile time a lot.

dgageot:
@nathanleclaire 

- Driver developers would compile a version of docker-machine that doesn't contain their driver as a core driver. I think it's basic enough.
- Let's see if we break something someday and then we can react.
- We could do that. Let's focus on testing the core drivers first. There's a lot to be done to improve our e2e tests.
- This PR make embedding `docker-machine` much more realistic. 125Mb is way too bug. Users have complained https://github.com/docker/machine/issues/2130, docker staff has complained cc @michaKFromParis. We have to fix that
- This also makes it trivial to test different versions of docker-machine without messing with the `PATH`

cmd/machine.go
```diff
@@ -133,6 +156,44 @@ func main() {
...
+	case "virtualbox":
+		plugin.RegisterDriver(virtualbox.NewDriver("", ""))
+	case "vmwarefusion":
+		plugin.RegisterDriver(vmwarefusion.NewDriver("", ""))
+	case "vmwarevcloudair":
+		plugin.RegisterDriver(vmwarevcloudair.NewDriver("", ""))
+	case "vmwarevsphere":
+		plugin.RegisterDriver(vmwarevsphere.NewDriver("", ""))
+	default:
+		fmt.Fprintf(os.Stderr, "Unsupported driver: %s\n", driverName)
```
nathanleclaire: Probably should just use `log.Fatal` here

nathanleclaire:
> Driver developers would compile a version of docker-machine that doesn't contain their driver as a core driver. I think it's basic enough.

No, I mean that when we move drivers like `rackspace` out, if the users have a version of `docker-machine` with the `rackspace` driver bundled, they'll have to update `docker-machine` (to get a version that _doesn't_ bundle `rackspace`) _and_ download the `docker-machine-driver-rackspace` binary.  Don't you feel that will be confusing?

> Let's see if we break something someday and then we can react.

Isn't that approach likely to result in a lot of broken code?

> This PR make embedding docker-machine much more realistic. 125Mb is way too bug. Users have complained #2130, docker staff has complained cc @michaKFromParis. We have to fix that

I agree that the excess disk space usage is bad.

I'm really skittish on this until we have better documentation around driver installation as well.  With the current installation method, it's clear to see that you are getting `docker-machine-driver-*` for each provider, and follow intuitively that installing new ones is done using the same method.  How are users supposed to know how to install plugins if some are bundled into the main binary and some are externally loaded?

That being said, the motivation for doing this is pretty reasonable, and I'm OK with merging it _if and only if_ we introduce regression testing for externally loaded plugins (something automated which is run in CI).

nathanleclaire:
This changes the release process as well so that document will need to be updated.

nathanleclaire:
@jeanlaurent and I discussed in standup today that, perhaps we can err on the side of merging this now, and then file issues for the remaining tasks.

nathanleclaire:
LGTM

PR #2468, opened 2015-12-01
Fix docker-machine config does not work on fish

### LINKED ISSUE #2415: On fish shell docker-machine config does not work
When running `docker-machine config <name>`, the output usually looks like:

```
--tlsverify --tlscacert="/Users/jake/.docker/machine/certs/ca.pem" --tlscert="/Users/jake/.docker/machine/certs/cert.pem" --tlskey="/Users/jake/.docker/machine/certs/key.pem" -H=tcp://192.168.99.100:2376
```

The primary use case for this feature is to use it in a command substitution, like this:

```
docker $(docker-machine config <name>) run ...
```

The trouble with space-separating is that not all shells perform word splitting on spaces by default. While `sh`, `bash`, and `zsh` split on all whitespace (`\s`, `\t`, or `\n`) by default, `fish` splits only on newlines. This is a design decision to help reduce script bugs caused by filenames containing spaces.

Since most common shells will split arguments on newlines (to my knowledge `ksh` and even `csh`/`tcsh` do this also), I think it's more portable to have `docker-machine config` output each argument separated by a newline instead of a space.

As a bonus, this makes grepping the output easier.

jeanlaurent:
Tiny PR to fix #2415 - Fix for `docker-machine config` broken on fish shell. Already covered by test

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -30,7 +30,7 @@ func cmdConfig(c CommandLine, api libmachine.API) error {
 
 	log.Debug(dockerHost)
 
-	fmt.Printf("--tlsverify --tlscacert=%q --tlscert=%q --tlskey=%q -H=%s",
+	fmt.Printf("--tlsverify\n--tlscacert=%q\n--tlscert=%q\n--tlskey=%q\n-H=%s\n",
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

PR #2473, opened 2015-12-02
Fixes #2451 - ensures ls filters `<key>` and `<value>` works case-insensitive

askb:
Fixes #2451 ensure filters `<key>` and `<value>` work when case-insensitive. Added tests to verify filters. Now the following filters should work properly. 

```
$ PATH="$PWD/bin:$PATH" bin/docker-machine ls --filter NAME=all*
NAME        ACTIVE   DRIVER   STATE     URL    SWARM   ERRORS
allstaged   -        none     Running   none           

$ PATH="$PWD/bin:$PATH" bin/docker-machine ls --filter dRiver=virtualBox
NAME   ACTIVE   DRIVER       STATE     URL   SWARM   ERRORS
dev    -        virtualbox   Stopped                 
dev9   -        virtualbox   Stopped                              

$ PATH="$PWD/bin:$PATH" bin/docker-machine ls --filter STATe=STOPPEd
NAME   ACTIVE   DRIVER       STATE     URL   SWARM   ERRORS
dev    -        virtualbox   Stopped                 
dev9   -        virtualbox   Stopped          
```

Signed-off-by: Anil Belur <askb23@gmail.com>

PR #2482, opened 2015-12-03
Now the returned `err` is being checked to ensure its `Nil`

askb:
Now the returned `err` is being checked to ensure its `Nil`

This is a follow up for PR #2473 


dgageot:
LGTM

PR #2483, opened 2015-12-03
Swarm container net should be host. If use bridge, sometime containers…

dongmx:
Swarm container network should be host. If use bridge, sometime containers ip addresses conflict.
For example, host A run a swarm agent C which ip is 10.10.0.1. Host B run a swarm agent D which ip is also 10.10.0.1. C and D all bridge to NICs which can connect to the switch. Then stop the iptables, conflict happend. With net=host is good. 

Each swarm agent or swarm agent master have to listen a port, no matter net=bridge or net=host. I think use net=host is more reasonable and can avoid potential ip address conflict. 

Signed-off-by: Mingxin Dong <dongmx@dmx.xyz>

### DIFF
--- a/libmachine/provision/configure_swarm.go
+++ b/libmachine/provision/configure_swarm.go
@@ -88,6 +88,7 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
 
 	swarmMasterCmdTemplate := `sudo docker run -d \
 --restart=always \
+--net=host \
 {{range .Env}} -e {{.}}{{end}} \
 --name swarm-agent-master \
 -p {{.Port}}:{{.Port}} \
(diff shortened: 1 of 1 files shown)


nathanleclaire:
@abronan @vieux @al What's your all's recommendations in this regard?

abronan:
This makes sense to me. We should use `host` networking if running swarm through the official Image. For the scenario mentioned above. We should also  update our docs on swarm to reflect that...

@vieux @al @dongluochen WDYT?

al:
You've got the wrong person @abronan ...

abronan:
My bad, sorry about that! I meant @aluzzardi (he is "*al*" almost everywhere so now we know why he uses aluzzardi on Github :smile:)

dongmx:
@abronan Thanks for your feedback. I can update the docs. So I should create another pull request or update this one ?

nathanleclaire:
Ahh lol sorry @al

nathanleclaire:
Since @abronan recommends, LGTM

dgageot:
LGTM

Issue #2485, opened 2015-12-03
In generic driver, if a permanent IP address and hostname has alread in /etc/hosts, docker-machine will insert a line "127.0.1.1 hostname" into /etc/hosts


dongmx:
I saw the source code, and find this comment in it. 
// ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

As the document says "For a system with a permanent IP address, that permanent IP address should be used here instead of 127.0.1.1." . But if there is already a permanent IP address and hostname in /etc/hosts, docker-machine with insert a line "127.0.1.1 hostname" into /etc/hosts which cause one hostname multi address.

I have fixed this issue and I will pull my code later.

dongmx:
I saw the source code, and find this comment in it. 
// ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

As the document says "For a system with a permanent IP address, that permanent IP address should be used here instead of 127.0.1.1." . But if there is already a permanent IP address and hostname in /etc/hosts, docker-machine with insert a line "127.0.1.1 hostname" into /etc/hosts which cause one hostname multi address. 

I have fixed this issue and I will pull my code later.

nathanleclaire:
Yeah, this line is somewhat problematic for a variety of reasons, and I've considered just removing it, but I'm curious to see what you're working on.

dongmx:
My pr is here https://github.com/docker/machine/pull/2486 :)

dongmx:
@nathanleclaire Thanks for your feedback. I have already create a pull request for this issue. And nobody comment this pull request, should I just wait or ping someone?

nathanleclaire:
@dongmx It's merged now ;D

PR #2486, opened 2015-12-03
issue #2485, In generic driver If a permanent IP address and hostname has alread in /etc/hosts

dongmx:
This comment in the source code. 
// ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

As the document says "For a system with a permanent IP address, that permanent IP address should be used here instead of 127.0.1.1."  
But if there is already a permanent IP address and hostname in /etc/hosts, docker-machine with insert a line "127.0.1.1 hostname" into /etc/hosts which cause one hostname multi addresses.

So I judge if the hostname has already in /etc/hosts, if true; do nothing, else do as usual.

Signed-off-by: Mingxin Dong <dongmx@dmx.xyz>


dgageot:
LGTM

nathanleclaire:
LGTM

@jeanlaurent ?

PR #2489, opened 2015-12-03
Replace custom logging system by logrus

jeanlaurent:
This PR Replace custom logging system by logrus.
This is the prerequiste to using bugsnag, having a per-machine logging system for debugging and modernize the way driver currently output logs.

Approach taken here :
- We use a single, simplified interface for all logging
- Each and every logs are currently redirected to logrus, no custom logs whatsoever.
- Remove more than 200 lines of code and add the logrus library used in other Docker projects.


libmachine/log/log.go
```diff
@@ -1,124 +1,70 @@
...
 func Warnf(fmtString string, args ...interface{}) {
-	l.Warnf(fmtString, args...)
+	logger.Warnf(fmtString, args...)
 }
 
-func WithField(fieldName string, field interface{}) Logger {
-	return l.WithFields(Fields{
-		fieldName: field,
-	})
+func GetStandardLogger() *logrus.Logger {
```
nathanleclaire: Probably can call this function simply `StandardLogger` as per Golang convention: https://golang.org/doc/effective_go.html#Getters

dgageot:
@jeanlaurent The vendoring is borked. Otherwise, the code looks good

jeanlaurent:
@dgageot Vendoring is unborked
@nathanleclaire I modified what you suggested

I modified the code to add testability, Ready for final review.

nathanleclaire:
LGTM

I have no idea what's going on with CI, it feels like Travis has been waiting on the build to start all day.

jeanlaurent:
About the travis/ci hang, it sounds like all tests are good on linux, but the darwin & win build have not started. Trying to trigger a new one somehow

dgageot:
LGTM

ahmetalpbalkan:
lol I totally wrote a driver using withfields everywhere and just realized it's gone. :)

ahmetalpbalkan:
@jeanlaurent I think I have an argument against this change: I was initializing a `lg := log.WithFields(...)` and I was calling `lg.Debug` far too many times, now with this change, this requires passing of log fields on every log statement –produces kinda bad code.

ahmetalpbalkan:
@jeanlaurent also I realized with this change, `log.Debugf` is simply doesn't exist. Consider this scenario:

```go
	lg := log.WithFields(r.LogFields())
	lg.Infof("Attempting to clean up %s resource...", resName)
```

let's say `resName = "IP Address" ` etc. I'll end up using fmt.Sprintf inside the log stmt. I think logrus got something right. Is there a way I can still use the logrus interface in the driver?

ahmetalpbalkan:
@jeanlaurent another one. logrus.WithFields takes a `map[string]interface{}`, the replacement methods are `map[string]string`... This is also increasing boilerplate code to convert everything strictly to string. My code looks a lot worse as I transition to the world without `WithFields()`.

ahmetalpbalkan:
Is this new `log.Debug(msg, map[string]string)` thread safe? Sometimes I'm getting a log like such as:

    (dev) %!(EXTRA map[string]string=map[vm:dev-firewall])Attempting to clean up Availability Set resource... map[name:docker-machine]

where `map[string]string=map[vm:dev-firewall]` actually belongs to an earlier statement and `map[name:docker-machine]` is for the current statement... My log line is simply:

     log.Info(fmt.Sprintf("Attempting to clean up %s resource...", r.ResourceType()), map[string]string{"name" : name})

Also this representation is 10x worse than logrus. Missing logrus so much already.

PR #2491, opened 2015-12-03
Revise SSH help flow and add 'docker-machine ssh' test

nathanleclaire:
Noticed a bug where a command like `docker-machine ssh default df -h` would trip up the help text.  This fixes it so that only happens if the first argument is a `--help` type flag, and adds the first test for `cmdSSH` to cover this case and others.


jeanlaurent:
LGTM

nathanleclaire:
@dgageot PTAL when you get a chance

nathanleclaire:
OK I've updated the test nice catch @dgageot

dgageot:
LGTM

Issue #2478, opened 2015-12-02
Machine should not use 'unsafe' characters in Openstack keypair


mstyne:
When executing `docker-machine create` using the `openstack` driver, machine will attempt to create a keypair with a name that uses the machine name concatenated with what appears to be a UUID/GUID.

For example, creating a machine `test.example.org` will result in the key name being:

`test.example.org--60bbb7c7c0a8...`

Unfortunately, it seems that Openstack does not consider periods to be valid ('unsafe') for the keypair name:

```
Error creating machine: Error in driver during machine creation: 
Expected HTTP response code [200] when accessing [POST https://my.openstack.endpoint.com], but got 400 instead
{"badRequest": {"message": "Keypair data is invalid: Keypair name contains unsafe characters", "code": 400}}
```
Reviewing the `nova` codebase, only underscores, hyphens, and alphanumeric characters are allowed for keypair names:

https://github.com/openstack/nova/blob/c44f711a0494db40ebbdacf85bfc3e7e2a357bf8/nova/compute/api.py#L3734

```python
    def _validate_new_key_pair(self, context, user_id, key_name, key_type):
        safe_chars = "_- " + string.digits + string.ascii_letters
        clean_value = "".join(x for x in key_name if x in safe_chars)
        if clean_value != key_name:
            raise exception.InvalidKeypair(
                reason=_("Keypair name contains unsafe characters"))
```

A possible solution for this would be to replace these spooky, unsafe characters with either a hyphen or an underscore, ex:

`test_example_org--60bbb7c7c0a8...`

```
✗ docker-machine -v
docker-machine version 0.5.2 ( 0456b9f )
```

nathanleclaire:
cc @ggiamarchi

dgageot:
Fixed with https://github.com/docker/machine/pull/2501

PR #2501, opened 2015-12-05
Sanitize keynames by replacing dots with underscores

fsoppelsa:
From https://github.com/docker/machine/issues/2478 makes sense only replace dots.
Probably for other separators, we can just print a message and Exit().


dgageot:
LGTM

jeanlaurent:
Would I loved to get some unit test for this.

LGTM

PR #2503, opened 2015-12-06
Fix detection of virtualization support on AMD CPU.

larskanis:
This also changes the amd test case to /proc/cpuinfo content of
a real AMD CPU.


larskanis:
Previous flag 'smd' is no valid flag in `/proc/cpuinfo` .

dgageot:
LGTM
@jeanlaurent do you know why we had `smd`?

jeanlaurent:
@dgageot I've just re-checked, we messed up the various vtx acronym.

LGTM

Issue #2490, opened 2015-12-03
Integration tests take too much time


dgageot:
The integration tests on VirtualBox take more than 20min on my machine.

Would you be ok if I changed the integration tests so that a single machine is shared across multiple tests files?
Of course some of the tests would still have to start a new distinct machine.

cc @docker/machine-maintainers

nathanleclaire:
I do really think it's a bit risky to start mixing results from different tests intertwined, but I understand the appeal.

How about this:

- We create a function in `helpers.bash`, something like `mcn_create`, which:
    - Creates the test machine if it doesn't exist already.
    - Skips creating the test machine if it _does_ exist already.

Then, we replace invocations of `machine create -d $DRIVER $NAME` with this function, but not anything which requires custom `create` flags.  Those will always run their own independent create.

And, it also includes support, to opt out of sharing machines by running with `NO_SHARE_MACHINES` environment or equivalent enabled.  That way, before releases and stuff, we can run it a few times that way to make sure the tests aren't depending on each other for their results.  Looks like these are the tests for which we could use such an approach:

```
test/integration/core/core-commands.bats
test/integration/core/env_shell.bats
test/integration/core/inspect_format.bats
test/integration/core/regenerate-certs.bats
test/integration/core/scp.bats
test/integration/core/ssh-backends.bats
test/integration/virtualbox/certs-checksum.bats
test/integration/virtualbox/pause-save-start.bats
```

How does that sound?

Also, how do you envision this working in tandem with the way that we uniquely tag the `NAME` using timestamp today?

PR #2511, opened 2015-12-07
Faster integration tests

dgageot:
Fixes https://github.com/docker/machine/issues/2490


test/integration/core/inspect_format.bats
```diff
@@ -24,7 +23,7 @@ load ${BASE_TEST_DIR}/helpers.bash
 
 @test "$DRIVER: check .Driver output is not flawed" {
   only_if_env DRIVER virtualbox
-  run docker-machine inspect -f '{{.Driver.SSHUser}}' $NAME
+  run machine inspect -f '{{.Driver.SSHUser}}' $NAME
```
nathanleclaire: Nice

nathanleclaire:
Looks great. Why not define a function in `helpers.bash`, e.g. `try_shared_machine` for the little `NAME` check instead of embedding it directly:

```bash
function try_shared_machine {
    if [[ -z "$NAME" ]]; then
        export NAME="$(unique_machine_name)"
    fi
}
```
Then each test would just have one `try_shared_machine` in front.

Would help to reduce duplication.

dgageot:
@nathanleclaire PTAL. I took your feedback into account.

nathanleclaire:
LGTM

jeanlaurent:
LGTM

PR #2513, opened 2015-12-07
Verbose ls

dgageot:
Following up the discussion on https://github.com/docker/machine/pull/1475, here's a version of `docker-machine ls` with a `verbose` flag.

The `DOCKER` and `ERROR` columns would be hidden by default and shown with this new flag.

@nathanleclaire, @thaJeztah, @hairyhenderson wdyt?

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -22,7 +22,7 @@ func cmdActive(c CommandLine, api libmachine.API) error {
 		return fmt.Errorf("Error getting active host: %s", err)
 	}
 
-	items := getHostListItems(hosts, hostsInError)
+	items := getHostListItems(hosts, hostsInError, false)
 
 	for _, item := range items {
 		if item.Active {
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -211,6 +211,10 @@ var Commands = []cli.Command{
 				Name:  "quiet, q",
 				Usage: "Enable quiet mode",
 			},
+			cli.BoolFlag{
+				Name:  "verbose, v",
+				Usage: "Enable verbose mode",
+			},
 			cli.StringSliceFlag{
 				Name:  "filter",
 				Usage: "Filter output based on conditions provided",
--- a/commands/create.go
+++ b/commands/create.go
@@ -293,7 +293,7 @@ func cmdCreateOuter(c CommandLine, api libmachine.API) error {
 	}
 
 	if _, ok := driver.(*errdriver.Driver); ok {
-		return errdriver.NotLoadable{driverName}
+		return errdriver.NotLoadable{Name: driverName}
 	}
 
 	// TODO: So much flag manipulation and voodoo here, it seems to be
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -167,9 +167,11 @@ func filterHosts(hosts []*host.Host, filters FilterOptions) []*host.Host {
 func getSwarmMasters(hosts []*host.Host) map[string]string {
 	swarmMasters := make(map[string]string)
 	for _, h := range hosts {
-		swarmOptions := h.HostOptions.SwarmOptions
-		if swarmOptions != nil && swarmOptions.Master {
-			swarmMasters[swarmOptions.Discovery] = h.Name
+		if h.HostOptions != nil {
+			swarmOptions := h.HostOptions.SwarmOptions
+			if swarmOptions != nil && swarmOptions.Master {
+				swarmMasters[swarmOptions.Discovery] = h.Name
+			}
 		}
 	}
 	return swarmMasters
@@ -189,7 +191,7 @@ func matchesSwarmName(host *host.Host, swarmNames []string, swarmMasters map[str
 		return true
 	}
 	for _, n := range swarmNames {
-		if host.HostOptions.SwarmOptions != nil {
+		if host.HostOptions != nil && host.HostOptions.SwarmOptions != nil {
 			if strings.EqualFold(n, swarmMasters[host.HostOptions.SwarmOptions.Discovery]) {
 				return true
 			}
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -44,7 +44,6 @@ type HostListItem struct {
 }
 
 func cmdLs(c CommandLine, api libmachine.API) error {
-	quiet := c.Bool("quiet")
 	filters, err := parseFilters(c.StringSlice("filter"))
 	if err != nil {
 		return err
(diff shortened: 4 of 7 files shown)


thaJeztah:
Thanks, @dgageot. I think this will fix the current requirements, just not sure if this will be the best solution in the longer term. I.e., it may be difficult to decide for the end-user what is considered "basic" and "verbose" fields; what will we do if more fields are added? `-vv` for "extra" verbose?

We had this discussion in docker/docker; https://github.com/docker/docker/issues/7477, https://github.com/docker/docker/pull/10255, and decided to go for a `--format` flag to put the user in control; https://github.com/docker/docker/pull/14699

So; yeah.. for a short term, but for consistency and longer term solution, I'd consider a `--format` flag

Just my 0.02c :+1:

hairyhenderson:
@dgageot I think I'm with @thaJeztah - this is a :+1: _for now_ but a `--format` eventually would probably be good.

pandrew:
I like the idea of `--format`. If so will there be a short version for this; `-f` ? That would speed up typing =)

nathanleclaire:
Maybe we can add in `RESPONSE` and then implement solutions such as this and/or `--format` when it is a concrete problem?  I don't know that we have much of an issue with wrap width today, for instance `docker ps --no-trunc` dwarfs us in terms of line length:

```console
$ docker ps --no-trunc
CONTAINER ID                                                       IMAGE                                                              COMMAND                                                                                              CREATED             STATUS              PORTS               NAMES
ca8584fb04b5394b814fb6ee0633c826e853d115c8657335803d099a9f31373d   0b12e7a4f3460b88c94a7bbf471f7e5722d28bef3b9e929c1e0ba669a4eb746b   "/bin/sh -c 'apt-get update && apt-get install -y --no-install-recommends openjdk-7-jre-headless'"   3 days ago          Up 3 days                               focused_yalow
$ dm ls
NAME      ACTIVE   DRIVER         STATE     URL                          SWARM   DOCKER   ERRORS
default   *        virtualbox     Running   tcp://192.168.99.100:2376            v1.9.1
droplet   -        digitalocean   Running   tcp://159.204.249.243:2376           v1.9.1
```

Maybe one more column is OK?  Any thoughts?

askb:
:+1: favouring `RESPONSE` in the default output would be very useful and also for `--format` and/or something like `--no-trunc` in the long term as suggested by @thaJeztah. :)

@nathanleclaire, most of the wrap width issue would seldom concern as the `ERRORS` column has been made optional, as we really may not want to see some long errors in `dm ls` output. (like in the output for issue #2518).

thaJeztah:
> Maybe we can add in RESPONSE and then implement solutions such as this and/or --format when it definitely known to be a problem instead of something we're concerned about abstractly

SGTM

dgageot:
@nathanleclaire @hairyhenderson @thaJeztah So, the plan is to close this PR, keep on adding more columns until it really hurts and fill a ticket to support `--format` flag? Right?

If so, I'll close the PR. @thaJeztah can you be in charge of creating the ticket?

thaJeztah:
> So, the plan is to close this PR, keep on adding more columns until it really hurts and fill a ticket to support --format flag? Right?

I'm interested to hear @hairyhenderson's opinion, who raised the issue of the output becoming too wide by default. I'm also good with using this (-v/--verbose) as a "for the time being" solution and add --format if even more columns are added.

hairyhenderson:
> I'm interested to hear @hairyhenderson's opinion

Dangerous words, @thaJeztah :grin:...

So, _right now_ the width hurts for me. Because I'm a sucker for punishment, I created a machine the other day with an FQDN instead of an IP address:

```
$ docker-machine ls
NAME             ACTIVE   DRIVER         STATE     URL                                                    SWARM                     DOCKER    ERRORS
archtest         -        generic        Running   tcp://ec2-11-11-111-111.compute-1.amazonaws.com:2376                             v1.9.1    
default          -        virtualbox     Stopped                                                                                    Unknown   
hairyhenderson   -        digitalocean   Running   tcp://11.11.111.111:2376                               hairyhenderson (master)   v1.9.1    
```

Here, the lines are 149 characters wide. Now, granted, it's a bit pathological because I could've just used an IP address. However, if I _had_, the lines would still be somewhere around 115 chars wide, owing to the `SWARM` column.

On my laptop, my terminals are usually either 101 or 116 characters wide. So for me, every time I run `docker-machine ls`, the output wraps. And when I get stuff in the `ERRORS` column, it's basically impossible to see what's going on at a glance.

One problem here is the `URL` column is subject to being super-wide, and I don't know how useful the output is in reality. It might be enough to truncate that column at something like 20 chars and display the full URL in a `--no-trunc` mode, or simply assume the user will `docker-machine inspect` if they really want to know the URL.

Another approach is to totally revisit the output and move from a column/row display to something a little different, like:

```
$ docker-machine ls
archtest:
  driver: generic
  state: Running
  url: tcp://ec2-11-11-111-111.compute-1.amazonaws.com:2376
  docker_version: v1.9.1    
default:
  driver: virtualbox
  state: Stopped
  docker_version: Unknown   
hairyhenderson:
  driver: digitalocean
  state: Running
  url: tcp://11.11.111.111:2376
  swarm: hairyhenderson (master)
  docker_version: v1.9.1
```

So, format it in a style less like `docker ps` and more like `docker info`... It _could_ be a way to totally avoid any of the width issues. Just spit-balling though... :grin:

thaJeztah:
> Dangerous words, @thaJeztah :grin:...

@hairyhenderson haha, well, let's see for whom;

> It might be enough to truncate that column at something like 20 chars
> Another approach is to totally revisit the output

Both those options should be possible if `--format` is implemented. Are you interested in contributing that? :innocent:

nathanleclaire:
Thanks for the feedback @hairyhenderson -- that's helpful to keep in mind.  FWIW, when `docker` output is too long like that, I usually pipe it into `less -S` (and almost always do this when using `--no-trunc`), but I agree addressing it first-class could be nice.

I guess my concern with `--format` is that it strikes me as a pretty power user feature (requiring at least some knowledge of Go templates and which properties they might be looking for) which is not particularly friendly, and I want most of the discussed information to be easily available to folks.  To that extent, David's solution here is really good -- just a simple flag to toggle on or off.

Truncating columns and adding a `--no-trunc` certainly seems like it could be a decent compromise for the immediate future, and would still be consistent with Docker's behavior.  Maybe we could do that, and see how we feel about potentially adding anything else after?

A good question is, are there any specific columns we can envision wanting to add in the near distant future?  If we can think of a lot, perhaps it's worth the effort to invest in something like `--format`.  If not, maybe we can just add `RESPONSE` and `--no-trunc`.  My feeling is there isn't a lot more that won't be covered by `inspect` already, but I could definitely be wrong.

thaJeztah:
Let's go with this then, and if somebody wants to contribute®, add a `--format` option for power users.

PR #2523, opened 2015-12-08
FIX 2510 Support DNS names

dgageot:


jeanlaurent:
Altough we should rename the IPAdress variable to something more meaningfull. Unfortunalty this is a huge task since it changes the json file format we use as well as most of the flags.

So LGTM for a quick win.

nathanleclaire:
:+1:

PR #2526, opened 2015-12-08
Monitor and report bug with bugsnag

jeanlaurent:
This enable crash reporting through bugsnag where we have the most errors : on 
* `docker-machine create`
* `docker-machine upgrade`

The crash report contains, docker-machine build & version, OS, ARCH, shell, uname ( for windows machine this identify clearly the shell running docker-machine ), stacktrace and history of the whole command.

To enable tracking one must export a `MACHINE_BUGSNAG_API_TOKEN` containing a valid bugsnag key to allow the crash reporting to occurs.

Opting-out :
The configuration check for a `no-error-report` file at the root of `.docker/machine` to disable any error reporting.

The code is ready for review

Things to discuss : 
- Toolbox could install the `no-error-report` file according to their GUI but should we do something to eqse the opt out tracking beside documenting the process in the doc ?
- Should we bundle the apikey with the code ?


libmachine/log/logrus_machine_logger.go
```diff
@@ -77,3 +77,8 @@ func (ml LogrusMachineLogger) Warn(args ...interface{}) {
 func (ml LogrusMachineLogger) Warnf(fmtString string, args ...interface{}) {
 	ml.logger.Warnf(fmtString, args...)
 }
+
+func (ml LogrusMachineLogger) MonitorCrash() {
+	hook, _ := NewBugsnagHook()
```
nathanleclaire: Let's please check this error (I'm assuming) here.

nathanleclaire:
What's going on with the `MonitorCrash` bit?  I think we can just register the hook in an `init` block, or wherever we create 

Also, since this is the PR introducing the actual feature of sending logs remotely, I feel like it should be the PR that also introduces a way to opt out.  My thinking on this was to have some kind of `~/.docker/NO_LOG` file which will cause us to skip registering the BugSnag hook, but we should talk to some other teams in Docker and see what they think.

libmachine/log/bugsnag.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/bugsnag/bugsnag-go"
+	"github.com/docker/machine/version"
+)
+
+var metaData bugsnag.MetaData
+
+func Setup(apiKey string) error {
+	if apiKey == "" {
+		apiKey = "e98db82ef639114001cd3397b2c9f885"
```
nathanleclaire: Think we should just make this a package-level `const`? I don't see why any invoker would ever call it with a different value.

libmachine/log/bugsnag.go
```diff
@@ -0,0 +1,54 @@
...
+	"runtime"
+	"strings"
+
+	"github.com/bugsnag/bugsnag-go"
+	"github.com/docker/machine/version"
+)
+
+var metaData bugsnag.MetaData
+
+func Setup(apiKey string) error {
```
nathanleclaire: Why not just make this `func init()`?

libmachine/log/bugsnag.go
```diff
@@ -0,0 +1,54 @@
...
+		Logger:          new(BugsnagLogger),
+	})
+
+	metaData = bugsnag.MetaData{}
+
+	metaData.Add("app", "compiler", fmt.Sprintf("%s (%s)", runtime.Compiler, runtime.Version()))
+	metaData.Add("device", "os", runtime.GOOS)
+	metaData.Add("device", "arch", runtime.GOARCH)
+
+	// Add environment variables
```
nathanleclaire: As mentioned in @dmp42 PR, we probably should not do this (report the user's environment variables) since it is highly likely to leak secrets -- for instance, the `DIGITALOCEAN_ACCESS_TOKEN` environment variable

libmachine/log/bugsnag_test.go
```diff
@@ -0,0 +1,11 @@
+package log
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestBugSnag(t *testing.T) {
```
nathanleclaire: Won't this test fail without network access and/or access to our particular Bugsnag API token?

nathanleclaire:
@jmorgan @mchiang @frenchben Can you chime in and talk a bit about how we track the "state" of users opting in or opting out of BugSnag reporting for Toolbox / KM?  It'd be really nice to have a unified layer for this.

libmachine/log/logrus_machine_logger.go
jeanlaurent: Of course yes.

libmachine/log/bugsnag.go
jeanlaurent: Yes, removed.

I'd like to add the current shell tough. But thats not that easy to detect.

libmachine/log/bugsnag_test.go
jeanlaurent: Yes, that more a test that help building and testing quickly, not sure what it will become

jeanlaurent:
ready for review @nathanleclaire

libmachine/drivers/plugin/localbinary/plugin_test.go
```diff
@@ -112,12 +111,10 @@ func TestExecServer(t *testing.T) {
 		finalErr <- lbp.execServer()
 	}()
 
-	expectedAddr := "127.0.0.1:12345"
-	expectedPluginOut := "Doing some fun plugin stuff..."
-	expectedPluginErr := "Uh oh, something in plugin went wrong..."
-
 	logScanner := bufio.NewScanner(logReader)
 
+	//
```
nathanleclaire: Looks like this are some accidentally left in comment lines

libmachine/crashreport/crash_report.go
```diff
@@ -0,0 +1,93 @@
...
+	optOutFilePath := filepath.Join(mcndirs.GetBaseDir(), ".notracking")
+	if _, err := os.Stat(optOutFilePath); os.IsNotExist(err) {
+		return false
+	}
+	log.Debug("Opting out of tracking")
+	return true
+}
+
+func detectRunningShell(metaData *bugsnag.MetaData) {
+	for _, v := range os.Environ() {
```
nathanleclaire: Why not just `os.Getenv` `SHELL` and `__fish_bin_dir` directly?

libmachine/host/host.go
```diff
@@ -134,11 +135,13 @@ func (h *Host) Upgrade() error {
 
 	provisioner, err := provision.DetectProvisioner(h.Driver)
 	if err != nil {
+		crashreport.ReportCrash(err)
```
nathanleclaire: Maybe you could just call this module `crash` and the method `Report`.  Then the invocation would be simply:

```golang
crash.Report(err)
```

nathanleclaire:
Looking like a pretty good start -- I have posted some comments in the code.

Any objection to calling the file something like `.no-error-report` instead?  "Tracking" has slightly negative connotations in this context.

Can you please add a blurb about error reporting to the documentation?

I have a proposal for how we might want to deal with "reported errors" (create failure) vs. "unreported errors" ("machine does not exist" type errors).  We might not be able to implement it right away, but it's something to think about:

Perhaps we could do something like create an interface, `Snagger`, with a simple `Snag()` method.  If errors (for which ideally we should always have custom types) want to indicate that they are the type which should be reported to BugSnag, they fulfill this interface (we could even have an easily embeddable struct for this).  Then `log.Warn` and `log.Fatal` calls will do a flush to BugSnag if one of the passed arguments has this type.  Something like:

```go
func Error(args ...interface{}) {
    for _, arg := range args {
        if _, ok := arg.(*crash.Snagger); ok {
            crash.Report(err)
        }
    }
    Logger.Error(args)
 }
```

What do you think about that idea?

jeanlaurent:
@nathanleclaire Thanks for your feedback, I implemented all your suggestions.

We believe we must go forward with this. The code is probably good to go now, but we must still discuss the opt-out and key bundling.

To ease things out and prevent countless rebase I disabled the bundled key by default. So as of now, it will report only if the `MACHINE_BUGSNAG_API_TOKEN` env variable is set.

As for the `snag()` method, let's discuss this, this sounds like bundling crash reporting with log handling which is something I'm a bit reluctant to do since they are IMHO very distinct concept. But let's open a new issue and discuss this there.

dgageot:
LGTM

libmachine/crashreport/crash_report.go
```diff
@@ -0,0 +1,111 @@
...
+		metaData.Add("device", "shell", shell)
+	}
+	shell = os.Getenv("__fish_bin_dir")
+	if shell != "" {
+		metaData.Add("device", "shell", shell)
+	}
+}
+
+func detectUname(metaData *bugsnag.MetaData) {
+	cmd := exec.Command("uname", "-s")
```
nathanleclaire: This will fail on systems without the `uname` command, i.e. Windows, which you seem aware of due to the bare `return` later on, but at any rate a `switch` on `runtime.GOOS` might be slightly safer.

nathanleclaire:
> As for the snag() method you suggested, let's discuss this, this sounds like bundling crash reporting with log handling which is something I'm a bit reluctant to do since they are IMHO distinct concepts.

I'm slightly confused -- I thought we were going to use the BugSnag hook for logrus?  https://github.com/Shopify/logrus-bugsnag New issue sounds fine though.

Nice stuff -- LGTM

libmachine/crashreport/crash_report.go
jeanlaurent: uname on windows in cmd or powershell will indeed fail. But not in the gazillion other shell. And this uname stuff is here right for this purpose it will tell you if the user is using MINGw, cygwin or gitbash.

jeanlaurent:
Well that what we did at first. But the hook is blunt. It triggers bugsnag notification based on the error level. It's hard to distinguish error from the user like forgetting to put a machine name in `docker-machine env` from a real `error.log` when catching `error`. So we took the decision to separate them and have few errors, but real one, rather than having tons of cli error we don't really care to monitor.

nathanleclaire:
OK sounds good.  I think we will probably know if there are other areas where we want to snag the bugs.  Create and upgrade are a good start.

dcramer:
Will these changes work generically for similar solutions?

Issue #1671, opened 2015-08-11
Create operation with vmwarefusion driver does not complete


robinbb:
On Mac OS 10.10.4, tried to create a VM using:

    docker-machine \
       -D \
       create \
       --driver vmwarefusion \
       --vmwarefusion-disk-size "12345" \
       --vmwarefusion-memory-size "1024" \
       spinzo-vm

This was with a docker-machine binary timestamped "Aug 11 15:50" downloaded from https://docker-machine-builds.evanhazlett.com/latest/.

Output was as seen in http://www.pastebin.ca/3099674

````
Creating SSH key...
Creating VM...
VixDiskLib: Invalid configuration file parameter. Failed to read configuration file.
Creating disk '/Users/robinbb/.docker/machine/machines/spinzo-vm/spinzo-vm.vmdk'
Virtual disk creation successful.
Starting spinzo-vm...
executing: /Applications/VMware Fusion.app/Contents/Library/vmrun start /Users/robinbb/.docker/machine/machines/spinzo-vm/spinzo-vm.vmx nogui
Waiting for VM to come online...
MAC address in VMX: 00:0c:29:87:83:87
IP found in DHCP lease table: 10.88.88.132
Got an ip: 10.88.88.132
Creating Tar key bundle...
executing: /Applications/VMware Fusion.app/Contents/Library/vmrun -gu docker -gp tcuser directoryExistsInGuest /Users/robinbb/.docker/machine/machines/spinzo-vm/spinzo-vm.vmx /var/lib/boot2docker
executing: /Applications/VMware Fusion.app/Contents/Library/vmrun -gu docker -gp tcuser CopyFileFromHostToGuest /Users/robinbb/.docker/machine/machines/spinzo-vm/spinzo-vm.vmx /Users/robinbb/.docker/machine/machines/spinzo-vm/userdata.tar /home/docker/userdata.tar
executing: /Applications/VMware Fusion.app/Contents/Library/vmrun -gu docker -gp tcuser runScriptInGuest /Users/robinbb/.docker/machine/machines/spinzo-vm/spinzo-vm.vmx /bin/sh sudo /bin/mv /home/docker/userdata.tar /var/lib/boot2docker/userdata.tar && sudo tar xf /var/lib/boot2docker/userdata.tar -C /home/docker/ > /var/log/userdata.log 2>&1 && sudo chown -R docker:staff /home/docker
[...]

PR #2528, opened 2015-12-08
Explicitly set umask before invoking vmrun in vmwarefusion

jhoos:
For some reason, invoking vmrun with the nogui option doesn't work right if umask is set to not allow world-readable permissions by default.  This seems to be true at least in VMware Fusion 7.0.x through 8.0.1 when running on OS X 10.10 (Yosemite) and possibly other versions (although at the moment it doesn't seem to be an issue in El Capitan with Fusion 8.0.2).  

As a workaround, this changes the driver to explicitly set the umask to 022 before invoking vmrun.

This might resolve issue #1671 for some people.

### DIFF
--- a/drivers/vmwarefusion/vmrun.go
+++ b/drivers/vmwarefusion/vmrun.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/docker/machine/libmachine/log"
 )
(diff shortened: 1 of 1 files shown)


nathanleclaire:
cc @frapposelli

frapposelli:
LGTM 

Don't think there are any drawbacks in implementing this, even though it shouldn't be necessary.

nathanleclaire:
LGTM cc @dgageot @jeanlaurent

dgageot:
LGTM

PR #2541, opened 2015-12-09
Add emacs support to env command

dougm:
This is useful for setting docker client env vars within Emacs, when using docker.el for example: https://github.com/dougm/docker.el/blob/machine/docker-machine.el#L36-L50

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -130,6 +130,10 @@ func shellCfgSet(c CommandLine, api libmachine.API) (*ShellConfig, error) {
 		shellCfg.Prefix = "SET "
 		shellCfg.Suffix = "\n"
 		shellCfg.Delimiter = "="
+	case "emacs":
+		shellCfg.Prefix = "(setenv \""
+		shellCfg.Suffix = "\")\n"
+		shellCfg.Delimiter = "\" \""
 	default:
 		shellCfg.Prefix = "export "
 		shellCfg.Suffix = "\"\n"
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Hey @dougm, thanks for the contribution.

Why not just `eval` the output for bash?  I'm not too familiar with e-lisp so bear with me.

dougm:
Thanks for the quick response @nathanleclaire .  In this case, I'm running Emacs.app and docker.el within that process that shells out to the 'docker' cli.  There is no bash shell in the mix.
More info on docker.el here: https://github.com/Silex/docker.el/blob/master/README.md#screenshots

jeanlaurent:
Thanks for your contribution @dougm, I have absolutely no idea how many users are using docker.el and docker-machine. I guess it's very tiny, but correct me if I'm wrong.

My concern here is that even if the code works right now, I don't really know how we can maintain it the long run without a community helping on those niche setup.

WDYT @docker/machine-maintainers ?

dougm:
There isn't any code in this PR specific to docker.el, that was just an example use.  The lisp code only depends on built-in Emacs functions (setenv, etc.) that have been around for years and won't change.  Agreed the number of users would be small relative to the other shells, but it would be a nice enhancement for us.  I'd be happy to help with any future maintenance for this.  Thanks for your consideration.

nathanleclaire:
@dougm My feeling is that if you're willing to bump up the test coverage in this PR (i.e. add cases for this to the other unit tests in `env_test.go` like the other shells have, and extended the `env` test in `test/integration/core` to include this) and are willing to commit to maintaining it, I'm willing to merge.

@jeanlaurent @dgageot Do you think this sounds agreeable?

dougm:
@nathanleclaire Sure, added additional unit test and integration tests, let me know if I missed anything.  I can commit to maintaining it, happy to.  Thanks.

nathanleclaire:
Nice -- LGTM

jeanlaurent:
counting on you @dougm 

LGTM

travisjeffery:
This is awesome. I could help out with maintaining this as well if you need help.

PR #2545, opened 2015-12-10
Google driver - Update firewall open ports

dgageot:
Now, each time a google host is created, the firewall rule is checked for missing opened ports. Then, the rule is either created or updated.


jeanlaurent:
Maybe you can squatch the first two commits.
apart from that LGTM

dgageot:
@jeanlaurent squashed!

PR #2546, opened 2015-12-10
add timeout override to ls command

### LINKED ISSUE #2534: docker-machine ls stateTimeoutDuration is hardcoded to 10sec
Hi, I'm trying to find out if we have any specific reason why stateTimeoutDuration in docker-machine ls is hard coded to 10sec?  It would be nice to make this adjustable from the plugins or from a command line option.   Has any thoughts been given for that?

jeanlaurent:
This add a `--timeout` flag for the `ls` command. 
The default stays the same at 10 seconds but can be overriden to suit a specific need.

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -21,8 +21,10 @@ import (
 	"github.com/skarademir/naturalsort"
 )
 
+const DefaultTimeoutDuration = 10
+
 var (
-	stateTimeoutDuration = 10 * time.Second
+	stateTimeoutDuration = DefaultTimeoutDuration * time.Second
 )
 
 // FilterOptions -
@@ -47,6 +49,9 @@ type HostListItem struct {
 }
 
 func cmdLs(c CommandLine, api libmachine.API) error {
+	stateTimeoutDuration = time.Duration(c.Int("timeout")) * time.Second
+	log.Debugf("ls timeout set to %d seconds", stateTimeoutDuration/time.Second)
+
 	quiet := c.Bool("quiet")
 	filters, err := parseFilters(c.StringSlice("filter"))
 	if err != nil {
--- a/docs/reference/ls.md
+++ b/docs/reference/ls.md
@@ -16,8 +16,21 @@ parent="smn_machine_subcmds"
 
     Options:
 
-       --quiet, -q					Enable quiet mode
+       --quiet, -q					                Enable quiet mode
        --filter [--filter option --filter option]	Filter output based on conditions provided
+       --timeout, -t				                Timeout in seconds, default to 10s
+
+## Timeout
+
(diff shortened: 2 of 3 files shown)


dgageot:
LGTM

jeanlaurent:

docs/reference/ls.md
```diff
@@ -16,8 +16,21 @@ parent="smn_machine_subcmds"
...
     Options:
 
-       --quiet, -q					Enable quiet mode
+       --quiet, -q					                Enable quiet mode
        --filter [--filter option --filter option]	Filter output based on conditions provided
+       --timeout, -t				                Timeout in seconds, default to 10s
+
+## Timeout
+
+The `ls` command tries to reach each host in parallel. If a given host does not answer in less than 10 seconds, the `ls` command
```
nathanleclaire: Pretty good just some small nitpicks on my end.  I'd write:

> The `ls` command tries to reach each host in parallel. If a given host does not answer in less than 10 seconds, the `ls` command will state that this host is in `Timeout` state. In some circumstances (poor connection, high load, or while troubleshooting), you may want to increase or decrease this value. You can use the -t flag for this purpose with a numerical value in seconds.

Also, can you please wrap the lines to 80 characters? (your editor probably has a way to do this automatically) It makes the diffs easier to read on a terminal.

commands/ls.go
```diff
@@ -47,6 +49,9 @@ type HostListItem struct {
 }
 
 func cmdLs(c CommandLine, api libmachine.API) error {
+	stateTimeoutDuration = time.Duration(c.Int("timeout")) * time.Second
+	log.Debugf("ls timeout set to %d seconds", stateTimeoutDuration/time.Second)
```
nathanleclaire: I think we can just move the `var stateTimeoutDuration` into this function fully now and it no longer has to be global-level, no?

commands/ls.go
```diff
@@ -21,8 +21,10 @@ import (
 	"github.com/skarademir/naturalsort"
 )
 
+const DefaultTimeoutDuration = 10
```
nathanleclaire: I don't think it's exported so it probably should be lowercase.  Also, I think indicating it's for `ls` in particular would be good: `lsTimeoutSeconds` ?

nathanleclaire:
I have a couple of nitpicks, I think I will just turn in a PR for them as it will be faster.  But I've posted comments so you can know my motivation.  

Nice job -- LGTM

commands/ls.go
nathanleclaire: Just saw there's a test this affects. Hm...

nathanleclaire: OK it's probably easier to just leave that bit as-is.

PR #2547, opened 2015-12-10
[WIP] Case insensitive filestore

dgageot:
Transform name to all lowercase before doing anything with a directory on disk. Therefore `A` and `a` always appear with their same original casing to users but reference only `a` on disk. Two machines with the same name but different casing are not be allowed.

**Big issue** this code is unable to read from the Filestore if a machine was created with uppercase letters.

Should we:
- Migrate all existing hosts
- Support hosts saved with a different case
- Forget about this PR


nathanleclaire:
Yeah, this is really a tough one.  I do want the feature, but we might have to sit on this one for a while and discuss the various implications.  Migrating existing hosts in this regard seems like it would be very difficult to do right.

dmp42:
(for context #1845 )

dgageot:
Closing for now until we find a solution. Keeping the issue (https://github.com/docker/machine/issues/1845) open of course

PR #2555, opened 2015-12-11
Add context, driver name & command to crash report

jeanlaurent:
This add Context, Driver name & command to bugsnag crash reporting.

Hopefully this should help in triaging errors.


dgageot:
LGTM

PR #2557, opened 2015-12-11
Fix the build on linux/windows by excluding fusion driver

dgageot:
- fusion driver should only be built on darwin
- CI uses the wrong target to cross build
- Added a test to check that the build actually produced binary files


jeanlaurent:
Apart from the slight comment, which IMHO is mandatory. 

LGTM

nathanleclaire:
:+1: :+1: :+1:

PR #2562, opened 2015-12-12
Strip certs and keys in log before sending to Bugsnag

nathanleclaire:
This is my only qualm with over-sharing in what we have so far.

cc @dgageot @jeanlaurent PTAL


dgageot:
LGTM

PR #2569, opened 2015-12-14
minor fix to formatting update for `dm config` output

askb:
minor fix to formatting update for `dm config` output


jeanlaurent:
bullseye @askb

PR #2576, opened 2015-12-14
Make integration test a bit more resilient

jeanlaurent:
adding ugly timeout to `docker-machine ls` after restart and start to make the integrations test a bit more *"resilient"*

### DIFF
--- a/test/integration/core/core-commands.bats
+++ b/test/integration/core/core-commands.bats
@@ -116,7 +116,7 @@ use_shared_machine
 }
 
 @test "$DRIVER: machine should show running after start" {
-  run machine ls
+  run machine ls --timeout 20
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

nathanleclaire:
So the issue is that the timeout is not long enough in its current form?  Is VirtualBox taking that long to respond or is it another driver which is problematic?

jeanlaurent:
They are not failing on David's Mac and it fails each time on mine.
Extending the timeout a bit ( well actually doubled ) make the tests passes each time on all machines.

nathanleclaire:
OK, odd to hear it takes so long but kicking the timeout up won't hurt anything.

LGTM

PR #2579, opened 2015-12-14
introduce dind (Docker in Docker) option to `config` command

bjaglin:
I am using [Crane](https://github.com/michaelsauter/crane) within a container to orchestrate a remote engine. If I setup my local client to use the remote engine, the orchestration container is spawned remotely, so I can't bind-mount any configuration into it. The `--dind` flag introduced in that PR allows me to run a local orchestrator onto which I can bind-mount, yet still pointing at my remote target.

I first thought about extending the `env` command, but I ended up putting in `config`, although neither of them feels 100% right...

I have not tested the functionality from a Windows client, but as far as I can tell (which is not much because command substitution is very limited, so even the existing `config` might be tricky), all is taken care in terms of spaces and specificities of the Windows filesystem mounts.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -151,6 +151,10 @@ var Commands = []cli.Command{
 				Name:  "swarm",
 				Usage: "Display the Swarm config instead of the Docker daemon",
 			},
+			cli.BoolFlag{
+				Name:  "dind",
+				Usage: "Generate CLI arguments to forward the config to a Docker in Docker client",
+			},
 		},
 	},
 	{
--- a/commands/config.go
+++ b/commands/config.go
@@ -2,11 +2,15 @@ package commands
 
 import (
 	"fmt"
+	"path/filepath"
 
+	"github.com/docker/machine/commands/mcndirs"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/check"
 	"github.com/docker/machine/libmachine/log"
 )
+
+const dindCertPath = "/etc/docker/cert"
 
 func cmdConfig(c CommandLine, api libmachine.API) error {
 	// Ensure that log messages always go to stderr when this command is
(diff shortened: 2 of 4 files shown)


nathanleclaire:
Hi @bjaglin, thanks a lot for the PR.  While I appreciate the contribution, I think it is a bit out of scope for what the `config` command in Machine today is intended to do.

Maybe a shell function + `inspect -f` might work for your use case?

Is there anything smaller in scope we could implement to help enable your use case?

bjaglin:
Thanks for having a look at it!

As for the fit with the current `config` command, I think it matches pretty well the semantics defined in the doc: "Show the Docker client configuration for a machine". It's just a case where the remote configuration is for an embedded client (potentially called indirectly via another binary) instead of the local client, therefore the introduction of a flag. In both cases, the configuration consists of options passed to the Docker CLI, which makes it shell-independant, and therefore doesn't fit with the `env` command. Compare https://github.com/docker/machine/pull/2579/files#diff-75243fefd5fbd9a3f5887c93e7560cb0R25 & https://github.com/docker/machine/pull/2579/files#diff-75243fefd5fbd9a3f5887c93e7560cb0R58.

I am not sure what you mean by leveraging `inspect -f` - the config we want to inject comes from machine, not a Docker container, so there is nothing to inspect for the setup. As for a smaller scope to enable my use-case: this whole thing started as an external custom shell script doing `seds` and escaping against the output of `config`, it's of course doable since all information is available (the same way `env` be implemented against the output of `config`), but I thought this was small and integration-testable enough to deserve a flag.

nathanleclaire:
> As for the fit with the current config command, I think it matches pretty well the semantics defined in the doc: "Show the Docker client configuration for a machine"

Well, it's lightly implied that it's to show the Docker client configuration for a Machine from the client computer where Machine is in use.  Also handling connection from things in a container expands the scope in my opinion.

> I am not sure what you mean by leveraging inspect -f - the config we want to inject comes from machine, not a Docker container, so there is nothing to inspect for the setup.

I mean why not use `docker-machine inspect -f`?  You could make a bash function like so which accomplishes largely the same thing:

```bash
function dindcfg () {
    docker-machine inspect -f '--volume="{{.HostOptions.AuthOptions.StorePath}}:/etc/docker/cert:ro"
--env "DOCKER_CERT_PATH={{.HostOptions.AuthOptions.CertDir}}"
--env "DOCKER_TLS_VERIFY=1"
--env "DOCKER_HOST=tcp://{{.Driver.IPAddress}}:2376"
' "$1"
}
```

Usage:

```console
$ dindcfg default
--volume="/Users/nathanleclaire/.docker/machine/machines/default:/etc/docker/cert:ro"
--env "DOCKER_CERT_PATH=/Users/nathanleclaire/.docker/machine/certs"
--env "DOCKER_TLS_VERIFY=1"
--env "DOCKER_HOST=tcp://192.168.99.188:2376"
```

bjaglin:
Oops, my bad, I wasn't aware of the `inspect -f` command on Machine, I thought you were referring to the Docker CLI. I'll have a look at it, thanks!

bjaglin:
That answers my needs completely, sorry for the noise and thanks for the guided tour!

For the record, there is a tiny mistake in the template above, it should be:
```
--volume="{{.HostOptions.AuthOptions.StorePath}}:/etc/docker/cert:ro"
--env "DOCKER_CERT_PATH=/etc/docker/cert"
--env "DOCKER_TLS_VERIFY=1"
--env "DOCKER_HOST=tcp://{{.Driver.IPAddress}}:2376"
```
(since the `--volume` bind-mounts the directory on a static path inside the container)

nathanleclaire:
Glad to hear it worked well for you!

PR #2584, opened 2015-12-15
FIX #2561 SSH pubkey should have username comment

### LINKED ISSUE #2561: [Google] SSH pubkey should have username comment
When creating a new machine using the Google driver, Docker-Machine is correctly uploading its public SSH key to the machine. But when editing the machine via the Google Cloud Console, the SSH key is shown as malformed, because Google expects "ssh-rsa [keymaerial] username". I have to manually add the trailing " docker-user" so that I can finish editing the instance.

It would be nice if Docker-Machine could append its username automatically when setting the SSH key.

dgageot:


jeanlaurent:
LGTM

PR #2599, opened 2015-12-16
Remove Logrus follow-up

jeanlaurent:
In the wake of removing logrus, here a couple of commits splitting responsability, making the code clearer and hardened testing.


nathanleclaire:
LGTM

PR #2605, opened 2015-12-17
update the bash completion with new options added.

askb:
This updates the bash completion scripts with new options added. Now changes should reflect the following:
```
dm ls --timeout
dm env --no-proxy
dm --bugsnag-api-token --github-api-token
```


jeanlaurent:
LGTM

PR #2609, opened 2015-12-17
Add vbox.log to crashreport

jeanlaurent:
Ready for review cc @docker/machine-maintainers

### DIFF
--- a/libmachine/crashreport/crash_report.go
+++ b/libmachine/crashreport/crash_report.go
@@ -13,6 +13,8 @@ import (
 
 	"errors"
 
+	"io/ioutil"
+
 	"github.com/bugsnag/bugsnag-go"
 	"github.com/docker/machine/commands/mcndirs"
 	"github.com/docker/machine/libmachine/log"
@@ -35,8 +37,7 @@ func Configure(key string) {
 	}
 }
 
-// Send through http the crash report to bugsnag need a call to Configure(apiKey) before
-func Send(err error, context string, driverName string, command string) error {
+func SendWithFile(err error, context string, driverName string, command string, path string) error {
 	if noReportFileExist() || apiKey == noreportAPIKey {
 		log.Debug("Opting out of crash reporting.")
 		return nil
@@ -67,6 +68,7 @@ func Send(err error, context string, driverName string, command string) error {
 	detectRunningShell(&metaData)
 	detectUname(&metaData)
 	detectOSVersion(&metaData)
+	addFile(path, &metaData)
 
 	var buffer bytes.Buffer
 	for _, message := range log.History() {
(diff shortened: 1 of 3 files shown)


libmachine/crashreport/crash_report_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+func TestFileIsNotReadWhenNotExisting(t *testing.T) {
+	metaData := bugsnag.MetaData{}
+	addFile("not existing", &metaData)
+	assert.Equal(t, 0, len(metaData))
+}
+
+func TestRead(t *testing.T) {
+	metaData := bugsnag.MetaData{}
+	addFile("../../test/_resources/vbox.log.txt", &metaData) // me no like this path :(
```
nathanleclaire: Instead of this hardcoded test file, why not use `ioutil.TempFile`? https://golang.org/pkg/io/ioutil/#TempDir

e.g., something like:

```go
f, err := ioutil.TempFile("", "")
if err != nil {
    t.Fatal(err)
}

logdata := []byte(`
VM started.
VM failling.
`)

if err := ioutil.WriteFile(f.Name(), logdata, 0644); err != nil {
    t.Fatal(err)
}

addFile(f.Name(), &metaData)
assert.Equal(t, logdata, metaData["logfile"][f.Name()])
```

I know it's slightly more code but it strongly decouples the test from having a magic file in the repo.

libmachine/crashreport/crash_report.go
```diff
@@ -76,6 +78,32 @@ func Send(err error, context string, driverName string, command string) error {
...
+	if path == "" {
+		return
+	}
+	file, err := os.Open(path)
+	if err != nil {
+		log.Debug(err)
+		return
+	}
+	defer file.Close()
+	var buffer bytes.Buffer
```
nathanleclaire: I think you probably just want `ioutil.ReadAll` instead of doing the scanning, right? https://golang.org/pkg/io/ioutil/#ReadAll

Then the code is just:

```go
data, err := ioutil.ReadAll(file)
if err != nil {
    log.Debug(err)
    return
}
metaData.Add("logfile", filepath.Base(path), string(data))
```

libmachine/libmachine.go
```diff
@@ -152,3 +152,13 @@ func (api *Client) performCreate(h *host.Host) error {
 	return nil
 
 }
+
+func sendCrashReport(err error, api *Client, host *host.Host) {
+	if host.DriverName == "virtualbox" {
+		// getting BaseDriver.ResolveStorePath here sounds like hell...
+		vboxlogPath := filepath.Join(api.Filestore.Path, "machines", host.Driver.GetMachineName(), host.Driver.GetMachineName(), "Logs", "VBox.log")
```
nathanleclaire: We have a method for this.  I think this line can be reduced to just:

```go
vboxLogPath := filepath.Join(api.GetMachinesDir(), host.Name, host.Name, "Logs", "VBox.log"
```

libmachine/libmachine.go
```diff
@@ -152,3 +152,13 @@ func (api *Client) performCreate(h *host.Host) error {
...
+		crashreport.SendWithFile(err, "api.performCreate", host.DriverName, "Create", vboxlogPath)
```
nathanleclaire: Instead of having a separate method `SendWithFile`, why not have a method like `crashreport.AttachVBoxLog` which does that operation on the `metaData`, and call that here, allowing it to always fall through to the `crashreport.Send`?  Would reduce the number of possible code paths.

nathanleclaire:
Looks pretty good -- there are a few things I've noted that would be good to get cleaned up before merging.  Excited about this.

jeanlaurent:
Integrated all your feedback. cc @docker/machine-maintainers

dgageot:
What about the idea of @nathanleclaire of not adding a `sendWithFile` function but rather to add a file to the next crash report?

dgageot:
LGTM

jeanlaurent:
@nathanleclaire I don't think putting the whole process stateful for saving a method is worth it.

I strongly believe we will still do something along those lines in the futures if we want to be more precise in the way we collect data for bugsnag. Still this is probably not the time right now.

I implemented all the other suggestions

nathanleclaire:
LGTM

PR #2611, opened 2015-12-17
Automate most of the release process

jeanlaurent:
This automate the release process.

Current Status : 
* [x] Version Bump
* [X] Compile & Generate Binaries
* [X] Tag
* [X] Create release on github
* [X] Push binaries to release page
* [X] Generate Checksums
* [X] Add Installation Instructions
* [X] Add Release Notes
* [X] Generate and Add Contributor List 
* [X] Update the CHANGELOG.md
* [X] Update the Documentation
* [X] Verify the Installation ( probably very hard to automate at 100% )
* [X] Order wine ( not sure how to do this one )

This is ready to try for the v5.0.5 release.
Things to consider for improvment :
- Runs everything ( not only the binary creation) in a container, so we don't rely on the local machine
- Do the final version bump back to dev automatically ( this needs to know the *next* version number
- Do not ask for the current version number since this can be done by carefully looking into `version/version.go`
- We could probably force push the docs branch but I'm not too familiar with the process, and prefer to live yet another release before doing it.


release.sh
```diff
@@ -0,0 +1,144 @@
...
+# Perform a docker-machine release
+# GITHUB_TOKEN=XXXXX ./release.sh 0.X.X
+
+#GITHUB_USER=docker
+GITHUB_USER=jeanlaurent
+GITHUB_REPO=machine
+
+
+function display {
+    echo "🐳   $1"
```
nathanleclaire: Won't this cause issues if someone uses the script from Windows or Linux ?  e.g. inside of a container

release.sh
```diff
@@ -0,0 +1,144 @@
...
+   rm -rdf $RELEASE_DIR
+   checkError "Can't clean up."
+ fi
+
+ display "Cloning into $RELEASE_DIR from $PROJECT_URL"
+
+ mkdir -p $RELEASE_DIR
+ checkError "Can't create $RELEASE_DIR, aborting."
+ git clone -q $PROJECT_URL $RELEASE_DIR
+ checkError "Can't clone into $RELEASE_DIR, aborting."
```
nathanleclaire: Why not put:

```bash
set -e
```

at the top of the script and remove these error checks?  It will cause `bash` to exit on any failed commands.

release.sh
```diff
@@ -0,0 +1,144 @@
...
+        echo "😡   $1"
+        exit 1
+    fi
+}
+
+VERSION=$1
+GITHUB_VERSION="v$1"
+PROJECT_URL="git@github.com:$GITHUB_USER/$GITHUB_REPO"
+RELEASE_DIR=../release-$VERSION
+DOCKER_MACHINE=$GOPATH/src/github.com/docker/machine/bin/docker-machine
```
nathanleclaire: Mine are usually in `bin/docker-machine` or `/usr/local/bin/docker-machine`-- why not simply refer to `docker-machine` directly, or just do one check at the beginning like `which docker-machine`?

release.sh
```diff
@@ -0,0 +1,144 @@
...
+  display "Checking if release already exist"
+  github-release info \
+      --security-token  ${GITHUB_TOKEN} \
+      --user ${GITHUB_USER} \
+      --repo ${GITHUB_REPO} \
+      --tag ${GITHUB_VERSION}
+
+  if [ "$?" -ne 1 ]; then
+    display "Release already exist, trying to delete it."
+    github-release delete \
```
nathanleclaire: I'm pretty wary of automatically deleting anything, just because it's very possible something unexpected might happen in the shell script and end up being a lot more destructive than anticipated.  Why not just count on removing things manually in this case?

release.sh
```diff
@@ -0,0 +1,144 @@
...
+
+  display "Creating release on github"
+
+  github-release release \
+      --security-token  ${GITHUB_TOKEN} \
+      --user ${GITHUB_USER} \
+      --repo ${GITHUB_REPO} \
+      --tag ${GITHUB_VERSION} \
+      --name ${GITHUB_VERSION} \
+      --description "" \
```
nathanleclaire: I'm pretty sure we'll eventually be able to automatically generate most of the release notes here as well which is pretty cool.

nathanleclaire:
Looking like a good start.  I posted some comments but this will be pretty useful once finished.

jeanlaurent:
ready for review @docker/machine-maintainers

dgageot:
LGTM

dgageot:
Will use this on Monday for the release!

dgageot:
Next step is to make it run in a container

Issue #1591, opened 2015-07-27
docker-machine create is hanging with driver virtualbox : SSH cmd err, output: exit status 255:


opskumu:
```
# docker -v && docker-machine -v && docker-compose -v
Docker version 1.6.2, build ba1f6c3/1.6.2
docker-machine version 0.3.1 (40ee236)
docker-compose version: 1.3.3
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
# rpm -qa | grep virtual -i
VirtualBox-4.3-4.3.30_101610_el7-1.x86_64
```
 
docker-machine create is hanging，the debug logs：

```
Getting to WaitForSSH function...
Testing TCP connection to: localhost:45105
Using SSH client type: external
About to run SSH command:
exit 0
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /root/.docker/machine/machines/dev2/id_rsa -p 45105 docker@localhost exit 0] []  <nil> <nil> <nil> [] <nil> <nil> <nil> <nil> false [] [] [] [] <nil>}
SSH cmd err, output: exit status 255: 
Error getting ssh command 'exit 0' : exit status 255
Getting to WaitForSSH function...
Testing TCP connection to: localhost:45105
Using SSH client type: external
About to run SSH command:
exit 0
&{/usr/bin/ssh [/usr/bin/ssh -o PasswordAuthentication=no -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -i /root/.docker/machine/machines/dev2/id_rsa -p 45105 docker@localhost exit 0] []  <nil> <nil> <nil> [] <nil> <nil> <nil> <nil> false [] [] [] [] <nil>}
SSH cmd err, output: exit status 255: 
Error getting ssh command 'exit 0' : exit status 255
... ...
```

tehmaspc:
I believe this is the same issue I'm having on MacOSX. Have tried a few times to create a new VirtualBox Dev VM from scratch. Have cleaned out .docker/ and even downgraded from VBOX 5. 

    % docker-machine --version
[...]

PR #2614, opened 2015-12-18
Disable ControlMaster and ControlPath on SSH connections

stelcheck:
Ref: https://github.com/docker/machine/issues/1591#issuecomment-126169020

I personally wouldn't merge this yet: this is meant to fix a very specific issue in a very specific place; I don't think we should have to trade-off a responsive create for a slower (or at least forcibly slower) ssh command.

However, I do believe those two flags should be added wherever we try to boot a machine and need to configure it. 

Suggestions welcome.

### DIFF
--- a/libmachine/ssh/client.go
+++ b/libmachine/ssh/client.go
@@ -50,6 +50,8 @@ var (
 	baseSSHArgs = []string{
 		"-o", "PasswordAuthentication=no",
 		"-o", "StrictHostKeyChecking=no",
+		"-o", "ControlMaster=no",
+		"-o", "ControlPath=none",
 		"-o", "UserKnownHostsFile=/dev/null",
 		"-o", "LogLevel=quiet", // suppress "Warning: Permanently added '[localhost]:2022' (ECDSA) to the list of known hosts."
 		"-o", "ConnectionAttempts=3", // retry 3 times if SSH connection fails


stelcheck:
Yes, saw the test result for that. Already have amended the commit.

libmachine/ssh/client.go
```diff
@@ -50,6 +50,8 @@ var (
 	baseSSHArgs = []string{
 		"-o", "PasswordAuthentication=no",
 		"-o", "StrictHostKeyChecking=no",
+		"-o", "ControlMaster=no",
```
nathanleclaire: Isn't `ControlMaster=no` already set on line 59 and `ControlPath=no`?

Probably it just needs `ControlPath` line updated to `ControlPath=none`?

stelcheck:
You're right. Not sure why I missed that :/

dgageot:
@stelcheck Can you update the PR then? Thanks a lot

dgageot:
@stelcheck To make sure this is included in next release in Monday, I took the liberty to modify your PR here: https://github.com/docker/machine/pull/2682

dgageot:
Merged with https://github.com/docker/machine/pull/2682

stelcheck:
Sorry for the insufficient responsiveness. And thank you for the quick response!

Issue #2619, opened 2015-12-18
Replace Hardware Virtualization warning with an error in PreCreateCheck()


dgageot:
Now that the check for Hardware Virtualization is battle tested, let's make it fail before the VM is created instead of just printing a warning.

Also, let's add a flag to disable this check in case somebody encounters a case where the check says the machine doesn't support Hardware Virtualization and VBox still manages to boot the VM.

dgageot:
Fixes https://github.com/docker/machine/issues/2619

dgageot:
Closed with https://github.com/docker/machine/pull/2620

PR #2620, opened 2015-12-18
Don't create the VM if no hardware virtualization

dgageot:


dgageot:
Fixes #2619

jeanlaurent:
A shorter name for the flag is better, a part from that, LGTM, a much needed feature.

dgageot:
@jeanlaurent fixed

jeanlaurent:
LGTM

dgageot:
@docker/compose-maintainers We might want to hold on this one until we test this behaviour: https://www.virtualbox.org/ticket/12350
It looks like a windows machine with Hyper-v installed will break the VT-X support in VirtualBox. There's nothing we can do but we could at least detect hyperv and tell the user that maybe it should be deactivated.

nathanleclaire:
The `hyper-v` interference has been brought up before.  I do agree a check for `hyper-v` would be awesome, but why would merging this PR be blocked on that?  It seems to just add an additional flag which users can use to skip the VTX check.  To me that falls into the "I really know what I'm doing" class of users.

Issue #2504, opened 2015-12-07
Unable to create/start/stop/remove machine on Windows.


kunalkushwaha:
With patch https://github.com/docker/machine/commit/1a3c3898c12d4a3d315d8460bb48617e8ef59c2c

A bug of windows is introduced, which restricts any creation/start/stop on machine.

```
$ docker-machine start swl-consul
(swl-consul) Starting VM...
Error saving host to store: rename C:\Users\kunal\.docker\machine\machines\swl-consul\config.json.tmp162410963 C:\Users\kunal\.docker\machine\machines\swl-consul\config.json: The process cannot access the file because it is being used by another process.
```

While debugging issue I found, [``os.Remove``](https://github.com/docker/machine/blob/master/libmachine/persist/filestore.go#L49) successfully removes the config.json

Not sure, How to fix this bug as code seems correct. Issue is with ``os.Rename`` function.

kunalkushwaha:
Reverting patch https://github.com/docker/machine/commit/1a3c3898c12d4a3d315d8460bb48617e8ef59c2c works fine.

dgageot:
Closed with https://github.com/docker/machine/pull/2505

PR #2623, opened 2015-12-18
Allow concurrently saving data to file (fixes #2612)

### LINKED ISSUE #2612: 'Error saving host to store' when running 'docker-machine ip default' concurrently
I get an 'Error saving host to store' error (https://github.com/docker/machine/blob/master/commands/commands.go#L91) when I run 'docker-machine ip default' multiple times concurrently on linux.

Easy to reproduce with: `for i in 1 2 3 4 5 6 ; do docker-machine ip default & ; done`

raboof:
The explicit os.Remove is not necessary because os.Rename will already
overwrite the target path (except perhaps on plan9, but docker-machine
doesn't support that anyway - see also https://github.com/golang/go/issues/13673).

This 'fixes' race conditions when 2 processes are saving the same file
concurrently. Of course if those 2 processes attempt to write different
data this might still lead to surprising results, but in practice 'last
writer wins' seems like a sane default strategy.

Signed-off-by: Arnout Engelen <arnout@bzzt.net>

### DIFF
--- a/libmachine/persist/filestore.go
+++ b/libmachine/persist/filestore.go
@@ -49,10 +49,6 @@ func (s Filestore) saveToFile(data []byte, file string) error {
 		return err
 	}
 
-	if err = os.Remove(file); err != nil {
-		return err
-	}
-
 	if err = os.Rename(tmpfi.Name(), file); err != nil {
 		return err
 	}


nathanleclaire:
Hi @raboof, thanks for the contribution.  LWW seems sane to me for Machine's use case.  As @dgageot mentions Windows is the reason we have the file saving semantics done this way, and it's actually a fairly sensitive operation due to this file's relative importance. We are in a position where we really need to treat cross-platform safety gingerly, so we need to tread lightly here.

That said, maybe we can treat the `.tmp` file as a sort of "mutex" and block on continuing with the filesave operation if it exists (with a timeout) to solve the issue of concurrent access between separately invoked Machine processes.  I'd be worried it would make an already complex operation even harder to reason about though.  @dgageot @jeanlaurent What do you feel like is the best tradeoff here?

raboof:
Agreed entirely that we need to be careful about cross-platform safety here, thanks for being prudent.

Are we sure a rename doesn't overwrite the file on Windows? I don't have a windows machine handy, but TestRenameOverwriteDest suggests it should overwrite the destination on all platforms (except plan9, https://github.com/golang/go/blob/master/src/os/os_test.go#L775).

Strange thing is https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L423 does not look like it actually would overwrite the target file, while https://github.com/golang/go/blob/master/src/internal/syscall/windows/syscall_windows.go#L130 does. Not sure what's going on there.

askb:
@nathanleclaire - Maybe use file simple file based lock - `Flock()` before rename/remove the tmp file, not sure if `flock` is portable on Windows. 
Doesn't frequent updates to any config file call for moving away from a flat-file into a light-weight back-end store (db) in a long run ?

dgageot:
@askb @raboof `flock` isn't working properly on windows if I recall correctly. 

I would rather leave the code as it is because:
 + it is working
 + fixing race conditions caused by multiple instances of `docker-machine` running on the same host at the same time is much more work than that

raboof:
@dgageot well, the thing is, it's not working reliably ;).

As mentioned in #2612, on linux `for i in 1 2 3 4 5 6 ; do docker-machine ip default & ; done` reliably causes an error for me.

dgageot:
@raboof Sure, I understand. My point is:
 + the issue is still valid
 + the PR doesn't address the issue. And removing the `os.Remove` would make the code fail on windows as pointed [here](https://github.com/docker/machine/issues/2504)

I would clearly like to see the issue fixed. We could have more actions run concurrently then.

Clearly a file lock on the filestore or the host folder would fix the issue. The problem is that the only way I know of to provide a portable global lock with golang is to open a tcp port. Which is kind of ugly.

raboof:
@dgageot ah, interesting - so while the target file is open it can be removed but it cannot be overwritten by a rename?

If we can't come up with a file locking strategy that works cross-platform perhaps we can introduce some kind of retry mechanism?

dgageot:
@raboof You could try to use this library that supports native/emulated lock files depending on the platform: https://github.com/camlistore/lock

dgageot:
wdyt @raboof? Should we keep the PR open?

raboof:
Yes, let's close it because apparently this solution causes trouble on Windows (though I'm surprised that it apparently you can remove a locked file but not overwrite it... shame I don't have a windows box handy to test on ;) ).

PR #2625, opened 2015-12-18
FIX #2624 reject duplicate hostonlyifs name/IP

### LINKED ISSUE #2624: Multiple host-only interfaces with same name can lead to generating a new useless host-only interface each time
Consider this output for `VBoxManage list hostonlyifs`:

```
Name:            VirtualBox NDIS 6.0 Miniport Driver
GUID:            6e31e280-558a-471e-89cd-d5144bf025f1
DHCP:            Disabled
IPAddress:       192.168.99.1
NetworkMask:     255.255.255.0
IPV6Address:     fe80:0000:0000:0000:c187:1d90:6de1:a85c
IPV6NetworkMaskPrefixLength: 64
HardwareAddress: 0a:00:27:00:00:00
MediumType:      Ethernet
Status:          Up
VBoxNetworkName: HostInterfaceNetworking-VirtualBox NDIS 6.0 Miniport Driver

Name:            VirtualBox NDIS 6.0 Miniport Driver
GUID:            f6050f89-457e-40a5-9c8c-7436fa4f299c
DHCP:            Disabled
IPAddress:       192.168.56.1
NetworkMask:     255.255.255.0
IPV6Address:     fe80:0000:0000:0000:e970:91cd:7ace:351f
IPV6NetworkMaskPrefixLength: 64
HardwareAddress: 0a:00:27:00:00:00
MediumType:      Ethernet
Status:          Up
VBoxNetworkName: HostInterfaceNetworking-VirtualBox NDIS 6.0 Miniport Driver
```

Searching for an host-only interface for IP 1921.68.99.1 will fail once the other time because the interfaces are stored in a map indexed by name.

When we see interfaces with duplicated name or duplicated IP, we should fail and ask the user to clean things up.

The test for borked hostonly interfaces could even be done in `PreCreateCheck()` so that the VM is not even created.

dgageot:

### DIFF
--- a/drivers/virtualbox/network.go
+++ b/drivers/virtualbox/network.go
@@ -13,9 +13,8 @@ const (
 )
 
 var (
-	reHostonlyInterfaceCreated            = regexp.MustCompile(`Interface '(.+)' was successfully created`)
-	errDuplicateHostOnlyInterfaceNetworks = errors.New("VirtualBox is configured with multiple host-only interfaces with the same IP. Please remove all of them but one.")
-	errNewHostOnlyInterfaceNotVisible     = errors.New("The host-only interface we just created is not visible. This is a well known bug of VirtualBox. You might want to uninstall it and reinstall the version listed here: https://www.virtualbox.org/ticket/14437?cversion=0&cnum_hist=42")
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Nice!  Needs a rebase but LGTM.

PR #2627, opened 2015-12-18
FIX #2626 Retrieve windows version on non-english OS

### LINKED ISSUE #2626: Retrieve Windows version on non english machine
On non-english machines, `systeminfo` does not list the `OS version`

dgageot:

### DIFF
--- a/libmachine/mcnutils/os_windows.go
+++ b/libmachine/mcnutils/os_windows.go
@@ -6,7 +6,7 @@ import (
 )
 
 func LocalOSVersion() string {
-	command := exec.Command(`systeminfo`)
+	command := exec.Command(`ver`)
 	output, err := command.Output()
 	if err != nil {
 		return ""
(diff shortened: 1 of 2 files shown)


libmachine/mcnutils/os_windows.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 func LocalOSVersion() string {
-	command := exec.Command(`systeminfo`)
+	command := exec.Command(`ver`)
```
nathanleclaire: Any particular reason for the backticks instead of `"`?  Seems a bit odd, probably fine though.

nathanleclaire:
LGTM

nathanleclaire:
Argh.  It's probably not immediately problematic but, FYI I have `systeminfo` in my git bash shell but not `ver`.  Since msysgit is our recommended way to use Machine on Windows, why not try both?

PR #2629, opened 2015-12-18
FIX #2617 Check hostonlyif after its creation

### LINKED ISSUE #2617: Virtualbox can't create a hostonlyif
Sometimes, depending on the OS and the version of Virtualbox, `VBoxManage list hostonlyifs` will always return an empty list and newly created host-only interfaces cannot be used.

 + Each time we create a host-only interface, we should check it's listed in `VBoxManage list hostonlyifs`.
 + We should give a clear message to the user and explain how Virtualbox could be reinstalled or upgraded to a known working version.
 + It would even be better if we could check that the host-only interface creation mechanism works properly in the `PreCreateCheck()` phase

see https://www.virtualbox.org/ticket/14437?cversion=0&cnum_hist=42

dgageot:


nathanleclaire:
_Supposedly_ the issue is fixed in VirtualBox 5.0.12, but I suppose it can't hurt to check.  We can just take it out later if we run into any issues.

LGTM

PR #2635, opened 2015-12-19
Add code to ensure RPC server cleanup in happy path

nathanleclaire:
I noticed in going over some of the RPC work that we rarely activate the cleanup of the plugin servers ourselves in commands.  This is, of course, not good (we should always cleanup after ourselves and let the heartbeat timeout take care of the rest if there is a panic / non-normal exit on client side).

This PR adds code to do so.  Combined with another PR I have in the pipe to fix truncated logs, it will help aid in diagnosing problems with plugins and with the RPC itself.

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -22,6 +22,8 @@ func cmdActive(c CommandLine, api libmachine.API) error {
 		return fmt.Errorf("Error getting active host: %s", err)
 	}
 
+	defer libmachine.CloseHosts(api, hosts)
+
 	items := getHostListItems(hosts, hostsInError)
 
 	for _, item := range items {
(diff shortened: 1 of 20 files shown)


nathanleclaire:
The `defer api.CloseHosts` idiom is intentionally similar to other Golang idioms such as `defer file.Close()`.

dgageot:
@nathanleclaire I would really prefer if we had a single place where this takes place. Like at the end of the `fatalOnError` function for example or in an `After` hook in codegangsta/cli. Wdyt?

dgageot:
@nathanleclaire the increased heartbeat timeout make it mandatory to implement proper closing in happy path. I came up with a simpler solution [here](https://github.com/docker/machine/pull/2681). I hope you won't hold it against me if I close this PR!

dgageot:
Fixed with https://github.com/docker/machine/pull/2681

PR #2636, opened 2015-12-19
Increase heartbeat interval and timeout

nathanleclaire:
@abronan suggested the current timeout is very aggressive and I'm inclined to agree.  This may help us with some of the issues we're seeing with RPC since it eases up a bit on the server.  If it seems to be too long, we can revert.


dgageot:
LGTM

Issue #2622, opened 2015-12-18
The output from the drivers is truncated when a create fails


dgageot:
When a Create fails, the end of the driver's output is not captured. This means that:
 + The user sometimes can't see what caused the creation failure
 + The bugsnag reports miss the most important part

nathanleclaire:
Dupe of: https://github.com/docker/machine/issues/2217

I have what seems to be a fix for this in the pipe.

dgageot:
Closing. Duplicate of @22217

Issue #2217, opened 2015-11-07
Exits from program sometimes cut off plugin output


nathanleclaire:
I'm not sure the exact solution since the concurrency around this is sort of tricky, but I notice that sometimes if the drivers exit with an error the scanners for STDOUT and STDERR don't have a chance to catch up.  So, if the plugin forces the main program to exit through an error etc., the logs will not be complete.  It's not a brutal bug but it's pretty annoying and I don't want to lose any information that we could be capturing for debugging. cc @docker/machine-maintainers

nathanleclaire:
@dgageot I think this is closed now, no?

PR #2637, opened 2015-12-19
Fix truncated plugin binary logs

nathanleclaire:
Closes https://github.com/docker/machine/issues/2622
Closes https://github.com/docker/machine/issues/2217

Simplifies some logic in the "plugin binary" client side code, and critically adds a `cmd.Wait()`, a few revisions to ensure things happening in the proper order, and a message to verify that we can see the successful exit of the plugin in the logs.

### DIFF
--- a/libmachine/drivers/rpc/client_driver.go
+++ b/libmachine/drivers/rpc/client_driver.go
@@ -173,12 +173,6 @@ func (c *RPCClientDriver) Close() error {
 	c.heartbeatDoneCh <- true
 	close(c.heartbeatDoneCh)
 
-	log.Debug("Making call to close connection to plugin binary")
-
-	if err := c.plugin.Close(); err != nil {
-		return err
-	}
-
 	log.Debug("Making call to close driver server")
 
 	if err := c.Client.Call(CloseMethod, struct{}{}, nil); err != nil {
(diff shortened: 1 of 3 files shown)


libmachine/drivers/rpc/client_driver.go
```diff
@@ -173,12 +174,6 @@ func (c *RPCClientDriver) Close() error {
...
 
-	log.Debug("Making call to close connection to plugin binary")
-
-	if err := c.plugin.Close(); err != nil {
-		return err
-	}
-
 	log.Debug("Making call to close driver server")
 
 	if err := c.Client.Call(CloseMethod, struct{}{}, nil); err != nil {
```
nathanleclaire: This RPC call should happen _first_, because otherwise we will actually exit the process and not be able to successfully make the call to the server.

nathanleclaire:
Any suggestions on how to test this (with `-race` it has no issues, seems to stream correctly, and I can always see the "ack" that the plugin is exiting) are welcome, it's definitely pretty sticky code.  There are some cases we might lose some data still but with this and the other PR to work on cleanly closing the servers we look to be in a lot better shape.

nathanleclaire:
OK, it's clear that I'm not actually going to be able to wrap this PR before I go on vacation next week unfortunately.  @dgageot @jeanlaurent you're welcome to take a crack at fixing the test if you like -- the logic in the PR code itself seems sound results-wise, though naturally it's a really tricky section of the code.  I think the concurrent logic in `plugin.go` is overly complex and will need to be refactored.  

If you do carry this PR, make sure to build the binary with `-race` flag and do some creates.  There are insidious race conditions in this code.

dgageot:
Merged with https://github.com/docker/machine/pull/2697

PR #2646, opened 2015-12-21
FIX #2643 Ignore empty ssh key

### LINKED ISSUE #2643: scp command requires a certificate path
We don't pass an SSH private key path when creating a machine.  Therefore there is no id_rsa in the docker machine directory. 

The scp command throws a warning because there is no key file supplied after the -i.

I think that the getInfoForScpArg in scp.go should check to see if GetSSHKeyPath is an empty string and not set the -i command.

ssh.go already does something similar.

    for _, privateKeyPath := range auth.Keys {
		if privateKeyPath != "" {
			args = append(args, "-i", privateKeyPath)
		}
	}

dgageot:


jeanlaurent:
LGTM

PR #2647, opened 2015-12-21
Try multiple commands to find windows version

dgageot:


jeanlaurent:
Bringing some code from me back from the dead... LGTM

PR #2654, opened 2015-12-22
Change host restart to use the driver implementation

dgageot:


dgageot:
That's a follow up to https://github.com/docker/machine/pull/2225

jeanlaurent:
Shouldn't we change restart for most driver then ?

dgageot:
I've tested `virtualbox`, `digitalocean`, `vmwarefusion`, `google` and `generic`. I've reviewed the other drivers and the code seems to make sense.

jeanlaurent:
Yeah did the same, the thing is, before we merge this PR, there is probably no-one beside a few hypothetic integration tests which did run this code. Code looks right but...

jeanlaurent:
LGTM

PR #2656, opened 2015-12-22
FIX #2655 Add quotes to the eval cmdline if there are spaces in the path

### LINKED ISSUE #2655: Missing quotes around docker-machine binary in `docker-machine env` hints
On windows, with bash, running `docker-machine env default` with docker-machine installed in `C:\Program Files\Docker Toolbox\docker-machine.exe`, gives this invalid hints:

```
$ eval "$(C:\Program Files\Docker Toolbox\docker-machine.exe env default)"
bash: C:Program: command not found
```

Same problem on linux/osx bash if `docker-machine` is in a path that contains spaces.

I believe, the hints should be:

```
eval $("C:\Program Files\Docker Toolbox\docker-machine.exe" env default)
```

with quotes around the binary and no quote around the `$(...)`

dgageot:

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -227,6 +227,11 @@ func (g *EnvUsageHintGenerator) GenerateUsageHint(userShell string, args []strin
 	cmd := ""
 	comment := "#"
 
+	dockerMachinePath := args[0]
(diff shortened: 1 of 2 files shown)


jeanlaurent:
LGTM

PR #2669, opened 2015-12-23
FIX #2663 reduce calls to the driver in ls

### LINKED ISSUE #2663: docker-machine ls in timeout with Hyper-v
Getting the ls information from a list a three hyperv hosts is so slow that the 10s timeout is not enough.
Could this be improved without having to increase the timeout.

dgageot:
Current code needs 5 calls to the driver to collect the state, the url and the docker version.

This PR creates a *fast path* that needs only 2 calls for a sane running machine.
Worst case will require 3 calls.


jeanlaurent:
\o/

LGTM

PR #2673, opened 2015-12-23
FIX #2271 don't build 32bits binaries for linux/osx

### LINKED ISSUE #2271: Still provide docker-machine with a GOARCH=386 for Linux & Darwin ?
We actually provide 6 binaries : `darwin`, `window` & `linux` each of them for two architectures `amd64` & `386`.

Two questions, open for discussion
* It sounds like docker engine project does [not provide](https://github.com/docker/docker/releases/tag/v1.9.0) `386` `linux` binaries, so why should we ? 
* Its sounds like the last mac running a 32bits processor has been out in [2005](http://apple.stackexchange.com/questions/99640/how-old-are-macs-that-cannot-run-64-bit-applications/99644#99644), why should still provide theses ? (Although the docker engine project still provide them, we should ask them)

If we have users on these, fair enough, let's keep them, if we don't maybe it's time to simplify this a bit. The main advantage is to speed up the cross-os projects, simplify the build and tests.

dgageot:

### DIFF
--- a/mk/build.mk
+++ b/mk/build.mk
@@ -1,11 +1,15 @@
 extension = $(patsubst windows,.exe,$(filter windows,$(1)))
 
+# Valid target combinations
+VALID_OS_ARCH := "[darwin/amd64][linux/amd64][windows/amd64][windows/386]"
+
 define gocross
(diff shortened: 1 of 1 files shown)


jeanlaurent:
LGTM

PR #2680, opened 2015-12-24
Better error when swarm is not started

dgageot:
When trying to check connection to a swarm cluster, we get an "invalid certs" errors if the problem comes from swarm containers being down. 
This PR gives a better error message if swarm connection is down but connection to the docker socket is up.


jeanlaurent:
I would have merged first and last commit together, as cleanup.

LGTM

PR #2681, opened 2015-12-24
Close drivers on exit

dgageot:
This is a simpler version of https://github.com/docker/machine/pull/2635
It makes sure we don't have to call close everywhere we load a host.


jeanlaurent:
LGTM

nathanleclaire:
Did you running (including `create`) after building with `-race`?  Similar solutions in the past for me have tripped the race detector (due to the shared memory access), but this seems to happen all in a single goroutine so maybe it is fine.

PR #2682, opened 2015-12-24
Disable ControlPath on SSH connections

dgageot:
This is a fix PR for https://github.com/docker/machine/pull/2614

It turns out that `ControlMaster was` already set to `no` and that `ControlPath` was set to `no` instead of `none` (as found in man ssh_config)

Signed-off-by: Marc Trudel <mtrudel@wizcorp.jp>


jeanlaurent:
LGTM

PR #2685, opened 2015-12-24
Release followup

jeanlaurent:
A couple of things to enhance the release process. Most notably the 76923dd which remove a workaround which was mandatory until the release template lands in master.


dgageot:
LGTM

PR #2691, opened 2015-12-26
Standardize --driver-memory-size flag including ENV vars

jgeiger:
- Some drivers use --driver-memory but have the ENV variable as
  DRIVER_MEMORY_SIZE. This request makes all drivers use the
  --driver-memory-size flag, but still allows for non-conforming
  drivers to use the --driver-memory flag until deprecated.
  In the case where --driver-memory-size matches the defaultMemory,
  use the --driver-memory flag. If neither flag is used, the
  memory is still set to defaultMemory.

### DIFF
--- a/docs/drivers/hyper-v.md
+++ b/docs/drivers/hyper-v.md
@@ -27,7 +27,7 @@ Options:
 -   `--hyperv-boot2docker-url`: The URL of the boot2docker ISO.
 -   `--hyperv-virtual-switch`: Name of the virtual switch to use.
 -   `--hyperv-disk-size`: Size of disk for the host in MB.
--   `--hyperv-memory`: Size of memory for the host in MB.
+-   `--hyperv-memory-size`: Size of memory for the host in MB.
 -   `--hyperv-cpu-count`: Number of CPUs for the host.
 
 Environment variables and default values:
@@ -37,5 +37,5 @@ Environment variables and default values:
 | `--hyperv-boot2docker-url` | `HYPERV_BOOT2DOCKER_URL` | _Latest boot2docker url_ |
 | `--hyperv-virtual-switch`  | `HYPERV_VIRTUAL_SWITCH`  | _first found_            |
 | `--hyperv-disk-size`       | `HYPERV_DISK_SIZE`       | `20000`                  |
-| `--hyperv-memory`          | `HYPERV_MEMORY`          | `1024`                   |
+| `--hyperv-memory-size`     | `HYPERV_MEMORY`          | `1024`                   |
 | `--hyperv-cpu-count`       | `HYPERV_CPU_COUNT`       | `1`                      |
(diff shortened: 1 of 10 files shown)


nathanleclaire:
Hi @jgeiger, thanks for the contribution. It's a breaking change to the CLI so I'm pretty hesitant about it.

jgeiger:
The new flags are added in addition to the old ones. If the new one isn't a default value, it gets used, otherwise it drops into the old flag and then default.

The non-size flags should be listed as deprecated at some point just to get things standardized.

thaJeztah:
@nathanleclaire one option would be to use the docker `mflag` package, which allows [outputting a deprecation message for updated flags](https://github.com/docker/docker/blob/927b334ebfc786276a039e45ec097e71bf9a104c/pkg/mflag/flag.go#L25-L34)

dgageot:
@jgeiger Like @nathanleclaire, I'm hesitant about this kind of improvements, it has little short term value, adds complexity to the code and can fail in ways we don't expect. On the other hand, improving the UX is very important.

This makes me think that every driver flag has the wrong name right now. All of them are prefixed with the driver name, which made sense before 0.5.0 but is useless and cumbersome now.

Could we come up with a global mechanism to support flag renaming?

/cc @thaJeztah

thaJeztah:
From a UX perspective, dropping the prefix would definitively be better. Some things to consider / take into account here;

- In docker "engine", we defined a [deprecation policy](https://docs.docker.com/engine/misc/#feature-deprecation-policy) that allows users to adopt; i.e., announce the old flags as deprecated, but keep them around for 2 more releases
- I suggested the mflags package; it served the engine well, but I haven't looked into the code quality, it may need some TLC (I think it hasn't been updated in a while :smile:)
- If the driver-prefix is dropped, it's important that the flags have the same behavior across drivers (accept the same kind of options / values) to prevent unexpected behavior for the end-user; can we guarantee that? We may need to define that behavior (describe in detail what a flag does, how it should work), so that plugin developers can take that into account.
- Flags that are specific to / only supported on a specific driver, should still be prefixed with the driver-name. Alternatively, introduce a "generic" flag that allows passing driver-specific options (e.g. `--driver-opt=somedriver.someopt=somevalue`)

...Just some thoughts

nathanleclaire:
My opinion is that, although the extra consistency would be nice, the amount of risk/effort in deprecating the old flags does not outweigh the reward.

dgageot:
@jgeiger I guess we are not ready to merge this PR. Sorry for the hard work. We do understand exactly your goal of bringing more consistency between drivers. But the risk/effort is clearly not in favour of merging as @nathanleclaire says some times ago.
I'd rather let you a chance to advocate this PR or close it by yourself, that's a bit more fair, I think!

jgeiger:
I figure this can be closed since it has little chance of being merged anyway. I do think there should be standards, and I like the idea of removing the specific prefixes for things that are common across drivers.

dgageot:
Thank you @jgeiger for your positive attitude. codegangsta/cli replacement is on my todo list. When it's done we should be able to improve flags consistency

PR #2704, opened 2015-12-28
Print a warning for Boot2Docker v1.9.1 (AUFS issue)

AkihiroSuda:
Print a warning for Boot2Docker v1.9.1, which is known to have an issue with AUFS (docker/docker#18180)

### DIFF
--- a/libmachine/mcnutils/b2d.go
+++ b/libmachine/mcnutils/b2d.go
@@ -34,6 +34,12 @@ var (
 You may be getting rate limited by Github.`)
 )
 
+var (
+	AUFSBugB2DVersions = map[string]string{
+		"v1.9.1": "https://github.com/docker/docker/issues/18180",
+	}
+)
+
 func defaultTimeout(network, addr string) (net.Conn, error) {
 	return net.Dial(network, addr)
 }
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Hm, it seems reasonable to do in the interim (until Docker >1.9.1 is release and hopefully a corresponding fixed patch with boot2docker).

libmachine/mcnutils/b2d.go
```diff
@@ -144,6 +144,12 @@ func (b *b2dReleaseGetter) getReleaseURL(apiURL string) (string, error) {
 	}
 
 	log.Infof("Latest release for %s/%s/%s is %s", host, org, repo, tag)
+	if tag == "v1.9.1" {
+		log.Warnf("Boot2Docker v1.9.1 has a known issue with AUFS. " +
```
nathanleclaire: Some nitpicks:

- Can you please make "v1.9.1" a `const` in top level of the file named something like `aufsBugB2DVersion` and use that here?
- Why not use backticks for the multi-line strings instead and only have one `Warnf` call? And also why not make it a suggestion instead of a command?  Also, since you're not interpolating anything, please use `Warn` instead of `Warnf`.  i.e.:

```go
log.Warn(`
Boot2Docker v1.9.1 has a known issue with AUFS.
See here for more details: https://github.com/docker/docker/issues/18180
Consider specifying another storage driver (e.g. 'overlay') using '--engine-storage-driver' instead.
`)
```

AkihiroSuda:
@nathanleclaire 
Thanks, updated PR.
I use `AUFSBugB2DVersions` map for future extensibility.

nathanleclaire:
LGTM thanks

cc @dgageot

dgageot:
LGTM

PR #2708, opened 2015-12-29
Add CI tests for an external driver

dgageot:

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,11 +1,9 @@
-# Pony-up!
 machine:
   pre:
   # Install gvm
     - bash < <(curl -s -S -L https://raw.githubusercontent.com/moovweb/gvm/1.0.22/binscripts/gvm-installer)
   # Install bats
     - git clone https://github.com/sstephenson/bats.git && cd bats && sudo ./install.sh /usr/local
-    # - sudo apt-get install -y virtualbox
 
   post:
     - gvm install go1.5.2 -B --name=stable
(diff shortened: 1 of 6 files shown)


circle.yml
```diff
@@ -31,23 +27,12 @@ test:
     - gvm use stable && go version
     - gvm use stable && make build:
         pwd: $BASE_STABLE
+    - gvm use stable && GO15VENDOREXPERIMENT=1 go get github.com/dgageot/docker-machine-driver-none-external
```
nathanleclaire: I know this is convenient, but can we find a way to do this without referencing a repository outside of this one? (e.g. using `cp -r` and `go install` from a subfolder of this repo)  If this repository needs to be updated it should be doable by anyone who has access to the Machine repo.

dgageot: I fear that having a folder in the `docker/machine` repo that contains a copy of some packages/dependencies of `docker/machine` itself would be a bit surprising. By having an external repo, we also protect ourselves from unwanted modifications during a refactoring. Could we have a repo under `docker` organization?

nathanleclaire: Yes, I think a repo under `docker` organization might be the best approach.

jeanlaurent:
LGTM

nathanleclaire:
Sent an email asking to get a new repo created for this functionality, after we accomplish that I'm good to go on this.  Thanks!

circle.yml
nathanleclaire: @dgageot https://github.com/docker/docker-machine-driver-ci-test (for reference)

nathanleclaire:
LGTM after repo bump :+1:

jeanlaurent:
@nathanleclaire I believe the repo has been up, I let you merge if you feel like it.

PR #2711, opened 2015-12-29
We must never ask for a password interactively

dgageot:


nathanleclaire:
LGTM, I'd like to support passphrase encoded SSH keys though :(

PR #2721, opened 2016-01-01
Fix #2717, Driver vmwarefusion fail because of soft link

### LINKED ISSUE #2717: Driver vmwarefusion fail because of soft link
- docker-machine 0.55 & 0.6.0-dev

```
docker-machine create -d vmwarefusion  vmdocker
```
This command will create a docker machine, but fail after waiting the machine up, this is because vmrun return absolute path, but fusion_darwin.go:GetState want  a path with soft link, so GetState() will always fail.

Source fusion_darwin.go
```
func (d *Driver) GetState() (state.State, error) {
	if stdout, _, _ := vmrun("list"); strings.Contains(stdout, d.vmxPath()) {
		return state.Running, nil
	}
	return state.Stopped, nil
}
```

Because I use external storage as my home directory, vmrun("list") return: 

```
Total running VMs: 1
/Volumes/exSSD/lijian/.docker/machine/machines/vmdocker.vmx
```

`d.vmxPath()` on my osx is:
```
/Users/lijian/.docker/machine/machines/vmdocker.vmx
```

I made a simple fix which only check the filename, work for me, but not the best solution: 

```
diff --git a/drivers/vmwarefusion/fusion_darwin.go b/drivers/vmwarefusion/fusion_darwin.go
index 8d17660..3bea878 100644
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -200,7 +200,7 @@ func (d *Driver) GetIP() (string, error) {

 func (d *Driver) GetState() (state.State, error) {
        // VMRUN only tells use if the vm is running or not
-       if stdout, _, _ := vmrun("list"); strings.Contains(stdout, d.vmxPath()) {
+       if stdout, _, _ := vmrun("list"); strings.Contains(stdout, fmt.Sprintf("%s.vmx", d.MachineName)) {
                return state.Running, nil
        }
        return state.Stopped, nil
```

rediceli:
Fix #2717

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strings"
@@ -200,7 +201,11 @@ func (d *Driver) GetIP() (string, error) {
 
 func (d *Driver) GetState() (state.State, error) {
 	// VMRUN only tells use if the vm is running or not
-	if stdout, _, _ := vmrun("list"); strings.Contains(stdout, d.vmxPath()) {
+	vmxp, err := filepath.EvalSymlinks(d.vmxPath())
+	if err != nil {
+		return state.Error, err
+	}
+	if stdout, _, _ := vmrun("list"); strings.Contains(stdout, vmxp) {
 		return state.Running, nil
 	}
 	return state.Stopped, nil


dgageot:
Thanks a lot @rediceli. I'd rather you don't ignore the error returned by `filepath.EvalSymlinks`. Can you fix the code?

dgageot:
LGTM

jeanlaurent:
@rediceli can you change the return state and we should be good to go.

thanks

dgageot:
@rediceli It looks like you pushed after @jeanlaurent's commit but with the initial code. Is it what you wanted?

rediceli:
It seems like I made a mistake.  :）

dgageot:
LGTM (for real now!)

jeanlaurent:
LGTM

PR #2724, opened 2016-01-03
Auto-detect shell on Windows

StefanScherer:
This PR adds auto-detection of the shell if running on Windows. So Windows users no longer have to add the `--shell cmd` or `--shell powershell` to the `docker-machine env` command.

If no `SHELL` environment is set (cygwin, fish, ...) it checks the parent process name to be one of `powershell` or `cmd` and returns this as the shell.

Users still can use the `--shell` parameter to specify the shell.

This is how it looks like within a plain PowerShell and a plain CMD shell:

bildschirmfoto 2016-01-03 um 02 38 32

The code to retrieve the parent process is already implemented in GO, but to retrieve the process name I had to re-implement the function [syscall.getProcessEntry()](https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L945-L965) that is only private. With that I could read the exe filename of the parent process.

Thanks to @inconshreveable (the author of ngrok) and his blog post [Sweat the small stuff](https://inconshreveable.com/09-09-2014/sweat-the-small-stuff/) where I found that he implemented the Windows version of the GO function `os.Getppid()`. The mousetrap repo is worth mentioning to work around the problem that Windows users unfamiliar with command line tools will often "double-click" the executable for a tool.

### DIFF
--- a/commands/detectshell.go
+++ b/commands/detectshell.go
@@ -1,0 +1,24 @@
+// +build !windows
+
+package commands
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func detectShell() (string, error) {
+	shell := os.Getenv("SHELL")
+
+	if shell == "" {
+		fmt.Printf("The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.\n\n")
+		return "", ErrUnknownShell
+	}
+
+	if os.Getenv("__fish_bin_dir") != "" {
+		return "fish", nil
+	}
+
+	return filepath.Base(shell), nil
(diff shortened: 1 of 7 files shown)


thaJeztah:
Awesome! Looking at this, I wonder if this should be a `pkg`; This looks like very generic code, and I can imagine this being useful for other projects as well :+1:

dgageot:
LGTM
Nice work @StefanScherer!
Tested git bash, cmd and powershell on win10. All good.

jeanlaurent:
LGTM

thanks  @StefanScherer

PR #2730, opened 2016-01-04
Close api

dgageot:
This makes the libmachine Api store/close the opened plugins, instead of relying on a global variable.


jeanlaurent:
I took the time to read it all. LGTM

Code change is quite complex. I think it's worth that @nathanleclaire gives is voice on it.

commands/commands.go
```diff
@@ -98,7 +97,8 @@ func runAction(actionName string, c CommandLine, api libmachine.API) error {
 
 func fatalOnError(command func(commandLine CommandLine, api libmachine.API) error) func(context *cli.Context) {
 	return func(context *cli.Context) {
-		api := libmachine.NewClient(mcndirs.GetBaseDir())
+		api := libmachine.NewClient(mcndirs.GetBaseDir(), mcndirs.GetMachineCertDir())
```
nathanleclaire: Why did the dir passed in here change?

nathanleclaire:
Nice. :+1: 

Any ideas how to unit test this to ensure that all plugin servers have indeed been shut down?

LGTM

nathanleclaire:
BTW running with `-race` this PR doesn't report any races on `create`, which is good news.

commands/commands.go
dgageot: Yes. To be honest, it has to be improved in a near future. I'm thinking about how we could pass a `MachinesDirs` struct that would be used in different places.

nathanleclaire:
Hey @dgageot, I'm confused about what happened in this PR.  It says that you merged in one commit above, but there are 4 commits in this PR, and it is still open.

Should this be rebased and/or merged?

dgageot:
I have no idea. I see all 4 commits merged in master and the PR is marked as closed for me...

nathanleclaire:
OK maybe Github hiccup or me mis-reading things.  It looks good now, thanks.

PR #2733, opened 2016-01-04
Run docker-machine from the PATH for core drivers

dgageot:
When external code uses `docker-machine` as a library, it should spawn core-drivers from the `docker-machine` binary that is in the PATH, not the current binary (`os.Args[0]`) which will not contain any driver.

Without this, no external code can use libmachine.

### DIFF
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -88,6 +88,8 @@ func main() {
 		return
 	}
 
+	localbinary.CurrentBinaryIsDockerMachine = true
+
 	setDebugOutputLevel()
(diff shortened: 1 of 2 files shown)


jeanlaurent:
Fix the issue, but gosh this is ugly.

LGTM

dgageot:
@jeanlaurent Yes, I couldn't find any other way to solve the issue.

jeanlaurent:
I don't have a better idea, let's merge this, and have some discussion around wether or not we should change our architecture on this.

PR #2743, opened 2016-01-05
Crash reporting shouldn't use global variables

dgageot:


nathanleclaire:
Oh, awesome!  Seems to be almost exactly what I was referencing in that one thread.

nathanleclaire:
LGTM :+1:

PR #2744, opened 2016-01-05
Tell the user that Hyperv could cause VBox to fail

dgageot:
Sometimes we can't check if VTX is disabled but we should fail anyway if hyperv is installed because we know it'll fail.


nathanleclaire:
LGTM

PR #2750, opened 2016-01-05
Add option to create amazon ec2 instances with the EbsOptimized setting.

### LINKED ISSUE #1904: ec2 driver should allow to pass flag to request EBSOptimized instance
The default ami used has an EBS file system.  When creating a instance AWS allows you to specify 

"EbsOptimized": [
        "true"
    ],

This results in a much better performing instance as I/O between the host and the disk is much faster.

jeffellin:
fixes #1904 Non EbsOptimized instances will be created by default but adding the flag
UseEbsOptimizedInstance will create and EBSOptimized Instance if the instance type supports it.

### DIFF
--- a/docs/drivers/aws.md
+++ b/docs/drivers/aws.md
@@ -54,6 +54,7 @@ This example assumes the VPC ID was found in the `a` availability zone. Use the`
 -   `--amazonec2-use-private-address`: Use the private IP address for docker-machine, but still create a public IP address.
 -   `--amazonec2-private-address-only`: Use the private IP address only.
 -   `--amazonec2-monitoring`: Enable CloudWatch Monitoring.
+-   `--amazonec2-use-ebs-optimized-instance`: Create an EBS Optimized Instance, instance type must support it.
 
 By default, the Amazon EC2 driver will use a daily image of Ubuntu 15.10 LTS.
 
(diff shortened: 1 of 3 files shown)


dgageot:
Thank you @jeffellin, can you update the documentation accordingly?

jeffellin:
Absolutely

jeffellin:
@dgageot let me know if there is anything else I can do to move this along.

jeffellin:
Not sure what this is all about 
```running hugo
Logging to /tmp/hugo322525699
INFO: 2016/01/06 Using config file: config.toml
INFO: 2016/01/06 syncing from /docs/themes/hugo-foundation/static to /docs/public/
INFO: 2016/01/06 syncing from /docs/static/ to /docs/public/
initialize & template prep:
	47.248985ms (47.762314ms)	    8.67 MB 	21975 Allocs
ERROR: 2016/01/06 loadData at 
load data:
	556.398µs (48.394501ms)	    0.02 MB 	185 Allocs
FATAL: 2016/01/06 mkdir /docs/public/engine/security/trust: no such file or directory
ERRORS = 1```

dgageot:
@jeffellin Looks like the problem is on our side. Can you try to change the amend your commit (to change its commit date) and force push to your branch? This should trigger checks again.

jeffellin:
@dgageot looks like its still having issues.

jeanlaurent:
I did some basic tests 
 1. image created with the flag are not EBS-optimized.
 1. the integration tests is non-functional.
- You should not put the test in `test/integration/drivers/amazonec2` but in `test/integration/amazonec2`
- `force_env` does not exist anymore and has been replaced by `only_if_env`
- `machine` should not be prefixed by `run`
- the `--amazonec2-subnet-id` is now a mandatory flag for the aws driver.

jeffellin:
@jeanlaurent I am going to make subnet set by env variable.  Does this make sense? I assume you can set it for your environment?

jeanlaurent:
@jeffellin yes

jeffellin:
@jeanlaurent please take another look,  thanks

jeanlaurent:
Code works, but IT test fails.

Going to LGTM this, and push a test fix right after.

dgageot:
LGTM. if the test is fixed in another PR

PR #2752, opened 2016-01-06
Support creating a machine from an existing google VM

dgageot:
This is very useful if you want to create a VM with options that are not supported by `docker-machine create`. In the future, all drivers could implement this and it would lower the urge on supporting every single option of each cloud provider.

### DIFF
--- a/docs/drivers/gce.md
+++ b/docs/drivers/gce.md
@@ -38,7 +38,7 @@ To create a machine instance, specify `--driver google`, the project id and the 
 
 ### Options
 
--   `--google-project`: **required** The id of your project to use when launching the instance.
+    -   `--google-project`: **required** The id of your project to use when launching the instance.
     -   `--google-zone`: The zone to launch the instance.
     -   `--google-machine-type`: The type of instance.
     -   `--google-machine-image`: The absolute URL to a base VM image to instantiate.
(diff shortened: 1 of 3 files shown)


jeanlaurent:
The original idea was to discuss was the inclusion of a `register` command which goal is to register and provision a previously existing host. This would be probably help in toning down the number of options we should support for each cloud driver as well as help register super specific created host.

The process would be that we support basic options to setup a host on the various cloud providers ( we can even maybe dream of sharing the same flag name ) and if people need very specific option, that we let them register AND provision (ie install docker ourselves) the host.

This implementation goes one way further, by allowing the create function to optionally support registering a machine. IMHO this sounds more pragmatic than adding a new command that each driver should implement. This is enforced by the fact that a lot of 'local' drivers would probably not need the 'register' function (vbox, fusion, hyper-v) and that the code is very very similar to the one of the create command.

Here's my LGTM, this needs discussion though with @docker/machine-maintainers.

nathanleclaire:
Interesting idea -- my preference would be to have a proposal first so we can discuss the approach and whether or not we want to include the feature (just to save you some work in case we decide against it).

I'm not opposed, but I do have a few concerns:

- Can we add some tests (at least unit, ideally integration) that ensure this type of workflow doesn't break in the future?  I feel like we should be protecting what we have before adding new functionality.
- Is the plan to support this in all drivers eventually?

jeanlaurent:
I was probably a bit too lyric on this.

IMHO This PR should be merged, we should learn from it's usage by the users and then make big plan on where we should go with this 'register' feature.

If we have to back down on this, this is not going to be a big deal.

LGTM

nathanleclaire:
I'm interested in the feature, but I think we should discuss the game plan (e.g. do we add such functionality for all drivers) before merging.  Let's count on discussing it at next standup.

nathanleclaire:
Oops, didn't discuss during standup today.  Tomorrow during planning perhaps.

dgageot:
@nathanleclaire @jeanlaurent Do you think we should rename the `--google-use-existing` to something like `--google-experimental-use-existing`? Or maybe just add a note to the documentation?

nathanleclaire:
@dgageot I think adding a note to the documentation and/or EXPERIMENTAL.md file should suffice.

dgageot:
@nathanleclaire Added a note in the gce doc.

jeanlaurent:
LGTM thx for the note in the doc

nathanleclaire:
Cool :+1:

PR #2759, opened 2016-01-06
Better windows version detection

dgageot:
Trying to improve windows version detection.
Also move the version detection code to a package that is not used by drivers


jeanlaurent:
LGTM

PR #2763, opened 2016-01-06
Retry VirtualBox commands

dgageot:

### DIFF
--- a/drivers/virtualbox/vbm.go
+++ b/drivers/virtualbox/vbm.go
@@ -11,11 +11,15 @@ import (
 
 	"strconv"
 
+	"time"
+
 	"github.com/docker/machine/libmachine/log"
 )
 
 const (
 	retryCountOnObjectNotReadyError = 5
+	objectNotReady                  = "error: The object is not ready"
+	retryDelay                      = 100 * time.Millisecond
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jeanlaurent:
By reading the title, I guessed that you would add some sort of wait time and maximum retry, but there is none.

Are we sure that we aren't going to have an infinite loop ?

dgageot:
No just a retry ;-) So, yes, the infinite loop can happen in theory but won't in real life. From my tests, the problem happens less than 1% of the time

jeanlaurent:
LGTM

nathanleclaire:
Can we please have a retry and break on error, even if it's only for a small amount?  I know the chances of an infinite loop seem small, but I think the awfulness of hitting an infinite loop is worth adding a few lines to prevent it.  After adding max retries, LGTM.

nathanleclaire:
Also, we should consider sleeping for a small amount before each retry, otherwise we will hammer `VBoxManage` any time we try to request an action and the object is locked.

nathanleclaire:
BTW, thanks for adding this.  It's very confusing to users when they get these kinds of messages.

dgageot:
Two against one ;-) OK, I'll had a retry count and a small delay between retries.

dgageot:
While I was at it, I added unit tests

jeanlaurent:
:+1: for the retries

LGTM

PR #2764, opened 2016-01-06
Replace samalba/dockerclient with docker/engine-api

dgageot:


jeanlaurent:
@dgageot Linting got you :smile: `var imageId should be imageID`

libmachine/provision/configure_swarm.go
```diff
@@ -41,70 +43,80 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
...
 	}
 
-	workerHostConfig := dockerclient.HostConfig{
-		RestartPolicy: dockerclient.RestartPolicy{
-			Name:              "always",
-			MaximumRetryCount: 0,
+	hostBind := fmt.Sprintf("%s:%s", dockerDir, dockerDir)
+	masterHostConfig := &container.HostConfig{
+		RestartPolicy: container.RestartPolicy{
+			Name: "always",
```
nathanleclaire: Would be worth checking if the restart policies have constants defined in `container` instead of simply strings.

libmachine/provision/configure_swarm.go
```diff
@@ -41,70 +43,80 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
...
+	cmd := fmt.Sprintf("manage --tlsverify --tlscacert=%s --tlscert=%s --tlskey=%s -H %s --strategy %s --advertise %s",
+		authOptions.CaCertRemotePath,
+		authOptions.ServerCertRemotePath,
+		authOptions.ServerKeyRemotePath,
+		swarmOptions.Host,
+		swarmOptions.Strategy,
+		advertiseInfo,
+	)
+
+	cmdMaster := strings.Fields(cmd)
```
nathanleclaire: Since these are going to become a string slice anyway now, maybe you could instantiate the `cmd` as a `[]string` directly instead of splitting it here?

nathanleclaire:
Nice!

Some minor feedback but LGTM whenever you fix the lint issue.

libmachine/provision/configure_swarm.go
dgageot: I checked already and I couldn't find one

libmachine/provision/configure_swarm.go
dgageot: Will do. Thanks for the feedback.

dgageot:
I'm going to close this PR. I fear that  the code is getting more and more complex each time we change our way of connecting to the Docker daemon: Docker command through ssh, then DockerClient, then Engine-api. 

I'm going to try simplify the code first and then we'll see

PR #2738, opened 2016-01-04
Enhance remove

jeanlaurent:
This PR fixes two points :
* Fix #2671
* When trying to remove multiple machine do not break on the first one which fails  but continue removing the other machines in the list.
* try to make the code a "bit" more readable.

PR #2767, opened 2016-01-06
Fix double error message while removing

jeanlaurent:
with #2738 we double the last error message to the console. This PR fixes this.

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -17,11 +17,11 @@ func cmdRm(c CommandLine, api libmachine.API) error {
 		return ErrNoMachineSpecified
 	}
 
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Yes, I think it would be far preferable to have a custom `Error` type which aggregates them all together (similar to https://github.com/docker/machine/blob/master/commands/commands.go#L388, but creating an actual type for it).  Returning `errors.New("")` seems very peculiar to me.

nathanleclaire:
Also, can you write a unit test that specifically checks to ensure that this regression is not happening?

dgageot:
LGTM

PR #2769, opened 2016-01-06
Set swarm manage to advertise on 3376

dustinblackman:
## Overview

Currently when you setup a new swarm master, it's advertising on `2376` instead of `3376`. When advertising on `2376` it causes issues with replication as replicas are forwarding to the standard docker port instead of swarm.

Fixes https://github.com/docker/swarm/issues/1576


dgageot:
LGTM

nathanleclaire:
LGTM

PR #2774, opened 2016-01-07
log.fatal is evil

jeanlaurent:
`log.Fatal` contains an implicit `os.Exit(1)`.
This is error prone, we've been recently caught red handed on this see https://github.com/dgageot/machine/commit/f96595d7d0515490c6b39ff3f22126eb495e7dd6#diff-b54b33b58b9d76939b70cd4d2f48f251L116

As we have no more than 2 calls, i did remove them all.


dgageot:
LGTM

PR #2784, opened 2016-01-08
AWS documentation for security group

jeanlaurent:


dgageot:
LGTM

mobsy74:
It would be nice if this modification of the security group was optional.  If you pass in your own, could you opt out of the opening of the ports?  I have a security group that has the necessary ports open ONLY to my IP, and would rather not have to go shut down the SSH and Docker ports (open to 0.0.0.0) from the console every time.

PR #2789, opened 2016-01-08
FIX #2411 On a btrfs system, don't default to aufs

### LINKED ISSUE #2411: generic driver: since aufs is not supported on btrfs, the default storage driver should be btrfs on such a system
When creating a docker machine using the generic driver and you do not pass the storage driver explicitly, aufs is set in the docker daemon configuration.

For example, on an ubuntu-systemd target system, the docker.service contains:
```
[Service]
ExecStart=/usr/bin/docker -d -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --stor
age-driver aufs --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.
pem --tlskey /etc/docker/server-key.pem --label provider=generic
```

This errors out since aufs on btrfs is a non supported setup:
```
docker[2969]: time="2015-11-25T16:49:49.636299049+01:00" level=fatal msg="Error starting daemon: error initializing graphdriver: backing file system is unsupported for this graph driver"
```

So instead of configuring the daemon to use aufs, it would be better to detect the btrfs and use that driver when creating a docker machine without having to explicitly set the storage driver to btrfs on the docker-machine commandline.

Version: master branch, released versions (tested with 0.5.0 and 0.5.1) do not work on systemd-ubuntu 15.10

robvanmieghem:
FIX #2411 On an Ubuntu system, don't default to aufs on a btrfs filesystem

Signed-off-by: Rob Van Mieghem <rob@vanmieghemcloud.com>

### DIFF
--- a/libmachine/provision/provisiontest/sshcommander.go
+++ b/libmachine/provision/provisiontest/sshcommander.go
@@ -1,0 +1,22 @@
+//Package provisiontest provides utilities for testing provisioners
+package provisiontest
+
+import (
+	"errors"
+	"strings"
+)
+
+//FakeSSHCommander is an implementation of provision.SSHCommander to provide predictable responses set by testing code
+//Extend it when needed
+type FakeSSHCommander struct {
+	//Result of the ssh command to look up the FilesystemType
+	FilesystemType string
+}
+
+//SSHCommand is an implementation of provision.SSHCommander.SSHCommand to provide predictable responses set by testing code
+func (sshCmder FakeSSHCommander) SSHCommand(args string) (string, error) {
+	if !strings.HasPrefix(args, "stat -f") {
+		return "", errors.New("Not implemented by FakeSSHCommander")
+	}
+	return sshCmder.FilesystemType + "\n", nil
+}
--- a/libmachine/provision/provisiontest/sshcommander.go
+++ b/libmachine/provision/provisiontest/sshcommander.go
@@ -1,22 +1,39 @@
 //Package provisiontest provides utilities for testing provisioners
 package provisiontest
 
-import (
-	"errors"
-	"strings"
-)
+import "errors"
 
-//FakeSSHCommander is an implementation of provision.SSHCommander to provide predictable responses set by testing code
-//Extend it when needed
-type FakeSSHCommander struct {
+//FakeSSHCommanderOptions is intended to create a FakeSSHCommander without actually knowing the underlying sshcommands by passing it to NewSSHCommander
+type FakeSSHCommanderOptions struct {
 	//Result of the ssh command to look up the FilesystemType
 	FilesystemType string
 }
 
-//SSHCommand is an implementation of provision.SSHCommander.SSHCommand to provide predictable responses set by testing code
-func (sshCmder FakeSSHCommander) SSHCommand(args string) (string, error) {
-	if !strings.HasPrefix(args, "stat -f") {
-		return "", errors.New("Not implemented by FakeSSHCommander")
+//FakeSSHCommander is an implementation of provision.SSHCommander to provide predictable responses set by testing code
+//Extend it when needed
+type FakeSSHCommander struct {
+	Responses map[string]string
+}
+
+//NewFakeSSHCommander creates a FakeSSHCommander without actually knowing the underlying sshcommands
+func NewFakeSSHCommander(options FakeSSHCommanderOptions) *FakeSSHCommander {
+	if options.FilesystemType == "" {
+		options.FilesystemType = "ext4"
 	}
-	return sshCmder.FilesystemType + "\n", nil
+	sshCmder := &FakeSSHCommander{
+		Responses: map[string]string{
+			"stat -f -c %T /var/lib": options.FilesystemType + "\n",
+		},
+	}
+
+	return sshCmder
+}
+
+//SSHCommand is an implementation of provision.SSHCommander.SSHCommand to provide predictable responses set by testing code
+func (sshCmder *FakeSSHCommander) SSHCommand(args string) (string, error) {
+	response, commandRegistered := sshCmder.Responses[args]
+	if !commandRegistered {
+		return "", errors.New("Command not registered in FakeSSHCommander")
+	}
+	return response, nil
 }
--- a/libmachine/provision/ubuntu_systemd_test.go
+++ b/libmachine/provision/ubuntu_systemd_test.go
@@ -2,6 +2,12 @@ package provision
 
 import (
 	"testing"
+
+	"github.com/docker/machine/drivers/fakedriver"
+	"github.com/docker/machine/libmachine/auth"
+	"github.com/docker/machine/libmachine/engine"
+	"github.com/docker/machine/libmachine/provision/provisiontest"
+	"github.com/docker/machine/libmachine/swarm"
 )
 
 func TestUbuntuSystemdCompatibleWithHost(t *testing.T) {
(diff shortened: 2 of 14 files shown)


nathanleclaire:
Seems pretty reasonable, although I wish we had better test coverage for the provisioners so we could add some for this case.

nathanleclaire:
If you're up to the challenge, it would be really great to get some unit tests for at least some of these code changes.

robvanmieghem:
@nathanleclaire you are absolutely right. Just added them.

nathanleclaire:
Nice thanks for adding the tests I have some feedback on the code but I am willing to add this feature once it gets addressed.  cc @docker/machine-maintainers also for review.

libmachine/provision/ubuntu_systemd_test.go
```diff
@@ -27,3 +32,34 @@ func TestUbuntuSystemdCompatibleWithHost(t *testing.T) {
 	}
 
 }
+
+func TestUbuntuSystemdStorageDriver(t *testing.T) {
+
+	p := NewUbuntuSystemdProvisioner(&fakedriver.Driver{}).(*UbuntuSystemdProvisioner)
```
nathanleclaire: Can you please convert this and the other test to use a [test table](https://github.com/golang/go/wiki/TableDrivenTests)?  It seems to be a good fit for what you're doing here, and there are a couple other examples of doing this across the codebase (e.g. `commands/env_test.go`) if you want to take a look.

libmachine/provision/utils.go
```diff
@@ -229,3 +238,17 @@ func waitForDocker(p Provisioner, dockerPort int) error {
 
 	return nil
 }
+
+//fakeSSHCommander is intended to be used by testing code
+//Extend it when needed
+type fakeSSHCommander struct {
```
nathanleclaire: Can you please make this exported (i.e. named `FakeSSHCommander` instead of `fakeSSHCommander`) and move it's definition into a test with `_test.go` name to delineate it from non-test code?  One suggestion is that you could add a `provisiontest` module inside of `provision` module (like how the standard library has `httptest`) and put it in a file there.

libmachine/provision/utils.go
```diff
@@ -208,6 +209,14 @@ func matchNetstatOut(reDaemonListening, netstatOut string) bool {
 	return false
 }
 
+func getFilesystemType(p Provisioner, directory string) (fstype string, err error) {
```
nathanleclaire: Nit: Can you please change the named return parameters to being explicitly defined in the function body?  That's convention in this codebase.

libmachine/provision/ubuntu_upstart.go
```diff
@@ -135,14 +135,30 @@ func (provisioner *UbuntuProvisioner) dockerDaemonResponding() bool {
 	return true
 }
 
+func (provisioner *UbuntuProvisioner) setStorageDriver() error {
```
nathanleclaire: Instead of duplicating this function for the two provisioners, why not make a helper function that you invoke from `Provision` method instead?  e.g. the invocation would look something like (in `Provision` method):

```golang
storageDriver, err := decideStorageDriver(provisioner, provisioner.EngineOptions.StorageDriver)
if err != nil {
    return err
}
```

nathanleclaire:
Thanks for your contribution @robvanmieghem! Left some comments on the code.

As a side note, I'd really like to switch the magic `"aufs"`, `"btrfs"`, etc. constants to use real `const`, but I think it's best addressed in a followup PR (since it was already that way before and is out of scope of this change).

robvanmieghem:
Thank you for reviewing and commenting on this pull request @nathanleclaire !

I modified the code in response to your notes. Since the storagedriver decision is now extracted from the Ubuntu provisioners, I added this this method call to the provisioners that set a default storagedriver as well. This should effectively only impact Debian since it uses aufs as a default too.

You might notice that I introduced a fakeProvisioner in the utils_test.go file. It's not in the `provisiontest` package since implementing the provisioner interface would cause a circular package dependency since this code would be used by the tests inside the `provision` package. 

I agree that refactoring these the magic "aufs", "btrfs", etc. constants to real consts is an easy improvement but might get this PR to get out of it's initial scope.

libmachine/provision/provisiontest/sshcommander.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"errors"
+	"strings"
+)
+
+//FakeSSHCommander is an implementation of provision.SSHCommander to provide predictable responses set by testing code
+//Extend it when needed
+type FakeSSHCommander struct {
+	//Result of the ssh command to look up the FilesystemType
+	FilesystemType string
```
nathanleclaire: Hm, instead of having a field specifically for this, could you maybe make some sort of lookup table field which is `map[string]string`?  Then you could define the result of a given call like so:

```console
cmder := &FakeSSHCommander{
    Responses: map[string]string{
        "stat -f -c %T /var/lib": "btrfs\n",
    },
}
```

WDYT?

nathanleclaire:
Nice everything is looking pretty good on my end, I posted one more comment with an additional ask.

@docker/machine-maintainers Please take a look at this and let us know your thoughts when you get a chance.

dgageot:
LGTM

jeanlaurent:
LGTM 

@nathanleclaire I let you give the final go/merge.

libmachine/provision/provisiontest/sshcommander.go
robvanmieghem: This would indeed be perfect for testing the output parsing.
For testing more high level functionality, like checking the default graphdriver in the RedHat provisioner for example, it does not really make sense to know that the underlying call is `stat -f -c %T /var/lib`, this would also prevent easy refactoring of the specific implementation of the filesystem checking without modifying all tests.
Both kind of tests are necessary so I'll refactor the FakeSSHCommander to use an internal command/resonse map for output parsing tests while a creationmethod with more high level configuration can be used by the other ones.

nathanleclaire: > For testing more high level functionality, like checking the default graphdriver in the RedHat provisioner for example, it does not really make sense to know that the underlying call is stat -f -c %T /var/lib, this would also prevent easy refactoring of the specific implementation of the filesystem checking without modifying all tests.

I'm not sure I understand.  If the underlying command sent to SSH changes (say, for Red Hat), I _want_ the tests to break.  We'll then know that the tests need to be updated as well, will have clearly defined expected behavior, and will only need to update one line in the tests.

nathanleclaire:
I thought of something else as well.  If user specifies something like `--engine-opt graph=/vol/docker`, the daemon will create and use the graph directory at `/vol/docker` instead of `/var/lib/docker`.  So, we should probably account for this case as well.

nathanleclaire:
Consequently, you should most likely pass in the entire `EngineOptions` to `decideStorageDriver`, loop to check if this set in `ArbitraryFlags` (`if strings.HasPrefix(flag, "graph=")`) and over-ride the default value of `/var/lib` if so.  It will require a bit of [path](https://golang.org/pkg/path/) parsing, most likely.

Maybe we should open another ticket and address it in a follow-up PR, but I wanted to make sure everyone is aware of the issue and a potential solution.

robvanmieghem:
@nathanleclaire , I modified the FakeSSHCommander as explained earlier. I hope this makes it a bit more clear and both ways of testing are supported this way. If you want to have it ammended in the previous commit, just let me know. Thought a separate commit makes the FakeSSHCommander for testing more clear.

robvanmieghem:
>I thought of something else as well. If user specifies something like --engine-opt graph=/vol/docker, the daemon will create and use the graph directory at /vol/docker instead of /var/lib/docker. So, we should probably account for this case as well.
You are right, did not think of that...

nathanleclaire:
> You are right, did not think of that...

Like I said, maybe we can just create a ticket to track the issue, and merge this (since I'm not aware of too many folks setting custom graph drivers today).  Then follow up later.

nathanleclaire:
LGTM

PR #2790, opened 2016-01-09
VMWARE Fusion get IP from multiple leases files

### LINKED ISSUE #2787: vmwarefusion driver hard coded to only look for docker-machine instances running on NAT vmnet interface
I've been having a hard time figuring out why when I changed the networking type to vmnet2 (custom network configuration) docker-machine was unable to communicate with the machines. 

This got me interested in how docker-machine resolves the host IP under normal circumstances when the IP addresses change (as they do with every reboot).

Digging through the code, I came upon the following [code](https://github.com/docker/machine/blob/master/drivers/vmwarefusion/fusion_darwin.go#L464):

	// DHCP lease table for NAT vmnet interface
	var dhcpfile = "/var/db/vmware/vmnet-dhcpd-vmnet8.leases"

	if vmxfh, err = os.Open(d.vmxPath()); err != nil {
		return "", err
	}

So that's how the driver resolves the changed IP address, by parsing that file. The problem of course is that the new dhcp lease will reside in the file "/var/db/vmware/vmnet-dhcpd-vmnet2.leases" hence docker-machine will be unable to determine the new IP address. 

I'd provide a patch, but unfortunately at this stage my golang programming skills are absolutely non-existent.

daehyeok:
Fixed #2787 .
Not a elegant solution. but changed to find ip from all of leases file in the /var/db/vmware/*.leases .

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -449,19 +449,10 @@ func (d *Driver) vmdkPath() string {
 }
 
 func (d *Driver) getIPfromDHCPLease() (string, error) {
-	var vmxfh *os.File
-	var dhcpfh *os.File
-	var vmxcontent []byte
-	var dhcpcontent []byte
-	var macaddr string
-	var err error
-	var lastipmatch string
-	var currentip string
-	var lastleaseendtime time.Time
-	var currentleadeendtime time.Time
 
-	// DHCP lease table for NAT vmnet interface
-	var dhcpfile = "/var/db/vmware/vmnet-dhcpd-vmnet8.leases"
+	var vmxfh *os.File
+	var vmxcontent []byte
+	var err error
 
 	if vmxfh, err = os.Open(d.vmxPath()); err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


jeanlaurent:
@frapposelli can you give us a hand on this one ?

frapposelli:
@daehyeok why is this needed? the VMX configuration should be statically set to the default NAT `vmnet` which is always present.

daehyeok:
@frapposelli Could you check issue #2787 ?. In that case how to know which leases file have to checked?

frapposelli:
I see nothing wrong with the patch (except for a call amplification effect, as getIP calls are numerous), but the behavior described in #2787 is not really kosher.

code is LGTM, but I'll follow up in the issue to better understand the use case.

jeanlaurent:
Sounds weird, but after reading #2787, it does the job and didn't break current setup.

LGTM

frapposelli:
@jeanlaurent the point of #2787 is that the setup he wanted could've been accomplished in a better way without resorting to custom networks (thus negating the need for this patch).

Although, I can see the need for custom network functionalities going forward so this PR is a step in that direction.

dgageot:
LGTM

PR #2797, opened 2016-01-11
Fix multiple VirtualBox bugs on Windows

dgageot:
Fixes a bug when an host-only adapter cannot be used
Fixes a bug when the host-only adapter uses a wrong IP and then multiple machines get the same IP
Fixes the error message when the VM cannot be started

### DIFF
--- a/drivers/virtualbox/network.go
+++ b/drivers/virtualbox/network.go
@@ -5,16 +5,20 @@ import (
 	"fmt"
 	"net"
 	"regexp"
-	"strconv"
+	"strings"
+	"time"
+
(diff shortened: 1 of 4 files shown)


jeanlaurent:
less host-only interfaces bug on virtualbox, are you kidding me ?

LGTM x10

nathanleclaire:
@Hito01 I don't think we have anything against it, it's just been for simplicity's sake. We're willing to take PRs (or at least an issue to track it as a desired feature).

akimserik:
Hi All!

Could you please explain for a dummy like me - how to apply these fixes on Windows?

Thanks!

PR #2799, opened 2016-01-11
Fix ec2 ebs optimized integration test

jeanlaurent:
Fix the ec2 IT test from #2750

### DIFF
--- a/test/integration/amazonec2/create-ebsinstance.bats
+++ b/test/integration/amazonec2/create-ebsinstance.bats
@@ -2,31 +2,26 @@
 
 load ${BASE_TEST_DIR}/helpers.bash
 
-force_env DRIVER amazonec2
+only_if_env DRIVER amazonec2
 
-#Use Instance Type that supports EBS Optimize 
-export AWS_INSTANCE_TYPE=m4.large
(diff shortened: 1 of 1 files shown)


jeffellin:
If I am reading this correctly, doesn't now not assert that the EBSOptimized = true?  Also, do you want to continue using the docker image or one that you control?

jeanlaurent:
Hmm it does, in the second "testcase". Bats is a bit tricky when your run multiple commands and one fails. By putting in one test the run ( and ensure it works) and putting the assertion in a second one, it's easier to read/write and more robust (in a bats way...)

I think this is fine for now. We have yet to have comprehensive and structured IT test for other driver than `none` and `virtualbox`.

This is a good start.

Thx for your help on this one @jeffellin

dgageot:
LGTM

jeffellin:
ok I see what you are doing now,  its just a little confusing because your second test label is check that its up, I probably would have done "check that its up and an ebs instance."

jeffellin:
Thanks @jeanlaurent for the help gettting this merged.

jeanlaurent:
Yep correct. Needs a rename.

PR #2807, opened 2016-01-11
Return exit code 3 on pre-create errors

jeanlaurent:
This will return exit code 1 for every errors, except pre-create check errors, which will return 3.

Kitematic/Toolbox can now filter those errors out.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -12,6 +12,7 @@ import (
 	"github.com/docker/machine/libmachine/crashreport"
 	"github.com/docker/machine/libmachine/host"
 	"github.com/docker/machine/libmachine/log"
+	"github.com/docker/machine/libmachine/mcnerror"
 	"github.com/docker/machine/libmachine/mcnutils"
 	"github.com/docker/machine/libmachine/persist"
 	"github.com/docker/machine/libmachine/ssh"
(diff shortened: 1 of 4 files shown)


jeanlaurent:
Ready for review @docker/machine-maintainers

nathanleclaire:
According to http://tldp.org/LDP/abs/html/exitcodes.html, exit code 2 is reserved for misuse of shell built-ins.  Exit Code 3 is not reserved for that so maybe we should use that instead?  

@tianon do you have any input?  For context, we are trying to differentiate the exit codes of pre-create failures (e.g. virtualization isn't enabled, API keys aren't set) from create failures (e.g. the API goes down in the middle of trying to create a machine).

nathanleclaire:
Code looks good though.

jeanlaurent:
Changed return code from 2 to 3

nathanleclaire:
Cool.  It'd be good to document this behavior, maybe I'll make a followup PR.

LGTM

nathanleclaire:
@dgageot Does it LGTY?

nathanleclaire:
Filed https://github.com/docker/machine/pull/2819 for a followup doc

dgageot:
LGTM

Issue #2781, opened 2016-01-08
Issue with docker-machine with rackspace driver


stushurik:
Recenlty, I have found these docs: 
https://docs.docker.com/machine/drivers/rackspace/ 
https://developer.rackspace.com/blog/using-docker-machine-to-deploy-your-docker-containers-on-rackspace/

I followed all steps from that instruction and get the next error:

olexandr@E5-551G:~/vboxes/autumn-docker-compose-files$ docker-machine create --driver rackspace test-host 
Running pre-create checks... 
Creating machine... 
Error creating machine: Error in driver during machine creation: Unable to find tenant named

Also if I list all machines:

olexandr@E5-551G:~/vboxes/autumn-docker-compose-files$ docker-machine ls 
NAME ACTIVE DRIVER STATE URL SWARM DOCKER ERRORS 
test-host - rackspace Error Unknown Expected HTTP response code [200 203] when accessing [GET https://lon.servers.api.rackspacecloud.com/v2/10045552/servers/], but got 404 instead

**UPD**. I use docker-machine v. 0.5.5

stushurik:
According to this (https://github.com/docker/machine/issues/2314) issue`s solution, I have downgrade docker-machine to 0.5.1 and it works

nathanleclaire:
I think it should be fixed with 0.7.0

PR #2809, opened 2016-01-11
Rackspace: Only derive tenant ID if tenant name is supplied.

smashwilson:
The Rackspace driver authenticates with username and API key rather than tenant name or ID. Modify the base OpenStack driver to only attempt to derive a tenant ID if a tenant name has been supplied, which will always be true for `-d openstack`, but never for `-d rackspace`.

Addresses #2781.

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -566,7 +566,7 @@ func (d *Driver) resolveIds() error {
 		})
 	}
 
-	if d.TenantId == "" {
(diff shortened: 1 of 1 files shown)


wouterla:
Just applied this fix and can confirm it fixes my issues creating docker-machines on rackspace. Thanks!

nathanleclaire:
cc @ggiamarchi

nathanleclaire:
LGTM

PR #2813, opened 2016-01-12
add private ip address for openstack driver

xinxian0458:
Signed-off-by: Zhang, Jie <zhangjie0220@gmail.com>


nathanleclaire:
cc @ggiamarchi WDYT

dgageot:
Hi @xinxian0458, I don't understand the purpose of this PR. Can you explain?

xinxian0458:
@dgageot, In my environment, I need to know both the floating ip address and private ip address of the machine created in the openstack. but the current openstack driver only provide one ip address, I can't get its private ip address if i set the openstack-floatingip-pool.

drivers/openstack/openstack.go
```diff
@@ -44,6 +44,7 @@ type Driver struct {
 	FloatingIpPoolId string
 	IpVersion        int
 	client           Client
+	PrivateIPAddress string
```
fsoppelsa: We already have `d.IPAddress`

drivers/openstack/openstack.go
```diff
@@ -318,6 +319,35 @@ func (d *Driver) GetIP() (string, error) {
 	return "", fmt.Errorf("No IP found for the machine")
 }
 
+func (d *Driver) GetPrivateIP() (string, error) {
```
fsoppelsa: Did you test this in conjunction with a floating IP of `assignFloatingIP()`?

How does the command look like?

drivers/openstack/openstack.go
```diff
@@ -318,6 +319,35 @@ func (d *Driver) GetIP() (string, error) {
...
+	log.Debug("Looking for the Private IP address...", map[string]string{"MachineId": d.MachineId})
+
+	if err := d.initCompute(); err != nil {
+		return "", err
+	}
+
+	addressType := Fixed
+
+	// Looking for the IP address in a retry loop to deal with OpenStack latency
+	for retryCount := 0; retryCount < 200; retryCount++ {
```
fsoppelsa: Where is `200` coming from?

drivers/openstack/openstack.go
xinxian0458: Yes, I see. but in my opinion, the machine need its private ip no matter whether set assignFloatingIP or not, so it should has another attribute.

drivers/openstack/openstack.go
xinxian0458: I just test it in the openstack which is using nova network.

docker-machine --debug create \
--driver openstack \
--openstack-auth-url http://172.19.0.2:5000/v2.0 \
--openstack-username demo \
--openstack-password password \
--openstack-tenant-name demo \
--openstack-flavor-id 4 \
--openstack-image-id 312a1b0c-e0b2-4968-a828-3e0b67c842d9 \
--openstack-ssh-user centos \
--openstack-sec-groups default \
--openstack-floatingip-pool public \
--openstack-nova-network=true \
openstack02

drivers/openstack/openstack.go
xinxian0458: Just copy it from GetIP() function.

drivers/openstack/openstack.go
fsoppelsa: Can you test that over Neutron too? (I suppose it should work)

dgageot:
I'm really confused by this PR. I understand that it's goal is to add the machine's private ip in the `config.json`. The `config.json` is mainly an internal format and shouldn't be used by external tools.
So, until I better understand the goal, I`m going to NotLGTM

fsoppelsa:
Easier, I would suggest:

Always use the private network.
If also a floating IP is needed, as the very latest step in machine's provisioning, allocate that additional port from the floating pool.

dgageot:
@fsoppelsa If I understand correctly, there's a way to achieve @xinxian0458's goal without this PR. Right?
In that case, I'd like to close the PR. @fsoppelsa @xinxian0458 wdyt?

fsoppelsa:
@dgageot well, my suggestion is to use the private network always for dm, and add some code in the very end to enable a floating IP.

This PR is not about this, it's adding things that IMHO are too much.

Getting the machine's floating or internal, is a matter of using the Nova APIs or even the CLI.

Issue #2318, opened 2015-11-17
Machine 0.5.0 fails to create with virtualbox driver on Win7


kpalf:
Using Toolbox 1.9 on Win7 Pro.

Launching the quickstart terminal fails. (I've added a -D to get extra info).  Tried hunting bug reports and forums but not seeing any reports.  I have a colleague with same issue on another laptop.

Creating Machine default...
Docker Machine Version: 0.5.0 (04cfa58)
Found binary path at .\docker-machine-driver-virtualbox.exe
Launching plugin server for driver virtualbox
Plugin server listening at address 127.0.0.1:51174
Error loading driver "virtualbox": dial-http tcp 127.0.0.1:51174: read tcp 127.0.0.1:51175->127.0.0.1:51174: wsarecv: An existing connection was forcibly closed by the remote host.
Looks like something went wrong... Press any key to continue...

Can occasionally see different output, but same result:
$ docker-machine -D create -d virtualbox default
Docker Machine Version: 0.5.0 (04cfa58)
Found binary path at C:\Program Files\Docker Toolbox\docker-machine-driver-virtualbox.exe
Launching plugin server for driver virtualbox
Plugin server listening at address 127.0.0.1:61709
() Calling RpcServerDriver.GetVersion
Using API Version 1
() Calling RpcServerDriver.SetConfigRaw
() Calling RpcServerDriver.GetMachineName
(flag-lookup) Calling RpcServerDriver.GetCreateFlags
Making call to close connection to plugin binary
Making call to close driver server
(flag-lookup) Calling RpcServerDriver.Close
Successfully made call to close driver server
Found binary path at C:\Program Files\Docker Toolbox\docker-machine-driver-virtualbox.exe
Launching plugin server for driver virtualbox
Plugin server listening at address 127.0.0.1:61713
Error loading driver "virtualbox": dial-http tcp 127.0.0.1:61713: read tcp 127.0.0.1:61714->127.0.0.1:61713: wsarecv: An existing connection was forcibly closed by the remote host.

nathanleclaire:
Any unusual networking configuration on your computer?  Proxy?  VPN?  Firewall?  Antivirus?

kpalf:
Apologies Nathan, I should have given more info. DockerCon != Sleep, but then from the time of your response I guess you are the same. ;-)
[...]

PR #2823, opened 2016-01-13
Add panic recovery to Create method for drivers

nathanleclaire:

Should hopefully help with some of the "connection reset by server" issues we've been seeing (I suspect some are due to panics) and is just generally a good idea (we have no way to catch panics in drivers today and they totally send Machine sideways due to the RPC server crashing).

This just does it for `Create` to get started, but there's a good argument that maybe we should `trapPanic` for every supported method.

### DIFF
--- a/libmachine/drivers/rpc/server_driver.go
+++ b/libmachine/drivers/rpc/server_driver.go
@@ -3,12 +3,28 @@ package rpcdriver
 import (
 	"encoding/gob"
 	"encoding/json"
+	"fmt"
+	"runtime/debug"
 
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/mcnflag"
 	"github.com/docker/machine/libmachine/state"
 	"github.com/docker/machine/libmachine/version"
+)
+
+type Stacker interface {
+	Stack() []byte
+}
+
+type StandardStack struct{}
+
+func (ss *StandardStack) Stack() []byte {
+	return debug.Stack()
+}
+
+var (
+	stdStacker Stacker = &StandardStack{}
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


dgageot:
LGTM

kpalf:
What's the easiest way to pick up a build containing this change so we can help test for you, and see if it helps resolve #2318 ?  I'm still on 0.5.4 from Toolbox 1.9.1f, but I can update to something else as its 100% broken for me currently.
Thanks, and keep up the fantastic work!
Kev

nathanleclaire:
@kpalf I can throw up a test build for you here if you give me a minute.

kpalf:
Just let me know, and if you need me to upgrade other parts at all first.
Kev (UK time)
 
Date: Wed, 13 Jan 2016 09:29:42 -0800

nathanleclaire:
@kpalf Can you please give https://www.dropbox.com/s/gfn4ow147e3xzbc/docker-machine_windows-amd64.exe?dl=1 a try and let us know how it goes?

sha256 for that binary is 9ba7f67beffff9774d62c1d33b5f08fb9d66759de6194d0420ef991aac2e2197

libmachine/drivers/rpc/server_driver.go
```diff
@@ -108,8 +124,22 @@ func (r *RPCServerDriver) SetConfigRaw(data []byte, _ *struct{}) error {
...
 
-func (r *RPCServerDriver) Create(_, _ *struct{}) error {
-	return r.ActualDriver.Create()
+func trapPanic(err *error) {
+	if r := recover(); r != nil {
+		*err = fmt.Errorf("Panic in the driver: %s\n%s", r.(error), stdStacker.Stack())
+	}
+}
+
+func (r *RPCServerDriver) Create(_, _ *struct{}) (err error) {
```
nathanleclaire: I should note for future reference that while I'm usually against named return parameters, it's actually needed here.  (otherwise the `defer`ed function can't modify the value of the parameter)

kpalf:
[5_debug_output_new_binary_060dev_4291b59_after_deleting_folders.txt](https://github.com/docker/machine/files/91655/5_debug_output_new_binary_060dev_4291b59_after_deleting_folders.txt)
Attached debug output when running under toolbox 1.9.1h (machine 0.5.5) but with the docker_machine binary replaced the one provided above (0.6.0-dev build 4291b59) .

I see the following errors in the output and the VM fails even though logging would suggest otherwise.

(default) DBG | COMMAND: C:\Program Files\Oracle\VirtualBox\VBoxManage.exe modifyvm default --natpf1 delete ssh
(default) DBG | STDOUT:
(default) DBG | {
(default) DBG | }
(default) DBG | STDERR:
(default) DBG | {
(default) DBG | VBoxManage.exe: error: Code E_INVALIDARG (0x80070057) - One or more arguments are invalid (extended info not available)
(default) DBG | VBoxManage.exe: error: Context: "RemoveRedirect(Bstr(ValueUnion.psz).raw())" at line 1767 of file VBoxManageModifyVM.cpp
(default) DBG | }

nathanleclaire:
@kpalf You don't get the `wsarecv: An existing connection was forcibly closed by the remote host.` message anymore though, right?

kpalf:
@nathanleclaire Apologies, no I can't get it reliably since later releases.  I switched back to putting detailed comments on the issue I originally opened #2318, and I'll try not to pollute this thread further.

PR #2828, opened 2016-01-13
ec2 Don't force user to use VPC_ID but use the default one when available

jeanlaurent:


jeanlaurent:
Removed duplication you hinted @dgageot ( it was dead code left during refactoring) and extracted a few errors.

dgageot:
@jeanlaurent I still see the duplication

jeanlaurent:
Yeah duplication was added in 0d2bc08 and removed in e37fcb5.
Now all is done in  0d2bc08...

dgageot:
LGTM

PR #2831, opened 2016-01-13
Check for nil before deferencing pointer in SG config

nathanleclaire:
You can see in debug output here https://ci.appveyor.com/project/eris-ltd/eris-cli/build/1.0.441-winTest that the `p.FromPort` seems to be dereferencing a pointer which is `nil`.  I'm not sure what state the SG could be in where `FromPort` is not specified but at any rate it seems to be prudent to check if it's `nil` first.

I'd like to add a unit test for this once https://github.com/docker/machine/pull/2828 (beginning of testable client interface) is in.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -866,13 +866,15 @@ func (d *Driver) configureSecurityGroupPermissions(group *ec2.SecurityGroup) []*
 	hasDockerPort := false
 	hasSwarmPort := false
 	for _, p := range group.IpPermissions {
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

jeanlaurent:
LGTM

ticean:
FWIW, I encountered this problem and my search led me to this PR. I had an ingress entry for "All traffic" from a given security group. After reading this code change I manually removed that single ingress entry from the SG and my problem was solved. Perhaps non-numeric entries such as this could be causing the nil state. Thanks for the fix and info.

PR #2838, opened 2016-01-14
libmachine/ssh: export the NativeClient.Session function

mssola:
This allows more flexibility to developers in regards to what one can
do with a SSH session. I've also added an example to showcase a situation in
which this exported function can be useful.

This commit fixes an issue that I was having in which I wanted to stream (and manipulate such stream) the stdout of the remote process.

Another solution would be to introduce a new function on the `ssh.Client` interface that returns two readers (stdout and stderr of the remote process). This would fix my situation, but it would introduce a new function on the API, and I'm not sure if that's desirable (even though this would then work for both the Native and the External clients).

Signed-off-by: Miquel Sabaté <msabate@suse.com>


nathanleclaire:
> Another solution would be to introduce a new function on the ssh.Client interface that returns two readers (stdout and stderr of the remote process). This would fix my situation, but it would introduce a new function on the API, and I'm not sure if that's desirable (even though this would then work for both the Native and the External clients).

I think I'd be more interested in this solution if you're willing :)

mssola:
@nathanleclaire sure, I'll update this patch whenever I can ;)

mssola:
@nathanleclaire done. Read the commit message in order to understand the changes that I've done ;)

mssola:
@nathanleclaire I've rebased to current master. Now everything should be fine

nathanleclaire:
I think the code looks OK, be advised that we are getting close to code freeze in preparation for the 0.6.0 / Docker 1.10 release so we will probably hold off on merging this for a little while.

mssola:
No problem, I can wait ;) Thanks for reviewing

dgageot:
@nathanleclaire Are you LGTM for 0.7.0?

nathanleclaire:
@dgageot Just one nit

libmachine/ssh/client.go
```diff
@@ -312,3 +356,34 @@ func (client ExternalClient) Shell(args ...string) error {
...
+
+	log.Debug(cmd)
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, nil, err
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		_ = stdout.Close()
```
nathanleclaire: Why swallow the errors here?  At least aggregating them with `fmt.Errorf("%s, %s", err, closeErr)` seems reasonable, no?

dgageot:
ping @mssola Can you have this fixed? Thanks a lot!

libmachine/ssh/client.go
mssola: Yes, makes sense. I'm on it.

mssola:
@nathanleclaire @dgageot done

nathanleclaire:
This PR LGTM

@dgageot @jeanlaurent PTAL

dgageot:
LGTM

Issue #1783, opened 2015-08-31
Discussion: Have some docker-machine commands default to "default" VM?


nathanleclaire:
In the original versions of Machine, commands like `docker-machine start` (without any arguments) would automatically run against the "active" machine.  This was fine when Machine was directly embedded inside of Docker (a historical curiosity) but ultimately ended up being confusing and problematic so we removed it in favor of always specifying a target machine.

With the introduction of the Toolbox and deprecation of boot2docker one of the pieces of feedback we are getting is that users miss simplicity of commands running automatically against a"default" machine.  Since it is likely that the majority of users will have only one machine, and we now have what is essentially a canonical default VM in `default`, I am wondering if we might want to just default most commands to be referencing `default`.  The ambiguity would potentially be a lot less than the active tracking method, since there's no, "Wait, which host is active again?" conundrum.

e.g.:

```console
$ docker-machine create -d virtualbox default
...
$ docker-machine stop
$ docker-machine start
$ docker-machine rm
```

One tricky angle on this is `ssh`.  Without arguments it works fine, but it's impractical for CLI to know if something like `docker-machine ssh ls` means "run `ls` command on `default` machine" or "SSH into machine named `ls`".  So, that would need some thought.

I don't really want to get into too much implicit behavior e.g. read the default host from a config file etc. because that's what we had before and it was damn confusing to debug when something went wonky.  This might give us 80% of the way there without introducing too much difficult to follow behavior.

WDYT @ehazlett @tianon @hairyhenderson @bfirsh

mhart:
Great, thanks for writing this up @nathanleclaire !

tianon:
This sounds reasonable to me.  For the `ssh` issue, could we utilize `--` to disambiguate?  ie, `docker-machine ssh dev ls` vs `docker-machine ssh -- ls`? (or is it even worth it vs just forcing users to explicitly state `default` if they want to run some other command?)

hairyhenderson:
+1 - this could be implemented super-simply by simply using `$DOCKER_MACHINE_NAME` when no machine name is present.

> Without arguments it works fine, but it's impractical for CLI to know if something like `docker-machine ssh ls` means "run `ls` command on `default` machine" or "SSH into machine named `ls`". So, that would need some thought.

Maybe naive, but couldn't we always assume the first argument is a machine name, except when there's no machine with that name, assume the active one instead?

[...]

PR #2843, opened 2016-01-15
Add ability to imply 'default' VM in commands

nathanleclaire:
This is one feature that people sorely missed from boot2docker and we now have the test infrastructure to make sure it's properly tested, so I'm excited to put this PR in.

This PR adds a "short form" to commands whereby if no name is specified, `default` is assumed to be the intended target.

Closes https://github.com/docker/machine/issues/1783

e.g.:

```console
$ docker-machine stop
Stopping "default"....
Machine "default" was stopped.

$ docker-machine start
Starting "default"...
(default) Waiting for an IP...
Machine "default" was started.
Started machines may have new IP addresses.  You may need to re-run the `docker-machine env` command.

$ docker-machine ip
192.168.99.100
```

cc @docker/machine-maintainers @londoncalling 

Possibly a few integration test failures that I'll fix due to the changed error message.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -62,7 +62,7 @@ Next, as noted in the output of the `docker-machine create` command, we have to
 tell Docker to talk to that machine. You can do this with the `docker-machine
 env` command. For example,
 
-    $ eval "$(docker-machine env dev)"
+    $ eval "$(docker-machine env default)"
     $ docker ps
 
 > **Note**: If you are using `fish`, or a Windows shell such as
@@ -74,15 +74,15 @@ This will set environment variables that the Docker client will read which
 specify the TLS settings. Note that you will need to do that every time you
 open a new tab or restart your machine.
 
-To see what will be set, run `docker-machine env dev`.
+To see what will be set, we can run `docker-machine env default`.
 
-    $ docker-machine env dev
+    $ docker-machine env default
     export DOCKER_TLS_VERIFY="1"
     export DOCKER_HOST="tcp://172.16.62.130:2376"
-    export DOCKER_CERT_PATH="/Users/<your username>/.docker/machine/machines/dev"
-    export DOCKER_MACHINE_NAME="dev"
+    export DOCKER_CERT_PATH="/Users/<yourusername>/.docker/machine/machines/default"
+    export DOCKER_MACHINE_NAME="default"
     # Run this command to configure your shell:
-    # eval "$(docker-machine env dev)"
+    # eval "$(docker-machine env default)"
 
 You can now run Docker commands on this host:
 
@@ -98,7 +98,7 @@ You can now run Docker commands on this host:
 Any exposed ports are available on the Docker host’s IP address, which you can
 get using the `docker-machine ip` command:
 
-    $ docker-machine ip dev
+    $ docker-machine ip default
     192.168.99.100
 
 For instance, you can try running a webserver ([nginx](https://www.nginx.com/)
@@ -109,7 +109,7 @@ in a container with the following command:
 When the image is finished pulling, you can hit the server at port 8000 on the
 IP address given to you by `docker-machine ip`. For instance:
 
-    $ curl $(docker-machine ip dev):8000
+    $ curl $(docker-machine ip default):8000
     <!DOCTYPE html>
     <html>
     <head>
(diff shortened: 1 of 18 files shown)


docs/get-started.md
```diff
@@ -74,15 +74,15 @@ This will set environment variables that the Docker client will read which
 specify the TLS settings. Note that you will need to do that every time you
 open a new tab or restart your machine.
 
-To see what will be set, run `docker-machine env dev`.
+To see what will be set, we can run `docker-machine env default`.
 
-    $ docker-machine env dev
+    $ docker-machine env default
```
nathanleclaire: I did change these since I noticed they were inconsistent, but if you feel it's out of scope for the PR let me know.

docs/get-started.md
```diff
@@ -142,7 +142,41 @@ output of `docker-machine ls`.
...
+
+    $ docker-machine stop
+    Stopping "default"....
+    Machine "default" was stopped.
+
+    $ docker-machine start
+    Starting "default"...
+    (default) Waiting for an IP...
+    Machine "default" was started.
+    Started machines may have new IP addresses.  You may need to re-run the `docker-machine env` command.
```
SvenDowideit: This implies that `docker-machine env` will work, but then the list below doesn't list it.

if `docker-machine env` doesn't work with out the machine name, then it needs to say so here too, else its confusing.

nathanleclaire: True... Probably easiest to just support it.

nathanleclaire:
Added the additional commands mentioned by @dgageot and will add some more tests.

docs/get-started.md
nathanleclaire: It works now.

```console
$ eval $(docker-machine env)
```

nathanleclaire:
Hm, actually I had some words mixed up in that original error message (it said "is VM" instead of "VM is").  Let me know how you feel about it now.

nathanleclaire:
Sweet, green.  Please take a look.  I think the test coverage in the PR (updates and additions) covers the most important bases -- let me know if there's anything you'd like me to add.

londoncalling:
Hey Nathan and Sven,
I read through all, makes sense. It's a friendly that if you don't specify
the machine in a command, it assumes "default" if there is one. (In fact, I
made this mistake today a couple of times, and had to re-type the commands
w/"default".)

On the other hand, since users can delete "default" and/or have other
machines with different names, it's probably just best practice to be in
the habit of specifying the machine name in all commands.

So what might work better for the docs is a mention of this, but not quite
so much detail and examples reinforcing it.

For the most part, it's nicely written, but I'll probably tweak a little
but of the phrasing toward active voice and so on while I'm in the process
or re-organizing and re-writing some of the Machine docs this weeks.
--Thanks, Vicky

dgageot:
LGTM

londoncalling:
Hey Nathan and Sven,
I read through all, makes sense. It's a friendly that if you don't specify the machine in a command, it assumes "default" if there is one. On the other hand, since users can delete "default" and/or have other machines with different names, it's probably just best practice to be in the habit of specifying the machine name in all commands. What might work better for the docs is a mention of this, but not quite so much detail and examples reinforcing it.   I'll probably tweak some of the phrasing while I'm in the process or re-organizing and re-writing some of the Machine docs this weeks. --Thanks, Vicky

kudos:
@londoncalling I'm using docker-machine to orchestrate local dev environments for an engineering team of 80. I really question how many of these users will ever juggle more than one Docker Machine.

dgageot:
Hi @kudos thank you for sharing your use case. We see a lot of developers using Docker Machine to provision multiple local or cloud hosts. One of the use cases is to install Docker Swarm.

kudos:
@dgageot we use it for more complex deployments too, but the ratio of engineers using it like boot2docker versus that is around ten or twenty to one.

londoncalling:
@nathanleclaire , follow-up question on this for clarification:

If you have only one docker machine but it is named something other than
"default" (e.g., "staging"), the commands will still require that you do
specify the machine name, right?

nathanleclaire:
@londoncalling Yes.  You have to do `docker-machine start staging`, `docker-machine ssh staging`, etc. just like before.

PR #2844, opened 2016-01-15
Fix docker daemon flag on systemd

zchee:
`-d` is no longer a valid flag.
`daemon` sub-command instead of `-d`.

As a result, provision will to succeed at least in `google` and `digitalocean` drivers.


dgageot:
LGTM

jeanlaurent:
LGTM

zchee:
@dgageot @jeanlaurent Thanks :)

PR #2847, opened 2016-01-15
Fixes #1898. Add parameter to pass private key path to use when creat…

### LINKED ISSUE #1898: Use existing keypair for new machines on EC2

jeffellin:
…ing a machine.

It is assumed that the corresponding public certificate will be the same file name + .pub


jeffellin:
@nathanleclaire hopefully this looks better.

jeanlaurent:
This code already got two LGTM don't know why it hasn't been finished in #1898.

But let's merge it.

nathanleclaire:
LGTM

PR #2849, opened 2016-01-15
Fix shared folder mount on restart

frapposelli:
Currently the `vmwarefusion` driver does not mount shared folders when restarting the VM.

Mentioned in boot2docker/boot2docker#1105


dgageot:
LGTM

sevikkk:
The problem is - if you restart VM outside of docker-machine (host reboot/etc) folder will be not mounted anyway

Original request was done in b2d because VBox driver does exactly the same - it mounts /Users using linux startup scripts.

jeanlaurent:
LGTM

PR #2853, opened 2016-01-17
get-started.md: use same machine name

henrik:
Instructions said to create a machine named "default" but then connect to one named "dev".


dgageot:
Hi @henrik, I guess this will conflict with https://github.com/docker/machine/pull/2843

henrik:
@dgageot Good catch. If that one won't be merged for a while maybe changing to "default" consistently would be worthwhile in the meanwhile then. And when it is merged the docs could rely on the implicit default. The current state of things is that the guide has commands that won't work due to this inconsistency.

nathanleclaire:
Hi @henrik, thanks for the contribution.  My opinion is that we should use "default" everywhere in this guide as it's heavily the convention elsewhere.  If you find any inconsistencies in there please let us know!

nathanleclaire:
FYI it's that way in master now that the mentioned PR's been merged: https://github.com/docker/machine/blob/master/docs/get-started.md

henrik:
Nice!

PR #2855, opened 2016-01-18
Fix the build

dgageot:


jeanlaurent:
LGTM

jeanlaurent:
@nathanleclaire you should probably check that. 
We had to do this, since the build was failing on `master`.

nathanleclaire:
Thanks for being so quick about fixing this.  I usually try to keep the test-related code (definitions of `Fake*`) inside test-specific modules like `provisiontest`, `sshtest`, etc. but I see no harm in having it this way for now.

PR #2856, opened 2016-01-18
Fix bugsnag

dgageot:
I broke Bugsnag reports in 0.5.6.

This PR fixes the report on a create error and disables reports on upgrade for now.

We need to find a way to:
+ Re-enable reports on upgrades
+ Write a proper integration tests that verifies BugSnag reports are OK

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -12,10 +12,13 @@ import (
 
 	"errors"
 
+	"time"
+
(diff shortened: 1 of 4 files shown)


jeanlaurent:
LGTM

nathanleclaire:
Good catch.  Let's file a few follow-up issues to address the two points you mention.

PR #2867, opened 2016-01-18
Bump changelog for 0.6.0-rc1

nathanleclaire:
@docker/machine-maintainers 

We haven't really done changelogs for RCs in the past so let me know if you want to go forward with this.  I don't think we've really defined a procedure around it yet.

One option is we could merge this now, then simply change the release version to "0.6.0" and add anything new if it comes up in the next RCs, but no new features should be added between now and then.


dgageot:
Let's merge it now!
LGTM

PR #2873, opened 2016-01-19
Fix #2870

### LINKED ISSUE #2870: docker-machine can not create VBox VMs with less than 300MB RAM
Might be OS X related, but i don't think so.

**Environment**
* OS X 10.10.5 (14F27)
* VirtualBox 5.0.12r104815
* docker-machine version 0.5.5, build 02c4254
* Docker version 1.9.1, build a34a1d5

**Steps to reproduce**
```$ docker-machine -D create -d virtualbox --virtualbox-memory 256 testerle```

**Effects**
* Machine hangs on WaitForSSH forever (see attached log)
* VM is created, but hangs with a kernel panic in VBox background (see attached screenshots)

**Should be**
If there is an arbitrary limit imposed by machine or VirtualBox for some reason, machine should
log and display a message to the user and exit gracefully.

**Gist**
https://gist.github.com/marsmensch/b8980dbc4b851dbfb11a

screen shot 2016-01-19 at 08 43 41
screen shot 2016-01-19 at 08 26 40

jeanlaurent:
This prints a warning message if the memory set by virtualbox is below 640mb.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -192,6 +192,9 @@ func (d *Driver) GetURL() (string, error) {
 func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 	d.CPU = flags.Int("virtualbox-cpu-count")
 	d.Memory = flags.Int("virtualbox-memory")
+	if d.Memory < 640 { // mininum amount is around 300Mb doubling that and rounding to the next IT-compliant number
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Hm, I think perhaps it might be better to document that we don't recommend running VMs with less than, say, 512MB of memory.  I think it's reasonable to draw a boundary somewhere, for instance in some situations we require proper network access for Machine to function properly.

jeanlaurent:
Let's do some doc then...

PR #2874, opened 2016-01-19
Fix infinite loop when deleting machine in ultra weird state

jeanlaurent:
If you try to force delete a machine in a weird state ( actually a boot2docker which kernel panicked, while the vm in virtualbox is still running). The remove command fall into an infinite loop, and the machine is not removed.

This PR kills the machine rather than wait for its state to be changing.


dgageot:
LGTM

PR #2875, opened 2016-01-19
Write CLI integration tests with Go

dgageot:
Writing Integration tests with bats is a pain and it won't run on Windows.
This prevents us from running ITs on our main target platform...

I'd like to write the tests in Go because they could then run anywhere.
This is a first pass that focuses on CLI test that don't really create VMs


nathanleclaire:
I'm +1 on the general direction.  I wonder if it's worth it to try and re-use code that docker/docker has for this purpose, or if it would be easier to just build our own.

dgageot:
@nathanleclaire I prefer creating something simple targeted at Docker Machine only.
For this PR to be ready to merge, I need to get stop using `assert` underneath and better deal with cases where `docker-machine` binary is not available.

Next step will be to migrate the remaining integration tests

dgageot:
It should be go to go now.

its/tester.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return fields
+}
+
+func (dmt *dockerMachineTest) Cmd(commandLine string) IntegrationTest {
+	if strings.HasPrefix(commandLine, "machine ") {
+		return dmt.cmd(dmt.testedBinary(), parseFields(dmt.replaceDriver(commandLine[len("machine "):]))...)
+	}
+	return dmt.cmd("bash", "-c", dmt.replaceMachinePath(dmt.replaceDriver(commandLine)))
```
nathanleclaire: What's the reasoning for wrapping the command in `bash -c`?

its/tester.go
```diff
@@ -0,0 +1,218 @@
...
+		if len(fields[i]) > 2 && strings.HasPrefix(fields[i], "'") && strings.HasSuffix(fields[i], "'") {
+			fields[i] = fields[i][1 : len(fields[i])-1]
+		}
+	}
+
+	return fields
+}
+
+func (dmt *dockerMachineTest) Cmd(commandLine string) IntegrationTest {
+	if strings.HasPrefix(commandLine, "machine ") {
```
nathanleclaire: Wouldn't it be pretty reasonable to assume that the invoked command is always `docker-machine`?  Why not remove the initial `machine` from the calls and skip this bit? e.g.:

```go
test.Cmd("ls -q --filter name=^t.*e[45]").ShouldSucceed()
```

Could even change the function name to `Machine` to be more clear:

```go
test.Machine("ls -q").ShouldSucceed().ShouldContainLines(0)
```

its/tester.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (dmt *dockerMachineTest) Run(description string, action func()) {
+	dmt.Description = description
+	dmt.RawOutput = ""
+	dmt.Lines = nil
+	dmt.Err = nil
+	dmt.Failed = false
+
+	fmt.Print("\033[1;33m[..]\033[0m " + description)
```
nathanleclaire: Why not make some small helper functions which make reading the use of these escape codes a bit more readable? e.g.:

```
func red(msg string) {
    return fmt.Sprintf("\e[31m%s\e[0m")
}
```

That said, the motivation for moving over to Go-based integration tests is largely for running on Windows, no?  Have you run these tests on git bash?  In my recollection the available Windows terminals don't play very well with escape codes, so maybe we should consider skipping them.

nathanleclaire:
Looking very good, I have a few suggestions which I'd appreciate you to take a look at.  Thanks for doing this.  I think it will be much more effective / straightforward than the current bats solution.

dgageot:
Thanks @nathanleclaire for the feedback. That's great!

+ I'll try to run the tests on Windows on Monday. If the color coding stuff fails, I'll just remove it. Otherwise, I'll create utility functions.
+ Regarding `Machine("ls ...)` vs `Cmd("machine ls ...)`, I tried the first one at the beginning and found out I preferred the second one after all. But it's a matter of taste. I'm happy to switch.
+ I use `bash -c` for the same reasons we use the in bats files. Sometimes the command contain pipes or redirections and it then needs to be wrapped in a `bash -c`. With bats, the developer has to think about it. Here, it's automatic.

I'll push an updated version today and ping everybody for a review.

jeanlaurent:
LGTM

dgageot:
@nathanleclaire PTAL

nathanleclaire:
Nice!

> I use bash -c for the same reasons we use the in bats files. Sometimes the command contain pipes or redirections and it then needs to be wrapped in a bash -c. With bats, the developer has to think about it. Here, it's automatic.

That makes sense -- I had forgotten how heavily we rely on IO redirection sometimes.

LGTM

(Although I would ask for a followup revision to the docs -- I'll make a new issue for that)

PR #2881, opened 2016-01-20
Replace public IPs with reserved ones

robbertkl:
In the documentation, I've noticed a few examples where real public (DigitalOcean owned) IPv4 addresses were used. When the droplet with this address is destroyed, the address will eventually be reassigned to a new customer, probably after a cool-down period. In my opinion, such IP addresses should not be used in public documentation, especially for a project of this popularity.

In fact, the IETF came up with a few IPv4 blocks, reserved especially for this purpose. [RFC 5737](https://tools.ietf.org/html/rfc5737) describes these blocks and what they're for. For this PR, I've picked the `203.0.113.0/24` prefix, since it's the most "public looking" of the three. For no particular reason, `203.0.113.81` is the IP I've used to change the DigitalOcean addresses in the documentation.

I realise the old IP was used in examples of the `digitalocean` driver, and this IP does not belong to DigitalOcean. However, I think it's more important to use addresses which aren't routable (and never will be), than to use an address that happens to assigned to DigitalOcean (which I doubt many people will notice anyway).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ To see how to connect Docker to this machine, run: docker-machine env staging
 $ docker-machine ls
 NAME      ACTIVE   DRIVER         STATE     URL                         SWARM   DOCKER   ERRORS
 default   -        virtualbox     Running   tcp://192.168.99.188:2376           v1.9.1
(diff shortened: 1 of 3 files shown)


dgageot:
LGTM

jeanlaurent:
LGTM

PR #2889, opened 2016-01-21
Always wait for the host-only adapter to appear

dgageot:
Sometimes it takes times for a host-only adapter to appear after it's created.
This is specially true the first time such an adapter is created.


jeanlaurent:
LGTM

nathanleclaire:
LGTM

Issue #2877, opened 2016-01-19
Use $DOCKER_MACHINE_NAME as default machine


robbertkl:
Love the default machine #2843 option in v0.6.0-rc1, but the first thing I wondered was: why use a hardcoded name "default". None of my machines are actually named default.

Usually when I interact with one of my docker machines, I have the `docker-machine env` variables set in my shell, so docker commands work. Wouldn't it make more sense if `$DOCKER_MACHINE_NAME` is used instead?

Just came across #1783, where @hairyhenderson actually suggest just this! Unfortunately, I don't see any discussion/acknowledgement around his proposal.

Since #2843 is now merged and #1783 closed, I'm starting this new issue to propose changing the default machine behaviour to use the value of `$DOCKER_MACHINE_NAME` instead of "default", unless `$DOCKER_MACHINE_NAME` is unset, in which case "default" can still be used as a fallback. This way, the environment variable can be seen as an override for `defaultMachineName`.

robbertkl:
For those interested, I'm working on a PR for this proposal. Expect it today or tomorrow.

dgageot:
Closing as won't do. See https://github.com/docker/machine/pull/2897

PR #2897, opened 2016-01-21
Override default machine name with $DOCKER_MACHINE_NAME

robbertkl:
See #2877 for a description of the proposed feature, or rather addition to the default machine feature introduced with #2843.

This is a pretty straightforward change to override `defaultMachineName` with the value of `$DOCKER_MACHINE_NAME`, including the related documentation changes. To keep the current tests from failing whenever `$DOCKER_MACHINE_NAME` is set, it (and related variables) are unset in the makefile.

To add proper testcases, the environment variable should be set (using `os.Setenv()`) before each testcase, and removed afterwards. While this would probably work, it does not feel like a particular clean solution to me. In my opinion the test cases should not depend on the outside environment (e.g. unsetting environment vars in the makefile should not be needed) and the test cases should not mess with the 'real' environment, even though they run in their own subprocess and are not exported to parent processes.

The ideal way to me would be to have all environment stuff in a separate map, which is filled from `os.Environ()` and have all other places interact with it. Then for test cases it can be faked, similar to what is done for the command line. Ideally, this would be incorporated for all places involving environment variables (there are a few more).

Would the above be the right way to go? If so, I can look into a solution along those lines and update the PR. Or perhaps any other suggestions or comments on the proposed changes currently in this PR? Thanks!

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -19,10 +19,13 @@ import (
 )
 
 const (
-	defaultMachineName = "default"
+	defaultMachineNameEnvVar   = "DOCKER_MACHINE_NAME"
+	defaultMachineNameFallback = "default"
 )
 
 var (
+	defaultMachineName = getEnvWithFallback(defaultMachineNameEnvVar, defaultMachineNameFallback)
+
 	ErrHostLoad           = errors.New("All specified hosts had errors loading their configuration.")
 	ErrNoDefault          = fmt.Errorf("Error: No machine name(s) specified and no %q machine exists.", defaultMachineName)
 	ErrNoMachineSpecified = errors.New("Error: Expected to get one or more machine names as arguments")
@@ -73,6 +76,14 @@ func (c *contextCommandLine) Application() *cli.App {
 	return c.App
 }
 
+func getEnvWithFallback(envVar string, fallback string) string {
+	value := os.Getenv(envVar)
+	if value == "" {
+		value = fallback
+	}
+	return value
+}
+
 // targetHost returns a specific host name if one is indicated by the first CLI
 // arg, or the default host name if no host is specified.
 func targetHost(c CommandLine, api libmachine.API) (string, error) {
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Hi @robbertkl.  Interesting change.  I'm quite skittish on introducing such a feature because a lot of confusion about the "active" machine was common in the past (and it made it 10x harder to debug output posted by users as it wasn't clear which machine they were pointed at).

I really like the simplicity of having the short form commands always refer to the "default" machine, and think we might want to stick with that for a while before introducing more complexity.  @docker/machine-maintainers What do you all think?

robbertkl:
Hi @nathanleclaire, thanks for your feedback. I definitely agree with the "active" machine, in this case they would partly "overlap". If you have the `docker-machine env` variables in your shell, the active machine will match the default machine, unless it's not running. If it's not running, `DOCKER_HOST` would not make any sense of course, but having `DOCKER_MACHINE_NAME` would still make it possible to `docker-machine start` the machine.

Albeit the default machine functionality being a great and highly anticipated (by me and many others) feature, in its current form it's quite useless to me. I have no "default" machine and, although I could rename one of my machines, I'm often working "in context" of a specific machine designated by the `DOCKER_` environment variables (together with colour tinted terminal windows :blush:).

I'm all for simplicity and realise the above might just be my own (quaint) setup. Love to hear from the others.

dgageot:
I'm really -1 on this. The whole `docker-machine env` is already so confusing.

nathanleclaire:
Yes, we have had a really bad time in the past with this sort of automatic behavior, so perhaps we should close this PR and move discussion of how to support the desired workflow on an issue / proposal?

nathanleclaire:
Going to close this.  @robbertkl, please feel free to make an issue for the use case you would like to advocate for.  Maybe we can all come up with something which is a bit less magic (configuration file?).

Thanks!

PR #2833, opened 2016-01-14
Configures swarm-master with failover support

kunalkushwaha:
If more then one swarm-master is created with same discovery backend,
Rest of swarm-master will be configured as replica of primary master

Signed-off-by: Kunal Kushwaha <kushwaha_kunal_v7@lab.ntt.co.jp>

PR #2904, opened 2016-01-22
Fix #2903, revert #2833 since it makes it impossible to start up a Swarm master with the token discovery method

### LINKED ISSUE #2903: Not possible to create a Swarm master with the token discovery method
In v0.6.0-rc1 it is no longer possible to create a Swarm master with the token discovery method. The `docker-machine create` command does not give any error, but the swarm master does not start up successfully. Running `docker logs` on the swarm master container shows the following error message:
`time="2016-01-21T15:11:25Z" level=fatal msg="Leader election is only supported with consul, etcd and zookeeper discovery."`.

How to reproduce: Create a Swarm with Docker Machine by following the instructions on this page: https://docs.docker.com/swarm/install-w-machine/

Also two of the integration tests in `integration/core/swarm-options.bats` are failing because of this:
```
$ DRIVER=virtualbox make test-integration test/integration/core/swarm-options.bats
test/integration/run-bats.sh test/integration/core/swarm-options.bats
INFO: Run the tests with B2D_CACHE=1 to avoid downloading the boot2docker iso each time.
=> test/integration/core/swarm-options.bats
 ✓ create swarm master
 ✓ create swarm node
 ✗ ensure strategy is correct
   (in test file test/integration/core/swarm-options.bats, line 21)
     `[[ "$strategy" == "binpack" ]]' failed
   Error running connection boilerplate: Connection to Swarm cannot be checked but the certs are valid. Maybe swarm is not started
   Cannot connect to the Docker daemon. Is the docker daemon running on this host?

 ✓ ensure heartbeat
 ✓ should not show as swarm active if normal active
 ✗ should show as swarm active
   (in test file test/integration/core/swarm-options.bats, line 40)
     `[[ ${lines[1]} == *"* (swarm)"*  ]]' failed
   Error checking TLS connection: Connection to Swarm cannot be checked but the certs are valid. Maybe swarm is not started
   NAME ACTIVE URL STATE URL SWARM DOCKER ERRORS queenbee - virtualbox Running tcp://192.168.99.104:2376 queenbee (master) v1.9.1

6 tests, 2 failures
```

patrikerdes:
I propose that the --replication option is removed since it makes it impossible to start up a Swarm master with the token discovery method, which is the default method described for getting started at https://docs.docker.com/swarm/install-w-machine/

(Another option would be to parse the discovery url and only add the `--replication` option if the swarm discovery url starts with etcd, consul or zk.)

There are also currently failing integration tests in test/integration/core/swarm-options.bats, which are passing again after this fix.

Signed-off-by: Patrik Erdes <patrik@erdes.se>

### DIFF
--- a/libmachine/provision/configure_swarm.go
+++ b/libmachine/provision/configure_swarm.go
@@ -42,7 +42,7 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
 
 	if swarmOptions.Master {
 		advertiseMasterInfo := fmt.Sprintf("%s:%s", ip, "3376")
-		cmd := fmt.Sprintf("manage --tlsverify --tlscacert=%s --tlscert=%s --tlskey=%s -H %s --strategy %s --replication  --advertise %s",
+		cmd := fmt.Sprintf("manage --tlsverify --tlscacert=%s --tlscert=%s --tlskey=%s -H %s --strategy %s --advertise %s",
 			authOptions.CaCertRemotePath,
 			authOptions.ServerCertRemotePath,
 			authOptions.ServerKeyRemotePath,


nathanleclaire:
LGTM

I think it might be good to add a note about this to the docs.  I'll file a separate issue for that.

PR #2909, opened 2016-01-23
Adds PrivateIPAddress in Digitalocean config.json

jgrowl:


dgageot:
Hi @jgrowl I don't understand the goal of this PR. Can you explain? Thanks a lot

jgrowl:
@dgageot Hi! Sorry, I'm not sure if anyone else would find this useful. I have been building tooling around docker-machine that sets up a consul server cluster. It is useful for me to have access to a machine's private ip address to be able to join servers to a bootstrapped server. I saw that the ec2 driver includes the PrivateIPAddress so that's why I added it in. Maybe it would be better to just rely on extracting it from a docker-machine ssh command?

dgageot:
Thanks for the feedback @jgrowl I think relying on `docker-machine ssh` is slower but reliable in time. As long as `PrivateIPAddress` is only exposed in `config.json` it's pretty sure it will disappear one day or another. @nathanleclaire wdyt?

nathanleclaire:
I have always been wary about including IP address information in the serialized configuration directly.  IP addresses can change over time and putting them into the serialized file adds an additional "consistency" burden to the driver that the current Machine model does not support well.

I think it's probably not the end of the world to allow it, but I am also very hesitant to keep going down this slippery slope.  Most likely over time configs will need to be able to have a way to indicate a property is some kind of computed property (that `inspect` can properly query for), e.g. struct tags.

jgrowl:
Thanks for the response.This PR can be closed then. I'll just use the ssh command to extract the private IP address for now.

nathanleclaire:
OK thanks.  We'll keep private IP address querying in mind for a future feature though !

PR #2911, opened 2016-01-24
return last vbox log instead of error state code

daehyeok:
Fixed TODO comment in source file.

return last line in the vbox log file instead of error state code when virtualbox vm start fail

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -247,10 +247,7 @@ func (d *Driver) PreCreateCheck() error {
 
 // IsVTXDisabledInTheVM checks if VT-X is disabled in the started vm.
 func (d *Driver) IsVTXDisabledInTheVM() (bool, error) {
-	logPath := filepath.Join(d.ResolveStorePath(d.MachineName), "Logs", "VBox.log")
-	log.Debugf("Checking vm logs: %s", logPath)
-
-	lines, err := d.logsReader.Read(logPath)
+	lines, err := d.readVBoxLog()
 	if err != nil {
 		return true, err
 	}
@@ -492,7 +489,9 @@ func (d *Driver) Start() error {
 		}
 
 		if err := d.vbm("startvm", d.MachineName, "--type", "headless"); err != nil {
-			// TODO: We could capture the last lines of the vbox log
+			if lines, readErr := d.readVBoxLog(); readErr == nil && len(lines) > 0 {
+				return fmt.Errorf("Unable to start the VM: %s\nDetails: %s", err, lines[len(lines)-1])
+			}
 			return fmt.Errorf("Unable to start the VM: %s", err)
 		}
 	case state.Paused:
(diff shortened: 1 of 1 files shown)


dgageot:
Thank you @daehyeok for fixing the TODOs! That's huge, I'll consider adding some more!

I've got some feedback though:
+ What if `lines` is empty?
+ Also, I think we should show the original `err` also.
+ Is one line enough? Here's a typical error:

```
00:00:01.582900 ERROR [COM]: aRC=E_FAIL (0x80004005) aIID={872da645-4a9b-1727-bee2-5585105b9eed} aComponent={ConsoleWrap} aText={Failed to open/create the internal network 'HostInterfaceNetworking-VirtualBox Host-Only Ethernet Adapter #2' (VERR_INTNET_FLT_IF_NOT_FOUND).
00:00:01.582939 Failed to attach the network LUN (VERR_INTNET_FLT_IF_NOT_FOUND)}, preserve=false aResultDetail=0
00:00:01.583842 Console: Machine state changed to 'PoweredOff'
00:00:01.600272 Power up failed (vrc=VERR_INTNET_FLT_IF_NOT_FOUND, rc=E_FAIL (0X80004005))
```

daehyeok:
@dgageot 
I think Return orginal error code when line empty is good idea. But i'm not sure original err code is useful informaion or not... In my case ir was just print 'return status 1'.
How can we extract error lines only related with start fail? Filter by time???

dgageot:
@daehyeok Yes, `return status 1` is not that useful. Still I think it's not a good practice to swallow the error.

Maybe we could display the last lines until the first one containing `ERROR`. But I'm not sure it covers all the cases. Or you could show the last line as you do and add a message to tell the use the location of the log file where more information can be found.

daehyeok:
@dgageot I changed code to return original status when lines is empty or error occur during read logs.

below is my log files. only `ERROR` string in my logs is last line.
It's looks like log file created each time when VM start. So tell the location is useless too(because almost every lines in files related with VM start fail. and log file will very short ..)
```
00:00:00.038376 Log opened 2016-01-24T05:26:51.493157000Z
00:00:00.038377 Build Type: release
00:00:00.038393 OS Product: Darwin
00:00:00.038400 OS Release: 15.3.0
00:00:00.038407 OS Version: Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64
00:00:00.038519 DMI Product Name: MacBookAir6,2
00:00:00.038575 DMI Product Version: 1.0
00:00:00.038583 Host RAM: 8192MB total, 2524MB available
00:00:00.038585 Executable: /Applications/VirtualBox.app/Contents/MacOS/VBoxHeadless
00:00:00.038585 Process ID: 45642
00:00:00.038586 Package type: DARWIN_64BITS_GENERIC
00:00:00.041067 Installed Extension Packs:
00:00:00.041082   None installed!
00:00:00.042591 Console: Machine state changed to 'Starting'
00:00:00.045906 Console: Machine state changed to 'PoweredOff'
00:00:00.054675 Power up failed (vrc=VINF_SUCCESS, rc=NS_ERROR_FAILURE (0X80004005))
```

daehyeok:
i re-modified. but `fmt.Errorf("Unable to start the VM %s : %s", err, lines[len(lines)-1])` is looks not a pretty error statement. 
 Have any idea? may be one of that just print using log.Debug.....

dgageot:
Something like `fmt.Errorf("Unable to start the VM: %s\nDetails: %s", err, lines[len(lines)-1])`?

daehyeok:
Applied!

dgageot:
LGTM

jeanlaurent:
LGTM

PR #2914, opened 2016-01-25
Dockerfile: use COPY

cyphar:
ADD is far too magical, and we should be steering people away from it,
so using it in an official Docker project sends a mixed message.

Signed-off-by: Aleksa Sarai <asarai@suse.com>


dgageot:
LGTM

jeanlaurent:
LGTM

PR #2915, opened 2016-01-25
FIX #494 - Use same binary filename scheme as other docker-project

### LINKED ISSUE #494: Please use `uname -s`-`uname -m` as suffix for the binary releases
This would make it easier to embed the download links in provisioning scripts.

Example:
```
curl -s -L https://github.com/docker/machine/releases/download/0.1.0-rc2/docker-machine-`uname -s`-`uname -m` > /usr/local/bin/docker-machine
```

### LINKED ISSUE #985: Release artefact names should follow covnetion used in Compose
To get Compose, I run
```
curl -L https://github.com/docker/compose/releases/download/1.2.0rc3/docker-compose-`uname -s`-`uname -m`
```

where expanded URL is
```
https://github.com/docker/compose/releases/download/1.1.0/docker-compose-Darwin-x86_64
```

With Machine, I cannot do the same, as URLs look like
```
https://github.com/docker/machine/releases/download/v0.2.0-rc3/docker-machine_darwin-amd64
```

So first of all it's the way OS/architecture prefix is. Secondly, there is the `1.2.0rc3` vs `v0.2.0-rc3`, which is somewhat less annoying.

jeanlaurent:
This PR puts docker-machine binaries in the same format as the other docker binaries (docker/docker and docker/compose).

This also update install instructions and release template.

btw the windows bash script instruction where badly failing under mingw.

This fix #985 and #494

### DIFF
--- a/docs/install-machine.md
+++ b/docs/install-machine.md
@@ -30,26 +30,23 @@ instructions in the next section.
 
 2.  Download the Docker Machine binary and extract it to your PATH.
 
-        Linux:
+    If you are running OS X or Linux:
 
-            $ curl -L https://github.com/docker/machine/releases/download/v0.5.5/docker-machine_linux-amd64 >/usr/local/bin/docker-machine && \
-            chmod +x /usr/local/bin/docker-machine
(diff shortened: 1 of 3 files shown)


dgageot:
LGTM

PR #2916, opened 2016-01-25
Add more logs, specially for windows users

dgageot:
On windows, Machine is sometimes waiting for the user to accept modifications to Virtualbox networks adapters. Sometimes those notifications are shown only in the taskbar which can be easily missed.


jeanlaurent:
LGTM

Issue #2920, opened 2016-01-25
docker-machine status outputs to stderr


lambdafu:
Srsly?

```
$ docker-machine status dev 2> /dev/null 
$ docker-machine status dev > /dev/null 
Running
```

jeanlaurent:
Sounds like a duplicate of #2587, with v0.6.1-rc1 it is fixed.

Please re-open if this occurs on the latest version.

PR #2922, opened 2016-01-26
Machine output redirected to stdout.

kunalkushwaha:
log functions are used both writing output and logging errors.
As raised in bug #2920, output redirection in script creates confusion.
This fixes the issue by redirecting error to stderr and output to stdout.

### DIFF
--- a/commands/config.go
+++ b/commands/config.go
@@ -10,9 +10,9 @@ import (
 )
 
 func cmdConfig(c CommandLine, api libmachine.API) error {
-	// Ensure that log messages always go to stderr when this command is
-	// being run (it is intended to be run in a subshell)
-	log.SetOutWriter(os.Stderr)
+
+	log.SetOutWriter(os.Stdout)
+	log.SetErrWriter(os.Stderr)
 
 	target, err := targetHost(c, api)
 	if err != nil {


dgageot:
Hi @kunalkushwaha Can you give an example of something fixed by this PR?

jeanlaurent:
In the `config` command, we take care of rerouting any log from stout to stderr because we protect ourselves of any logs that can messed up the output.

We do the same in the `env` command.

nathanleclaire:
Hi @kunalkushwaha, thanks for the PR.

It is deliberate that any log output should go to STDERR for `config` and for `env`.  The reason for this is that those commands are meant to be executed within subshells (`$(...)` or backticks), and anything printed to STDOUT from them will be interpreted by the shell as a command. (STDERR will "escape" to be displayed on the terminal)  So, we are intentionally being deliberate about what we send to STDOUT.

I think we should probably not merge this.

kunalkushwaha:
@dgageot @jeanlaurent @nathanleclaire I understood the reason for output redirection to stderr.

I saw bug assigned for 0.6.0 release, so tried to fix it. I admit it was not well thought solution though.
Closing the PR.

Thanks

nathanleclaire:
Thanks for following up to close @kunalkushwaha !

PR #2924, opened 2016-01-26
added hyperv static mac address and vlan option for machine creation

sipicsg:
For networks using dhcp and/or vlan organized subnets


dgageot:
Hi @sipicsg You need to format `drivers/hyperv/hyperv.go` for the tests to pass. Thanks a lot!

PR #2925, opened 2016-01-26
Fix #2918 and match doc with current code

### LINKED ISSUE #2918: undocumented flag: --vmwarevsphere-hostsystem/VSPHERE_HOSTSYSTEM for vmwarevsphere driver
Ran into an issue using the vSphere driver.

```Error creating machine: Error with pre-create check: default host resolves to multiple instances, please specify```

After some lengthy searching, I finally discovered there was a flag I could pass to specify which hypervisor to connect to, but this info was not documented on the driver info page...

https://docs.docker.com/machine/drivers/vsphere/

jeanlaurent:
vsphere doc was not matched with the state of the code. 
Added 2 missing flags and removed one.

### DIFF
--- a/docs/drivers/vsphere.md
+++ b/docs/drivers/vsphere.md
@@ -14,18 +14,19 @@ Creates machines on a [VMware vSphere](http://www.vmware.com/products/vsphere) V
 
 Options:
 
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

PR #2929, opened 2016-01-26
FIX #2705 Add a command line sample for each driver

### LINKED ISSUE #2705: Missing documentation for --driver names?
With older versions of docker-machine I believe "help create" would list all the possible options for drivers. I just updated from the older version to the latest edition and it wasn't clear that the hyper-v driver was renamed hyperv. Can the available driver names be printed with "help create", or at least documented somewhere? https://docs.docker.com/machine/drivers/ doesn't list the driver name and the hyper-v page doesn't give an example that includes --driver "hyperv". There's a few other pages that also omit the full example.

dgageot:


jeanlaurent:
LGTM

PR #2931, opened 2016-01-26
Remove spurious log that adds no value and make ITs fail sometimes because of an additional line

dgageot:
sometimes fail because of an additional line.


jeanlaurent:
LGTM

PR #2933, opened 2016-01-26
[vmwarevsphere] Start using govmomi provided helpers

caglar10ur:
Also update govmomi to latest revision and call Logout method when needed

Signed-off-by: S.Çağlar Onur <conur@vmware.com>


frapposelli:
Awesome work @caglar10ur!

@dgageot this LGTM

caglar10ur:
Removed the part says "The vSphere driver depends on [`govc`]" from the docs as it's no longer true. Single ESXi host is happy without "ha-datacenter" parameter so that's gone too.

nathanleclaire:
Pretty close before the release for such a large change of dependency but it doesn't seem to add any new features, and @frapposelli said it's alright so maybe it's OK.

@caglar10ur Could you elaborate just a bit on what the motivation behind the PR is?  Does it just update the dep to use a slightly better / newer version of the library?

caglar10ur:
@nathanleclaire correct. This is just a cleanup for the driver code with no additional features. We are updating the govmomi dependency to the latest version because that's the one provides new helpers, also it's important to note that we only use/test latest govmomi version.

I think the change is pretty isolated from the rest of the docker machine code so it should be safe to merge (ahem - famous last words) but I understand if you decide to wait and merge it after the release.

nathanleclaire:
> I think the change is pretty isolated from the rest of the docker machine code so it should be safe to merge (ahem - famous last words) but I understand if you decide to wait and merge it after the release.

Haha.

> This is just a cleanup for the driver code with no additional features. We are updating the govmomi dependency to the latest version because that's the one provides new helpers, also it's important to note that we only use/test latest govmomi version.

OK -- just making sure my understanding of this was correct -- in this case my judgment call is that it's OK to merge and LGTM.

@dgageot @jeanlaurent I'll let you all be the final decider / LGTM though.

dgageot:
LGTM

frapposelli:
:tada:

PR #2936, opened 2016-01-27
Wait for Docker daemon on host start

nathanleclaire:

Closes https://github.com/docker/machine/issues/2175

(some of the instances of the problem, at least)

You should be able to duplicate this bug by doing something like this with the most recent Machine RC:

```console
$ docker-machine stop
$ docker-machine start; docker-machine env
Error checking TLS connection: Error checking and/or regenerating the certs: There was an error validating certificates for host "192.168.99.100:2376": dial tcp 192.168.99.100:2376: getsockopt: connection refused
You can attempt to regenerate them using 'docker-machine regenerate-certs [name]'.
Be advised that this will trigger a Docker daemon restart which will stop running containers.
```

This patch fixes the issue by waiting for the Docker daemon as part of the `Host` `Start`.

### DIFF
--- a/libmachine/engine/engine.go
+++ b/libmachine/engine/engine.go
@@ -1,5 +1,9 @@
 package engine
 
+const (
+	DefaultPort = 2376
+)
+
 type Options struct {
 	ArbitraryFlags   []string
 	DNS              []string `json:"Dns"`
--- a/libmachine/host/host.go
+++ b/libmachine/host/host.go
@@ -111,6 +111,17 @@ func (h *Host) Start() error {
 	}
 
 	log.Infof("Machine %q was started.", h.Name)
+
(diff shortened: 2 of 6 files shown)


libmachine/engine/engine.go
```diff
@@ -1,5 +1,9 @@
 package engine
 
+const (
+	DefaultPort = 2376
```
nathanleclaire: This was pretty long overdue anyway so I've gone ahead and started pulling instances of this magic variable into a constant.

nathanleclaire:
@vikstrous FYI

dgageot:
@nathanleclaire I'm going to take a look this morning

dgageot:
LGTM

dgageot:
Will follow up with something similar on `Restart()`

PR #2953, opened 2016-01-28
Update release script

jeanlaurent:
Various fixes on release script due to binaries change and the new IT test in go


dgageot:
LGTM

PR #2954, opened 2016-01-28
Fix #2029 - local drivers dont run with engine-install-url

### LINKED ISSUE #2029: Drivers which don't use `--engine-install-url` should at least warn if the option is used
If someone wants to install a RC or experimental version, they can do something like:

```
docker-machine create -d amazonec2 --engine-install-url test.docker.com mytestbox
```

However, if they try to do that with a local VM (`-d virtualbox`) the option is silently ignored (since the `virtualbox` driver uses boot2docker images rather than the SSH-based provisioning system).

Instead, the `--virtualbox-boot2docker-url` has to be used.

It would be nice if Machine could display a warning when the `--engine-install-url` (and potentially other options) are used with a driver that doesn't use them.

Also: I couldn't find much info in the docs about this; but maybe I didn't look hard enough (I figured it out with the command-line help, mostly).

/cc @ManoMarks

jeanlaurent:
Got caught on this today, should not be caught twice now...


dgageot:
and now lint!

jeanlaurent:
@dgageot PTAL

jeanlaurent:
@dgageot PTAL again.

dgageot:
LGTM

PR #2961, opened 2016-01-29
Update release script for GNU sed compatibility

nathanleclaire:
I changed my default OSX `sed` to GNU `sed`, so this line blows up on me.

This change should work with both BSD `sed` and GNU `sed`.


dgageot:
LGTM

PR #2965, opened 2016-01-29
Fix non-templated release value

nathanleclaire:
The value was not templated in this file -- this fixes the issue.


jeanlaurent:
Good catch. 

LGTM

PR #2979, opened 2016-02-02
Exit with code 3 if error is during pre-create check

nathanleclaire:

Would be good to get a test for this, but also could be pretty drawn out (need to have a plugin with a reliable precreate fail, and write an integration test for it).


nathanleclaire:
OK fixed the failing test -- will keep an eye on this and merge if it LGTY when CI is finished.

jeanlaurent:
LGTM

PR #2981, opened 2016-02-03
drivers/exoscale: fix configuration of exoscale endpoint

vincentbernat:
The environment variable was `EXOSCALE_ENDPOINT` but the flag was
`--exoscale-url`. When configuring from flags, the wrong name was
used. Therefore, a user was unable to use an alternate endpoint.


nathanleclaire:
LGTM

PR #2984, opened 2016-02-03
FIX #2982 Fix shell auto detection

### LINKED ISSUE #2982: Shell auto detection is broken
It always default to `bash` even on windows and even if a `SHELL` env variable is set

dgageot:


nathanleclaire:
LGTM

PR #3018, opened 2016-02-08
Improve --virtualbox-no-vtx-check option to allow starting the VM

StefanScherer:
The VirtualBox driver option `--virtualbox-no-vtx-check` can be used to turn off the VT-X checks. The flag is used only once, but there is a second place in the code to allow to start VM without this check.

With this PR the VM can be started without any VT-X checks.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -507,14 +507,16 @@ func (d *Driver) Start() error {
 		log.Infof("VM not in restartable state")
 	}
 
-	// Verify that VT-X is not disabled in the started VM
-	vtxIsDisabled, err := d.IsVTXDisabledInTheVM()
-	if err != nil {
-		return fmt.Errorf("Checking if hardware virtualization is enabled failed: %s", err)
-	}
+	if !d.NoVTXCheck {
+		// Verify that VT-X is not disabled in the started VM
+		vtxIsDisabled, err := d.IsVTXDisabledInTheVM()
+		if err != nil {
+			return fmt.Errorf("Checking if hardware virtualization is enabled failed: %s", err)
+		}
 
-	if vtxIsDisabled {
-		return ErrMustEnableVTX
+		if vtxIsDisabled {
+			return ErrMustEnableVTX
+		}
 	}
 
 	log.Infof("Waiting for an IP...")


nathanleclaire:
Why not move the actual check of the option into the function `IsVTXDisabledInTheVM`?

dgageot:
@StefanScherer @nathanleclaire I don't understand what this PR is supposed to do.
`IsVTXDisabledInTheVM` is supposed to check if the vm could boot or not. This has to be done whatever the `--virtualbox-no-vtx-check` flag value is. 
The flag disables pre-checks that might fail for a wrong reason.
I've never seen post-checks fail with a successfully started VM. Have you?

StefanScherer:
@dgageot Yes, I have, but "far out in the uncharted backwaters of the unfashionable end of the western spiral arm of the Galaxy" where a [32bit boot2docker.iso](https://github.com/StefanScherer/boot2docker/pull/1#issuecomment-181556366) exists that could be started without VT-X support.

I encountered the problem spinning up a VirtualBox Docker Machine VM inside [Travis](https://travis-ci.org/StefanScherer/travis-docker-check-config/builds/107842825) or [AppVeyor](https://ci.appveyor.com/project/StefanScherer/boot2docker/build/36) where no nested hypervisor is available. I found the `--virtualbox-no-vtx-check` option but still docker-machine aborted before starting the VM. So I added the check for the option there as well. Now docker-machine doesn't "stand in the way" of doing this.

I have to confess that this is like exploring a new continent, but I was curious to see how far we could go with an end-to-end integration test for the VirtualBox driver.

dgageot:
@StefanScherer Could you share the Vbox.log of a 32bits vm that succeeds to start but that `IsVTXDisabledInTheVM` will recognise as failed?

StefanScherer:
@dgageot Here is a gist:

* [description of setup](https://gist.github.com/StefanScherer/f576009c5b5c51dc912f#file-01-docs-md)
  * [VBox.log of 64bit VM](https://gist.github.com/StefanScherer/f576009c5b5c51dc912f#file-02-test-vbox-log)
* [spin up 32bit vm](https://gist.github.com/StefanScherer/f576009c5b5c51dc912f#file-03-docs-md)
  * [VBox.log of 32bit VM](https://gist.github.com/StefanScherer/f576009c5b5c51dc912f#file-04-test32-vbox-log)

StefanScherer:
This is what it looks like:

bildschirmfoto 2016-02-11 um 18 37 29

dgageot:
Thanks for the logs. Now I understand better and I guess what we should do is improve the recognition of a VM that couldn't start because of a VT-X issue.

Currently, it searches for:
 + VT-x is disabled
 + the host CPU does NOT support HW virtualization
 + VERR_VMX_UNABLE_TO_START_VM

in the VBox.log.

Because it sees `00:00:00.569994 WARNING! 64-bit guest type selected but the host CPU does NOT support HW virtualization.` in you log, it thinks the VM failed to start.

I need to find the original log that led us to add the search for this string. Maybe it can sometimes be an error and sometimes just a warning.

Still, ATEOD, I tend to not think it's a good thing to disable the `IsVTXDisabledInTheVM` altogether when `--virtualbox-no-vtx-check` is set.

dgageot:
see #2127

StefanScherer:
Sure, spinning up a 32bit VM / boot2docker.iso also should set the `OSType="Linux26"` instead of `OSType="Linux26_64"` in the .vbox file so VirtualBox doesn't come up with that warning.

So instead of this PR someone would need an option `--virtualbox-ostype=Linux26` :angel:

dgageot:
@StefanScherer I completely changed my mind! I've seen a couple of other corner cases where disabling the post-create check is helpful. So I'm ready to merge your PR!

LGTM

StefanScherer:
@dgageot OK. I'm glad I could help ;-)

jeanlaurent:
LGTM

PR #3019, opened 2016-02-08
Warn that Boot2Docker ISO won't be auto upgraded if its URL is explicitly set

pdalpra:
This PR adds a log to warn the user that automatic upgrade won't happen if trying to upgrade a machine created with an explicitly set Boot2Docker ISO URL.

### DIFF
--- a/libmachine/mcnutils/b2d.go
+++ b/libmachine/mcnutils/b2d.go
@@ -374,12 +374,17 @@ func (b *B2dUtils) UpdateISOCache(isoURL string) error {
 		}
 	}
 
+	exists := b.exists()
+
 	if isoURL != "" {
+		if exists {
+			// Warn that the b2d iso won't be updated if isoURL is set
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Thanks.  Just have some small nitpicks.

libmachine/mcnutils/b2d.go
```diff
@@ -374,12 +374,17 @@ func (b *B2dUtils) UpdateISOCache(isoURL string) error {
 		}
 	}
 
+	exists := b.exists()
+
 	if isoURL != "" {
+		if exists {
+			// Warn that the b2d iso won't be updated if isoURL is set
+			log.Infof("Boot2Docker URL was explictly set to %s, automatic upgrade is disabled.", isoURL)
```
nathanleclaire: This should probably be set to `log.Warnf` instead and I'd make the copy say: "Boot2Docker URL was explicitly set to %q at create time, so Docker Machine cannot upgrade this machine to the latest version."

pdalpra: fixed !

nathanleclaire:
Cool -- LGTM

PR #3021, opened 2016-02-09
Add retry support to AWS driver

dhiltgen:
Some AWS API failures are retryable, so we should give the option to attempt
retries in these cases.  One particular example is the API rate limiting that
kicks in if you try to create many machines in parallel.  Without this change,
I can only create ~3 machines in parallel, but with this change I can easily
create 10 or more machines in parallel.

While this change helps `ls`, given the current fan-out model for plugins, once
you get a large number of AWS machines, you'll still wind up seeing
"RequestLimitExceeded: Request limit exceeded." errors.

Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>


jeanlaurent:
Thx @dhiltgen could you update your PR and update the [doc](https://github.com/docker/machine/blob/master/docs/drivers/aws.md) accordingly ?
( Try to squash your commit too )

dhiltgen:
Added docs.

jeanlaurent:
LGTM

nathanleclaire:
Oh, nice!

LGTM

PR #3027, opened 2016-02-10
Fix #2995

### LINKED ISSUE #2995: hyperv driver should be more explicit when the vswitch is not created
We should output a link to the documentation explaning how it works.

jeanlaurent:
Add a pointer to the driver doc to setup a vswitch.


nathanleclaire:
LGTM

PR #3029, opened 2016-02-10
Remove `MountFlags=slave` from systemd unit definitions

errordeveloper:
This prevents one from using mount propagation flags in Docker v1.10 (see https://github.com/docker/docker/issues/19625).

### DIFF
--- a/libmachine/provision/coreos.go
+++ b/libmachine/provision/coreos.go
@@ -72,7 +72,6 @@ Requires=docker.socket early-docker.target
 [Service]
 Environment=TMPDIR=/var/tmp
 EnvironmentFile=-/run/flannel_docker_opts.env
-MountFlags=slave
 LimitNOFILE=1048576
 LimitNPROC=1048576
 ExecStart=/usr/lib/coreos/dockerd daemon --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:{{.DockerPort}} --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}}{{ range .EngineOptions.Labels }} --label {{.}}{{ end }}{{ range .EngineOptions.InsecureRegistry }} --insecure-registry {{.}}{{ end }}{{ range .EngineOptions.RegistryMirror }} --registry-mirror {{.}}{{ end }}{{ range .EngineOptions.ArbitraryFlags }} --{{.}}{{ end }} \$DOCKER_OPTS \$DOCKER_OPT_BIP \$DOCKER_OPT_MTU \$DOCKER_OPT_IPMASQ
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Thanks @errordeveloper -- Can you elaborate a bit on your use case?  e.g. some steps to reproduce that don't work without this patch.  My notion of how this ties together is fairly hazy.  If `MountFlags=slave` is what's set by the default Docker packages maybe we should consider an additional `--engine-*` type flag to enable this conditionally as well.

errordeveloper:
@nathanleclaire what I'm working on is a thing makes Kubernetes deployment very easy.

Here is the project itself: https://github.com/weaveworks/weave-kubernetes-anywhere

And here is the Docker Machine example: https://github.com/weaveworks/weave-kubernetes-anywhere/tree/master/examples/docker-machine

Feel free to look at the real thing, the failure surfaced for me on DigitalOcean, and seemed extremely obscure for a few hours. The code that actually fails is in [`setup-kubelet-volumes.sh`] and is being called from [`create-cluster.sh`](https://github.com/weaveworks/weave-kubernetes-anywhere/blob/6439026f07bf64598e411c82b5f1cb2bc14bafe8/docker-images/setup-kubelet-volumes.sh#L31-L32), which is called from `https://github.com/weaveworks/weave-kubernetes-anywhere/blob/6439026f07bf64598e411c82b5f1cb2bc14bafe8/examples/docker-machine/create-cluster.sh#L153-L157). The reason this is required is that `kubelet` happens to mount secrets in a `tmpfs`, for some extra security.

The basics test that fails on Ubuntu is this:
```
mount --make-rshared /
docker run -v /mnt:/host-mnt:rw,rshared busybox
```

errordeveloper:
It would be helpful to know when and why `MountFlags=slave` first appeared. It is certainly still shipped in official packages. I have asked https://github.com/docker/docker/issues/19625#issuecomment-182284053 and haven't got an answer just yet. I bet this a kind of thing that was added one day and people don't touch it cause they don't know what it is exactly, it really needs to be clarified not that it stand on the way for someone needing to use a feature available in the latest release.

AkihiroSuda:
@errordeveloper

Without the flag, the host-side /dev/* can be unexpectedly unmounted.

https://github.com/docker/docker/issues/20670

AkihiroSuda:
FYI, this commit introduced the flag
(You can find such information by clicking "blame" button in 
GitHub web UI)
https://github.com/docker/docker/commit/eb76cb2301fc883941bc4ca2d9ebc3a486ab8e0a

nathanleclaire:
Yeah, I think we need to err on the side of sticking with Docker upstream here, I'd be willing to consider proposals to customize turning these types of features on/off during provisioning, however it's a slippery slope so we need to be careful about what kind of customization we enable (supporting on all provisioners etc. is difficult).

errordeveloper:
I think there should be a more explicit way to fix the issue with `/dev`, which might lie in the Docker daemon rather then the systemd unit...

zhulinhong:
Remove `MountFlags=slave` 
The systemd default flags is  `shared `,  mounts and unmounts are propagated from the host to the container and vice versa. Is there has the security issue.  
Is user can unmount  the volume in the container?

errordeveloper:
I've submitted this change to Docker upstream, how is it mean to propagate to Machine?

zhulinhong:
Probably  the  mounts and unmounts operations  on the host will effect the containers.  
And the  mounts and unmounts operations  in container will effect the host.

errordeveloper:
@zhulinhong not that, I mean https://github.com/docker/docker/pull/22806 has been merged, so I still need to make a PR to this repo or what?

zhulinhong:
No need, thx a lot. It's configurable, I can change the config if I need.

nathanleclaire:
I think we'll still need it, to ensure the upstream changes aren't over-written to be excluded, if you'd liek to re-open.

errordeveloper:
Ok, that's what I was wondering, i.e. whether I should copy upstream
version of the units here, there may be other flags too.

nathanleclaire:
Ah, yes, let's stick as closely as possible to upstream please.

pbecotte:
@errordeveloper @nathanleclaire This PR is still needed- the systemd units from machine overwrite the ones from upstream, so this is still setting this flag.

PR #3057, opened 2016-02-14
env emits DOCKER_HOST_IP

msabramo:
This makes the `env` command emit a new environment variable called `DOCKER_HOST_IP`.

E.g.:

```
$ bin/docker-machine env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.101:2376"
export DOCKER_HOST_IP="192.168.99.101"
export DOCKER_CERT_PATH="/Users/marca/.docker/machine/machines/default"
export DOCKER_MACHINE_NAME="default"
# eval $(bin/docker-machine env)
```

This is useful because it can be used in `docker-compose.yml` to make a container's external port bind to the Docker host's public IP address.

```yaml
...
    ports:
      - "${DOCKER_HOST_IP}::8000"
...
```

I prefer to bind to the `DOCKER_HOST_IP` rather than letting it default to `0.0.0.0`, because then I get useful output from `docker-compose port` and can do stuff like:

```
$ docker-compose port anonweb 8000
192.168.99.101:32775

$ curl -I http://$(docker-compose port anonweb 8000)/status/pid
HTTP/1.1 200 OK
Server: gunicorn/19.4.5
Date: Sun, 14 Feb 2016 05:41:58 GMT
Connection: close
Content-Type: application/json; charset=UTF-8
Content-Length: 113
SM-Request-ID: 37e72936-0c2a-44d3-8494-e0d6280387fd
```

See:
https://github.com/docker/compose/issues/2915#issuecomment-183816763

Cc: @sudarkoff


msabramo:
Hmm, why did the `documentation` check fail?

SvenDowideit:
retest this pleaes

SvenDowideit:
test this please

msabramo:
@SvenDowideit Oh cool! Thanks! Good to know!

dgageot:
@msabramo Sorry, we are clearly not going to add another environment variable to the `env` command. cc @docker/machine-maintainers wdyt?

jeanlaurent:
@dgageot no, we are not. And in this case, this is clearly duplication for a very specific use case.

msabramo:
Sure, your call of course. What is the drawback of adding an environment variable?

Maybe someone can suggest an alternative way to get a usable IP address/port pair for accessing the container from a Mac? The best I can come up with is using a combo of `docker-machine ip` with `docker-compose port` or `docker port` and that's fine but it feels a bit weird to use 2 different tools.

dgageot:
@msabramo the role of the `env` command is to point `docker` CLI to the right machine. It sets the environment variables that are needed by the CLI. 
Additional variables, that are not used by docker CLI, are out of scope for the `env` command.

Sorry @msabramo and thanks again for the PR. Next time, you have something you'd like to see implemented, please open a ticket, discuss about the feature and then, when everybody agree that it's a good idea to move forward, you or somebody else can open a PR.

msabramo:
Ah ok thanks for the explanation. 

Technically I am using it in a `docker-compose.yml` file and so Docker Compose is passing the IP address to Docker which is using it. But it's not strictly necessary for connecting to Docker.

Unfortunately this was the best idea I had for making it really easy for folks to run containers from Macs and easily be able to access them. 

I wonder if there is some other solution for this that I have overlooked?

dnephin:
I've seen a few stackoverflow issues with this exact problem, and at least one Compose issue. 

I think this would be useful for developers, is there a disadvantage to provide this information in a pre-parsed form (instead of requiring a developer to parse out the IP address from the `DOCKER_HOST` var) ?

dgageot:
@dnephin `docker-machine ip $DOCKER_MACHINE_NAME` should do the trick, right?

dnephin:
I guess someone would need to do
```
export DOCKER_HOST_IP=$(docker-machine ip default)
```

in a script somewhere, which doesn't seem that unreasonable to me.

PR #3064, opened 2016-02-17
Docs typo, generic driver uses _public_ key

robert-wallis:
This is a docs update, they should say `public` not private.
The public key works, the private key does not.

This PR updates the documentation for the `--generic-ssh-key` flag to ask for the `public` key instead of the private key.

## How to reproduce ... the confusion:
Given a server where your `~/.ssh/id_rsa.pub` key is an `authorized_key`.
When you use create a new docker-machine with the generic driver with the following command.
```
docker-machine create --driver=generic --generic-ip-address=example.com --generic-ssh-key=/Users/robert/.ssh/id_rsa failing-machine
```
(note: the above command is using the private key `id_rsa`, not `id_rsa.pub` as directed by the documentation)
Then the command will hang and eventually fail with the following output.
```
Running pre-create checks...
Creating machine...
(failing-machine) Importing SSH key...
```

However if you simply use the public key in the `id_rsa.pub` file instead the driver functions correctly.

### DIFF
--- a/docs/drivers/generic.md
+++ b/docs/drivers/generic.md
@@ -28,7 +28,7 @@ The driver will perform a list of tasks on create:
 Options:
 
 -   `--generic-ip-address`: **required** IP Address of host.
--   `--generic-ssh-key`: Path to the SSH user private key.
+-   `--generic-ssh-key`: Path to the SSH user public key.
 -   `--generic-ssh-user`: SSH username used to connect.
 -   `--generic-ssh-port`: Port to use for SSH.
 
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -41,7 +41,7 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 		},
 		mcnflag.StringFlag{
 			Name:   "generic-ssh-key",
-			Usage:  "SSH private key path (if not provided, identities in ssh-agent will be used)",
+			Usage:  "SSH public key path (if not provided, identities in ssh-agent will be used)",
 			Value:  "",
 			EnvVar: "GENERIC_SSH_KEY",
 		},


dgageot:
Hi @robert-wallis, I'm sorry but you must have something broken in your configuration because it's the private key that needs to be passed. Not the public key. The public key is, by definition, public and cannot be used to authenticate the user.

robert-wallis:
Thank you, I did some extra research and found out we are both right and both wrong ha ha :)

The main thing that causes the behavior I see is my private key is password protected.  Which somehow makes the public key work in docker-machine after I unlock it in my keychain.

### How to reproduce:

```bash
ssh-keygen -b 2048 -t rsa -N "keypass" -f testkey 

cat > Dockerfile <<EOF
# based on https://docs.docker.com/engine/examples/running_ssh_service/
FROM ubuntu

RUN apt-get install -y openssh-server
RUN mkdir /var/run/sshd

# SSH login fix. Otherwise user is kicked off after login
RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd

RUN useradd docker-machine
RUN echo "docker-machine ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers.d/docker-machine

RUN mkdir -p /home/docker-machine/.ssh/
COPY ./testkey.pub /home/docker-machine/.ssh/authorized_keys

EXPOSE 22

CMD ["/usr/sbin/sshd", "-D"]

EOF

docker build -t sshd .
docker run -d -p 12345:22 --name inst sshd

INST_IP=$(docker-machine ip)

echo "=== 'keypass' is the password to unlock testkey ================="
ssh docker-machine@${INST_IP} -p 12345 -i testkey 'sudo ls'

echo "=== private key =================="
docker-machine create --driver generic --generic-ip-address=${INST_IP} --generic-ssh-port=12345 --generic-ssh-key=testkey --generic-ssh-user=docker-machine testPrivate

echo "=== public key ==================="
docker-machine create --driver generic --generic-ip-address=${INST_IP} --generic-ssh-port=12345 --generic-ssh-key=testkey.pub --generic-ssh-user=docker-machine testPublic

echo "==== cleaning up ================="
docker-machine rm -y testPrivate
docker-machine rm -y testPublic
docker stop inst
docker rm inst
#docker rmi sshd
#rm testkey testkey.pub Dockerfile
```

###  Output (truncated)
```
Successfully built dbb8fb8c2e98
c48209776708e3775393852b0812466aeec7bb86b1c19c27e09aadc1e3d3111c
=== 'keypass' is the password to unlock testkey =================
Saving password to keychain failed
Identity added: testkey ((null))
=== private key ==================
Running pre-create checks...
Creating machine...
(testPrivate) Importing SSH key...
=== public key ===================
Running pre-create checks...
Creating machine...
(testPublic) Importing SSH key...
Detecting the provisioner...
Provisioning with ubuntu(upstart)...
Error creating machine: Error running provisioning: Something went wrong running an SSH command!
command : sudo hostname testPublic && echo "testPublic" | sudo tee /etc/hostname
err     : exit status 1
output  : hostname: you must be root to change the host name


==== cleaning up =================
About to remove testPrivate
Successfully removed testPrivate
About to remove testPublic
Successfully removed testPublic
inst
inst
```
(note that using the public key it got to "Provisioning with ubuntu" which means it worked)

dgageot:
If you use a password protected key, then the only way to make generic driver properly is to use an ssh agent as stated in the documentation. Have you tried?

robert-wallis:
Yes, I'll close this PR. It breaks the more common non-protected key case.

PR #1069, opened 2015-04-28
Disable problematic VirtualBox DNS proxy

jfieber:
Ran into DNS problems, research found them identified and corrected in boot2docker.

References:

* https://github.com/boot2docker/boot2docker/issues/451
* https://github.com/boot2docker/boot2docker-cli/pull/320
* https://www.virtualbox.org/ticket/11911

PR #3069, opened 2016-02-17
VirtualBox: use dns proxy mode

dgageot:


nathanleclaire:
I'm not sure it's a good idea. I know that having to `restart` to fix DNS is annoying, but IIRC the alternative (we originally introduced this settings in https://github.com/docker/machine/pull/1069) is worse (unfixable).

dgageot:
We've never tried this settings. We used dnsresolver, not dnsproxy. Dnsresolver is known to have bugs. It seems that dnsproxy has no such problem

nathanleclaire:
Maybe it's fixed in VirtualBox > 5, but https://www.virtualbox.org/ticket/13002 (linked to from the PR mentioned above) suggests that setting `--natdnsproxy on` causes DNS lookups to take a very long time (~5 seconds).  I'm going to fiddle a bit and see if the issue still exists.

nathanleclaire:
I'm not seeing any issues after some testing so I'm willing to give it a shot (maybe there were issues in 4.*.* that were fixed in 5).  LGTM

PR #3075, opened 2016-02-17
Enable syncing time for VMWare Fusion guests

synfinatic:
By default, VMWare Fusion does not sync the guest OS's time to the host
and when running Fusion+docker-machine on a laptop the guest clock will
drift and not resync which can create a variety of problems (such as
authentication issues with AWS, etc)


dgageot:
ping @frapposelli

dgageot:
LGTM

frapposelli:
Isn't boot2docker using `ntpd` to keep time in sync? using VMware tools to sync time is ok but it will lead to sudden time jumps (as it doesn't drift time).

frapposelli:
Besides that comment, LGTM

synfinatic:
The problem is when I suspend my laptop, the clock falls behind so far ntp won't resync in a reasonable amount of time.  In my case I'm seeing > 5min skew just by taking my laptop home after work which leads to AWS API auth failures. In my case, I'd definitely choose clock jumping to waiting hours for ntp to slew the clock.

frapposelli:
@synfinatic fair enough, to be honest, we do have scripts to handle that scenario, but they're not included in boot2docker (I'll make it a todo item :) ).

@dgageot proper LGTM

PR #3081, opened 2016-02-18
docs: remove reference to lxc-docker

thaJeztah:
The package on apt.dockerproject.org is now called 'docker-engine'.

Also replaced 'dev' with 'default', because that's the default name for machines.


dgageot:
LGTM

PR #3082, opened 2016-02-18
Fix #3015

### LINKED ISSUE #3015: The 'docker-machine active' command should print a different message than "No active host found" on timeout
I sometimes get "No active host found" when running the command `docker-machine active`, even though there is an active host. This is because of the default 10 s timeout. When there is a timeout, I think that the the error message should somehow state that there was a timeout. Just saying "No active host found" is misleading.

How to reproduce:
I have 6 machines on AWS and one virtualbox, I can reproduce this issue quite easily. When the command take less than 10 seconds, it is successful. When it takes more than 10 seconds, it sometimes fail. (I have changed line breaks below, and removed "user" and "sys" time for readability)
```
# eval $(docker-machine env aws-swarm-node-3)
#for i in {1..10}; do time docker-machine active; done
No active host found
real	0m10.127s

aws-swarm-node-3
real	0m5.710s

aws-swarm-node-3
real	0m5.749s

aws-swarm-node-3
real	0m5.802s

aws-swarm-node-3
real	0m5.564s

No active host found
real	0m10.125s

aws-swarm-node-3
real	0m6.753s

aws-swarm-node-3
real	0m10.134s

aws-swarm-node-3
real	0m5.773s

aws-swarm-node-3
real	0m10.139s
```

patrikerdes:
After this fix, the logic for determining the outcome of `docker-machine active` will be:
1. If any active docker host or swarm is found, return it as active
2. If no active host is found, and any host is in state `state.Timeout`, then the error is `errActiveTimeout`: "Error getting active host: timeout"
3. If no active host is found, and no host is in state `state.Timeout`, then the error is `errNoActiveHost`: "No active host found"

I have implemented step 2, and I have added one unit test and modified old unit tests to check 1-3 above.

Signed-off-by: Patrik Erdes <patrik@erdes.se>

### DIFF
--- a/commands/active.go
+++ b/commands/active.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/persist"
+	"github.com/docker/machine/libmachine/state"
 )
 
 const (
@@ -15,7 +16,8 @@ const (
 )
 
 var (
-	errNoActiveHost = errors.New("No active host found")
+	errNoActiveHost  = errors.New("No active host found")
+	errActiveTimeout = errors.New("Error getting active host: timeout")
 )
 
 func cmdActive(c CommandLine, api libmachine.API) error {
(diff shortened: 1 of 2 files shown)


dgageot:
LGTM

nathanleclaire:
LGTM

PR #3088, opened 2016-02-19
Add OpenNebula Driver Plugin

jmelis:
We have developed a functional OpenNebula Plugin for docker-machine. We would like it if you could accept this PR in order to include it in the list of plugins.

Thanks!


dgageot:
LGTM

jeanlaurent:
LGTM

PR #3097, opened 2016-02-22
FIX #1858 Add engine port

### LINKED ISSUE #1858: Specifying another port than 2376 when machine checks docker daemon presence at creation time
I am looking for a way to specify another port than `2376` when docker machine waits for the presence of the docker daemon on the remote host, after it has installed the docker daemon and arrange for it to start.

My use case is that I am trying to support remote machines that are behind a NAT. I have setup a persistent [reverse ssh tunnel](https://raymii.org/s/tutorials/Autossh_persistent_tunnels.html) to be able to access them and am trying to use the generic driver from [machinery](https://github.com/efrecon/machinery). Using the tunnel I can provide `machine` with options like `--generic-ip-address 127.0.0.1 --generic-ssh-port 10000` (note the use of the `localhost` IP address and, thus, a local port, tunneled to the remote ssh port). I could setup a second tunnel to access port `2376` on the remote as locally on the server running machine/machinery, but this requires me to be able to pinpoint something else that `2376` and hand this port to docker machine. Any ideas?

tpires:
Added flag `--generic-engine-port` to `generic` driver in order to specify other port than `2376` on
Docker engine.

Signed-off-by: Tiago Pires <tandrepires@gmail.com>

### DIFF
--- a/docs/drivers/generic.md
+++ b/docs/drivers/generic.md
@@ -62,6 +62,7 @@ as a sudoer with `NOPASSWD`. See https://help.ubuntu.com/community/Sudoers.
 
 ### Options
 
+-   `--generic-engine-port`: Port to use for Docker Daemon (Note: This flag will not work with boot2docker).
 -   `--generic-ip-address`: **required** IP Address of host.
 -   `--generic-ssh-key`: Path to the SSH user private key.
 -   `--generic-ssh-user`: SSH username used to connect.
@@ -73,6 +74,7 @@ Environment variables and default values:
 
 | CLI option                 | Environment variable | Default                   |
 | -------------------------- | -------------------- | ------------------------- |
+| `--generic-engine-port`    | `GENERIC_ENGINE_PORT`| `2376`                    |
 | **`--generic-ip-address`** | `GENERIC_IP_ADDRESS` | -                         |
 | `--generic-ssh-key`        | `GENERIC_SSH_KEY`    | _(defers to `ssh-agent`)_ |
 | `--generic-ssh-user`       | `GENERIC_SSH_USER`   | `root`                    |
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/engine"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/mcnflag"
 	"github.com/docker/machine/libmachine/mcnutils"
@@ -17,7 +18,8 @@ import (
 
 type Driver struct {
 	*drivers.BaseDriver
-	SSHKey string
+	EnginePort int
+	SSHKey     string
 }
 
 const (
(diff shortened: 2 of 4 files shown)


nathanleclaire:
Won't the provisioner code still try to use `:2376` for the engine port?

tpires:
No, because dockerPort will be overwritten by the driver.GetURL(). In provision/utils.go the dockerPort value is replaced if the driver.GetURL has ":" with a valid port.

nathanleclaire:
There's assumptions like this that hardcode the port though: https://github.com/docker/machine/blob/15b5a1d10df275a0d1acd179ee1f2c0985ea5a97/libmachine/host/host.go#L113

Unfortunately I think the scope of the change needs to be a bit bigger (eliminating hardcoding of the engine port in all provisioning code) to make it really viable :(

tpires:
Ok :+1:

So assumptions like https://github.com/docker/machine/blob/15b5a1d10df275a0d1acd179ee1f2c0985ea5a97/libmachine/provision/configure_swarm.go#L38 and https://github.com/docker/machine/blob/15b5a1d10df275a0d1acd179ee1f2c0985ea5a97/libmachine/provision/configure_swarm.go#L41 are also making it harder to make it viable?

nathanleclaire:
@tpires Yes -- I would really love to enable this feature but those hardcoded assumptions need to be worked around before it's really viable.

tpires:
@nathanleclaire ok, going to have a better look to avoid those hardcoded assumptions. I'll ping you after rebase.

nathanleclaire:
OK thanks!

tpires:
@nathanleclaire code rebased, I've updated `libmachine/provision/configure_swarm.go` to enable support for `--generic-engine-port` with `--swarm` and `--swarm-master`.

I've searched through the code and it seems that `libmachine/host/host.go` (hardcoded `2376`) doesn't interfere with `generic` driver specific engine port (but I'm probably wrong about it).

I've tested both `generic` and `virtualbox` driver with and without `generic-engine-port` and it's :+1:.

nathanleclaire:
Oh, very cool.  I'll take a look soon.

nathanleclaire:
`netstat` check code needs also to be updated.  Otherwise Machine will never succeed in checking for listening daemon.

nathanleclaire:
Hm, code seems like it should work for this but I'm getting netstat showing Docker listening only on 2376 on Boot2Docker.  I'll investigate

nathanleclaire:
Blergh, https://github.com/boot2docker/boot2docker/blob/057e76a8cd6755415b42dea7b4b9185c5aae7ae4/rootfs/rootfs/usr/local/etc/init.d/docker#L70 over-writes `DOCKER_HOST`.  Maybe we can just open a new issue for that and treat it as a known issue, since I don't think using `generic` with boot2docker will be a very common use case.

tpires:
@nathanleclaire I agree, using `generic` driver with boot2docker isn't a common use case. But my main focus on this PR was to be able to install and access docker on remote machines (that aren't on a supported cloud provider) even when you're behind a strong firewall (i.e. only outgoing connections on 80,443,8080 and 4443).

nathanleclaire:
@tpires Can I get you to please add a note to the documentation that this feature will not work with boot2docker? After that, this PR LGTM

tpires:
@nathanleclaire added the note to the documentation and rebased.

nathanleclaire:
Thanks!

LGTM

@dgageot Please take a look and merge if it seems OK to you?

dgageot:
LGTM

PR #3110, opened 2016-02-24
Update Godeps.

jsirois:
This was a mechanical change:
```
make dep-restore
godep update -goversion
make dep-save
```

Signed-off-by: John Sirois <john.sirois@gmail.com>

Issue #3109, opened 2016-02-24
 Support multiple `--amazonec2-security-group`s in `create`


jsirois:
Currently you can only specify one security group by name.  In the process of beta testing DTR/UCP I ran across the need to specify 2 pre-existing security groups when spinning up extra UCP controllers to form an HA cluster.

jsirois:
Noting that a team-mate and I have already worked up a branch with the fix and tests, I'm hoping to get this issue self-assigned and driver/ec2 labelled.  I think I'll break the fix up into a PR that just brings Godeps up2date (large diff), then the follow-on that adds the feature and tests.

jsirois:
This is now in master.

PR #3111, opened 2016-02-24
Support multiple `--amazonec2-security-group`s.

jsirois:
This change also bumps to latest stretchr/testify, in the course of
adding a dep on stretchr/testify/mock to help test the modified
`configureSecurityGroups` code.

The unused `deleteSecurityGroup` is dropped as well.

Signed-off-by: John Sirois <john.sirois@gmail.com>
Signed-off-by: Bill Farner <terasurfer@gmail.com>


jsirois:
NB: This is part 2 of 2 for #3109.

NB: This is intended for merge after #3110 which forms the 1st commit  @ 5896ecf.  Not sure how you like this to be handled.  I can re-base this PR against master if/when #3110 lands to reduce the PRs commit set to just the contents of the 2nd commit 9979e85.

jsirois:
This PR was run through `make clean build test validate` successfully but I had neglected to run the integration tests.  For the record, I just ran these and they went green:
```
DRIVER=amazonec2 \
AWS_ACCESS_KEY_ID=*** \
AWS_SECRET_ACCESS_KEY=*** \
AWS_VPC_ID=*** \
AWS_SSH_USER=ubuntu \
AWS_SSH_KEYPATH=$HOME/.ssh/*** \
AWS_INSTANCE_TYPE=t2.micro \
AWS_DEFAULT_REGION=us-west-2 \
AWS_ZONE=a AWS_SECURITY_GROUP=*** \
make test-integration test/integration/amazonec2/
  test/integration/run-bats.sh test/integration/amazonec2/
=> test/integration/amazonec2/createwithkeypair.bats
 ✓ amazonec2: Should Create Instance with Pre existing SSH Key

1 test, 0 failures

=> test/integration/amazonec2/create-ebsinstance.bats
 ✓ amazonec2: Should Create an EBS Optimized Instance
 ✓ amazonec2: Check the machine is up

2 tests, 0 failures

=> test/integration/amazonec2/amazon.bats
 ✓ amazonec2: Should Create a default host

1 test, 0 failures

make: Nothing to be done for 'test/integration/amazonec2/'.
```

nathanleclaire:
Needs a rebase -- thanks

drivers/amazonec2/stub_test.go
```diff
@@ -73,7 +77,9 @@ type fakeEC2WithLogin struct {
 	*fakeEC2
 }
 
-func (f *fakeEC2WithLogin) DescribeAccountAttributes(input *ec2.DescribeAccountAttributesInput) (*ec2.DescribeAccountAttributesOutput, error) {
+func (f *fakeEC2WithLogin) DescribeAccountAttributes(
```
nathanleclaire: `gofmt` allows this?  Seems a bizarre wrap.  I'd rather have it all on one line even if lengthy

jsirois: Yes, `gofmt` does allow this, ie: this is a no-diff `gofmt -s -w drivers/` - that said - 2 go noobs here who are used to wrapping at 80 or 100 cols by long convention in other langs.

Happy to undo this.

jsirois: Undone.  The 2 `DescribeAccountAttributes` methods are now back to the formatting on master.

jsirois:
@nathanleclaire PTAL - rebased, formatting changes to untouched code reverted, and CI green.

dgageot:
@jsirois Have you tried creating an machine with 0.6.0 and then use it with this PR merged? Changing the `Driver` structure which is stored in json format can have random effects.

jsirois:
Aha - thanks for pointing this out @dgageot.  I see the [host migration code](https://github.com/docker/machine/blob/ff54bf1b0c684c81e838db9d22c4b20a91428ca3/libmachine/persist/filestore.go#L112) now, although none of the migration funcs (V0 -> V1, ...etc) seem to make any affordance for migration of the `Host.Driver` field.  So, IIUC this means the 2 options are:
+ Add 2 new slice fields in amazonec2.Driver instead of turning the 2 existing fields into slices and create accessor functions to present a slice view over the 2 fields.  For example:
```go
type Driver struct {
        ...
       
        // NB: SecurityGroupId expanded from single value to slice on 26 Feb 2016 - we maintain both
        // for host storage backwards compatibility.
        SecurityGroupId string
        SecurityGroupIds []string

        ...
}

func (d *Driver) GetSecurityGroupIds() (ids []string) {
        if d.SecurityGroupId != "" {
                // Old singular format
                ids = append(ids, d.SecurityGroupId)
        } else {
                ids = d.SecurityGroupIds
        }
        return
}
```
+ Make some affordance in the host migration for `Host.Driver` migration and implement this in the `amazonec2.Driver`.

I'll noodle on this a bit and dive in on a proposed solution.

nathanleclaire:
Good call out @dgageot -- yes, we can't simply change the type of the field or it will break existing hosts.

Messing around with migration tends to be fairly erratic, so if we can work around without having to do that it would be preferable.

The simplest (if not elegant) solution seems to be adding a new field to the driver.

jsirois:
K - I implemented the simple solution 1st before moving on to try introducing a `Metadata.DriverVersion` field and migration machinery for `Driver` structs.  I'll halt work on the more comprehensive solution and post the added fields diff for you all to vet.

jsirois:
PTAL - all good now wrt old configs stored with singular `SecurityGroupName` field.
I'm happy to take a spin back down the larger `Metadata.DriverVersion` support path as well if just to learn the codebase and golang a bit better.

dgageot:
@jsirois sorry I didn't have time to tal today. Ping @nathanleclaire

dgageot:
@jsirois can you squash your commits to make it easier to review? Thanks a lot!

jsirois:
@dgageot - now re-based against upstream and squashed into 1 commit.

dgageot:
LGTM. Could use a second reader @nathanleclaire

nathanleclaire:
Will take a look thanks!

drivers/amazonec2/stub_test.go
```diff
@@ -89,6 +91,35 @@ func (f *fakeEC2WithLogin) DescribeAccountAttributes(input *ec2.DescribeAccountA
...
+func (f *fakeEC2SecurityGroupTestRecorder) CreateSecurityGroup(
+	input *ec2.CreateSecurityGroupInput) (*ec2.CreateSecurityGroupOutput, error) {
+
+	result := f.Called(input)
+	value, _ := result.Get(0).(*ec2.CreateSecurityGroupOutput)
+	return value, result.Error(1)
+}
+
+func (f *fakeEC2SecurityGroupTestRecorder) AuthorizeSecurityGroupIngress(
+	input *ec2.AuthorizeSecurityGroupIngressInput) (*ec2.AuthorizeSecurityGroupIngressOutput, error) {
```
nathanleclaire: As discussed before, my preference would be to have this method definition on one line

drivers/amazonec2/stub_test.go
```diff
@@ -89,6 +91,35 @@ func (f *fakeEC2WithLogin) DescribeAccountAttributes(input *ec2.DescribeAccountA
...
+
+	result := f.Called(input)
+	value, _ := result.Get(0).(*ec2.AuthorizeSecurityGroupIngressOutput)
```
nathanleclaire: Why swallow the error instead of just returning it?

nathanleclaire:
Just a couple of nitpicks but other than that it looks good.

nathanleclaire:
I'll merge and fix the nitpicks myself in a follow up PR

drivers/amazonec2/stub_test.go
jsirois: Fair enough.  My preference would be to use gomock - I think - to avoid the boilerplate, but that seemed an inappropriate change to mix in with this fix.  If that seems like a reasonable path to you though, I'd be happy to retrofit both it and `go generate` into the Makefile setup / CONTRIBUTING docs.

PR #3112, opened 2016-02-24
Determine host-only interface dynamically, stop assuming eth1

### LINKED ISSUE #3108: Machine can grab wrong IP for host-only network in VirtualBox
We run Machine with VirtualBox to support our local development environments. Recently, we've been seeing a lot of networking failures with Machine 0.6.0 and VirtualBox 5.14. Basically, operations like `docker-machine regenerate-certs` will periodically be unable to connect to the VM and will time out. This problem has sometimes been solved by VM restarts, though it does not always fix itself on the first restart.

I believe I've tracked the bug down to this line in the VirtualBox driver:

https://github.com/docker/machine/blob/c40f9f7a215dd2cbc1bc31b0ace861a0e22b82f4/drivers/virtualbox/virtualbox.go#L666

This line is making an assumption that the host-only network will always be mounted as `eth1` inside the VM. It appears that this assumption does not (edit: always) hold. The NAT and host-only networks seem to get mounted semi-randomly as `eth0` and `eth1` inside our VMs. When the NAT network gets mounted on `eth1` instead of the host-only network, we see `docker-machine ip` returning the wrong IP (the 10.x NAT IP instead of the 192.x host-only IP) and commands reliant on that IP, such as `docker-machine regenerate-certs`, experience timeouts and failures. Restarting seems to sometimes re-assign the interfaces and allow subsequent commands to succeed.

I'm trying to get up to speed on how these interfaces get assigned and hopefully contribute a fix. I wanted to post this issue here in case anyone else is having this problem or could shed some light on the assumptions being made.

thieman:
Fixes #3108.

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -652,6 +652,67 @@ func (d *Driver) GetState() (state.State, error) {
 	return state.None, nil
 }
 
+func (d *Driver) getHostOnlyMACAddress() (string, error) {
+	// Return the MAC address of the host-only adapter
+	// assigned to this machine. The returned address
+	// is lower-cased and does not contain colons.
+
+	stdout, stderr, err := d.vbmOutErr("showvminfo", d.MachineName, "--machinereadable")
+	if err != nil {
+		if reMachineNotFound.FindString(stderr) != "" {
+			return "", ErrMachineNotExist
+		}
+		return "", err
+	}
+
+	// First, we get the number of the host-only interface
+	re := regexp.MustCompile(`(?m)^hostonlyadapter([\d]+)`)
+	groups := re.FindStringSubmatch(stdout)
+	if len(groups) < 2 {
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Cool!  Want to take a whack at some unit tests?

thieman:
Sure, I'll see if I can base anything off what's already there.

thieman:
@nathanleclaire Added unit tests for the two new methods I added, thanks!

nathanleclaire:
Awesome!  Will take this for a spin today.

nathanleclaire:
LGTM

nathanleclaire:
Considering doing a patch release (Machine 0.6.1) for this.  @dgageot @jeanlaurent WDYT

dgageot:
@nathanleclaire ok

nathanleclaire:
I'll do some integration testing etc. and think about cutting a patch release next week.

Issue #2876, opened 2016-01-19
google: --google-use-internal-ip creates external IP too


ahmetalpbalkan:
Hi, I'm trying to understand the [`--google-use-internal-ip`](https://docs.docker.com/machine/drivers/gce/) argument. Is it created for using private IP instead of public IP in machine OR not having an external IP for the machine altogether?

I assumed this would not create a public IP and make the machine secure within the private network but it doesn't work that way. So I'm trying to understand the motive here.

dgageot:
Hi @ahmetalpbalkan, if I recall correctly, VMs on GCE have an internal IP and have an optional ephemeral or reserved public IP. 
The ` --google-use-internal-ip` flag configures the VM to use its private IP only.

The motive here is to create private VMs that are not reachable from the Internet.

Did you encounter a different behaviour?

ahmetalpbalkan:
Yes I encountered a different behavior. External IP gets created for the instance from what I tell from the management console.

daehyeok:
I'm quick look the source code.  ```--google-use-internal-ip``` option not affect when instance created. I think it's easy to patch to create inter only machine. but how to progress provision after create machine without reachable ip when it executed in the external network(internet)?

ahmetalpbalkan:
I think --use-internal-ip is only usable when you're inside a Google Cloud
instance and in that network you are trying to create the new machine. That
is the only way provisioning would proceed.
On Wed, Jan 20, 2016 at 1:39 PM Daehyeok Mun <notifications@github.com>
wrote:

> I'm quick look the source code. --google-use-internal-ip option not
> affect when instance created. I think it's easy to patch to create inter
> only machine. but how to progress provision after create machine without
> reachable ip when it executed in the external network(internet)?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/machine/issues/2876#issuecomment-173368190>.
>
-- 
—Ahmet Alp Balkan

dgageot:
@ahmetalpbalkan yes, that's it. When you create Docker Machine hosts from inside Google's network, maybe you to reach them through their internal IP but still have an external IP to expose services.

daehyeok:
Okay. If nobody interesting implement or fix code. I will try.

ahmetalpbalkan:
@daehyeok I don't think there is anything to fix here.

[...]

PR #3133, opened 2016-03-01
Google driver: add `--google-use-internal-ip-only` flag

otherjason:
This addresses previously-closed issue #2876, which points out that instances created with the `--google-use-internal-ip` command-line flag are still assigned an external IP address. The new flag (which implies the presence of `--google-use-internal-ip` if it isn't specified) will cause the new instance to have no externally-accessible IP address.


nathanleclaire:
@dgageot @jeffmendoza What do you think

dgageot:
+1 @otherjason please update the markdown doc also

otherjason:
@dgageot How does that look?

jeffmendoza:
Looks good, not sure about the UX

otherjason:
@jeffmendoza For what it's worth, this is essentially identical to a feature already present in the AWS driver.

jeffmendoza:
@otherjason consistency is good =)

dgageot:
LGTM. ping @nathanleclaire

nathanleclaire:
LGTM!  Thanks all

PR #3134, opened 2016-03-01
fix: --engine-env was not effective when provisioning on CoreOS

otherjason:
The `--engine-env` command-line arguments used to set environment variables in the Docker engine are not effective when provisioning hosts on CoreOS. The provisioning code generates a systemd script that invokes the daemon; it appears that the line needed to insert the environment variables was omitted by accident. This change (copied from the Red Hat provisioner) inserts the environment variables properly.


nathanleclaire:
LGTM

nathanleclaire:
Thanks!

PR #3140, opened 2016-03-02
Updated VMWare Fusion driver to attempt to bind mount the share directory, to fix #2050.

### LINKED ISSUE #2050: Symlinks corrupt in shared folders
Symlinks that are in (nested) shared folder, get corrupt with `vmwarefusion` driver, e.g.:

Local symlink:
```
lrwxr-xr-x   1 ain  staff    20 Jul  3 20:46 recaptcha.yml -> recaptcha.yml.docker
```

Same symlink in container:
```
lrwxr-xr-x 1 501 dialout   20 Jul  3 18:46 recaptcha.yml -> cha.yml.docker
```

Platform info:
```
Mac OS X 10.11
VMWare Fusion Version 8.0.1 (3094680)

$ docker version
Client:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 18:01:15 UTC 2015
 OS/Arch:      darwin/amd64

Server:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 18:01:15 UTC 2015
 OS/Arch:      linux/amd64

$ docker-machine --version
docker-machine version 0.5.0-rc2 (6cac8cc)

$ docker-compose --version
docker-compose version: 1.4.2
```

SeerUK:
Issue link: #2050 

Using VMWare Fusion and a project that has symlinks was not entirely possible without strange workarounds that may not be either reliable, or cross-platform friendly (from say, Compose; i.e. using `/mnt/hgfs` in the Compose configuration which would tie the configuration to OS X).

The issue was that symlink's paths would be truncated, so if you had a relative path like say `../foo` it would end up pointing at `oo` in the machine. One other workaround I found without this modification was to prefix the path with some `./` padding, i.e. `././././././../foo` would work, but is less than ideal. It would just truncate some of that padding, and still work.

Apologies if I've missed anything in the contribution docs, I'm not actually a Go dev, but tried to make sure I did everything!

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -376,7 +376,7 @@ func (d *Driver) Create() error {
 		} else if !os.IsNotExist(err) {
 			// add shared folder, create mountpoint and mount it.
 			vmrun("-gu", B2DUser, "-gp", B2DPass, "addSharedFolder", d.vmxPath(), shareName, shareDir)
-			command := "[ ! -d " + shareDir + " ]&& sudo mkdir " + shareDir + "; [ -f /usr/local/bin/vmhgfs-fuse ]&& sudo /usr/local/bin/vmhgfs-fuse -o allow_other .host:/" + shareName + " " + shareDir + " || sudo mount -t vmhgfs -o uid=$(id -u),gid=$(id -g) .host:/" + shareName + " " + shareDir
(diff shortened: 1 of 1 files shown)


dgageot:
ping @frapposelli

frapposelli:
Reasonable workaround, LGTM

nathanleclaire:
LGTM

PR #3144, opened 2016-03-03
[DigitalOcean] Allow to use an existing SSH key (closes #1445)

### LINKED ISSUE #1445: Digital Ocean Driver creates a new ssh key everytime.
Currently there is no way to tell the digital ocean driver use a preexisting ssh key. This leads to a lot of cruft with new keys for every instance launch

bamarni:
I also suggest to use the key fingerprint for this feature, it is more meaningful than digital ocean's internal id. It can then be used to avoid deleting the key when removing the machine, as it's most likely undesirable.


dgageot:
LGTM

nathanleclaire:
LGTM

PR #3147, opened 2016-03-04
Fix amazonec2 test nitpicks

nathanleclaire:
FYI @jsirois 

### DIFF
--- a/drivers/amazonec2/stub_test.go
+++ b/drivers/amazonec2/stub_test.go
@@ -96,28 +96,34 @@ type fakeEC2SecurityGroupTestRecorder struct {
 	mock.Mock
 }
 
-func (f *fakeEC2SecurityGroupTestRecorder) DescribeSecurityGroups(
-	input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
-
+func (f *fakeEC2SecurityGroupTestRecorder) DescribeSecurityGroups(input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
 	result := f.Called(input)
-	value, _ := result.Get(0).(*ec2.DescribeSecurityGroupsOutput)
-	return value, result.Error(1)
+	err := result.Error(1)
+	value, ok := result.Get(0).(*ec2.DescribeSecurityGroupsOutput)
+	if !ok && err == nil {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Hm, looks like checking the type assertion is leading to an unforeseen issue with the tests.  Will look into it.

drivers/amazonec2/stub_test.go
```diff
@@ -96,27 +96,30 @@ type fakeEC2SecurityGroupTestRecorder struct {
 	mock.Mock
 }
 
-func (f *fakeEC2SecurityGroupTestRecorder) DescribeSecurityGroups(
-	input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
-
+func (f *fakeEC2SecurityGroupTestRecorder) DescribeSecurityGroups(input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
 	result := f.Called(input)
-	value, _ := result.Get(0).(*ec2.DescribeSecurityGroupsOutput)
+	value, ok := result.Get(0).(*ec2.DescribeSecurityGroupsOutput)
```
jsirois: The issue is that this type assertion need only be `ok` when err is `nil`.  When err is non-`nil`, the value is expected to be `nil` and thus fails the type assertion.

In general you'd need something like this it seems to me if you want to have a controlled err instead of a panic:
```go
func (f *fakeEC2SecurityGroupTestRecorder) DescribeSecurityGroups(input *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
	result := f.Called(input)
	value, ok := result.Get(0).(*ec2.DescribeSecurityGroupsOutput)
	err := result.Error(1)
	if err == nil && !ok {
		err = errors.New("Type assertion to DescribeSecurityGroupsOutput failed")
	}
	return value, err
}
```

That said, do you think the type assertion check is worth it?  Either way you get a test failure and the type assertion check here is more a sanity check the mock library is being used correctly than a test of the SUT.  As such I'll argue the extra checking here obscures the intent of the test.  I am arguing from a go noob perspective though!

nathanleclaire: Probably best being ignored for now then.  It's my instinct to never swallow `err` or `ok` but since it's for a test I think it's probably OK.

PR #3157, opened 2016-03-07
Remove azure driver temporarily + Godeps for new azure driver

ahmetalpbalkan:
This commit temporarily removes Azure driver and its dependencies from the source tree and adds dependencies for the new Azure driver (so that Azure driver PR will not have godeps changes and will be
easier to review).

Very shortly after this is merged I will send the actual PR for the new azure driver.


nathanleclaire:
LGTM

ahmetalpbalkan:
thanks aaand here's the actual PR #3159

nathanleclaire:
OK @dgageot @jeanlaurent I went ahead and merged this but it occurs to me that there there are probably at least a few instances of existing `azure` machines in the wild.  Based on the numbers I see it's not very common with probably about 88 failed `create` errors (compare to ~3000 for the most popular cloud drivers) in the past month over ~10 different hosts.  I wonder what we should do, since @ahmetalpbalkan new driver and this PR seems definitely to be the right direction in general.  Any ideas?

friism:
Maybe Azure just works way better than the other clouds and that's why there are orders of magnitude fever errors. :trollface:

nathanleclaire:
@friism

http://3.bp.blogspot.com/-3Y4_Y-v6cug/UYqb0V__6LI/AAAAAAAALuI/NwmRVgUGtZQ/s1600/nodding.gif

ahmetalpbalkan:
@nathanleclaire haha I wouldn't be surprised if those errors were me while trying to develop this. :stuck_out_tongue: Azure is not super big on Linux just yet (we're about a 25% Linux cloud) but that's growing rapidly. I'm constantly getting emails from customers using docker-machine on Azure, it appears like there is a demand for it (that's why we did this after all).

I think the new driver is solid from Azure perspective, there might be still some flakiness that we can address over time but yesterday I created about 250 machines and all failures were due to non-Azure issues.

nathanleclaire:
@ahmetalpbalkan Yup, no doubt that it could be a very popular driver -- my concern is largely just that we shouldn't leave existing users of the `azure` driver in the lurch.  Since it doesn't seem very common, we may not need to take drastic action, e.g. a large note about breaking changes in the release notes for the next version might suffice, but we should come up with and execute a game plan on it.

ahmetalpbalkan:
@nathanleclaire Yes, a large note about the breaking change will be really useful. We can suggest sticking to older versions of machine for existing VMs or re-importing the VM to the `generic` driver perhaps.

nathanleclaire:
@ahmetalpbalkan If possible, we should add a "safety gate" to refuse to move forward with any Machine operations given the new `azure` driver in the presence of an old config, thereby protecting the user's config and allowing them to switch back to an older version of Machine until they are ready to upgrade.

Issue #3024, opened 2016-02-09
PSA: Upgrade from docker-machine 0.5.3 to 0.6.0 broke vmwarevsphere plugin for existing machines


ljrittle:
$ docker-machine version
    docker-machine version 0.5.3, build 4d39a66
    $ docker-machine ls
    NAME                    ACTIVE   DRIVER          STATE     URL                       SWARM                DOCKER    ERRORS
    arc-dev-00              *        vmwarevsphere   Running   tcp://10.26.67.187:2376                        v1.9.0    
    [...]

Upgrade to docker-machine-0.6.0.

    $ docker-machine version
    docker-machine version 0.6.0, build e27fb87
    $ NAME                    ACTIVE   DRIVER          STATE     URL                         SWARM                DOCKER    ERRORS
    arc-dev-00              -        vmwarevsphere   Error                                                      Unknown   
    [...]

Here is what I see with `status` post upgrade:

    $ docker-machine status arc-dev-00
    error getting state for host arc-dev-00: Post https://10.26.66.118:0/sdk: dial tcp 10.26.66.118:0: connect: can't assign requested address
    Error

I am looking into how to manually touch the existing e.g. `.docker/machine/machines/arc-dev-00/config.json` to fix this without manually removing and recreating the machine.  I will report the workaround if/when I find it.

nathanleclaire:
cc @frapposelli

frapposelli:
@ljrittle your problem seems to be related to the vCenter port (shown as `0` in the failed command).

@nathanleclaire was there any particular change in `v0.6.0` related to provider configuration?

nathanleclaire:
> @nathanleclaire was there any particular change in v0.6.0 related to provider configuration?

Hm, I'm not sure exactly what you mean.  There were no config migrations and the host loading => plugin mechanic has been largely unchanged since the 0.5.* series of releases.

frapposelli:
@nathanleclaire that's what I've thought, I've tried to reproduce this issue to no avail.

@ljrittle do you have repro steps for this?

nathanleclaire:
Please post a minimally reproducible example on a new ticket if this problem is still occurring.  Thanks!

PR #3158, opened 2016-03-07
Default to port 443 for all vsphere actions, not just creation

### LINKED ISSUE #3150: Machine made prior to 0.5.5 missing vSphere Port Driver config
If you try to manage (e.g. docker-machine ls) vSphere machines made before 0.5.5, with that or newer releases, you'll encounter an error like the below:

> Post https://XXXXXXX:0/sdk: dial tcp XXXXXXX:0: connectex: The requested address is not valid in its context.

The issue is the 0.5.5 govmomi code added a "Port" config to the "Driver" section of the machine's config.json, defaulting to 443 at build time, 0 upon other uses such as above.  The simplest fix is to add this numeric value to your json manually.

Alternatively, initialising Port to 443 in NewDriver (drivers/vmwarevsphere/vsphere.go) will do the same for all.  Would a PR be appropriate?

tyrken:
Fixes #3150 and probably #3024 following the new gvmomi code in 0.5.5.

Signed-off-by: Tristan Keen <tristan.keen@gmail.com>

### DIFF
--- a/drivers/vmwarevsphere/vsphere.go
+++ b/drivers/vmwarevsphere/vsphere.go
@@ -69,6 +69,7 @@ const (
 	defaultCpus     = 2
 	defaultMemory   = 2048
 	defaultDiskSize = 20480
+	defaultSDKPort  = 443
 )
 
 // GetCreateFlags registers the flags this driver adds to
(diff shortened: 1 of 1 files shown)


nathanleclaire:
cc @frapposelli

nathanleclaire:
LGTM

frapposelli:
looks great :+1:

PR #3159, opened 2016-03-07
New Microsoft Azure docker-machine driver

### LINKED ISSUE #774: Azure driver - Error creating machine: exit status 1
Hi,
I'm using:
```
docker-machine version 0.1.0
```
I run command:
```
docker-machine create -d azure --azure-subscription-id=[SUB ID] --azure-subscription-cert=mycert.pem docker02
```
but getting error:
```
←[31mERRO←[0m[0032] Error creating machine: exit status 1
```
I checked on azure portal:
New storage account was created but there's no VM.

Any idea what am I doing wrong?
I run that command also with --debug
```
←[34mINFO←[0m[0001] Creating Azure machine...
←[34mDEBU←[0m[0011] Generating certificate for Azure...
←[34mDEBU←[0m[0011] executing: C:\Git\bin\ssh-keygen.exe ssh-keygen -t rsa -N  -f C:\Users\...

Your identification has been saved in C:\Users\...
Your public key has been saved in C:\Users\...
The key fingerprint is:
[...]
The key's randomart image is:
+--[ RSA 2048]----+
[...]
←[31mERRO←[0m[0013] Error creating machine: exit status 1
←[33mWARN←[0m[0013] You will want to check the provider to make sure the machine
 and associated resources were properly removed.
←[31mFATA←[0m[0013] Error creating machine
```

### LINKED ISSUE #1126: Azure Driver - add (optional) cli flag for azure hostname
The azure driver uses the machine name as a hostname during creation. 

This leads to problems because hostnames on azure have to be globally unique, so the same applies for our machine name.

### LINKED ISSUE #1142: Azure driver storage account contention for VHDs
Right now Azure driver picks some storage account in the specified region under the picked subscription either randomly or alphabetically first one and places VHD of the VM in there –which might land the disk blob in some undesired place really...

Instead of that we could look for a storage account in region prefixed with 'docker' and use that. (similar to what Azure Portal does when VM is created via UI, creates a storage account `portal$RANDOM` and uses that for subsequent VM creations in the same region.)

Self-assigning.

### LINKED ISSUE #1368: Azure: update Go lib to current
The Azure Go lib is quite a bit behind.  When trying to update it appears to have changed drastically.  We should get this up to date.

### LINKED ISSUE #2236: Azure Driver: cannot create vm; fails with isavailable internal error (yosemite/ubuntu 14.04 LTS)
I've tried to figure this out, but I've had bad luck on two machines now, so filing the issue to make sure I'm doing something incorrectly.

On mac, Docker Tools installation fresh. d-m verison 0.5.0 (04cfa58) on both machines.
- `docker-machine create --driver virtualbox` works perfectly. 
- `docker-machine create --driver azure` fails with an internal error on both the mac and the ubuntu. 

Mac:

>
☁  .docker  docker-machine -D create -d azure --azure-subscription-id=f7f09258-xxxx-xxxx-xxxx-xxxxxxxxxxxx --azure-subscription-cert=cert.pem --azure-location="Central US" --azure-image=2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-766.5.0  yodocker
Docker Machine Version: 0.5.0 (04cfa58)
Found binary path at /usr/local/bin/docker-machine-driver-azure
Launching plugin server for driver azure
Plugin server listening at address 127.0.0.1:56600
Using API Version 1
(flag-lookup) Calling RpcServerDriver.GetCreateFlags
Making call to close connection to plugin binary
Making call to close driver server
(flag-lookup) Calling RpcServerDriver.Close
Successfully made call to close driver server
Found binary path at /usr/local/bin/docker-machine-driver-azure
Launching plugin server for driver azure
Plugin server listening at address 127.0.0.1:56605
Using API Version 1
(yodocker) Calling RpcServerDriver.GetMachineName
(yodocker) Calling RpcServerDriver.DriverName
(yodocker) Calling RpcServerDriver.GetCreateFlags
(yodocker) Calling RpcServerDriver.SetConfigFromFlags
Running pre-create checks...
(yodocker) Calling RpcServerDriver.PreCreateCheck
Error creating machine: Error with pre-create check: Get https://management.core.windows.net/f7f09258-xxxx-xxxx-xxxx-xxxxxxxxx/services/hostedservices/operations/isavailable/yodocker: local error: internal error
>

And the ubuntu is identical:

>
ralph@local:~ 
 $ docker-machine -D create \                                                                                                           1
-d azure \
--azure-subscription-id="f7f09258-xxxx-xxxx-xxxx-xxxxxxxxxxxxx" \
--azure-subscription-cert="/home/ralph/.ssh/myCert.pem" \
yodocker
Docker Machine Version: 0.5.0 (04cfa58)
Found binary path at /usr/local/bin/docker-machine-driver-azure
Launching plugin server for driver azure
Plugin server listening at address 127.0.0.1:41644
Using API Version 1
(flag-lookup) Calling RpcServerDriver.GetCreateFlags
Making call to close connection to plugin binary
Making call to close driver server
(flag-lookup) Calling RpcServerDriver.Close
Successfully made call to close driver server
Found binary path at /usr/local/bin/docker-machine-driver-azure
Launching plugin server for driver azure
Plugin server listening at address 127.0.0.1:46952
Using API Version 1
(yodocker) Calling RpcServerDriver.GetMachineName
(yodocker) Calling RpcServerDriver.DriverName
(yodocker) Calling RpcServerDriver.GetCreateFlags
(yodocker) Calling RpcServerDriver.SetConfigFromFlags
Running pre-create checks...
(yodocker) Calling RpcServerDriver.PreCreateCheck
Error creating machine: Error with pre-create check: Get https://management.core.windows.net/-xxx-xxxx-xxx-/services/hostedservices/operations/isavailable/yodocker: local error: internal error

>

### LINKED ISSUE #2408: Additional Azure VM configuration options
Hello

I believe we are missing some Azure VM configuration options that are needed.

1. Each VM is created with its own Cloud Service.  If I need multiple VMs to be able to communicate with each other using the internal IP which is assigned to the local interface then I need them to be in the same Cloud Service.  There are other scenarios where I would want multiple VMs in the same or manually pre-configured Cloud Service too.  Therefore, I would like the option of being able to specify an existing cloud service to create the VM against.  The Azure Management Portal also allows this option.

2. The option of specifying the Region/Affinity Group/Azure Virtual Network along with Virtual network Subnets to connect to the created VM.  The Azure Management Portal allows this option when creating a VM also.  If I want to do this with a VM created using Docker Machine, I have to delete the VM while keeping the VHD and recreate it manually.

3. The option of specifying additional endpoint/ports to open when creating the Virtual Machine.  For example, I might want 8500 open for Consul service discovery, especially common now using Docker 1.9 overlay networking functionality. Unless I use the internal IP which I cannot because the VMs are in separate Cloud Services.  The Azure Management Portal also allows configuration of endpoints/port forwards when creating a VM.

4. The option of specifying an Availability Set.  The Azure Management Portal also allows configuration of this during VM creation.

In principal, I need to be able to set the same configuration options for VMs as I can when using the Azure Management Portal to manually create VMs.

Thanks

James

### LINKED ISSUE #2742: Azure driver - resource manager version
I think I'm right in saying that the current support for docker-machine on Azure uses the Classic deployment model. It would be great to support the Resource Manager model too (since that's what MS is recommending for new deployments).

ahmetalpbalkan:
The new driver uses Azure Resource Manager APIs and offers a lot
more functionality compared to the old Azure driver. It is also
easier to authenticate and does not require user to create and place
certificate files. It only has a single required argument.

**This is a breaking change:** The new driver cannot work with machines
created with the older Azure driver and vice versa (as the APIs are
entirely different and resources are not shared between old/new azure
APIs).

The new driver addresses many issues about the azure driver reported
so far. Please read `azure.md` for docs about what's new.

This resolves #2742, resolves #1368, resolves #1142, resolves #2236,
resolves #2408, resolves #1126, resolves #774.

### DIFF
--- a/docs/drivers/azure.md
+++ b/docs/drivers/azure.md
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
 
 # Microsoft Azure
 
-Create machines on [Microsoft Azure](http://azure.microsoft.com/).
+You will need an Azure Subscription to use this Docker Machine driver.
+[Sign up for a free trial.][trial]
 
-You need to create a subscription with a cert. Run these commands and answer the questions:
+> **NOTE:** This documentation is for the new version of  Azure driver started
+> shipping with v0.7.0 and it is not backwards-compatible with the old Azure
+> driver. If you like to manage your existing Azure machines, please download
+> and use machine versions prior to v0.7.0.
 
-    $ openssl req -x509 -nodes -days 365 -newkey rsa:1024 -keyout mycert.pem -out mycert.pem
-    $ openssl pkcs12 -export -out mycert.pfx -in mycert.pem -name "My Certificate"
-    $ openssl x509 -inform pem -in mycert.pem -outform der -out mycert.cer
+[azure]: http://azure.microsoft.com/
+[trial]: https://azure.microsoft.com/free/
 
-Go to the Azure portal, go to the "Settings" page (you can find the link at the bottom of the
-left sidebar - you need to scroll), then "Management Certificates" and upload `mycert.cer`.
+## Authentication
 
-Grab your subscription ID from the portal, then run `docker-machine create` with these details:
+First time you try to create a machine, Azure driver will ask you to
+authenticate:
 
-    $ docker-machine create -d azure --azure-subscription-id="SUB_ID" --azure-subscription-cert="mycert.pem" A-VERY-UNIQUE-NAME
+    $ docker-machine create --driver azure <machine-name>
+    Running pre-create checks...
+    Microsoft Azure: To sign in, use a web browser to open the page https://aka.ms/devicelogin.
+    Enter the code [...] to authenticate.
 
-The Azure driver uses the `b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB`
-image by default. Note, this image is not available in the Chinese regions. In China you should
- specify `b549f4301d0b4295b8e76ceb65df47d4__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB`.
+After authenticating, the driver will remember your credentials up to two weeks.
 
-You may need to `machine ssh` in to the virtual machine and reboot to ensure that the OS is updated.
+## Options
 
-Options:
+Azure driver only has a single required argument to make things easier. Please
+read the optional flags to configure machine details and placement further.
 
--   `--azure-docker-port`: Port for Docker daemon.
--   `--azure-image`: Azure image name. See [How to: Get the Windows Azure Image Name](https://msdn.microsoft.com/en-us/library/dn135249%28v=nav.70%29.aspx)
--   `--azure-location`: Machine instance location.
--   `--azure-password`: Your Azure password.
--   `--azure-publish-settings-file`: Azure setting file. See [How to: Download and Import Publish Settings and Subscription Information](https://msdn.microsoft.com/en-us/library/dn385850%28v=nav.70%29.aspx)
--   `--azure-size`: Azure disk size.
--   `--azure-ssh-port`: Azure SSH port.
--   `--azure-subscription-id`: **required** Your Azure subscription ID (A GUID like `d255d8d7-5af0-4f5c-8a3e-1545044b861e`).
--   `--azure-subscription-cert`: **required** Your Azure subscription cert.
--   `--azure-username`: Azure login user name.
+Required:
+
+- `--azure-subscription-id`: **(required)** Your Azure Subscription ID.
+
+Optional:
+
+- `--azure-image`: Azure virtual machine image. [[?][vm-image]]
+- `--azure-location`: Azure region to create the virtual machine. [[?][location]]
+- `--azure-resource-group`: Azure Resource Group name to create the resources in.
+- `--azure-size`: Size for Azure Virtual Machine. [[?][vm-size]]
+- `--azure-ssh-user`: Username for SSH login.
+- `--azure-vnet`: Azure Virtual Network name to connect the virtual machine. [[?][vnet]]
+- `--azure-subnet`: Azure Subnet Name to be used within the Virtual Network.
+- `--azure-subnet-prefix`: Private CIDR block to be used for the new subnet.
+- `--azure-availability-set`: Azure Availability Set to place the virtual machine into. [[?][av-set]]
+- `--azure-open-port`: Make additional port number(s) accessible from the Internet [[?][nsg]]
+- `--azure-private-ip-address`: Specify a static private IP address for the machine.
+- `--azure-use-private-ip`: Use private IP address of the machine to connect.
+- `--azure-no-public-ip`: Do not create a public IP address for the machine.
+- `--azure-docker-port`: Port number for Docker engine [$AZURE_DOCKER_PORT]
+- `--azure-environment`: Azure environment (e.g. `AzurePublicCloud`, `AzureChinaCloud`).
+
+[vm-image]: https://azure.microsoft.com/en-us/documentation/articles/resource-groups-vm-searching/
+[location]: https://azure.microsoft.com/en-us/regions/
+[vm-size]:  https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-size-specs/
+[vnet]:     https://azure.microsoft.com/en-us/documentation/articles/virtual-networks-overview/
+[av-set]:   https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-manage-availability/
 
 Environment variables and default values:
 
 | CLI option                      | Environment variable          | Default            |
 | ------------------------------- | ----------------------------- | ------------------ |
-| `--azure-docker-port`           | -                             | `2376`             |
-| `--azure-image`                 | `AZURE_IMAGE`                 | _Ubuntu 15.10 x64_ |
-| `--azure-location`              | `AZURE_LOCATION`              | `West US`          |
-| `--azure-password`              | -                             | -                  |
-| `--azure-publish-settings-file` | `AZURE_PUBLISH_SETTINGS_FILE` | -                  |
-| `--azure-size`                  | `AZURE_SIZE`                  | `Small`            |
-| `--azure-ssh-port`              | -                             | `22`               |
-| **`--azure-subscription-cert`** | `AZURE_SUBSCRIPTION_CERT`     | -                  |
 | **`--azure-subscription-id`**   | `AZURE_SUBSCRIPTION_ID`       | -                  |
-| `--azure-username`              | -                             | `ubuntu`           |
+| `--azure-environment`           | `AZURE_ENVIRONMENT`           | `AzurePublicCloud` |
+| `--azure-image`                 | `AZURE_IMAGE`                 | `canonical:UbuntuServer:14.04.3-LTS:latest` |
+| `--azure-location`              | `AZURE_LOCATION`              | `westus`           |
+| `--azure-resource-group`        | `AZURE_RESOURCE_GROUP`        | `docker-machine`   |
+| `--azure-size`                  | `AZURE_SIZE`                  | `Standard_A2`      |
+| `--azure-ssh-user`              | `AZURE_SSH_USER`              | `ubuntu`           |
+| `--azure-vnet`                  | `AZURE_VNET`                  | `docker-machine`   |
+| `--azure-subnet`                | `AZURE_SUBNET`                | `docker-machine`   |
+| `--azure-subnet-prefix`         | `AZURE_SUBNET_PREFIX`         | `192.168.0.0/16`   |
+| `--azure-availability-set`      | `AZURE_AVAILABILITY_SET`      | `docker-machine`   |
+| `--azure-open-port`             | -                             | -                  |
+| `--azure-private-ip-address`    | -                             | -                  |
+| `--azure-use-private-ip`        | -                             | -                  |
+| `--azure-no-public-ip`          | -                             | -                  |
+| `--azure-docker-port`           | `AZURE_DOCKER_PORT`           | `2376`             |
+
+## Notes
+
+Azure runs fully on the new [Azure Resource Manager (ARM)][arm] stack. Each
+machine created comes with a few more Azure resources associated with it:
+
+* A [Virtual Network][vnet] and a subnet under it is created to place your
+machines into. This establishes a local network between your docker machines.
+* An [Availability Set][av-set] is created to maximize availability of your
+machines.
+
+These are created once when the first machine is created and reused afterwards.
+Although they are free resources, driver does a best effort to clean them up
+after the last machine using these resources is removed.
+
+Each machine is created with a public dynamic IP address for external
+connectivity. All its ports (except Docker and SSH) are closed by default. You
+can use `--azure-open-port` argument to specify multiple port numbers to be
+accessible from Internet. 
+
+Once the machine is created, you can modify [Network Security Group][nsg]
+rules and open ports of the machine from the [Azure Portal][portal].
+
+[arm]:    https://azure.microsoft.com/en-us/documentation/articles/resource-group-overview/
+[nsg]:    https://azure.microsoft.com/en-us/documentation/articles/virtual-networks-nsg/
+[portal]: https://portal.azure.com/
--- a/docs/drivers/azure.md
+++ b/docs/drivers/azure.md
@@ -72,11 +72,11 @@ Environment variables and default values:
 | ------------------------------- | ----------------------------- | ------------------ |
 | **`--azure-subscription-id`**   | `AZURE_SUBSCRIPTION_ID`       | -                  |
 | `--azure-environment`           | `AZURE_ENVIRONMENT`           | `AzurePublicCloud` |
-| `--azure-image`                 | `AZURE_IMAGE`                 | `canonical:UbuntuServer:14.04.3-LTS:latest` |
+| `--azure-image`                 | `AZURE_IMAGE`                 | `canonical:UbuntuServer:15.10:latest` |
 | `--azure-location`              | `AZURE_LOCATION`              | `westus`           |
 | `--azure-resource-group`        | `AZURE_RESOURCE_GROUP`        | `docker-machine`   |
 | `--azure-size`                  | `AZURE_SIZE`                  | `Standard_A2`      |
-| `--azure-ssh-user`              | `AZURE_SSH_USER`              | `ubuntu`           |
+| `--azure-ssh-user`              | `AZURE_SSH_USER`              | `docker-user`      |
 | `--azure-vnet`                  | `AZURE_VNET`                  | `docker-machine`   |
 | `--azure-subnet`                | `AZURE_SUBNET`                | `docker-machine`   |
 | `--azure-subnet-prefix`         | `AZURE_SUBNET_PREFIX`         | `192.168.0.0/16`   |
--- a/docs/drivers/azure.md
+++ b/docs/drivers/azure.md
@@ -13,17 +13,18 @@ parent="smn_machine_drivers"
 You will need an Azure Subscription to use this Docker Machine driver.
 [Sign up for a free trial.][trial]
 
-> **NOTE:** This documentation is for the new version of  Azure driver started
-> shipping with v0.7.0 and it is not backwards-compatible with the old Azure
-> driver. If you like to manage your existing Azure machines, please download
-> and use machine versions prior to v0.7.0.
+
+> **NOTE:** This documentation is for the new version of the Azure driver, which started
+> shipping with v0.7.0. This driver is not backwards-compatible with the old
+> Azure driver. If you want to continue managing your existing Azure machines, please
+> download and use machine versions prior to v0.7.0.
 
 [azure]: http://azure.microsoft.com/
 [trial]: https://azure.microsoft.com/free/
 
 ## Authentication
 
-First time you try to create a machine, Azure driver will ask you to
+The first time you try to create a machine, Azure driver will ask you to
 authenticate:
 
     $ docker-machine create --driver azure --azure-subscription-id <subs-id> <machine-name>
--- a/docs/drivers/azure.md
+++ b/docs/drivers/azure.md
@@ -26,7 +26,7 @@ You will need an Azure Subscription to use this Docker Machine driver.
 First time you try to create a machine, Azure driver will ask you to
 authenticate:
 
-    $ docker-machine create --driver azure <machine-name>
+    $ docker-machine create --driver azure --azure-subscription-id <subs-id> <machine-name>
     Running pre-create checks...
     Microsoft Azure: To sign in, use a web browser to open the page https://aka.ms/devicelogin.
     Enter the code [...] to authenticate.
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -1,6 +1,7 @@
 package azure
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -329,6 +330,10 @@ func (d *Driver) Create() error {
 
 // Remove deletes the virtual machine and resources associated to it.
 func (d *Driver) Remove() error {
+	if err := d.checkLegacyDriver(false); err != nil {
+		return err
+	}
+
 	// NOTE(ahmetalpbalkan):
 	//   - remove attemps are best effort and if a resource is already gone, we
 	//     continue removing other resources instead of failing.
@@ -367,6 +372,10 @@ func (d *Driver) Remove() error {
 
 // GetIP returns public IP address or hostname of the machine instance.
 func (d *Driver) GetIP() (string, error) {
+	if err := d.checkLegacyDriver(true); err != nil {
+		return "", err
+	}
+
 	if d.resolvedIP == "" {
 		ip, err := d.ipAddress()
 		if err != nil {
@@ -407,6 +416,10 @@ func (d *Driver) GetURL() (string, error) {
 
 // GetState returns the state of the virtual machine role instance.
 func (d *Driver) GetState() (state.State, error) {
+	if err := d.checkLegacyDriver(true); err != nil {
+		return state.None, err
+	}
+
 	c, err := d.newAzureClient()
 	if err != nil {
 		return state.None, err
@@ -425,6 +438,10 @@ func (d *Driver) GetState() (state.State, error) {
 
 // Start issues a power on for the virtual machine instance.
 func (d *Driver) Start() error {
+	if err := d.checkLegacyDriver(true); err != nil {
+		return err
+	}
+
 	c, err := d.newAzureClient()
 	if err != nil {
 		return err
@@ -434,6 +451,10 @@ func (d *Driver) Start() error {
 
 // Stop issues a power off for the virtual machine instance.
 func (d *Driver) Stop() error {
+	if err := d.checkLegacyDriver(true); err != nil {
+		return err
+	}
+
 	c, err := d.newAzureClient()
 	if err != nil {
 		return err
@@ -445,6 +466,10 @@ func (d *Driver) Stop() error {
 
 // Restart reboots the virtual machine instance.
 func (d *Driver) Restart() error {
+	if err := d.checkLegacyDriver(true); err != nil {
+		return err
+	}
+
 	// NOTE(ahmetalpbalkan) Azure will always keep the VM in Running state
 	// during the restart operation. Hence we rely on returned async operation
 	// polling to make sure the reboot is waited upon.
@@ -461,4 +486,16 @@ func (d *Driver) Kill() error {
 	// machines, Stop() is the closest option.
 	log.Debug("Azure does not implement kill. Calling Stop instead.")
 	return d.Stop()
+}
+
+// checkLegacyDriver errors out if it encounters an Azure VM created with the
+// legacy (<=0.6.0) docker-machine Azure driver.
+func (d *Driver) checkLegacyDriver(short bool) error {
+	if d.ResourceGroup == "" {
+		if short {
+			return errors.New("New azure driver cannot manage old VMs, downgrade to v0.6.0")
+		}
+		return errors.New("New azure driver uses the new Azure Resource Manager APIs and therefore cannot manage this existing machine created with old azure driver. Please downgrade to docker-machine 0.6.0 to continue using these machines or to remove them.")
+	}
+	return nil
 }
--- a/drivers/azure/azure.go
+++ b/drivers/azure/azure.go
@@ -1,0 +1,464 @@
+package azure
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+
+	"github.com/docker/machine/drivers/azure/azureutil"
+	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/log"
+	"github.com/docker/machine/libmachine/mcnflag"
+	"github.com/docker/machine/libmachine/state"
+
+	"github.com/Azure/azure-sdk-for-go/arm/storage"
+)
+
+const (
+	defaultAzureEnvironment     = "AzurePublicCloud"
+	defaultAzureResourceGroup   = "docker-machine"
+	defaultAzureSize            = "Standard_A2"
+	defaultAzureLocation        = "westus"
+	defaultSSHUser              = "ubuntu" // 'root' not allowed on Azure
+	defaultDockerPort           = 2376
+	defaultAzureImage           = "canonical:UbuntuServer:14.04.3-LTS:latest"
+	defaultAzureVNet            = "docker-machine-vnet"
+	defaultAzureSubnet          = "docker-machine"
+	defaultAzureSubnetPrefix    = "192.168.0.0/16"
+	defaultStorageType          = storage.StandardLRS
+	defaultAzureAvailabilitySet = "docker-machine"
+)
+
+const (
+	flAzureEnvironment     = "azure-environment"
+	flAzureSubscriptionID  = "azure-subscription-id"
+	flAzureResourceGroup   = "azure-resource-group"
+	flAzureSSHUser         = "azure-ssh-user"
+	flAzureDockerPort      = "azure-docker-port"
+	flAzureLocation        = "azure-location"
+	flAzureSize            = "azure-size"
+	flAzureImage           = "azure-image"
+	flAzureVNet            = "azure-vnet"
+	flAzureSubnet          = "azure-subnet"
+	flAzureSubnetPrefix    = "azure-subnet-prefix"
+	flAzureAvailabilitySet = "azure-availability-set"
+	flAzurePorts           = "azure-open-port"
+	flAzurePrivateIPAddr   = "azure-private-ip-address"
+	flAzureUsePrivateIP    = "azure-use-private-ip"
+	flAzureNoPublicIP      = "azure-no-public-ip"
+)
+
+const (
+	driverName = "azure"
+	sshPort    = 22
+)
+
+// Driver represents Azure Docker Machine Driver.
+type Driver struct {
+	*drivers.BaseDriver
+
+	Environment    string
+	SubscriptionID string
+	ResourceGroup  string
+
+	DockerPort      int
+	Location        string
+	Size            string
+	Image           string
+	VirtualNetwork  string
+	SubnetName      string
+	SubnetPrefix    string
+	AvailabilitySet string
+
+	OpenPorts     []string
+	PrivateIPAddr string
+	UsePrivateIP  bool
+	NoPublicIP    bool
+
+	// Ephemeral fields
+	ctx        *azureutil.DeploymentContext
+	resolvedIP string // cache
+}
+
+// NewDriver returns a new driver instance.
+func NewDriver(hostName, storePath string) drivers.Driver {
+	// NOTE(ahmetalpbalkan): any driver initialization I do here gets lost
+	// afterwards, especially for non-Create RPC calls. Therefore I am mostly
+	// making rest of the driver stateless by just relying on the following
+	// piece of info.
+	d := &Driver{
+		BaseDriver: &drivers.BaseDriver{
+			SSHUser:     defaultSSHUser,
+			MachineName: hostName,
+			StorePath:   storePath,
+		},
+	}
+	return d
+}
+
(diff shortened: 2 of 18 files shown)


drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	ctx        *azureutil.DeploymentContext
+	resolvedIP string // cache
+}
+
+// NewDriver returns a new driver instance.
+func NewDriver(hostName, storePath string) drivers.Driver {
+	// NOTE(ahmetalpbalkan): any driver initialization I do here gets lost
+	// afterwards, especially for non-Create RPC calls. Therefore I am mostly
+	// making rest of the driver stateless by just relying on the following
+	// piece of info.
```
nathanleclaire: Which pieces of info are you missing here?  On operations subsequent to `create` invocation, `[]byte` from disk (`config.json` file) for driver will be unmarshaled into the `struct` directly, _after_ the `NewDriver` function returns the initial struct, so it's not unexpected behavior.

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+			SSHUser:     defaultSSHUser,
+			MachineName: hostName,
+			StorePath:   storePath,
+		},
+	}
+	return d
+}
+
+// GetCreateFlags registers the flags this d adds to
+// "docker hosts create"
```
nathanleclaire: Probably best to remote this (probably historical) allusion to `docker hosts` :stuck_out_tongue:

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	log.Info("Completed machine pre-create checks.")
+	return nil
+}
+
+// Create creates the virtual machine.
+func (d *Driver) Create() error {
+	// NOTE(ahmetalpbalkan): We can probably parallelize the sh*t out of this.
+	// However that would lead to a concurrency logic and while creation of a
+	// resource fails, other ones would be kicked off, which could lead to a
+	// resource leak. This is slower but safer.
```
nathanleclaire: Safest == best for now.

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	c, err := d.newAzureClient()
+	if err != nil {
+		return err
+	}
+	steps := []func() error{
```
nathanleclaire: What's the reasoning for creating this slice of functions here?  Seems to me that they should just be invoked in order, right?

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	//   - we can probably do a lot of parallelization here but a sequential
+	//     logic works fine too. If we were to detach the NIC from the VM and
+	//     then delete the VM, this could enable some parallelization.
+
+	log.Info("NOTICE: Please check Azure portal/CLI to make sure you have no leftover resources to avoid unexpected charges.")
+	c, err := d.newAzureClient()
+	if err != nil {
+		return err
+	}
+	for _, f := range []func() error{
```
nathanleclaire: Same comment here as above.  Why?

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	// NOTE (ahmetalpbalkan) I noticed that this is not used until machine is
+	// actually created and provisioned. By then GetIP() should be returning
+	// a non-empty IP address as the VM is already allocated and connected to.
+	ip, err := d.GetIP()
+	if err != nil {
+		return "", err
+	}
+	u := (&url.URL{
```
nathanleclaire: Seems a bit belabored to make a `url.URL` and then convert it to string anyway, any reason not to just use `fmt.Sprintf` directly?

drivers/azure/azure.go
```diff
@@ -0,0 +1,455 @@
...
+	return c.StartVirtualMachine(d.ResourceGroup, d.naming().VM())
+}
+
+// Stop issues a power off for the virtual machine instance.
+func (d *Driver) Stop() error {
+	c, err := d.newAzureClient()
+	if err != nil {
+		return err
+	}
+	log.Info("NOTICE: Stopping an Azure Virtual Machine is just going to power it off, not deallocate.")
```
nathanleclaire: So there's a difference between stop and deallocate?  Why not make `Stop` deallocate? (not saying it's better just wondering)

nathanleclaire:
Made a few comments / questions but it's looking really good!  Thanks for taking the initiative on this.

My general feeling is that I want to err on the side of merging this soon, and testing-as-we-go.

Also, with this branch of development, we're breaking backwards compatibility with old `azure`-created hosts completely, no?  What's our transition plan?

drivers/azure/azure.go
ahmetalpbalkan: it's just to avoid the if err := f(); err != nil { } song and dance.

drivers/azure/azure.go
ahmetalpbalkan: @nathanleclaire hehe we can have a long argument about this. I'm very defensive about repeating the logic of stdlibs. For instance, people using `:` character to split IP/port are essentially going to be broken with IPv6. If should've used net.SplitHostPort, they'd be fine. So I'm basically using the logic in net/url to be on the safe side. :)

drivers/azure/azure.go
ahmetalpbalkan: @nathanleclaire hmm, so deallocationg will stop charging. Stopping on the other hand, will keep charging but will keep the IP address the same. (for instance if you stop/start google machines, you run into cert issues because IP address is ephemeral; not static, and I took the same approach here). Also starting deallocated VM will be slightly slower to start; starting a stopped VM on the other hand is way faster.

I'm open to any suggestions about what drivers should do in this case.

ahmetalpbalkan:
@nathanleclaire I share the same feelings, I feel like it's in a usable state already. 

> Also, with this branch of development, we're breaking backwards compatibility with old azure-created hosts completely, no? What's our transition plan?

It it completely breaking (I have a comment about this in `azure.md`), there's no way you could even delete Azure machines created with the old driver; and that's because new Azure APIs are completely different and they don’t share any data. There's not much we can do about that other than making the docs better, I believe.

drivers/azure/azure.go
ahmetalpbalkan: I'll fix if that is more idiomatic or readable to you. This could've looked better with lambdas. (/sigh).

drivers/azure/azure.go
ahmetalpbalkan: @nathanleclaire I believe from what I tried, assigning certain objects during various points of this object's lifetime caused those instances to get lost. So it is not very clear to me at which exact point this config data gets marshaled and unmarshaled.

For instance, when I assign something here during create or pre-create, I saw that it totally gets lost during GetState/GetIP calls that occur later during the machine creation.

For quite a moment I thought maybe it is reinitializing the driver for those calls –but this is not an external driver, I was assuming it'd be in-process and there'd be only a single azure Driver instance throughout the lifetime of a docker-machine execution process; but it appears like due to RPC that's not the case.

I'm still not clear about inner workings of RPC model in docker-machine. I  found my way around it but guess I still appreciate any explanation about lifecycle of Driver instance and when/how config gets (un/)marshaled.

drivers/azure/azure.go
ahmetalpbalkan: ohman this is very old. I wish we could leave it here, this might be the last reference to `docker hosts`.

drivers/azure/azure.go
ahmetalpbalkan: fixed this by unrolling funcs into `if...;err!=nil{}` statements.

drivers/azure/azure.go
ahmetalpbalkan: fixed this by unrolling funcs into `if...;err!=nil{}` statements.

drivers/azure/azure.go
ahmetalpbalkan: fixed. btw some other drivers have the outdated comment too.

ahmetalpbalkan:
Sorry about this gordon noise. **Protip:** never push this "Update branch" button:

>

docs/drivers/azure.md
```diff
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
...
-    $ docker-machine create -d azure --azure-subscription-id="SUB_ID" --azure-subscription-cert="mycert.pem" A-VERY-UNIQUE-NAME
+    $ docker-machine create --driver azure <machine-name>
+    Running pre-create checks...
+    Microsoft Azure: To sign in, use a web browser to open the page https://aka.ms/devicelogin.
+    Enter the code [...] to authenticate.
 
-The Azure driver uses the `b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB`
-image by default. Note, this image is not available in the Chinese regions. In China you should
- specify `b549f4301d0b4295b8e76ceb65df47d4__Ubuntu-15_10-amd64-server-20151116.1-en-us-30GB`.
+After authenticating, the driver will remember your credentials up to two weeks.
```
ppadala: It will be good to add a comment saying that, if someone wants to automate using docker-machine, they will have to provide publishsettings or subscriptionID, cert pair

docs/drivers/azure.md
```diff
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
...
--   `--azure-password`: Your Azure password.
--   `--azure-publish-settings-file`: Azure setting file. See [How to: Download and Import Publish Settings and Subscription Information](https://msdn.microsoft.com/en-us/library/dn385850%28v=nav.70%29.aspx)
--   `--azure-size`: Azure disk size.
--   `--azure-ssh-port`: Azure SSH port.
--   `--azure-subscription-id`: **required** Your Azure subscription ID (A GUID like `d255d8d7-5af0-4f5c-8a3e-1545044b861e`).
--   `--azure-subscription-cert`: **required** Your Azure subscription cert.
--   `--azure-username`: Azure login user name.
+Required:
+
+- `--azure-subscription-id`: **(required)** Your Azure Subscription ID.
```
ppadala: No certs? I am a bit confused. My understanding was that subID + cert or publishsettings file has to be provided.

Is the browser based login mandatory now?

ahmetalpbalkan: @ppadala there are no certs or publishSettings files in new Azure APIs. Browser auth is the easiest way for user, the next easiest way is like creating a service principal account in 63 steps, so we went ahead with this.

drivers/azure/azure.go
```diff
@@ -0,0 +1,464 @@
...
+	c, err := d.newAzureClient()
+	if err != nil {
+		return err
+	}
+
+	if err := c.CreateResourceGroup(d.ResourceGroup, d.Location); err != nil {
+		return err
+	}
+	if err := c.CreateAvailabilitySetIfNotExists(d.ctx, d.ResourceGroup, d.AvailabilitySet, d.Location); err != nil {
+		return err
```
ppadala: Don't have to unwind the resource group creation?

drivers/azure/azure.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+	if err := c.CreateNetworkSecurityGroup(d.ctx, d.ResourceGroup, d.naming().NSG(), d.Location, d.ctx.FirewallRules); err != nil {
+		return err
```
ppadala: Same here. Don't you have to remove the objects created earlier?

drivers/azure/azure.go
```diff
@@ -0,0 +1,464 @@
...
+	//   - we can probably do a lot of parallelization here but a sequential
+	//     logic works fine too. If we were to detach the NIC from the VM and
+	//     then delete the VM, this could enable some parallelization.
+
+	log.Info("NOTICE: Please check Azure portal/CLI to make sure you have no leftover resources to avoid unexpected charges.")
+	c, err := d.newAzureClient()
+	if err != nil {
+		return err
+	}
+	if err := c.DeleteVirtualMachineIfExists(d.ResourceGroup, d.naming().VM()); err != nil {
```
ppadala: Is it better to do this at the end of removing other objects?

docs/drivers/azure.md
ppadala: How can this be done non-interactively?

drivers/azure/azure.go
```diff
@@ -0,0 +1,464 @@
...
+	flAzureUsePrivateIP    = "azure-use-private-ip"
+	flAzureNoPublicIP      = "azure-no-public-ip"
+)
+
+const (
+	driverName = "azure"
+	sshPort    = 22
+)
+
+// Driver represents Azure Docker Machine Driver.
```
ppadala: Can you add a comment for information for Environment, ResourceGroup etc.

drivers/azure/azure.go
ahmetalpbalkan: I am not sure what you mean?

drivers/azure/azure.go
ahmetalpbalkan: @ppadala  No, I don't think so. We create two types of resources:

1. Resources who are attached to VM: (e.g. Network interface, IP address, Firewall) you can't delete those without deleting the VM or detaching those resources first. Deleting VM just disassociates these resources so that Delete calls below work just fine.

2. Resources that hold the VM: (e.g. Resource Group, Availability Set, Virtual Network, Subnet) You can't remove those without removing the VM first, it's like deleting a non-empty folder. The API does not allow it.

drivers/azure/azure.go
ahmetalpbalkan: @ppadala umm not sure why I should comment each field. I haven't seen any other driver doing it and they're sort of self-documenting through (a) naming (b) existing documentation.

drivers/azure/azure.go
ahmetalpbalkan: like what?

docs/drivers/azure.md
ahmetalpbalkan: @ppadala it cannot be done interactively and it is by design. I have discussed this with machine maintainers some time ago and this seems like a more practical solution as it is a lot less tedious for the users compared to the other means of authentication.

docs/drivers/azure.md
ahmetalpbalkan: Not anymore. Please look at the azure.md as a whole, not a diff. The new driver has nothing to do with the old driver, including the authentication model.

docs/drivers/azure.md
ppadala: Makes sense for interactive users, but we (and a few others like Rancher and DCHQ) are using this as part of our automated workflows to create Docker hosts.

Is there a solution here?

drivers/azure/azure.go
ppadala: No need to comment each field, just link at the top of the structure or driver file, that points to relevant Azure APIs. Note that, for people outside Microsoft, there's good amount of learning, cross-referencing with go-sdk etc.

drivers/azure/azure.go
ppadala: Ok, thanks for the clarification.

drivers/azure/azure.go
ppadala: Line 291: creates a resource group
Line 294: tries to create availabilityset and fails, don't we have to remove the resource group created above?

drivers/azure/azure.go
ppadala: See above comment. I am trying to understand, if the objects created earlier need to be removed in the case of failures.

drivers/azure/azure.go
ahmetalpbalkan: @ppadala all these Azure concepts can be googled very easily and they are not cryptic (so that only msft employees would understand). I don't see the point of adding a ton of "how Azure works" comments here. godoc might not be the right place for an Azure lecture. :)

drivers/azure/azure.go
ahmetalpbalkan: @ppadala I don't think so. Could you please point me to any driver that does that?

docs/drivers/azure.md
ahmetalpbalkan: @ppadala I was not aware of automated use case of docker-machine by other programs/scripts. I'd need to chat with maintainers about this. It looks like we will go ahead with this model for a while, if it is a supported use case, we may add support for that 63-step authentication model.

drivers/azure/azure.go
ahmetalpbalkan: @nathanleclaire what do you think should be the desired behavior here? faster (that costs money and keeps the ephemeral IP address) or slower (that stops charges but changes the ephemeral IP address).

docs/drivers/azure.md
ppadala: @ahmetalpbalkan we (ContainerX) use it as a building block for provisioning on Azure. We are going support Windows server Docker hosts on Azure and automation is required. I would hate to stay with older version of azure driver for this reason. I can't speak for Rancher, but they posted this last year http://rancher.com/architercutre-of-ranchers-docker-machine-integration/.

I have seen many customers who are using docker-machine in Jenkins/Bamboo workflows. If docker-machine Azure doesn't support for automation, where it can be invoked in a script, it's going to be a problem and customers are going to choose different cloud providers that allow easy automation. Also, reentering your credentials every two weeks is just not going work. Consider this scenario, let's say we open a browser window the very first time an IT admin wants to setup Azure as a backend and store the credentials. How are we going to ask them to reenter again? e-mail them to come back and reenter? It's just an ugly workflow.

Can you point me to the documentation for the other authentication model? Is it really 63 steps?

ahmetalpbalkan: @ppadala I agree, as I said, I'll chat with maintainers and implement it. It's not 63 steps but it is some 20 steps, you can search for "azure service principal account", but you won't be able to use it today with this driver. Let's move this discussion to an issue once this PR is merged. I don't see this as a blocking step.

nathanleclaire:
Is there any way to provide users a script or something that will migrate old versions of VMs created with the `azure` driver to this new driver?  I'm starting to feel that I was a bit hasty with the merge removing the old driver and considering reverting it.  Our deprecation policy is [two releases](https://docs.docker.com/engine/#feature-deprecation-policy) so dropping support for a driver so abruptly has me concerned.

Let's talk about options to preserve and migrate existing `azure` VMs, since I'm sure at least some are out there.  Maybe we can rename and include the existing driver as `azure-legacy` and bump config versions to do a migration of existing `azure` VMs (at config version 3) to have the driver name `azure-legacy`.

ahmetalpbalkan:
@nathanleclaire fair point. I was not aware of two versions policy, though I remember asking about it in our earlier conversations (:P).

A migration is just not possible. What I can offer is, maybe we keep azure driver around for another release and call this `azure-new then fix it later?

Basically if you ask me, I don't think people will be disappointed if we break with a nice error message saying "hey this version doesn't support old Azure VMs, please go back to v0.6.0", but up to you.

ahmetalpbalkan:
Or we can call this `azure` and when an old VM is triggered, we can offer a link to migration script that modifies existing VMs to `azure-old`. Feel free to pick one of these solutions @nathanleclaire, I will happily implement.

nathanleclaire:
So, consensus on our end after talking with @dmp42 is that:

1. We are fine with moving forward using the new Azure driver, however-
2. Any existing `azure` machines machines should error out when any operation  with the next released version of Machine (after merging this code in) is attempted.  A "deprecated" type warning suggesting the user switch to the new model in order to use the new version of Machine should be shown.
3. Likewise, no destructive updates to the `config.json` should be made.  It _must_ be possible for the user to use Machine 0.6.0 to tie up loose ends (e.g. `docker-machine rm`) with the legacy driver before moving to the new one.
4. We will also put a notice about this in the release notes.

How does that sound?  Are you willing to work with us on verifying that the above works as expected? I've not had much success with the legacy Azure driver in the past unfortunately.  The tricky bit in the above will be preventing damage to the `config.json` -- it gets loaded/saved in quite a few places.

ahmetalpbalkan:
@nathanleclaire That sounds good. It looks like I just have to make sure we don't damage config.json and terminate early when we detect the old `azure` driver (and tell users about deprecation and downgrading to `rm` old machines).

Actually this new driver already errors out and doesn't damage config.json for old Azure VMs but I will make the error clearer.

nathanleclaire:
> Actually this new driver already errors out and doesn't damage config.json for old Azure VMs but I will make the error clearer.

OK sounds great. Thanks Ahmet.

ahmetalpbalkan:
Added the commit that handles the gracefully erroring out machines of created with the old driver, as a separate commit. I intend to undo it after two versions. It looks like this:

image

drivers/azure/azure.go
ppadala: This seems to be a bad design in docker-machine overall. Leaving the cleanup of dangling cloud objects to users seems bad. It's up to you whether you want to add this to Azure driver (we have made changes to vSphere driver (to be submitted upstream) to take care of some of the cleanup problems). 

If there's a way to delete all these objects in one shot with docker-machine rm -f <machine name>, that would be fine instead of unwinding for every failure, which can become complicated and error prone.

docs/drivers/azure.md
ppadala: Ok, sounds good. I checked azure service principal account and it seems to require AAD. We can discuss this further in a separate issue.

drivers/azure/azure.go
ahmetalpbalkan: @ppadala yes, an interrupted `create` can be cleaned up perfectly with `rm` in this new azure driver.

ppadala: Great!

nathanleclaire:
Have you tested `start`, `stop`, `ssh`, `env` etc. with Machine 0.6.0 still works properly after doing some invocations such as the above using your new driver?

nathanleclaire:
Nice shell prompt btw :wink: 
screen shot 2016-03-10 at 4 00 50 pm

ahmetalpbalkan:
@nathanleclaire yup, it errors out and exits with non-zero exit code properly.

nathanleclaire:
Did you test with `start` and/or `stop` family of commands specifically (not just `rm`)?  The reason is that those commands are known to be ones that modify `config.json` (unlike `rm`).

ahmetalpbalkan:
@nathanleclaire yep just tested manually, it errors out on all commands immediately and machine is still usable after running all these commands (`dm`: this PR, `docker-machine`:v0.6.0):

```


➜ dm ls
NAME           ACTIVE   DRIVER   STATE   URL   SWARM   DOCKER    ERRORS
azure-old   -           azure    Error                 Unknown   New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm ssh azure-old
New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm start azure-old
Starting "azure-old"...
New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm stop azure-old
Stopping "azure-old"...
New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm ip azure-old
Error getting IP address: New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm env azure-old
Error checking TLS connection: New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm kill azure-old
Killing "azure-old"...
New azure driver cannot manage old VMs, downgrade to v0.6.0

➜ dm rm azure-old
Are you sure? (y/n): y
Error removing host "azure-old": New azure driver uses the new Azure Resource Manager APIs and therefore cannot manage this existing machine created with old azure driver. Please downgrade to docker-machine 0.6.0 to continue using these machines or to remove them.
```

and it still works:

```
➜ docker-machine ip azure-old
azure-old.cloudapp.net

➜ eval $(docker-machine env azure-old)
➜ docker ps
(it works)

➜ docker-machine restart azure-old
Restarting "azure-old"...

➜ docker ps
(it works)
```

nathanleclaire:
OK great.  Can I ask you to please update the `docs/RELEASE.md` in this PR to have a small note about calling out the breaking change to `azure` driver?  So that we remember in the next release, we can remove it after.

nathanleclaire:
@dgageot @jeanlaurent Please take a look at this PR when you get a chance.  I'd like to merge as soon as we feel comfortable.

ahmetalpbalkan:
@nathanleclaire added 029ed63 updating docs/RELEASE.md.

dgageot:
Huge work everyone! I'll take a look tomorrow

nathanleclaire:
Hey @ahmetalpbalkan don't thumbs down the tortoise!  He works hard OK :stuck_out_tongue:

drivers/azure/azure.go
```diff
@@ -0,0 +1,501 @@
...
+	resolvedIP string // cache
+}
+
+// NewDriver returns a new driver instance.
+func NewDriver(hostName, storePath string) drivers.Driver {
+	// NOTE(ahmetalpbalkan): any driver initialization I do here gets lost
+	// afterwards, especially for non-Create RPC calls. Therefore I am mostly
+	// making rest of the driver stateless by just relying on the following
+	// piece of info.
+	d := &Driver{
```
dgageot: @ahmetalpbalkan I'd return that directly. No need for `d` variable/

drivers/azure/azure.go
```diff
@@ -0,0 +1,501 @@
...
+			Value:  defaultAzureSubnetPrefix,
+		},
+		mcnflag.StringFlag{
+			Name:   flAzureAvailabilitySet,
+			Usage:  "Azure Availability Set to place the virtual machine into",
+			EnvVar: "AZURE_AVAILABILITY_SET",
+			Value:  defaultAzureAvailabilitySet,
+		},
+		mcnflag.StringFlag{
+			Name:  flAzurePrivateIPAddr,
```
dgageot: @ahmetalpbalkan Is it on purpose that you don't specify env variables for these last 4 flags?

ahmetalpbalkan: @dgageot yes, they either cannot really be used on multiple VMs, or they should be consciously set as an argument, or they're bool/slices so it's not really worth the trouble making them env vars.

drivers/azure/azure.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+// Remove deletes the virtual machine and resources associated to it.
+func (d *Driver) Remove() error {
+	if err := d.checkLegacyDriver(false); err != nil {
+		return err
+	}
+
+	// NOTE(ahmetalpbalkan):
+	//   - remove attemps are best effort and if a resource is already gone, we
```
dgageot: attempts?

drivers/azure/azure.go
dgageot: Ok. I thought so, wanted to check.

drivers/azure/azure.go
ahmetalpbalkan: well every http request or API call is an attempt, it may fail, as rest of the sentence says :) I don't really know what you mean.

drivers/azure/azure.go
```diff
@@ -0,0 +1,501 @@
...
+	// NOTE(ahmetalpbalkan) In Azure, there is no kill option for virtual
+	// machines, Stop() is the closest option.
+	log.Debug("Azure does not implement kill. Calling Stop instead.")
+	return d.Stop()
+}
+
+// checkLegacyDriver errors out if it encounters an Azure VM created with the
+// legacy (<=0.6.0) docker-machine Azure driver.
+func (d *Driver) checkLegacyDriver(short bool) error {
+	if d.ResourceGroup == "" {
```
dgageot: I'd remove some nesting here with:
```
if d.ResourceGroup != "" {
  return nil
}
```

drivers/azure/azure.go
dgageot: I meant I guess there's a typo. The question marks is there because I'm not sure of my english ;-)

dgageot:
I'm getting this message just after the auth part:
```
Error with pre-create check: "Unknown resource provider \"Microsoft.Compute\""
```

As a user, I don't know what to do with that. Couldn't find in the doc either (maybe I missed something). We might want to give directions on how to solve this issue. Don't you think?

(Also, the quoting is a bit unfortunate)

drivers/azure/azure.go
ahmetalpbalkan: ah alright! :) let's keep it as is, I'll try to address in a future commit.

dgageot:
@ahmetalpbalkan I've created a fresh new account and it works better. Testing in progress...
Since it takes some time, let me say that it's an impressive PR. Thanks a lot for the contribution.

ahmetalpbalkan:
@dgageot hmm that looks certainly interesting. I'd love to investigate that. If you still have the Azure subscription, could you please use `azure` CLI to run `azure provider list` command with that subscription? I just haven't encountered a case where the Azure subscription can't see the Compute system. It should be enabled for everybody. The error comes from https://github.com/docker/machine/pull/3159/files#diff-93a074128ed8940ea0dafbd000270978R249

dgageot:
@ahmetalpbalkan @nathanleclaire except for this `Unknown resource provider \"Microsoft.Compute\""` issue that seems more connected to my subscription than anything, it all LGTM.
I'm going to investigate the issue based on @ahmetalpbalkan instructions.

LGTM!

dgageot:
@ahmetalpbalkan I can't find a `azure provider list` command with azure cli 0.9.16

ahmetalpbalkan:
@dgageot argh our cli uses old APIs by default. Try running "azure config mode arm" first.

dgageot:
Thanks

dgageot:
I think the issue it that checks on providers are case sensitive. 
```
data:    microsoft.compute                Registered
```
Currently investigating if I can make it work.

ahmetalpbalkan:
Wow that is massively weird. Can you please email me the full output? I'll follow up with the product team and I'll send out a fix commit in the morning PST.

dgageot:
It works with case insensitive checks. Il will send you the full output.

Here's my take on the issue:
```
func (a AzureClient) RegisterResourceProviders(namespaces ...string) error {
	l, err := a.providersClient().List(nil)
	if err != nil {
		return err
	}
	if l.Value == nil {
		return errors.New("Resource Providers list is returned as nil.")
	}

	m := make(map[string]bool)
	for _, p := range *l.Value {
		m[strings.ToLower(to.String(p.Namespace))] = to.String(p.RegistrationState) == "Registered"
	}

	for _, ns := range namespaces {
		registered, ok := m[strings.ToLower(ns)]
		if !ok {
			return fmt.Errorf("Unknown resource provider %q", ns)
		}
		if registered {
			log.Debugf("Already registered for %q", ns)
		} else {
			log.Info("Registering subscription to resource provider.", logutil.Fields{
				"ns":   ns,
				"subs": a.subscriptionID,
			})
			if _, err := a.providersClient().Register(ns); err != nil {
				return err
			}
		}
	}
	return nil
}
```

ahmetalpbalkan:
@dgageot just pushed a commit fixing this. Sorry for the trouble. API folks are looking at why some namespaces might be returned as lowercase for some users. Thanks for reporting this.

dgageot:
LGTM

ahmetalpbalkan:
@dgageot thanks for reviewing and testing, I really appreciate we found that bug before releasing this.  We currently don't know how many users are getting this different uppercase/lowercase output from the APIs but we got it covered for now.

dgageot:
ping @nathanleclaire

nathanleclaire:
Going to take it for a test drive, then most likely merge.  cc @londoncalling for docs review; if there are any problems, I suggest we open and issue and do a follow-up PR.

docs/drivers/azure.md
```diff
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
...
-    $ openssl x509 -inform pem -in mycert.pem -outform der -out mycert.cer
+[azure]: http://azure.microsoft.com/
+[trial]: https://azure.microsoft.com/free/
 
-Go to the Azure portal, go to the "Settings" page (you can find the link at the bottom of the
-left sidebar - you need to scroll), then "Management Certificates" and upload `mycert.cer`.
+## Authentication
 
-Grab your subscription ID from the portal, then run `docker-machine create` with these details:
+First time you try to create a machine, Azure driver will ask you to
```
londoncalling: The first time you try to ...etc.

docs/drivers/azure.md
```diff
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
 
 # Microsoft Azure
 
-Create machines on [Microsoft Azure](http://azure.microsoft.com/).
+You will need an Azure Subscription to use this Docker Machine driver.
+[Sign up for a free trial.][trial]
 
-You need to create a subscription with a cert. Run these commands and answer the questions:
+> **NOTE:** This documentation is for the new version of  Azure driver started
```
londoncalling: is there an extra space between "of" and "Azure"?

docs/drivers/azure.md
```diff
@@ -10,51 +10,105 @@ parent="smn_machine_drivers"
...
+> shipping with v0.7.0 and it is not backwards-compatible with the old Azure
+> driver. If you like to manage your existing Azure machines, please download
```
londoncalling: If you want to continue managing your existing Azure machines, please download ...

londoncalling:
I made some wording corrections, but haven't had time to test the driver yet. If you like, I can update for flow of the text in the azure.md file after this is merged. There's just a lot of nitpicky things I would change throughout about the wording.

ahmetalpbalkan:
@londoncalling thanks for reviewing, just addressed docs comments in a separate commit (a5ad617)

nathanleclaire:
@ahmetalpbalkan How exactly does one get an Azure subscription ID?  This screen doesn't have any and can't figure out how to get one.

screen shot 2016-03-14 at 5 35 57 pm

ahmetalpbalkan:
@nathanleclaire hmm that's weird I didn't know you could have an Azure account without having a subscription. Try http://azure.com/free to get a Trial Subscription.

nathanleclaire:
Guess I had a free trial one that expired so maybe Azure got confused.  I am starting the whole funnel again from that link you sent me -- think this should work.

nathanleclaire:
Create and other expected operations works smoothly for me (and create is miles ahead of the old driver in terms of user experience -- excellent job!).

LGTM

Thanks and congratulations on all your good work Ahmet and everyone else involved in bringing this to fruition.

ahmetalpbalkan:
@nathanleclaire @dgageot It's very delightful to hear these, thank you for taking time to review this as well. I will be sending the improvements incrementally.

ppadala:
Great work @ahmetalpbalkan! I have started rebasing my change. Also, filed an issue based on our discussion above about automation.

edevil:
Great work!

PR #3168, opened 2016-03-09
[Issue-2989] Allow non administrators to create docker-machines

jtnord:
If the user has "hyper-v administrator" permissions but not Windows "Administrator" permissions they should still be able to create a docker machine.

Unfortunately resizing a VHD still requires windows administrator permissions so in the case the user only has Hyper-V Administrator permissions then create the disk with the correct size upfront.  (this is a much slower process but it is one that will allow non admins to create machines)

### DIFF
--- a/drivers/hyperv/hyperv.go
+++ b/drivers/hyperv/hyperv.go
@@ -428,8 +428,19 @@ func (d *Driver) generateDiskImage() (string, error) {
 	diskImage := d.ResolveStorePath("disk.vhd")
 	fixed := d.ResolveStorePath("fixed.vhd")
 
+	// Resizing vhds requires administrator priviledges
+	// incase the user is only a hyper-v admin then create the disk at the target size to avoid resizing.
+	isWindowsAdmin, err := isWindowsAdministrator()
+	if err != nil {
+		return "", err
+	}
+	fixedDiskSize := "10MB"
+	if !isWindowsAdmin {
+		fixedDiskSize = toMb(d.DiskSize)
+	}
+
 	log.Infof("Creating VHD")
-	if err := cmd("New-VHD", "-Path", quote(fixed), "-SizeBytes", "10MB", "-Fixed"); err != nil {
+	if err := cmd("New-VHD", "-Path", quote(fixed), "-SizeBytes", fixedDiskSize, "-Fixed"); err != nil {
 		return "", err
 	}
 
@@ -451,12 +462,14 @@ func (d *Driver) generateDiskImage() (string, error) {
 	}
 	file.Close()
 
-	if err := cmd("Convert-VHD", "-Path", quote(fixed), "-DestinationPath", quote(diskImage), "-VHDType", "Dynamic"); err != nil {
+	if err := cmd("Convert-VHD", "-Path", quote(fixed), "-DestinationPath", quote(diskImage), "-VHDType", "Dynamic", "-DeleteSource"); err != nil {
 		return "", err
 	}
 
-	if err := cmd("Resize-VHD", "-Path", quote(diskImage), "-SizeBytes", toMb(d.DiskSize)); err != nil {
-		return "", err
+	if isWindowsAdmin {
+		if err := cmd("Resize-VHD", "-Path", quote(diskImage), "-SizeBytes", toMb(d.DiskSize)); err != nil {
+			return "", err
+		}
 	}
 
 	return diskImage, nil
--- a/drivers/hyperv/powershell.go
+++ b/drivers/hyperv/powershell.go
@@ -69,6 +69,36 @@ func hypervAvailable() error {
 }
 
 func isAdministrator() (bool, error) {
+	var hypervAdmin bool
+	var windowsAdmin bool
+	var windowsErr error
+
+	hypervAdmin = isHypervAdministrator()
+
+	if hypervAdmin == true {
+		return true, nil
+	}
+
+	windowsAdmin, windowsErr = isWindowsAdministrator()
+
+	if windowsErr != nil {
+		return false, windowsErr
+	}
+
+	return windowsAdmin, nil
+}
+
+func isHypervAdministrator() bool {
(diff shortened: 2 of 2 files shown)


jtnord:
@peschkaj can you sign the initial commit that this was based off - then I can do the same.

peschkaj:
@jtnord do you have a guide that shows me how to do it without modifying history? When I run `git --amend --signoff`, the push to GitHub is rejected. Or, if changing history is fine, I can just push that.

jtnord:
@peschkaj changing history is fine.

nathanleclaire:
cc @dgageot please take a look

nathanleclaire:
We allow simply one commit signed by multiple parties as well -- that would be our preference here.

dgageot:
This looks ok. @jtnord can you please sort out the commits?

jtnord:
@dgageot what is wrong with the commits - they are all signed?  Merging them into one will show it as all my work (a single commit = single author/committer) which is not correct as the first commit was done by someone else.

nathanleclaire:
@jtnord You can sign a commit as multiple authors.

jtnord:
nathanleclaire I don;t care what is in the commit message.
I care about the actually commit (ie the author and the committer).  I should not be marked as an author for someone elses work!

nathanleclaire:
> nathanleclaire I don;t care what is in the commit message.
> I care about the actually commit (ie the author and the committer). I should not be marked as an author for someone elses work!

It's precedent on this project to have as few commits as possible when merging PRs in order to make rebases as easy as possible for everyone.  Sometimes that involves authoring or signing a commit as multiple parties, e.g. when working with another person before submitting the final work to upstream, or when a maintainer has to carry a PR to make a few small revisions themselves before merging it.  The authorship and signing trail is still explicit.

Multiple `Signed-off-by` lines seem to have a precedent in other projects [such as the Linux kernel](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=d6b7eaeb03421139e32800324ef04ab50bba886d).

I hope that helps to explain about our preference and practice of this on the project.

At any rate, the build isn't failing because of the way the commits are signed.  The TravisCI is failing because it seems that `gofmt` needs to be run.

```
Congratulations! All commits are properly signed with the DCO!
drivers/hyperv/hyperv.go
mk/validate.mk:12: recipe for target 'fmt' failed
make: *** [fmt] Error 1
make: *** [build validate coverage-send] Error 2
```

Can you please take a look at that?  Thanks!

jtnord:
still don't know what all the errors where for (useful tool!) so just blindly formatted all the code, and squashed this down to 2 commits and it made me feel dirty.

nathanleclaire:
@StefanScherer Looks like the AppVeyor build is failing due to the Go source tarball 503ing... Do you know if we can re-run the build?

drivers/hyperv/powershell.go
```diff
@@ -69,6 +69,36 @@ func hypervAvailable() error {
 }
 
 func isAdministrator() (bool, error) {
+	var hypervAdmin bool
```
nathanleclaire: Why not skip the `var` declarations entirely by using the `:=` operator?

```go
hypervAdmin := isHypervAdministrator()
if hypervAdmin {
    return true, nil
}

windowsAdmin, err := isWindowsAdministrator()
if err != nil {
    return false, err
}

return windowsAdmin, nil
```

StefanScherer:
@nathanleclaire Sure you can "RE-BUILD COMMIT" if you are logged in and have access rights to do so. Here is a sample of the top right corner with the buttons you are searching for:

bildschirmfoto 2016-03-21 um 21 41 06

drivers/hyperv/hyperv.go
```diff
@@ -428,8 +428,16 @@ func (d *Driver) generateDiskImage() (string, error) {
 	diskImage := d.ResolveStorePath("disk.vhd")
 	fixed := d.ResolveStorePath("fixed.vhd")
 
+	// Resizing vhds requires administrator priviledges
+	// incase the user is only a hyper-v admin then create the disk at the target size to avoid resizing.
+	isWindowsAdmin, err := isWindowsAdministrator()
```
nathanleclaire: Where does this `err` get checked?

drivers/hyperv/powershell.go
```diff
@@ -69,6 +69,36 @@ func hypervAvailable() error {
...
+	if windowsErr != nil {
+		return false, windowsErr
+	}
+
+	return windowsAdmin, nil
+}
+
+func isHypervAdministrator() bool {
+	stdout, err := cmdOut(`@([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole("Hyper-V Administrators")`)
+	if err != nil {
```
nathanleclaire: Why not also `log.Debug(err)`?  There are quite a few reasons this might return an error which don't indicate permissions status.

drivers/hyperv/powershell.go
jtnord: possibly because the authors (well this one at least) are new to go and this was a drive by submission :smile:   Made the changes - but not tested, assuming they are functionally equivalent.

drivers/hyperv/hyperv.go
jtnord: it doesn't but you won't have got this far if the function would fail - as it is run earlier in the are you admin or hyper-v admin check. 

Also I don't really get how multiple return values and assignments work - as this was a drive by submission as I wanted a working docker-machine on my machine I did not stop to attend the go-102 (I did part of the go-101 course).

drivers/hyperv/powershell.go
jtnord: added

nathanleclaire:
Cool -- code looks good, thanks -- hopefully the AppVeyor build goes smoothly this go-round.

nathanleclaire:
LGTM thanks @jtnord @peschkaj

jtnord:
thanks @nathanleclaire

PR #3182, opened 2016-03-11
added scpi command for copying images

marconi:
This adds copy image (`scpi`) command for copying images from one machine to another. Helpful if you have huge image that pulling it again will take long or if you don't have internet access.

This is my first PR, any feedback would be appreciated. :)

### DIFF
--- a/commands/scpi.go
+++ b/commands/scpi.go
@@ -3,10 +3,10 @@ package commands
 import (
 	"errors"
 	"fmt"
+	"os/exec"
 	"strings"
 
 	"github.com/docker/machine/libmachine"
-	"github.com/docker/machine/libmachine/host"
 )
 
 func cmdScpi(c CommandLine, api libmachine.API) error {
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Very interesting.  I guess this would allow for things like:

```console
$ docker $(docker-machine config swarmnode-0) pull me/image
$ docker-machine scpi swarmnode-0:me/image swarmnode-1
$ docker-machine scpi swarmnode-0:me/image swarmnode-2
```

?

marconi:
@nathanleclaire yep, sorry forgot to add an example. The syntax is basically:

```
docker-machine scpi [machine]:[image][:tag] [machine]
```
The `tag` is optional, if omitted, defaults to `latest`.

commands/scpi.go
```diff
@@ -0,0 +1,106 @@
...
+	// Remote path
+	host, err := api.Load(hostName)
+	if err != nil {
+		return nil, "", fmt.Errorf("Error loading host: %s", err)
+	}
+
+	return host, image, nil
+}
+
+func generateSaveFilename(srcImage string) string {
```
nathanleclaire: Could you make a test for this method as well please?

commands/scpi.go
```diff
@@ -0,0 +1,106 @@
...
+	rmCmd := fmt.Sprintf("rm %s", saveFilename)
+	srcClient.Shell(rmCmd)
+	destClient.Shell(rmCmd)
+
+	return nil
+}
+
+func getScpiHostAndImage(hostAndPath string, isSource bool, api libmachine.API) (*host.Host, string, error) {
+	var (
+		hostName = hostAndPath
```
nathanleclaire: If this is going to be set to different value later on, why not leave off this initial assignment ?

nathanleclaire:
Cool use case -- I'm definitely interested.  Why not use the Docker client API that machine has (this is how we manage swarm containers) directly instead of running `docker` commands over SSH?

commands/scpi.go
marconi: @nathanleclaire yep, sure.

commands/scpi.go
marconi: Because it acts as default value when this function is called for destination where it doesn't need parsing since destination only contains machine name directly unlike source where it has machine and image name.

marconi:
@nathanleclaire ah right, I think I can make that work. Will keep you posted.

marconi:
@nathanleclaire I made some progress and I thought it even became simpler, let me know what you think. ;)

nathanleclaire:
Hm, I think I have not been clear in communicating what I mean. 

I mean that instead of shelling out to `docker` and `docker-machine` commands, it should use the built-in `DockerClient` from `mcndockerclient` module: https://github.com/docker/machine/blob/master/libmachine/mcndockerclient/docker_client.go#L11

Example usage in Swarm code here: https://github.com/docker/machine/blob/master/libmachine/provision/configure_swarm.go#L110

Can you update it to use that please?

Likewise, it should probably invoke `scp` directly.  I'd suggest trying to use the existing `docker-machine scp`-related methods.

Thanks!

marconi:
@nathanleclaire ah, I see. Sure will update, I was actually initially looking for something like it but wasn't sure how `libmachine` worked.

marconi:
Only see one problem, the `DockerClient` doesn't have support for saving image. But the `engine-api`'s `client` [does](https://godoc.org/github.com/docker/engine-api/client#Client.ImageSave), I saw some discussion with `dockerclient` moving to `engine-api` so its prolly not worth it adding an image save endpoint for it esp. since machine has [this](https://github.com/docker/machine/pull/2764) as well.

marconi:
Close this for now.

PR #3194, opened 2016-03-15
azure: Add --azure-static-public-ip flag

ahmetalpbalkan:
This flag creates Azure VMs with static public IP addresses that don’t get released when the VM gets deallocated.


dgageot:
LGTM

nathanleclaire:
LGTM

PR #3200, opened 2016-03-16
Specify the format for azure-image specification

ppadala:
The format for azure-image is not specified anywhere in the docs. The link for more information talks about how to get information on publisher,offer,sku and version, but doesn't specify format for azuere-image option for azure driver.

cc @ahmetalpbalkan

### DIFF
--- a/docs/drivers/azure.md
+++ b/docs/drivers/azure.md
@@ -45,7 +45,7 @@ Required:
 
 Optional:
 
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
Lgtm. The default value and the "[?]" link in the docs probably help a bit, but this is better. Users still have to run the azure CLI to find the image name, SKU, publisher etc. it's hard to guess. And cli actually prints the full image name.

dgageot:
LGTM

Issue #3181, opened 2016-03-11
--native-ssh does not seem to be recognised by `docker-machine env` (v0.6/Docker Toolbox)


efrecon:
I have engaged in the process of porting [machinery](https://github.com/efrecon/machinery) to windows. Most of it works flawlessly, but I've had to turn on --native-ssh to rely on the internal ssh implementation. I am getting the following error when running from within my terminal (mobaXTerm):

```
[emmanuel.thinny] ➤ docker-machine.exe --native-ssh env default
Error checking TLS connection: Something went wrong running an SSH command!
command : ip addr show dev eth1
err     : fork/exec C:\Users\emmanuel\DOCUME~1\MOBAXT~1\slash\bin\ssh.exe: This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
output  :
```

Other commands, e.g. `docker-machine.exe --native-ssh ssh` work as they should in the terminal.

Does `env` respect the `--native-ssh` option? If it does not, could it? Why does it need to?

nathanleclaire:
I can dupe... Will follow up

nathanleclaire:
Seems like it's not just `env`, but _all_ commands which aren't `ssh` (`restart` etc. shows same ignoring of `--native-ssh` flag).  I suspect it's related to `SkipFlagParsing` setting.

nathanleclaire:
Mmm, I have a guess: https://github.com/docker/machine/issues/2296

Since the SSH is invoked by the plugin, not main binary, it defiantly ignores the flag.

We need to communicate these flags to plugins processes as well, fixing #2296.

nathanleclaire:
One possible solution (albeit kind of a hack) is to `os.Setenv` the environment variables in the client process, then have a block similar to https://github.com/docker/machine/blob/master/commands/commands.go#L157-L159 in `RegisterDriver` for plugins.

nathanleclaire:
Ideal solution would be to beam the `libmachine` configuration over the wire to plugin process after it gets invoked.

nathanleclaire:
Working on a solution for this, it will take a second though.

nathanleclaire:
https://github.com/docker/machine/pull/3209 should fix this.

botmtl:
Was there a definitive follow-up/fix to this issue?

PR #3209, opened 2016-03-19
Propogate given configuration to plugin servers

nathanleclaire:
Closes https://github.com/docker/machine/issues/3181

cc @dgageot for review -- this one is a bit tricky and we need to define what our expectations are around API breakage / versioning.  Because this would introduce a call to a RPC endpoint which won't be present in all cases.  I'm open to other ideas if you have them of course.

I guess my feeling is that in this case, we are creating a brand new method, and if we try to call it but it's not there, nothing is different than it was before (when the plugin servers weren't getting the propagated config information anyway) and we can keep moving forward.  It doesn't break existing plugins (shouldn't at least), just adds a new method. 

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -13,7 +13,7 @@ import (
 	"github.com/docker/machine/libmachine/host"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/mcnerror"
-	"github.com/docker/machine/libmachine/mcnutils"
+	"github.com/docker/machine/libmachine/opt"
 	"github.com/docker/machine/libmachine/persist"
 	"github.com/docker/machine/libmachine/ssh"
 )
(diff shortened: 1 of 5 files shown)


nathanleclaire:
What do you think best way to unit test this is?  Through `runCommand`?  `RegisterPlugin`?

nathanleclaire:
Heh... the 3rd party plugin compatibility test I was adamant about is working as intended :stuck_out_tongue: wonder if I should submit a change to the upstream repo as well

nathanleclaire:
WDYT @dgageot ?  Should I add to https://github.com/docker/docker-machine-driver-ci-test/blob/master/driver.go ?  Or should we adopt a different approach?

dgageot:
@nathanleclaire adding a function that every driver should implement is really cumbersome, right?

nathanleclaire:
> @nathanleclaire adding a function that every driver should implement is really cumbersome, right?

Yes, in this case however it is not needed to make an update to the `Driver` itself.  The problem is around adding a new method to the `RPCServerDriver` API (which mostly wraps a Driver, but also has some other methods on top such as `GetVersion` for API management layer).  If you were a 3rd party plugin maintainer you'd simply have to re-compile with the latest Machine master and do a new release of your binary.

Still, I'm not sure it's the best approach exactly, and might end up throwing in a different, hackier (but more compatible) solution for the release.

claygriffiths:
What's the status of this? 

Using `--native-ssh` would most likely help for issues such as:

* https://github.com/docker/machine/issues/4305
* https://github.com/docker/machine/issues/3955
* https://github.com/docker/machine/issues/2737

PR #3215, opened 2016-03-21
Add --rackspace-active-timeout argument to rackspace driver

sivel:
This PR adds a new argument `--rackspace-active-timeout` to the `rackspace` driver to match that of the `openstack` driver.

Additionally, I have updated the description for `rackspace-image-id` to reflect the correct `Ubuntu 15.10` default image.  The image ID that is the default is documented and validated as being 15.10, the description however mentioned that it was 14.04.


nathanleclaire:
cc @ggiamarchi what do you think?

nathanleclaire:
LGTM

Issue #3166, opened 2016-03-08
bootlocal.sh gets munged when machine is restarted following creation in /var/lib/boot2docker


theotherian:
Just spent a good couple hours scratching my head on this until I realized what I was doing different between two machines.

Mac OSX 10.11.3
docker-machine version 0.6.0, build e27fb87
Docker version 1.10.1, build 9e83765

Steps to reproduce:
1) create a new machine (in my case I was using `$ docker-machine create -d virtualbox --virtualbox-memory="4096" --virtualbox-cpu-count="2" --virtualbox-disk-size="10000" --virtualbox-host-dns-resolver=true MyMachine`)
1a) this likely doesn't matter, but in the interest of sharing everything I did in case it's of relevance:
`$ VBoxManage sharedfolder add MyMachine --name my-data --hostpath /private/var/lib/my-data/ --automount`
`$ VBoxManage sharedfolder add MyMachine --name my-logs --hostpath /private/var/log/my-logs/ --automount`
2) `$ docker-machine ssh MyMachine`
3) create `/var/lib/boot2docker/bootlocal.sh`. probably doesn't matter what you have here, but for reference I had:
```
#! /bin/bash
 
mkdir -p /var/lib/my-data
mkdir -p /var/log/my-logs
sudo mount -t vboxsf -o uid=999,gid=999 my-data /var/lib/my-data
sudo mount -t vboxsf -o uid=999,gid=999 my-logs /var/log/my-logs
```
4) `$ docker-machine restart MyMachine`

After doing this, the `bootlocal.sh` file still retains the same number of bytes per `ls -lrth`, but is just gibberish when opened. Also, none of the commands executed on the VM when it started.

If I instead run `$ docker-machine stop MyMachine` and `$ docker-machine start MyMachine`, the file is preserved.

nathanleclaire:
cc @tianon FYI

tianon:
Interesting -- is `docker-machine restart` doing some kind of "hard" reboot instead of graceful?  Do you see the same gibberish if you do a `sync` inside the VM after writing the file contents?

theotherian:
I may have time to try that today @tianon; I haven't tried doing that just yet.

nathanleclaire:
I'm going to investigate this right now to see if I can dupe

nathanleclaire:
Oh yeah, I can definitely dupe.  `/var/lib/boot2docker/bootlocal.sh` is toast after a reboot.

nathanleclaire:
[...]

PR #3221, opened 2016-03-22
Prevent /var/lib/boot2docker/bootlocal.sh munging on restart

nathanleclaire:
cc @dgageot @tianon FYI

Closes https://github.com/docker/machine/issues/3166

Seems the `reset` is problematic.  We could maybe look more into why, but this fixes the problem for now.


nathanleclaire:
LGTM

PR #3233, opened 2016-03-24
Silences docker-machine env command on Windows.

### LINKED ISSUE #3232: env command should be silent on Windows
When I run `docker-machine env ...`, I get this suggestion:

```
REM Run this command to configure your shell:
REM     FOR /f "tokens=*" %i IN ('docker-machine env rdvbox') DO %i
```

This works, but it echoes each command including the two comments, which is slightly annoying.

On OSX (and Git Bash for Windows), the suggestion is to run this command:

```
eval $(docker-machine env rdvbox)
```

This one is nicer because it's silent.

If we change the suggestion to run this command in a command prompt, we'll get the same behavior as we get in Git Bash for Windows and OSX:

```
@FOR /f "tokens=*" %i IN ('docker-machine env rdvbox') DO @%i
```

AndreLouisCaron:
Fixes #3232.


StefanScherer:
Nice catch @AndreLouisCaron. Haven't thought about that once adding the cmd support. LGTM.

nathanleclaire:
LGTM

PR #3243, opened 2016-03-28
FIX #3141 Refactor netstat command and tests

### LINKED ISSUE #3141: Better way to detect if docker host is running
Hello

I'm running into a problem using the generic driver (I believe this applies to other drivers too). docker-machine fails to detect that my docker host is running because it checks the output of ``netstat -an``, which is locale dependent. I'm currently writing from Brazil, and sometimes the default linux distro language pack is downloaded according to the Locale.

I know that this is an easy fix on my part (just re-install ubuntu using english language pack), but I believe that docker-machine can use a more robust way to tell if the docker host is running or not.

I've located the piece of code which is responsible for this error, it happens that the "LISTEN" parameter is hardcoded on the regex match:

https://github.com/docker/machine/blob/8046e13683f6c40cfa55ae106bba5f4f4d87f7ea/libmachine/provision/utils.go#L253

Since the docker host port is until now hardcoded, can we run the regex based on the port it's listening on, instead of the status?

Thanks

tpires:
Updated `netstat` command to minimize its output with `-t` and `-l`
flags.
Refactored regex on `libmachine/provision/utils.go`:`checkDaemonUp` to
avoid `matchNetstatOut` returning `true` when remote machine has a
similar port to dockerPort listening (e.g. dockerPort := 2376 and
remote machine has port 23760 listening).
Refactored `utils_test.go` with more accurate `netstat` output.

Signed-off-by: Tiago Pires <tandrepires@gmail.com>

### DIFF
--- a/libmachine/provision/utils.go
+++ b/libmachine/provision/utils.go
@@ -250,10 +250,10 @@ func getFilesystemType(p Provisioner, directory string) (string, error) {
 }
 
 func checkDaemonUp(p Provisioner, dockerPort int) func() bool {
-	reDaemonListening := fmt.Sprintf(":%d.*LISTEN", dockerPort)
+	reDaemonListening := fmt.Sprintf(":%d\\s+.*:.*", dockerPort)
 	return func() bool {
 		// HACK: Check netstat's output to see if anyone's listening on the Docker API port.
-		netstatOut, err := p.SSHCommand("netstat -an")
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Nice!

nathanleclaire:
LGTM

nathanleclaire:
Thanks for putting this together @tpires !

PR #3248, opened 2016-03-29
adding hypervisor check to virtualbox pre-create check

theodthompson:
Based on feedback from https://github.com/docker/docker/pull/21042. 

The Hyper-V hypervisor and VirtualBox are currently incompatible. However, Hyper-V can be installed without running the hypervisor.

Added a check to see if the hypervisor is running. If it is, we should point people to use Hyper-V as a driver, or disable the hypervisor if they want to use VirtualBox.

### DIFF
--- a/drivers/virtualbox/virtualbox_windows.go
+++ b/drivers/virtualbox/virtualbox_windows.go
@@ -95,6 +95,25 @@ func getShareDriveAndName() (string, string) {
 }
 
 func isHyperVInstalled() bool {
+	// check if hyper-v is installed
 	_, err := exec.LookPath("vmms.exe")
-	return err == nil
+	if err != nil {
+		errmsg := "Hyper-V is not installed."
+		log.Debugf(errmsg, err)
+		return false
+	}
+
+	// check to see if a hypervisor is present. if hyper-v is installed and enabled,
+	// display an error explaining the incompatibility between virtualbox and hyper-v.
+	output, err := cmdOutput("wmic", "computersystem", "get", "hypervisorpresent")
+
+	if err != nil {
+		errmsg := "Could not check to see if Hyper-V is running."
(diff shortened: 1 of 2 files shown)


drivers/virtualbox/virtualbox_windows.go
```diff
@@ -95,6 +95,22 @@ func getShareDriveAndName() (string, string) {
...
+
+		if err != nil {
+			errmsg := "Could not check to see if Hyper-V is running."
+			log.Debugf(errmsg, err)
+			return false
+		}
+
+		enabled := strings.Contains(output, "TRUE")
+		return enabled
+	} else{
```
dgageot: @theodthompson  You don't need the `else` here. And it's better to deal with errors first with a guard.

drivers/virtualbox/virtualbox_windows.go
```diff
@@ -95,6 +95,25 @@ func getShareDriveAndName() (string, string) {
...
+		errmsg := "Hyper-V is not installed."
+		log.Debugf(errmsg, err)
+		return false
+	}
+
+	// check to see if a hypervisor is present. if hyper-v is installed and enabled,
+	// display an error explaining the incompatibility between virtualbox and hyper-v.
+	output, err := cmdOutput("wmic", "computersystem", "get", "hypervisorpresent")
+
+	if err != nil {
```
theodthompson: @dgageot is this what you had in mind?

nathanleclaire:
@dgageot, does this LGTY?  I wasn't involved in the original discussion

dgageot:
@nathanleclaire please don't merge. We've seen issues with this command on docker for Windows. I'll check tomorrow and come back to you

nathanleclaire:
Thanks David, will stay tuned.

dgageot:
LGTM

PR #3252, opened 2016-03-30
driver/azure: Performance improvements

ahmetalpbalkan:
Making Azure operation polling intervals more frequent to
detect completions faster (and thus making machine creations
faster).

Reduce one Azure API call for each machine command by caching
the discovered Azure Active Directory tenant ID of the user’s
Azure subscription into a local file. This also makes Azure
driver commands faster.

Removing the hacky polling around `azureutil.go#createStorageAccount`
by fixing it in Azure Go SDK.

Signed-off-by: Ahmet Alp Balkan <ahmetalpbalkan@gmail.com>

### DIFF
--- a/drivers/azure/azureutil/clients.go
+++ b/drivers/azure/azureutil/clients.go
@@ -2,6 +2,7 @@ package azureutil
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/docker/machine/version"
 
@@ -31,6 +32,7 @@ func subscriptionsClient(baseURI string) subscriptions.Client {
 	c.Client.UserAgent += fmt.Sprintf(";docker-machine/%s", version.Version)
 	c.RequestInspector = withInspection()
 	c.ResponseInspector = byInspecting()
+	c.PollingDelay = time.Second * 5
 	return c
 }
 
@@ -40,6 +42,7 @@ func (a AzureClient) providersClient() resources.ProvidersClient {
 	c.Client.UserAgent += fmt.Sprintf(";docker-machine/%s", version.Version)
 	c.RequestInspector = withInspection()
 	c.ResponseInspector = byInspecting()
+	c.PollingDelay = time.Second * 5
 	return c
 }
 
@@ -49,6 +52,7 @@ func (a AzureClient) resourceGroupsClient() resources.GroupsClient {
 	c.Client.UserAgent += fmt.Sprintf(";docker-machine/%s", version.Version)
 	c.RequestInspector = withInspection()
 	c.ResponseInspector = byInspecting()
+	c.PollingDelay = time.Second * 5
 	return c
 }
 
@@ -58,6 +62,7 @@ func (a AzureClient) securityGroupsClient() network.SecurityGroupsClient {
 	c.Client.UserAgent += fmt.Sprintf(";docker-machine/%s", version.Version)
 	c.RequestInspector = withInspection()
 	c.ResponseInspector = byInspecting()
+	c.PollingDelay = time.Second * 5
 	return c
 }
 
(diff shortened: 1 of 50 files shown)


drivers/azure/azureutil/tenantid.go
```diff
@@ -2,12 +2,55 @@ package azureutil
...
+func loadOrFindTenantID(env azure.Environment, subscriptionID string) (string, error) {
+	var tenantID string
+
+	// Load from cache
+	fp := tenantIDPath(subscriptionID)
+	b, err := ioutil.ReadFile(fp)
+	if err == nil {
+		tenantID = strings.TrimSpace(string(b))
+		log.Debugf("Tenant ID loaded from file: %s", fp)
+	} else {
```
dgageot: Could this be an `else if`?

drivers/azure/azureutil/tenantid.go
```diff
@@ -41,3 +84,7 @@ func findTenantID(env azure.Environment, subscriptionID string) (string, error)
 	}
 	return m[1], nil
 }
+
+func saveTenantID(path string, tenantID string) error {
+	return ioutil.WriteFile(path, []byte(tenantID), 0600)
```
dgageot: Can we have multiple go routines writing at the same time the same file? Can this be problematic?

ahmetalpbalkan: makes sense, I forgot about this as our `saveToken` handles this already. I'll turn this into write to tempfile and move to here.

drivers/azure/azureutil/tenantid.go
ahmetalpbalkan: fair point

drivers/azure/azureutil/tenantid.go
dgageot: Be careful, moving/renaming doesn't behave the same on linux/osx and windows.

dgageot:
Other than the nitpicks, it LGTM!

Nice speedup. Creating an second azure machine:
Driver on master: 6m3
With this PR: 4m23

ahmetalpbalkan:
@dgageot added another commit refactoring if/else and updating `saveTenantID` for atomic writes.

ahmetalpbalkan:
@dgageot I've seen some 4m1s before. I know we can do better by potentially parallelizing creation of various independent resources required for the machines but this should be fair enough. :)

ahmetalpbalkan:
@dgageot btw about your comment:

> Be careful, moving/renaming doesn't behave the same on linux/osx and windows.

what do you particularly refer to? We've been using it in storing azure tokens in docker-machine already and it worked fine on Windows and other OSes as far as I can tell.

nathanleclaire:
Nice!  Looks fun.

drivers/azure/azureutil/clients.go
```diff
@@ -49,6 +52,7 @@ func (a AzureClient) resourceGroupsClient() resources.GroupsClient {
 	c.Client.UserAgent += fmt.Sprintf(";docker-machine/%s", version.Version)
 	c.RequestInspector = withInspection()
```
ahmetalpbalkan: @nathanleclaire  that could actually work, though as my note indicates, we're planning to make this stuff part of the SDK sometime soon via a factory pattern or something. So things will get better. I may actually submit a separate PR for that.

ahmetalpbalkan: @nathanleclaire hmm I think the common base on all clients is `autorest.Client`, if we were to pass it as reference, that trick might do it, let's merge this and I'll take a stab at it next.

nathanleclaire:
@ahmetalpbalkan Sure, sounds reasonable to address as a followup.

LGTM

PR #3258, opened 2016-03-31
Added support for local_ssd and metadata (from files) to the google driver

cpopescu:
- add extra local ssd disks (number provided in flag, also the interface)
- added support for extra metadata (the interesting one to use is startup_script, which can be quite long, so while the key is provided in the flag, the value is read from a file)  
Thanks for looking over this !

cp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: bash
 services: docker
 env:
   matrix:
-    - TARGET_OS=linux TARGET_ARCH=amd64 TARGETS="build validate coverage-send"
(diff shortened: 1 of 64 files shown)


nathanleclaire:
Looks pretty good, we are in a code freeze window right now (no new features allowed), but after the release I will take an in-depth look (and most likely merge).

cpopescu:
Cool - thx - cp

cpopescu:
I think I messed up my repo - will close this one open a new

Issue #3257, opened 2016-03-31
Typo in Digital Ocean example


moxiegirl:
User reports typo in Digital Ocean:

Hello, unless I am mistaking myself, there's a typo on the following page:
https://docs.docker.com/machine/examples/ocean/

The new docker-sandbox machine is running, and it is the active host as indicated by the asterisk (*). When you create a new machine, your command shell automatically connects it. If for some reason your new machine is not the active host, you’ll need to run docker-machine env aws-sandbox <- (we aren't talking about aws-sandbox here, are we..?) , followed by eval $(docker-machine env docker-sandbox) to connect to it.

londoncalling:
@moxiegirl thanks, this is fixed by PR #3259

moxiegirl:
Thanks.

PR #3259, opened 2016-03-31
fixed typos which were referencing the wrong example machines

londoncalling:
Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>


londoncalling:
This fixes issue #3257 reported by Kevin Carr <kevin@evolition.io> via @moxiegirl .

londoncalling:
@moxiegirl can you review and let me know if it LGTYou and/or merge when build succeeds?

moxiegirl:
Hey, I see your corrections.  I thought you weren't going to use "machine" in a generic sense? Catch me up when you have a chance.

londoncalling:
Hi,
I was hoping first that the intro would clear up some of the confusion, and
the goal was to avoid using "machine" frequently in a generic sense.

On the other hand, I didn't want to totally avoid it in cases where that
would be forced and  awkward. I skimmed the examples where I fixed the
typos ... were you seeing inappropriate use of "machine" in those docs? Can
you show me where?

Thanks, Vicky

Issue #3250, opened 2016-03-30
New azure driver prints out "Creating subnet..." when specifying an existing subnet.


tmds:
I am using the 0.7.0-rc1 to create machines in azure. When an existing network is passed as an option, the command prints out "Virtual network already exists.".. For the subnet, it always prints a message "Creating subnet..." regardless of the existence of the subnet. For consistency it would be better to print a similar message when the subnet exists: "Subnet already exists".

nathanleclaire:
cc @ahmetalpbalkan

ahmetalpbalkan:
@tmds that's because we check existence of virtual network but we do a CreateOrUpdate operation for subnet without checking its existence, so the message is accurate. This is to reduce number of HTTP requests we make. Also CreateOrUpdate call to subnet can extend the specified CIDR range, but we can't update virtual network, so it’s by design.

Does this impact the desired behavior? I  assume it works just fine?

tmds:
Yes, it works fine. It is just a a bit confusing.

What I expected the command to do: create a subnet if it did not exist. In case it existed and a subnet range is specified, check whether they are the same. If they aren't, give an error message and stop.

I can't tell whether these expectations make sense to most of the users and I don't know if they are consistent with the behavior of other drivers.

Since the command takes a while to execute, you don't want to run it multiple time to figure out the arguments.

ahmetalpbalkan:
@tmds If the subnet exists, we still issue a HTTP request but as long as the range is the same, API doesn't do anything about it, if not, it gives an error just as you said. And if the subnet does not exist, the same request is used in creating the subnet. The Azure APIs don't have plain "Create", we have "CreateOrUpdate" mostly.

I think the stuff that could fail pretty much executes within the first 10-20 secs so it should be reasonable. Also, I think the upsert logic we have is pretty much alright. If you run into issues please let me know.

tmds:
As a suggestion: maybe the text can change to "Configuring subnet".
For my information, do you have a reference that describes the behavior you've explained?
Thanks!

ahmetalpbalkan:
@tmds calling it "Configuring" is certainly a good idea, I'll submit a PR.

PR #3269, opened 2016-04-04
azure: more accurate messages for resources created

ahmetalpbalkan:
Addresses #3250. We now print 'Configuring' instead of 'Creating' for upsert operations.


nathanleclaire:
LGTM

nathanleclaire:
:+1:

PR #3275, opened 2016-04-05
Fix RC version mismatch bug

nathanleclaire:
The new code I implemented doesn't match the magic ISO string correctly for RCs, resulting in a double-download of the ISO on every create.  This should fix the issue by allowing a proper match.

### DIFF
--- a/libmachine/mcnutils/b2d.go
+++ b/libmachine/mcnutils/b2d.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/version"
@@ -23,6 +24,7 @@ const (
 	defaultURL            = "https://api.github.com/repos/boot2docker/boot2docker/releases"
 	defaultISOFilename    = "boot2docker.iso"
 	defaultVolumeIDOffset = int64(0x8028)
+	versionPrefix         = "-v"
 	defaultVolumeIDLength = 32
 )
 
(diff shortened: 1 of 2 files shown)


libmachine/mcnutils/b2d.go
```diff
@@ -304,10 +306,12 @@ func (b *b2dISO) version() (string, error) {
 		return "", err
 	}
 
-	verRegex := regexp.MustCompile(`v\d+\.\d+\.\d+`)
-	ver := string(verRegex.Find(isoMetadata))
-	log.Debug("local Boot2Docker ISO version: ", ver)
-	return ver, nil
+	// Original magic file string looks similar to this: "Boot2Docker-v0.1.0              "
+	// This will return "v0.1.0" given the above string
+	vers := strings.TrimSpace(string(isoMetadata))[prefixOffset:]
```
tianon: What happens if the ISO is a freak and we don't have at least twelve characters in this metadata?

tianon: Would IndexOf "-v" or something be more resilient?

nathanleclaire: > Would IndexOf "-v" or something be more resilient?

Yeah, actually I like that, I had considered expanding the existing regex but I think I like that better.

nathanleclaire:
LGTM

libmachine/mcnutils/b2d.go
```diff
@@ -304,10 +306,19 @@ func (b *b2dISO) version() (string, error) {
...
+	fullVersion := string(isoMetadata)
+
+	versionIndex := strings.Index(fullVersion, versionPrefix)
+	if versionIndex == -1 {
+		return "", fmt.Errorf("Did not find prefix %q in version string", versionPrefix)
+	}
+
+	// Original magic file string looks similar to this: "Boot2Docker-v0.1.0              "
+	// This will return "v0.1.0" given the above string
+	vers := strings.TrimSpace(fullVersion)[versionIndex+1:]
```
tianon: If you come back to this code, it might be a good idea to move the `[]` inside the `()` here to make sure we don't get messed up by leading whitespace in the future. :+1:

(ie, `fullVersion = " something-v1.0.0 "` would lead to `1.0.0` instead of `v1.0.0`)

nathanleclaire: Yeah, that's a good point.

PR #3281, opened 2016-04-06
Several additions in support of google cloud engine:

cpopescu:
- added support for specifying machine metadata via a --google-metadata `<key>=<filename>[,<key>=<filename>]` flag (interesting here are the startup and shutdown scripts.
- added support for local-ssd (number of disks and interface type flags with --google-local-ssd-num and --google-local-ssd-interface)
-- added upstart based provisioner for debian + added a base upstart provisioner on the model of systemd.go. The differentiation between main debian provisioner (which is systemd based) and the debian/upstart is done in CompatibleWithHost based on the presence of systemctl
Usage case for the provisioner: this seems to be the best way to deal with some google compute engine machines (debian based w/ nvme local ssd) for which conversion to systemd is too cumbersome.

Signed-off-by: Catalin Popescu <catalin@gmail.com>

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -261,6 +261,44 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 		instance.NetworkInterfaces[0].AccessConfigs = append(instance.NetworkInterfaces[0].AccessConfigs, cfg)
 	}
 
+	if d.LocalSsdNum > 0 {
+		for i := 0; i < d.LocalSsdNum; i++ {
+			localSsd := &raw.AttachedDisk{
+				AutoDelete: true,      // Cannot create local SSD without autoDelete
+				Type:       "SCRATCH", // Again, necessary for local SSD
+				Mode:       "READ_WRITE",
+				Interface:  d.LocalSsdInterface,
+				InitializeParams: &raw.AttachedDiskInitializeParams{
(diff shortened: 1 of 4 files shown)


cpopescu:
Had another pull request with just a part of these, wanted to add more but messed some merge at sign-off phase, so created a new pull request - cleaner. thx.

nathanleclaire:
How common is it for default Debian images to use upstart?  What's the motivation for that addition?

cpopescu:
For example the default debian images w/ Google compute engine come w/ upstart (e.g. nvme-backports-debian-7-wheezy-xxxx for using local ssds w/ nvme) ..

nathanleclaire:
Do we not default to Debian Jessie on GCE?  It should be.

cpopescu:
But if you want to use the fast local disks on gce (w/ nvme) you have to use the google packed wheezy images: https://cloud.google.com/compute/docs/disks/local-ssd
Probably support will come in some time (dunno when)

cpopescu:
Any news on this ? thx

nathanleclaire:
Is it needed to have the Debian/upstart provisioner to support the additional GCE parameters?  I really would prefer to separate that into another PR if possible.

cpopescu:
Ok - will do something
thx

nathanleclaire:
Hi @cpopescu, going to close this due to lack of follow-up, feel free to open a new PR and ping me if you'd like to continue with the path we discussed

PR #3287, opened 2016-04-08
Add support for --swarm-opt to agent nodes

### LINKED ISSUE #3285: docker-machine ignores --swarm-opt for swarm agent
docker-machine seems to take into account the `--swarm-opt` arbitrary options only if `--swarm-master` is specified.

This blocks me when trying to configure a swarm agent with TLS discovery.

Full command line :
```
docker-machine create --driver generic --swarm \
--swarm-opt="discovery-opt=\"kv.keyfile=/etc/docker/server-key.pem\"" \
--swarm-opt="discovery-opt=\"kv.certfile=/etc/docker/server.pem\"" \
--swarm-opt="discovery-opt=\"kv.cacertfile=/etc/docker/ca.pem\"" \
--swarm-discovery consul://discovery:8500  \
--engine-opt="cluster-store=consul://discovery:8500" \
--engine-opt="cluster-advertise=myip:2376" \
--engine-opt="cluster-store-opt=\"kv.keyfile=/etc/docker/server-key.pem\"" \
--engine-opt="cluster-store-opt=\"kv.certfile=/etc/docker/server.pem\"" \
--engine-opt="cluster-store-opt=\"kv.cacertfile=/etc/docker/ca.pem\"" \
--engine-label "type=node" \
--generic-ip-address  myip   mymachine
```

With this configuration, the engine manages to connect to the consul through TLS but the swarm agent tries to connect without TLS.
Result of `docker inspect swarm-agent` :

```
"Args": [
            "join",
            "--advertise",
            "195.154.47.76:2376",
            "consul://discovery:8500"
        ]
```

`docker-machine --version `
`docker-machine version 0.6.0, build e27fb87`

LBaali:
Closes #3285 

This modification permits the swarm-agent created by docker-machine to take into account the `—swarm-opt `arguments.

Signed-off-by: Lucien Gougerot lucien.gougerot@gmail.com

### DIFF
--- a/libmachine/provision/configure_swarm.go
+++ b/libmachine/provision/configure_swarm.go
@@ -119,15 +119,20 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
 		},
 	}
 
+	cmdWorker := []string{
+		"join",
+		"--advertise",
+		advertiseInfo,
+	}
+	for _, option := range swarmOptions.ArbitraryFlags {
+		cmdWorker = append(cmdWorker, "--"+option)
+	}
+	cmdWorker = append(cmdWorker, swarmOptions.Discovery)
+
 	swarmWorkerConfig := &dockerclient.ContainerConfig{
-		Image: swarmOptions.Image,
-		Env:   swarmOptions.Env,
-		Cmd: []string{
-			"join",
-			"--advertise",
-			advertiseInfo,
-			swarmOptions.Discovery,
-		},
+		Image:      swarmOptions.Image,
+		Env:        swarmOptions.Env,
+		Cmd:        cmdWorker,
 		HostConfig: workerHostConfig,
 	}
 	if swarmOptions.IsExperimental {


nathanleclaire:
Looks pretty good -- can't merge right now as we're in a code freeze window before the 0.7.0 release, but will take a closer look after that's over.  If you're feeling brave you might want to take a whack at unit testing this via `configureSwarm` or by splitting some of that code off into other sub-routines.

nathanleclaire:
Needs a rebase, but code looks good.

libmachine/provision/configure_swarm.go
```diff
@@ -117,17 +118,23 @@ func configureSwarm(p Provisioner, swarmOptions swarm.Options, authOptions auth.
 			Name:              "always",
 			MaximumRetryCount: 0,
 		},
+		Binds: []string{hostBind},
```
nathanleclaire: Why add this?  The `join`ed nodes don't need to access the certs IIRC.

LBaali: In my case, the nodes need to access the certs for authenticating with the consul. I either need to bind them like this or be able to specify a volume option (-v /mycerts:...) and this seemed like the easiest way (I also understand that this isn't the best approach for a long term solution).

nathanleclaire: I see... Seems that this is out of the scope of the original feature request (supporting `--swarm-opt` for join nodes), how do you feel about making a separate proposal and PR for that?

LBaali: Ok, i'll remove this from this PR and open up a new issue, thanks

nathanleclaire:
LGTM

ndeloof:
How can one pass options to master _and_ agent ? 
For sample, I'd like to pass `--heartbeat` and `--ttl` to agent, but those options will be rejected by the `manage` command. It would be ok if they are just ignored, but this actually makes master to fail.

So I guess we'd need `--swarm-agent-opt` (and maybe rename `--swarm-opt` to `--swarm-master-opt` for clarity), wdyt @nathanleclaire ?

ndeloof:
@nathanleclaire IIUC with this change, I can't use `--swarm-opt` anymore with options that are not supported by `swarm join` command, or the later will fail to start. 

I think we need some way to distinct options passed to agent and master. I'd really like to use `--swarm-opt` and `--swarm-master-opt` to match https://github.com/docker/machine/pull/3365 but this would then change this option meaning :'(. I'd be happy to prepare a PR for this, please let me know your opinion on this.

nathanleclaire:
Oof, you have a good point @ndeloof.  Maybe we should submit a followup to change the use of `--swarm-opt` here to `--swarm-join-opt` (that should preserve BC, and also kind of makes sense, right? I think of the manager options as "for the whole swarm").

ndeloof:
@nathanleclaire ok will do.

LBaali:
Ok, so should this be another PR or will it be added by @ndeloof  to #3365 ? Wouldn't it make more sense to deprecate

ndeloof:
I'll do a separate PR to fix this, #3365 is another topic to be discussed on it's own.

nathanleclaire:
👍

ndeloof:
see https://github.com/docker/machine/pull/3387

Issue #3292, opened 2016-04-11
Windows 10: blocking ssh keys management (docker-machine 0.7.0-rc4)


VonC:
docker-machine 0.7.0-rc4, docker 11-rc3 on Windows 10 with VirtualBox 5.0.14 or 5.0.16

I can no longer list machines with docker-machine:

````
C:\prgs\dm>dm ls
NAME       ACTIVE   DRIVER       STATE     URL   SWARM   DOCKER    ERRORS
test       *        virtualbox   Running                 Unknown   Permissions 0666 for 'C:\Users\vonc\.docker\machine\machines\blessed\id_rsa' are too open.
default    *        virtualbox   Running                 Unknown   Permissions 0666 for 'C:\Users\vonc\.docker\machine\machines\default\id_rsa' are too open.
````

That is because of https://github.com/docker/machine/blame/5b9663f8960af73c048e14cc60566ae487c07728/libmachine/ssh/client.go#L335, commit https://github.com/docker/machine/commit/f2acfa9492d9352edeb722ba625adaf86bcdee7e

It uses `os.Stat()` (https://golang.org/src/os/stat_windows.go) which wraps a calls to https://msdn.microsoft.com/en-us/library/windows/desktop/aa364946%28v=vs.85%29.aspx

That maps poorly with POSIX permission, and setting a Windows file to 600 (instead of the default 644) is difficult/impossible (I tried http://stackoverflow.com/q/5264595/6309 or http://superuser.com/a/318176/141)

Note: openssh (from git-for-windows or for cygwin) does work even if 'ls' (also packaged with git-for-windows) sees the certificates as 644, not 600: I can ssh with ssh.exe, not with docker-machine 0.7.0-rc4 (I can with docker-machine 0.6.0 obviously, since commit https://github.com/docker/machine/commit/f2acfa9492d9352edeb722ba625adaf86bcdee7e was not there)

Even if I recompile docker-machine by removing that test, docker-machine ls still does not work (but any ssh access/copy does work). It will list the machines with the warning/error `certificate signed by unknown authority` message. (I'll open another issue for that: regenerating the certs does not change the SAN in the tls/server.pem)

nathanleclaire:
I've seen this issue as well.  Should we set the permissions to 644 for Windows?

VonC:
@nathanleclaire I am not sure you can set any specific protection mode on Windows. I just removed that test (for Windows only) in my compiled version.

nathanleclaire:
[...]

PR #3294, opened 2016-04-11
Revert "Correct error handling for external SSH client"

nathanleclaire:
This reverts commit f2acfa9492d9352edeb722ba625adaf86bcdee7e due to https://github.com/docker/machine/issues/3292 breaking Windows machines.

I'm happy to add it back in later once the regression is fixed, but we're getting down to the wire on the release, so I'm erring on the side of caution.


nathanleclaire:
Urgh, it seems signed just fine, but Gordon doesn't seem to realize.

Issue #3310, opened 2016-04-14
Generic driver: `docker-machine create` stalls when using SSH ProxyCommand


brettdh:
I have this setting in my `~/.ssh/config` for accessing a remote site:

```
Host <remote-hostname>
  ProxyCommand corkscrew <proxy> 3128 %h %p ~/.ssh/.corkscrew-auth
```

And when I run `docker-machine -D create remote --driver generic --generic-ip-address <remote-hostname> ...`, it gets stuck thusly:

```
Host does not exist: "remote"
Docker Machine Version:  0.6.0, build e27fb87
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver generic
Plugin server listening at address 127.0.0.1:59640
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(flag-lookup) Calling .GetMachineName
(flag-lookup) Calling .DriverName
(flag-lookup) Calling .GetCreateFlags
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver generic
Plugin server listening at address 127.0.0.1:59644
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(remote) Calling .GetMachineName
(remote) Calling .DriverName
(remote) Calling .GetCreateFlags
(remote) Calling .SetConfigFromFlags
Running pre-create checks...
(remote) Calling .PreCreateCheck
(remote) Calling .GetConfigRaw
Creating machine...
(remote) Calling .Create
(remote) Calling .GetConfigRaw
(remote) No SSH key specified. Connecting to this machine now and in the future will require the ssh agent to contain the appropriate key.
(remote) DBG | IP: remote-hostname
(remote) Calling .DriverName
(remote) Calling .DriverName
Waiting for machine to be running, this may take a few minutes...
(remote) Calling .GetState
(remote) Calling .GetState
(remote) Calling .GetState
(remote) Calling .GetState
(remote) Calling .GetState
<snip>
(remote) Calling .GetState
(remote) Calling .GetState
Error creating machine: Error waiting for machine to be running: Maximum number of retries (60) exceeded
Opting out of crash reporting.
```

Looking at [.GetState for the generic driver][GetState], it appears to be trying to connect directly to the SSH port, which doesn't respect proxy settings.

I'm not sure if there are other things that would prevent this configuration from working, but this is my current roadblock.

[...]

PR #3319, opened 2016-04-15
Generic driver: flags to provide SSH config file

### LINKED ISSUE #3318: enhancement: add flag to provide SSH config file for generic driver
Follow-on from #3310

> @nathanleclaire 
> Machine ignores ~/.ssh/config. This is deliberate, as existing users' SSH configs were causing all sorts of mysterious and unintended side effects in Machineland.

> I'd be happy to consider proposals or PRs to specify an SSH config file to use (via flag), though. Seems perfectly fine to make configurable for those who know what they're doing.

This was pretty easy to implement, so I just did it. :-) Posting a PR shortly, after I test it out a bit in my environment.

brettdh:
This adds two flags to the `generic` driver:

    --generic-ssh-config-file <path/to/config>

        Use the provided config file with the external SSH client (-F)

    --generic-use-user-ssh-config

        Use ~/.ssh/config as the config file

The flags are mutually exclusive in order to avoid ambiguity.

This also adds the SSHConfigFile parameter to BaseDriver and a
GetSSHConfigFile method to all drivers. It seems like there should
be a way to reduce this duplication, since most of the implementations
are identical (return "/dev/null"). I'm a golang noob, though, so if
you know of a better way to accomplish this, please let me know.

Tried this out in my environment, which requires a `ProxyCommand` to
access a remote site via SSH. Before, the issue described in #3310
occurred, preventing me from using `docker-machine` to configure this
host. With this, the `--generic-use-user-ssh-config` option adds
the config file in place of the previous-default `/dev/null` and all
seems well.

Also added a couple unit test cases.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -713,6 +713,13 @@ func (d *Driver) GetSSHUsername() string {
 	return d.SSHUser
 }
 
+func (d *Driver) GetSSHConfigFile() string {
+	if d.SSHConfigFile == "" {
+		d.SSHConfigFile = "/dev/null"
+	}
+	return d.SSHConfigFile
+}
+
 func (d *Driver) Start() error {
 	_, err := d.getClient().StartInstances(&ec2.StartInstancesInput{
 		InstanceIds: []*string{&d.InstanceId},
--- a/drivers/errdriver/error.go
+++ b/drivers/errdriver/error.go
@@ -71,6 +71,10 @@ func (d *Driver) GetSSHUsername() string {
 	return ""
 }
 
+func (d *Driver) GetSSHConfigFile() string {
+	return "/dev/null"
+}
+
 func (d *Driver) GetState() (state.State, error) {
 	return state.Error, NotLoadable{d.Name}
 }
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -128,6 +128,13 @@ func (d *Driver) GetSSHUsername() string {
 	return d.SSHUser
 }
 
+func (d *Driver) GetSSHConfigFile() string {
+	if d.SSHConfigFile == "" {
+		d.SSHConfigFile = "/dev/null"
+	}
+	return d.SSHConfigFile
+}
+
 // DriverName returns the name of the driver
 func (d *Driver) DriverName() string {
 	return "exoscale"
--- a/drivers/fakedriver/fakedriver.go
+++ b/drivers/fakedriver/fakedriver.go
@@ -72,6 +72,10 @@ func (d *Driver) GetSSHUsername() string {
 	return ""
 }
 
+func (d *Driver) GetSSHConfigFile() string {
+	return "/dev/null"
+}
+
 func (d *Driver) GetState() (state.State, error) {
 	return d.MockState, nil
 }
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"strconv"
 	"time"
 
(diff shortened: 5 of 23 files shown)


brettdh:
@nathanleclaire I'm noticing that another place in the code still makes a bare TCP connection, skirting around proxy settings: https://github.com/docker/machine/blob/master/libmachine/cert/cert.go#L250

In the generic driver, I got around this by running an SSH command instead of just making a TCP connection. I should be able to do similarly here, just making an HTTP request to the ping endpoint, as I seem to recall that go respects `http_proxy`. What's the appropriate equivalent for an http client in go, that machine prefers to use?

brettdh:
Commented before `grep`ing. `net/http` seems pretty common. Going to test that out.

brettdh:
Augh, keep mistaking what I'm seeing. All the `net/http` usage in machine is in `vendor/`. So now I am unsure again. 😬

nathanleclaire:
> @nathanleclaire I'm noticing that another place in the code still makes a bare TCP connection, skirting around proxy settings: https://github.com/docker/machine/blob/master/libmachine/cert/cert.go#L250

That's to test the connection to the daemon though.  It's testing a TLS/HTTP connection, not SSH.  If you have `HTTP_PROXY` environment variable set for your proxy that should be adequate to route that through there IIRC, although the custom HTTP client might throw things off.

nathanleclaire:
And yes, we rely on `net/http` for requests pretty much ubiquitously (you kind of have to).

nathanleclaire:
Thanks for the pull request -- there are some issues with this implementation that I will outline -- however the basic feature seems acceptable.

brettdh:
@nathanleclaire yeah, I had to use a custom `Transport` with `Proxy: http.ProxyFromEnvironment` and `TLSClientConfig`, as using one with just `TLSClientConfig` overrode the default `Proxy` setting.

There are a couple other places that need to change, which I'm discovering as I finish stepping through my end-to-end test. For instance, I just discovered that I'll need to change the machine-local docker client somehow to use the proxy settings; it doesn't seem to do so by default. Same reason as above: https://github.com/docker/machine/blob/master/vendor/github.com/samalba/dockerclient/utils.go#L12

nathanleclaire:
I don't think the current approach in this PR will work per se, unfortunately.  For starters we can't change the Driver API without bumping the version and/or getting all the plugin authors to re-compile with the new version, which is painful, so we would be a lot better off avoiding that if possible.

I think this work may be blocked by implementing something like [this PR](https://github.com/docker/machine/pull/3209).  Basically we might as well have a universal way to set configuration across the wire without needing to add in new methods every time.

brettdh:
Ah yes - I was surprised to find that adding this to one driver required adding a method to all the drivers, but I don't think I realized I was breaking an API in doing so.

Perhaps a reasonable workaround would be to just add a new function (`utils.GetSSHClientFromDriverWithOptions` or similar) that allows the options override per-driver? That would decouple this feature from the universal-config effort and also isolate it to the generic driver as I originally intended.

brettdh:
Ok, so clearly that doesn't work either - there are a lot of SSH operations that take place outside the driver and simply use data from the driver to set params. (This should have occurred to me up front, but hey, at least it only took a few minutes of trying it out to come to the realization.)

I could imagine a single new method that just returns a `[]string` of args to pass to the ssh command. This would require one version bump but would then isolate future option changes from requiring another bump. But, at that point, if I understand it correctly, you might as well go with something like #3209, as you mentioned. (Took me a bit to understand what was going on there.)

I notice that #3209 hasn't had any discussion for a while, though, and was left a bit unresolved. Where should we go from here?

nathanleclaire:
> I notice that #3209 hasn't had any discussion for a while, though, and was left a bit unresolved. Where should we go from here?

I'd like to rebase / whip that into shape, which could hopefully then make this PR doable by adding an additional configuration option (applicable for all drivers).  Since we've never done an API version bump before, I've been a bit skittish about it, but it looks like we're headed towards wanting to do one anyway, and now's the time since we just released.  I'll keep you posted.

brettdh:
From what I can gather from #3209, I think it's intended to be used for options that are provided on every invocation of `docker-machine`. What about options that could be set once on `create` and then applied to every subsequent operation on that machine? It looks like those options should be put in `HostOptions`, and perhaps looked up via `libmachine.Client.Load` in `commands/commands.go:runCommand` in order to get the additional options to pass to `mcnopt.SetOpts`?

Still just trying to get the lay of the land in an unfamiliar codebase. :smile: Any advice is appreciated; I'm going to take a crack at this privately, and I'll plan to rebase on whatever #3209 ends up becoming when the time comes.

brettdh:
Closing in favor of #3410.

PR #3328, opened 2016-04-17
SSH - don't enable BatchMode (fixes #3306)

### LINKED ISSUE #3306: "exit status 255" on docker-machine ssh (when ssh key not yet in ssh-agent)
Description of problem:

`docker-machine ssh` run after mac startup fails, seemingly because my passphrase-protected SSH key hasn't been added to the auth agent. After the first manual SSH with this key, `docker-machine ssh` succeeds again.

`uname -a`:

`Darwin hobbes.local 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64`

Environment details (AWS, VirtualBox, physical, etc.):

`remote-machine` below uses the `generic` driver.

Steps to Reproduce:
```
# after first login to mac
$ docker-machine ssh remote-machine
exit status 255
$ ssh remote-machine
<dialog pops up, prompts for passphrase for SSH key>
remote-machine $ exit
$ docker-machine ssh remote-machine
remote-machine $
```

Actual Results: above.

Expected Results: the first command should succeed, or at least print a more helpful error message. about the underlying ssh failure.

bamarni:
When a passphrase protected key is not yet in ssh-agent, it gives a
chance to the user to fill a prompt instead of silently failing.

### DIFF
--- a/libmachine/ssh/client.go
+++ b/libmachine/ssh/client.go
@@ -65,7 +65,6 @@ const (
 var (
 	baseSSHArgs = []string{
 		"-F", "/dev/null",
-		"-o", "BatchMode=yes",
 		"-o", "PasswordAuthentication=no",
 		"-o", "StrictHostKeyChecking=no",
 		"-o", "UserKnownHostsFile=/dev/null",


libmachine/ssh/client.go
```diff
@@ -65,7 +65,6 @@ const (
 var (
 	baseSSHArgs = []string{
 		"-F", "/dev/null",
-		"-o", "BatchMode=yes",
 		"-o", "PasswordAuthentication=no",
```
nathanleclaire: Does this line not also prevent password auth?

nathanleclaire:
Have you tested this with `virtualbox` driver?

libmachine/ssh/client.go
bamarni: Yes it does, my intention was only to allow prompts for ssh key passphrases. Do you think password auth should be supported too? I'm not sure about the impacts.

bamarni:
@nathanleclaire : I've tested it with the generic driver. Is there anything special with the virtualbox driver that I should check?

nathanleclaire:
Just wondering if it may have been included to help something there but I'm trying it and it seems to work without issues.

LGTM

ivandir:
Still getting this with v17.03.0-ce and generic driver on Ubuntu 16.04 VM via virtualBox Version 5.1.18 r114002 (Qt5.6.2)

bamarni:
@ivandir : what do you mean? I think the version that matters here is docker-machine's one.

colthreepv:
As @ivandir reports, I am receiving this exit status 255 on windows 10x64, docker-machine v 0.10, trying to connect to a remote "generic" docker host

Of course classic ssh command works

Issue #2907, opened 2016-01-22
Connect to Windows Docker engine


StefanScherer:
Has anybody started to add or think about a docker-machine driver and provisioner for Windows Docker engines?
So anybody can use docker-machine to secure and connect to a Windows Server 2016 TP4 as easy as we are used for a Linux Docker Engine?

Something like:

```
docker-machine create -d generic --generic-winrm-user "administrator" ... dev
docker-machine create -d azure --azure-username "administrator" --azure-winrm-port 5985 ... dev
```

Using WinRM (HTTP/HTTPS) instead of SSH to access the remote Windows Server and put the TLS certs into it and reconfigures the Windows Docker service and other things.
I know that packer has communicators for SSH and WinRM built in Go for provisioning.
Just connecting the dots...

ping @ahmetalpbalkan :smile:

nathanleclaire:
Due to the way the provisioner model currently works, it might not be _too_ crazy to implement such functionality, although it would require re-thinking or papering over some pretty fundamental assumptions (e.g. the presence of SSH).

I have a feeling that the way it will need to be implemented is by having some flags which specifically indicate that it's a Windows machine.  e.g. `--generic-winrm-userpass`.

ahmetalpbalkan:
I don't know much about Windows Server containers, however as Nathan said most of the work probably would be getting the SSH protocol on Windows either by waiting Microsoft to do it (it's happening) or providing a polyfill using WinRM maybe (I don't know how it works at all).

I remember some discussion about this was happening internally too. cc: @jstarks as he is on the windows containers team.

enderb-ms:
Hi @StefanScherer, I'm a PM in the Hyper-V team and I'm actually looking into how we could make this possible. One of the things we're thinking about is maybe we could use [this](http://blogs.technet.com/b/nanoserver/archive/2016/01/07/download-just-nano-server-in-a-vhd-windows-server-2016-technical-preview-4.aspx) as the OS that Docker Machine could pull. If we use Nano, then it would stay as a small download. The repo for the Open SSH in Windows project is [here](https://github.com/PowerShell/Win32-OpenSSH).

Any ideas are appreciated and if you're going to prototype something please let me know. I'll keep posting back here if I find out more about how to make this work.

cc: @dgageot

StefanScherer:
[...]

Issue #3199, opened 2016-03-16
[Azure] Non-interactive way of authenticating with Azure is needed


ppadala:
The new Azure driver uses the device login authentication method, which requires the user to enter a code in a browser and then login with Microsoft OneLogin credentials. Though, this may work for interactive users, scripts or management platforms automating management of Azure VMs using docker-machine need a way to automate this.

@ahmetalpbalkan and I discussed this a little bit at https://github.com/docker/machine/pull/3159#discussion_r55563081. Filing this to track it.

P.S. I also noticed that every time docker-machine is run, the user is asked to enter the code. Can this be avoided by entering only once and caching it for certain number of days (I thought that's what docs suggest)?

ahmetalpbalkan:
@ppadala thanks for filing the issue. We are internally looking for a clean way to provide a credentials file all our CLIs (including docker-machine driver) and SDKs can share. There’s currently no ETA for now, however my intention is to find a solution to this in the short term.

> I also noticed that every time docker-machine is run, the user is asked to enter the code. Can this be avoided by entering only once and caching it for certain number of days

This is actually a bug. I'm currently debugging this with one of users. Normally we refresh the token as it expires (every 1 hour) and we can do this for 15 days, and then user is asked to re-authenticate after that time. So it is a manual operation that ought to happen every 2 weeks.

In your case, are you getting asked to authenticate on every single time you run docker-machine? If so, please open a separate issue, I’d love to debug that.

ppadala:
> In your case, are you getting the authentication prompt every single time you run docker-machine? If so, please open a separate issue, I’d love to debug that.

I am running this on Mac, the browser doesn't open automatically, but I get this message. 

Pradeeps-MacBook-Air-2:machine ppadala$ ./bin/docker-machine -D create -d azure --azure-subscription-id $(cat ~/azure/subid) --azure-os "windows" --azure-image a699494373c04fc0bc8f2bb1389d6106__WindowsServerCore_en-us_TP4_Container_Azure-20151118 vhost-win-test
Docker Machine Version:  0.7.0-dev, build a38276a
Found binary path at ./bin/docker-machine
Launching plugin server for driver azure
Plugin server listening at address 127.0.0.1:57286
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(flag-lookup) Calling .GetMachineName
(flag-lookup) Calling .DriverName
(flag-lookup) Calling .GetCreateFlags
[...]

Issue #3062, opened 2016-02-16
implement support for daemon --config-file option (/etc/docker/daemon.json by default)


dweomer:
This is a proposal to simplify the divergence of daemon options persistence across providers by leveraging the new `--config-file` option. Simply put: consistent `DaemonOptionsFile` persistence into the  location specified by `--config-file`.

# Problem
Prior to Docker 1.10 and the advent of the `--config-file` daemon option, all `--engine-*` options passed to `docker-machine create` are persisted in ways specific to the provisioner (unit files for systemd, rc files for upstart, ... whatever suse seems to be doing). The [`--config-file`](https://docs.docker.com/engine/reference/commandline/daemon/#daemon-configuration-file) option allows for us to generate this file in a consistent way across provisioners.

Additionally, any conflict in daemon CLI options with options found in the location specified by `--config-file` will prevent the daemon from reloading and/or starting. This means that if you want to leverage the new [configuration reloading](https://docs.docker.com/engine/reference/commandline/daemon/#configuration-reloading) capabilities of the daemon for nodes provisioned by `docker-machine` you will need to fiddle with them by hand after creation is complete.

# Proposal
Implement first-class support for persisting all daemon options to this new daemon config file. This would mean supporting a new `--engine-*` option implied by such an entry in `commands/create.go`:
```
cli.StringFlag{
	Name:   "engine-config-file",
	Usage:  "Specify Docker Daemon configuration file",
	EnvVar: "ENGINE_CONFIG_FILE",
},
```
With the default value assumed to be overridden by provisioner implementations. This would simplify the template definitions for generic/upstart, systemd, suse, etc such that they would only container the pointer to the daemon configuration file. *More importantly*, this would also for nodes created via `docker-machine` to be able to leverage the new configuration readling feature.

# Corollary: Signalling and/or Configuration Reloading 
Implement something like `docker-machine signal` to send a `SIGHUP` (by default, overridable via switch) to the daemon to easily capitalize on this new functionality.

# Corrollary: `DockerOptionsDir`
If `--engine-config-file` is specified, should the value of DockerOptionsDir be overridden to be the base path of such?

-
###### Yes, this is a shameless attempt to copy @nathanleclaire's proposal format as embodied by #2516

nathanleclaire:
I think that ideally what should be done is to migrate the existing `--engine-*` flags to use the new config file, and pull as much as possible out of the existing service / upstart configuration files.

[...]

PR #3329, opened 2016-04-17
Support for Windows Docker hosts on Azure

ppadala:
Issue #2907.

This pull request adds support for spinning up Windows 2016 servers on Azure. Brief overview of all the steps

1. Spin up a [Windows 2016 TP4 with containers image](https://azure.microsoft.com/en-us/marketplace/partners/microsoft/windowsserver2016technicalpreviewwithcontainers/). 
2. Enable WinRM using deploy template. WinRM is used to execute commands on the VM, once it is up
3. Stop docker-engine
3. Upload certs and keys using WinRM
4. Start docker-engine

Usage: Three new parameters are added to Azure driver

```
--azure-os "linux"			        OS for the Azure VM (Windows|Linux)
--azure-winrm-user "docker-user"	Username for WinRM login [$AZURE_WINRM_USER]
--azure-winrm-password 		        Password for WinRM login [$AZURE_WINRM_PASSWORD]
```

Sample command

```
docker-machine -D create -d azure --azure-subscription-id $(cat ~/azure/subid) --azure-os "windows" --azure-image MicrosoftWindowsServer:WindowsServer:2016-Technical-Preview-with-Containers:2016.0.20151118 --azure-winrm-password "mysecretpassword" vhost-win-test
```


ppadala:
cc @ahmetalpbalkan and @nathanleclaire for comments

drivers/azure/azure.go
```diff
@@ -35,10 +36,13 @@ const (
 	flAzureSubscriptionID  = "azure-subscription-id"
 	flAzureResourceGroup   = "azure-resource-group"
 	flAzureSSHUser         = "azure-ssh-user"
+	flAzureWinRMUser       = "azure-winrm-user"
+	flAzureWinRMPassword   = "azure-winrm-password"
 	flAzureDockerPort      = "azure-docker-port"
 	flAzureLocation        = "azure-location"
 	flAzureSize            = "azure-size"
 	flAzureImage           = "azure-image"
+	flAzureOS              = "azure-os"
```
ahmetalpbalkan: I'd say let's group flags for windows together it'll make it easy to refactor to other drivers tomorrow

drivers/azure/azure.go
```diff
@@ -125,6 +130,18 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 			EnvVar: "AZURE_SSH_USER",
 			Value:  defaultSSHUser,
 		},
+		mcnflag.StringFlag{
+			Name:   flAzureWinRMUser,
+			Usage:  "Username for WinRM login",
```
ahmetalpbalkan: let's prefix Usage for all new flags with `(Windows-only)` otherwise it'll be kind of confusing to linux machine users

drivers/azure/azure.go
```diff
@@ -233,6 +257,11 @@ func (d *Driver) SetConfigFromFlags(fl drivers.DriverOptions) error {
 	d.NoPublicIP = fl.Bool(flAzureNoPublicIP)
 	d.StaticPublicIP = fl.Bool(flAzureStaticPublicIP)
 	d.DockerPort = fl.Int(flAzureDockerPort)
+	d.OS = fl.String(flAzureOS)
+	if d.OS == "windows" {
+		// Open WinRM port
+		d.OpenPorts = append(d.OpenPorts, "5986")
```
ahmetalpbalkan: I suggest doing this in `getSecurityRules` method. https://github.com/docker/machine/blob/9718d6c266a767d09cc7521370d5a49c7307d596/drivers/azure/util.go#L73 so that you can give it a proper name

drivers/azure/azureutil/azureutil.go
```diff
@@ -159,6 +160,9 @@ func (a AzureClient) CreatePublicIPAddress(ctx *DeploymentContext, resourceGroup
 			Location: to.StringPtr(location),
 			Properties: &network.PublicIPAddressPropertiesFormat{
 				PublicIPAllocationMethod: ipType,
+				DNSSettings: &network.PublicIPAddressDNSSettings{
+					DomainNameLabel: to.StringPtr(name),
```
ahmetalpbalkan: why do you need to add a FQDN? if the ip-name is taken "globally" this will fail. For instance if two of us ask for a machine named foo in westus, one of us will get a failure because FQDN is used. That'll just surface more errors to the users, unless you need a FQDN, let's avoid it.

drivers/azure/azureutil/azureutil.go
```diff
@@ -467,6 +471,31 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
...
+			},
+		}
+	} else if os == drivers.WINDOWS {
+		osProfile = &compute.OSProfile{
+			ComputerName:         to.StringPtr(name),
+			AdminUsername:        to.StringPtr(username),
+			AdminPassword:        to.StringPtr(password),
+			WindowsConfiguration: &compute.WindowsConfiguration{},
+		}
+	}
```
ahmetalpbalkan: if you could handle `else` as well (by erroring out) that would be great. I feel like we should have an osutil file where we host the windows/linux distinction codes. Otherwise inline if/else OS-handling code will get longer and longer.

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +535,37 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
...
+		Type:     to.StringPtr("Microsoft.Compute/virtualMachines/extensions"),
+		Name:     to.StringPtr("winrm"),
+		Location: to.StringPtr(location),
+		Properties: &compute.VirtualMachineExtensionProperties{
+			Publisher:          to.StringPtr("Microsoft.Compute"),
+			Type:               to.StringPtr("CustomScriptExtension"),
+			TypeHandlerVersion: to.StringPtr("1.4"),
+			Settings: &map[string]interface{}{
+				"fileUris": []string{
+					"https://raw.githubusercontent.com/ppadala/machine/azure-windows/vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/ConfigureWinRM.ps1",
```
ahmetalpbalkan: we should probably check in these scripts/binaries to this repo :P

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +535,37 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
...
+	}
+
+	vmExtension := compute.VirtualMachineExtension{
+		Type:     to.StringPtr("Microsoft.Compute/virtualMachines/extensions"),
+		Name:     to.StringPtr("winrm"),
+		Location: to.StringPtr(location),
+		Properties: &compute.VirtualMachineExtensionProperties{
+			Publisher:          to.StringPtr("Microsoft.Compute"),
+			Type:               to.StringPtr("CustomScriptExtension"),
+			TypeHandlerVersion: to.StringPtr("1.4"),
```
ahmetalpbalkan: try to use 1.0 with autoMinorVersionUpgrade:true, that'll give you highest 1.*

vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/azuredeploy.json
```diff
@@ -0,0 +1,194 @@
+{
```
ahmetalpbalkan: why do you need to vendor `azure-quickstart-templates` repo?

ahmetalpbalkan:
I like this! General code quality is really good @ppadala. I left a few comments. Going forward the Linux/Windows distinction is going to grow, therefore I'm concerned with having inline if/else statements to handle those. Our general approach should be better writing different methods/types where we can factor out these if/else statements and have type system do the work for us.

We should also get machine maintainers on board to add `--winrm-username`, `--winrm-password` arguments directly to machine rather than individual drivers. For now as long as we document them correctly (by adding `(Windows-only)` prefix), we can keep ’em in the azure driver.

drivers/azure/azure.go
ppadala: Done

drivers/azure/azure.go
ppadala: Agree, done

drivers/azure/azure.go
ppadala: Done

drivers/azure/azureutil/azureutil.go
ppadala: Long story. The ConfigureWinRM.PS1 script needed FQDN, which is used as the CN in the cert. Later I realized that I can make the client ignore the CN, so this is not needed. Will remove.

drivers/azure/azureutil/azureutil.go
ppadala: I will add a top-level check, where flags are set.

Regarding if/else, it bothers me too. We do need a separate library that hides OS specific stuff. I think, as we onboard more drivers, we can start doing this.

drivers/azure/azureutil/azureutil.go
ppadala: Yes, done.

drivers/azure/azureutil/azureutil.go
ppadala: Done

vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/azuredeploy.json
ppadala: I made changes to the files, as there are a couple of bugs. 

- makecert is generating negative serial numbers ocassionally
- WinRM was not setup for basic auth.

Both of the above are required for the golang WinRM client to connect. Submitted upstream, not merged yet. When it is merged, I will remove the vendored version.

ppadala:
> I like this! General code quality is really good @ppadala. I left a few comments. Going forward the Linux/Windows distinction is going to grow, therefore I'm concerned with having inline if/else statements to handle those. Our general approach should be better writing different methods/types where we can factor out these if/else statements and have type system do the work for us.

Agree. Maybe a simple OS interface and implementation for Linux and Windows. As I mentioned above, let's do this as it gets clearer how other drivers will work with Windows.

> We should also get machine maintainers on board to add --winrm-username, --winrm-password arguments directly to machine rather than individual drivers. For now as long as we document them correctly (by adding (Windows-only) prefix), we can keep ’em in the azure driver.

I was thinking the same, but it looks like even the -ssh-username is per driver. May be @nathanleclaire can comment.

nathanleclaire:
Wow, epic.  It will take me a while to review in detail but thanks for this!

> I was thinking the same, but it looks like even the -ssh-username is per driver. May be @nathanleclaire can comment.

Ahmet is right -- they should be global if possible.  I think Amazon and other clouds support Windows machines, so it makes sense to me.  Making the `--ssh-*` parameters per-driver happened mostly incidentally, not as a deliberate design decision (and really I'd like to pull them all into one common set of flags), I'd like to get things right from the start on this one if possible.

drivers/azure/azure.go
ahmetalpbalkan: also extract the port number to const if you can :)

nathanleclaire:
Just FYI, since it changes the RPC API, it will most likely require a version bump to the API.  Otherwise it may try to call methods that don't exist on the other end.  We haven't handled such a transition before, but we can discuss how we might handle it in this case and others (https://github.com/docker/machine/pull/3209, for instance, really should bump the API version as well).

drivers/azure/azure.go
```diff
@@ -325,13 +354,28 @@ func (d *Driver) Create() error {
...
+			return err
+		}
+	} else if d.OS == drivers.WINDOWS {
+		if err := c.CreateVirtualMachine(d.OS, d.ResourceGroup, d.naming().VM(), d.Location, d.Size, d.ctx.AvailabilitySetID,
+			d.ctx.NetworkInterfaceID, d.BaseDriver.WinRMUser, d.BaseDriver.WinRMPassword, d.ctx.SSHPublicKey, d.Image, d.ctx.StorageAccount); err != nil {
+			return err
+		}
+		if err := c.CreateVirtualMachineExtension(d.OS, d.ResourceGroup, d.naming().VM(), d.Location); err != nil {
+			return err
+		}
```
ahmetalpbalkan: let's handle `else` here (and everhywhere else applicable). 

also for these stuff, `switch` statement is more idiomatic.

ahmetalpbalkan:
+3200 file addition seems a bit weird, did you use godep to vendor the files? normally it should include only the go packages needed for the build (namely only *.go files), but it appears like this just included the entire `packer` repo

ppadala:
> +3200 file addition seems a bit weird, did you use godep to vendor the files? normally it should include only the go packages needed for the build (namely only *.go files), but it appears like this just included the entire packer repo

Yeah, I was not sure how to do this, included packer repo. Are there instructions somewhere on how to do this?

drivers/azure/azureutil/azureutil.go
ahmetalpbalkan: alright, lmk when u push the new commits

ahmetalpbalkan:
@ppadala first you need to undo the changes you did to `vendor/`, then `make dep-restore`, then `go get` the dependencies you need in your code, and then `make dep-save`. Then only the packages you need and only *.go files should be copied to `vendor/`.

drivers/azure/azure.go
```diff
@@ -150,6 +170,11 @@ func (d *Driver) GetCreateFlags() []mcnflag.Flag {
 			Value:  defaultAzureImage,
 		},
 		mcnflag.StringFlag{
+			Name:  flAzureOS,
+			Usage: "OS for the Azure VM (Windows|Linux)",
```
ahmetalpbalkan: I think this field is case-sensitive, so these docs are going to be wrong. Let's just say `(allowed values: "windows", "linux")`

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +532,33 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
...
+					"https://raw.githubusercontent.com/docker/machine/master/vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/ConfigureWinRM.ps1",
+					"https://raw.githubusercontent.com/docker/machine/master/vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/makecert.exe",
+					"https://raw.githubusercontent.com/docker/machine/master/vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/winrmconf.cmd",
+				},
+				"commandToExecute": "powershell -ExecutionPolicy Unrestricted -file ConfigureWinRM.ps1 " + name,
+			},
+		},
+	}
+
+	log.Debug("Creating VM extensions")
```
ahmetalpbalkan: more accurately: `("Adding CustomScriptExtension to the VM")`

drivers/azure/azureutil/azureutil.go
```diff
@@ -587,6 +626,20 @@ func (a AzureClient) GetPrivateIPAddress(resourceGroup, name string) (string, er
 	return to.String((*nic.Properties.IPConfigurations)[0].Properties.PrivateIPAddress), nil
 }
 
+func (a AzureClient) GetFQDN(resourceGroup, name string) (string, error) {
```
ahmetalpbalkan: per your earlier comment, this method should be unused

drivers/azure/util.go
```diff
@@ -96,6 +97,11 @@ func (d *Driver) getSecurityRules(extraPorts []string) (*[]network.SecurityRule,
 		mkRule(300, "DockerAllowAny", "Allow docker engine access (TLS-protected)", "*", fmt.Sprintf("%d", d.DockerPort)),
 	}
 
+	if d.OS == drivers.WINDOWS {
```
ahmetalpbalkan: we shouldn't be enabling SSH port for windows, so this should be

    if linux {add ssh rule } elseif windows {add winrm rule}

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +532,33 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
...
+		Name:     to.StringPtr("winrm"),
+		Location: to.StringPtr(location),
+		Properties: &compute.VirtualMachineExtensionProperties{
+			Publisher:               to.StringPtr("Microsoft.Compute"),
+			Type:                    to.StringPtr("CustomScriptExtension"),
+			TypeHandlerVersion:      to.StringPtr("1.0"),
+			AutoUpgradeMinorVersion: to.BoolPtr(true),
+			Settings: &map[string]interface{}{
+				"fileUris": []string{
+					"https://raw.githubusercontent.com/docker/machine/master/vendor/github.com/Azure/azure-quickstart-templates/201-vm-winrm-windows/ConfigureWinRM.ps1",
```
ahmetalpbalkan: I guess when you correctly run godep, these files will no longer be here. So we need to have a discussion where to check in these 3 files. I think a directory under `contrib` would be fine.

Also as we keep fetching these from github, users may run into throttling as github aggressively throttles raw file downloads. Perhaps if docker can offer a better location to store these, that would be the best solution.

drivers/azure/azure.go
ppadala: Already done.

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +532,33 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
 	return err
 }
 
+func (a AzureClient) CreateVirtualMachineExtension(os, resourceGroup, name, location string) error {
+	vmExtension := compute.VirtualMachineExtension{
+		Type:     to.StringPtr("Microsoft.Compute/virtualMachines/extensions"),
+		Name:     to.StringPtr("winrm"),
```
ahmetalpbalkan: I suggest calling this `"CustomScript"` or `"CustomScript-winrm"`

ahmetalpbalkan: actually you can probably drop this Name field here. You're already specifying it all the way at the end of this method win `CreateOrUpdate`, API will just use that name.

drivers/azure/azureutil/azureutil.go
ppadala: I removed the call, was thinking of leaving it there for future usage, if required. I can remove, if you feel strongly about it.

drivers/azure/util.go
ppadala: Done

drivers/azure/azureutil/azureutil.go
ppadala: I am ok with contrib. @nathanleclaire is that ok? and is there a place we can put these files instead of github?

drivers/azure/azureutil/azureutil.go
ppadala: Done

drivers/azure/azureutil/azureutil.go
ppadala: Done.

ppadala:
> Just FYI, since it changes the RPC API, it will most likely require a version bump to the API. Otherwise it may try to call methods that don't exist on the other end. We haven't handled such a transition before, but we can discuss how we might handle it in this case and others (#3209, for instance, really should bump the API version as well).

What's the process for bumping the version?

drivers/azure/azureutil/azureutil.go
ahmetalpbalkan: It's dead code. Removing it would be better.

nathanleclaire:
> What's the process for bumping the version?

Well, we haven't actually done it so far, so we'd be in new territory (and it's why we should tread carefully), but it'd be something like:

1. Bump the `APIVersion` located in `libmachine/version/version.go`
2. Convert RPC client initiation code to check the version, and use the relevant client struct.  We might be able to avoid breaking backward compat here by simply defaulting to the older version if the API is V1.  This will probably be the lion's share of the work, because it will entail pulling the common methods into a struct which both can embed.
3. Send out an e-mail to the 3rd party plugin authors that we're going to be bumping the version, and urging them to update before we release a version with this bump.

It's definitely non-trivial, but also not impossible.  Also, should get as many eyes on this possible transition as possible, including @dgageot @dmp42.

One thing I'd like to do, if we're bumping the version anyway, is to maybe try and clean up / consolidate some of these methods related to SSH / remote access, for instance I wonder if we could abstract SSH / WinRM behind a common interface, or if that would be too annoying due to differences between them.

drivers/azure/azureutil/azureutil.go
ppadala: Ok, gone.

ppadala:
@nathanleclaire is it better for docker-machine core maintainers to bump the change, after this code is merged? It would take me much longer to do this and it also makes sense for core maintainers to reach out to plugin authors (don't even know how to contact).

On a related note, is there anyone using docker-machine in a client-server model? Looks like server is the plugin? Any pointer to architecture also would be useful.

nathanleclaire:
> is it better for docker-machine core maintainers to bump the change, after this code is merged? It would take me much longer to do this and it also makes sense for core maintainers to reach out to plugin authors (don't even know how to contact).

Hm, well, we're all in this together -- if we need to bump the version API we can work together on it (in fact we have to, I'm not willing to merge any known backwards-incompatible code unless the circumstance is exceptional).  I'm happy to help with it, I'm sure we can figure it out.  But, we should design carefully and get as many eyeballs on it as possible before merging.

As for plugin authors, that I'll take care of, one of the reasons we made a document for it in this repo was so that we could have a definitive list of 3rd party plugins and the maintainers' contact info.

> On a related note, is there anyone using docker-machine in a client-server model? Looks like server is the plugin? Any pointer to architecture also would be useful.

Yes the plugin model is client-server, basically the way it works is that for each driver Machine execs a binary which spins up an RPC server and then talks to it to perform operations.  We are pretty overdue on some actual docs in this regard, maybe I should put something together.

nathanleclaire:
We need to ensure that the license on the Packer code (MPL v2.0) is compatible with our license (Apache) as well.

ppadala:
> @ppadala first you need to undo the changes you did to vendor/, then make dep-restore, then go get the dependencies you need in your code, and then make dep-save. Then only the packages the build needs and only the *.go files should be copied to vendor/.

I tried this.

1. Started with a clean GOPATH dir
2. make dep-restore works successfully
3. make build fails because packer is missing
4. go get github.com/mitchellh/packer/communicator/winrm
5. make dep-save 

Step 5 fails with 

```
Pradeeps-MacBook-Air-2:machine ppadala$ make dep-save
...
godep: Package (github.com/dylanmei/iso8601) not found
```

I am using go1.6.1 on MacOS amd64, which I am guessing is supported. Any thoughts?

ppadala:
Scratch my previous comment, I have to get all the dependencies and then run godep save.

ppadala:
> Hm, well, we're all in this together -- if we need to bump the version API we can work together on it (in fact we have to, I'm not willing to merge any known backwards-incompatible code unless the circumstance is exceptional). I'm happy to help with it, I'm sure we can figure it out. But, we should design carefully and get as many eyeballs on it as possible before merging.

@nathanleclaire I am happy to take care of bumping the API, if there are clear directions and arch documentation on how to do this. I suggested the core maintainers to do this, because you have all the knowledge of design decisions over time and what you mean by backward compatibility.

Looking forward to docs on this.

ppadala:
> We need to ensure that the license on the Packer code (MPL v2.0) is compatible with our license (Apache) as well.

Wrestling with dependencies, once I have them all figured out, will check. Quick check on Packer is MPL v2.0

nathanleclaire:
Asked around a bit and upstream docker/docker vendors some MPL code, so we're probably good to follow their example (we cannot, of course, change any of the vendored code).

ppadala:
Alright, I took care of all the comments including vendoring in only the files needed (found a problem on Mac #3337, also added instructions #3336). Checked licenses, all the deps have either MPL v2 or Apache 2.0. There's one package with LGPLv3, does that matter?

ppadala:
@nathanleclaire, I have been checking the RPC client/server and the plugin code. Here's what I understand so far and some questions.

1. The plugins are expected to binaries compiled separately, which will be setup as a rpc server by executing them.
    - How to compile these plugins?
    - Is there a place these binaries need to placed for docker-machine to pick them up?
    - Can the plugins run on a different machine?
    - What's the expectation by plugin authors? Do they expect to stay with old API, compiled server binary and have the new code work? Why can't they recompile their binaries?

2. Client initiation code. I already see there are two versions, but the handling is for clients. switchToV0 seems just change the name that's passed. 
```go
	if err := c.Client.Call(GetVersionMethod, struct{}{}, &serverVersion); err != nil {
		// this is the first call we make to the server. We try to play nice with old pre 0.5.1 client,
		// by gracefully trying old RPCServiceName, we do this only once, and keep the result for future calls.
		log.Debugf(err.Error())
		log.Debugf("Client (%s) with %s does not work, re-attempting with %s", c.Client.MachineName, RPCServiceNameV1, RPCServiceNameV0)
		c.Client.switchToV0()
		if err := c.Client.Call(GetVersionMethod, struct{}{}, &serverVersion); err != nil {
			return nil, err
		}
	}
```
 - What is the expected behavior from a client that sees older server? There are two options
     -  Let the calls proceed, catch the error and return an error to caller with IncompatibleVersions and fail. New code never works with old server. This make it easier, since the new code already has calls to GetOS().
     - Have two interfaces, switch to the old one to call the old server, provide sane defaults for GetOS(), which is reasonable and easy to do.

I am singling out GetOS() because that's the only new function. GetWinRMUsername() and GetWinRMPassword() won't get invoked unless GetOS() returns "windows", which won't happen with old servers.

Regarding refactoring SSH and WinRM, I suggest a function called GetCredentials() that returns a structure with Linux and Windows credentials configuration. If this is done, there's going to be more work to wrestle with old servers.

nathanleclaire:
I think we are pretty skittish on LGPL, I don't think there's much precedent for including in in Docker projects.  What's the component / how important is it?

nathanleclaire:
> How to compile these plugins?

Usually a `package main` file which calls `plugin.RegisterDriver()` with an instantiated `Driver` struct, e.g.  https://github.com/nathanleclaire/docker-machine-driver-dind/blob/master/bin/main.go

> Is there a place these binaries need to placed for docker-machine to pick them up?

Anywhere in `$PATH`

> Can the plugins run on a different machine?

You mean like the server run on a different machine?  It's not supported today, but theoretically possible.  However, the tricky bit is the filesystem, which is largely assumed to be shared: persistence of SSH keys, VM disks, etc.

> What's the expectation by plugin authors? 

That their currently deployed binaries keep working

> Do they expect to stay with old API, compiled server binary and have the new code work?

Yes

> Why can't they recompile their binaries?

They can, but we can't make them and/or make sure that we bring it to their attention.  It's just a nuisance if you wanted to deploy some code and have it stay working.  That said, we pretty much have to bump the API at some point, so that's why I started the `AVAILABLE_DRIVER_PLUGINS.md`, partially to have a list of emails where we could reach them to alert them to changes like this.

> Client initiation code. I already see there are two versions, but the handling is for clients. switchToV0 seems to just change the name that's passed.

Yeah, you're looking at some cleanup code due to an accidental API break we made.  We might be best off with a "clean break" model, it's still up in the air how we should handle this.  (I think this answers your next question as well -- basically, "I'm not sure, let's evaluate the merits of each approach").

> I am singling out GetOS() because that's the only new function. GetWinRMUsername() and GetWinRMPassword() won't get invoked unless GetOS() returns "windows", which won't happen with old servers.

Hmmm.... Theoretically, but never say never.

> Regarding refactoring SSH and WinRM, I suggest a function called GetCredentials() that returns a structure with Linux and Windows credentials configuration. If this is done, there's going to be more work to wrestle with old servers.

Yeah, my concern is mostly around the explosion of individual methods on the `Driver` interface, which is already extremely overloaded.  If we could re-design it from the beginning, at the very least I'd want to group all of the SSH methods together in some fashion (common interface / struct, I'm not 100% sure).

ppadala:
> Why can't they recompile their binaries?
>
> They can, but we can't make them and/or make sure that we bring it to their attention. It's just a nuisance if you wanted to deploy some code and have it stay working. That said, we pretty much have to bump the API at some point, so that's why I started the AVAILABLE_DRIVER_PLUGINS.md, partially to have a list of emails where we could reach them to alert them to changes like this.

I understand this, we should definitely inform them. But, I am puzzled that breaking changes in drivers are allowed like changing arguments, auth model etc.

Regarding the approaches,

>1. Let the calls proceed, catch the error and return an error to caller with IncompatibleVersions and fail. New code never works with old server. This make it easier, since the new code already has calls to GetOS().
>
> 2. Have two interfaces, switch to the old one to call the old server, provide sane defaults for GetOS(), which is reasonable and easy to do.

I am going to prepare a change for option (2) above, and see where it goes.

nathanleclaire:
> I understand this, we should definitely inform them. But, I am puzzled that breaking changes in drivers are allowed like changing arguments, auth model etc.

What do you mean?

The Drivers are free to change as much as they want as long as the API they expose remains constant.

> I am going to prepare a change for option (2) above, and see where it goes.

Nice!

ppadala:
> I understand this, we should definitely inform them. But, I am puzzled that breaking changes in drivers are allowed like changing arguments, auth model etc.
>
> What do you mean?

I am talking from the end user perspective. Simple example is how Azure driver moved from classic to ARM. I am actually in favor of changing this, but for an end user (or a company bundling docker-machine), this is a huge change. As an end-user, I can't even access my old (Azure) machines that are spun up.

If that's allowed, then why worry about driver interface changing? Plugins don't live on their own anyway, if they are bundling docker-machine client, they can always stay with old client code until they recompile.

I am not sure, what you are achieving by this backward compatibility.

Also, can you provide instructions on how to compile a plugin binary and setting it up so that I can test? I might be misunderstanding something in terms of how plugins are created and used, this should clarify for me.

ppadala:
Another way to look at it is that

- For a given released major version, minor version changes always maintain backward compatibility with Driver interface
- For bump in release numbers, no such contract

If that's the case, we can time the merge such that this happens when major version bump happens. That allows refactoring of Driver interface and other breaking interface changes much easier.

nathanleclaire:
> I am talking from the end user perspective. Simple example is how Azure driver moved from classic to ARM. I am actually in favor of changing this, but for an end user (or a company bundling docker-machine), this is a huge change. As an end-user, I can't even access my old (Azure) machines that are spun up.

Because myself and @ahmetalpbalkan considered it worth the risk, there weren't many active users of the existing `azure` driver, and we went to great lengths to ensure that the change was highly visible and a clear migration strategy was present, i.e. that users wouldn't be totally hosed if they happened to invoke a `docker-machine` command with the new version and still had existing hosts  (I've heard almost zero complaints about the `azure` driver move so far, whereas when we've broken plugin API in the past we start hearing about it right away).

> Plugins don't live on their own anyway

[Yes, they do](https://github.com/docker/machine/blob/master/docs/AVAILABLE_DRIVER_PLUGINS.md).  Note all of those plugins are distributed as binaries independent of the Docker Machine client.  We simply happen to bundle the core ones in the same binary as the client.

> I am not sure, what you are achieving by this backward compatibility.

Principle of least surprise.  Don't you hate it when an upstream breaks things on you without warning?  "That which is working today, continues to work, unless I have been alerted to the fact that it might break as far ahead of time as possible".  

Now, I know in this case it _seems_ to be that there is minimal risk to simply add some new API methods that _probably_ won't get invoked on anything but the `azure` driver, at least for the immediate future.  But, don't you feel that the more tolerant and less structured that backwards compatibility breaks in the project are, no matter how small the risk seems, the more that folks will use it the precedent as leverage to try and justify _more_ backwards compatibility breaks, without a cogent model to do it safely?

> Also, can you provide instructions on how to compile a plugin binary and setting it up so that I can test? I might be misunderstanding something in terms of how plugins are created and used, this should clarify for me.

1. Make a `package main` file such as [this one](https://github.com/nathanleclaire/docker-machine-driver-dind/blob/master/bin/main.go) which calls `RegisterDriver` with an instance of your `Driver` struct.
2. (Let's say the plugin is named `foo`) Compile this binary with the name `docker-machine-driver-foo` and drop it into your `PATH`.
3. Now, invoking `docker-machine create -d foo` should work.

ppadala:
> Because myself and @ahmetalpbalkan considered it worth the risk, there weren't many active users of the existing azure driver, and we went to great lengths to ensure that the change was highly visible and a clear migration strategy was present, i.e. that users wouldn't be totally hosed if they happened to invoke a docker-machine command with the new version and still had existing hosts (I've heard almost zero complaints about the azure driver move so far, whereas when we've broken plugin API in the past we start hearing about it right away).

I mentioned this in the review for @ahmetalpbalkan's change. We (as ContainerX company) are using docker-machine to spin up Azure VMs. We have released beta version of our full product (which bundles a specific docker-machine version) that is used by customers. Azure ARM move is definitely a breakage for us. Not only that, the new model doesn't allow automation, which we have an outstanding issue #3199.

May be I should have made more noise about this. But, I am ok with the change at this point, because ARM is the right path forward for failure handling and other issues.

ppadala:
Regarding backward compatibility, I know very well why we need it. But, your definition of backward compatibility of Driver interface is a narrow one. You gotta think about end users and people trying to build stuff on top of docker-machine.

As I understand, what we are preserving is the following workflow for someone who doesn't want to re-compile and change their plugin code. My only gripe is that true backward compatibility is about end user usage. You should also preserve backward compatibility for third party plugins, but that should be done at major versions.

Anyway, we seem to be agreeing on top level issue of backward compatibility. Let me see, if I can whip out a change for versioning and see where we go.

ppadala:
> Plugins don't live on their own anyway
>
> Yes, they do. Note all of those plugins are distributed as binaries independent of the Docker Machine client. We simply happen to bundle the core ones in the same binary as the client.

Thanks for this clarification. It also makes it clear to me how the plugins are expected to built and distributed.

nathanleclaire:
Yeah, I think we are more agreeing than not.  My point is not "let's preserve backwards compatibility at all costs, even if it results in hackier code or prevents new features from shipping", but rather "let's make a calculated risk in breaking backwards compatibility and ensure that downstreams are as safe/aware as possible".  And the way to do that sanely is by versioning and pinging maintainers.  Personally I wince a little bit every time I realize we have broken or changed something for users unintentionally / unannounced, so I've become more and more adamant over time about ensuring we're as safe as we can be (even though mistakes are inevitable).

ppadala:
> I think we are pretty skittish on LGPL, I don't think there's much precedent for including in in Docker projects. What's the component / how important is it?

Packer seems to include masterzen/winrm (which is the main library that provides winrm support) which depends on masterzen/xmlpath, which LGPLv3.

I hate the fact that we need to include all these libraries to get winrm functionality. Will check, if there's an alternative. @ahmetalpbalkan, is there any official library (from Microsoft or others) in golang for winrm?

ahmetalpbalkan:
@ppadala I believe there are no Go libraries we developed or used for WinRM.

ppadala:
@nathanleclaire, I came up with a simple change that uses a default when versions are different. See https://github.com/ppadala/machine/commit/3cc3785db8d419387f663438c4d0f96fb91e24eb. Tested this with the DIND driver and it works fine.

I was originally thinking of creating a new name space like RPCServerDriverV2 with corresponding code changes in server_driver.go and massaging of calling specific functions, but that seemed like an overkill for this.

Thoughts?

drivers/azure/azureutil/azureutil.go
```diff
@@ -520,6 +534,32 @@ func (a AzureClient) CreateVirtualMachine(resourceGroup, name, location, size, a
 	return err
 }
 
+func (a AzureClient) CreateVirtualMachineExtension(os, resourceGroup, name, location string) error {
+	vmExtension := compute.VirtualMachineExtension{
```
ahmetalpbalkan: can we add here a logging statement like the following:

    Adding CustomScriptExtension to virtual machine

otherwise the users won't be able to see the progress or feel that it is proceeding very clearly.

ppadala: Already added in line 557

ahmetalpbalkan: Hmm I suggest moving it up here. B/c if say the next statement gets a nil dereference panic, we'll know where it comes from very clearly.

ppadala: Ok, good point. Done.

ppadala:
@nathanleclaire any comments on https://github.com/ppadala/machine/commit/3cc3785db8d419387f663438c4d0f96fb91e24eb ?

On a side note, my upstream change for Azure quick-start-template for WinRM is merged, so I removed the bundling of that into this change.

ahmetalpbalkan:
Azure driver changes **LGTM**. However when it is time to update the docs, I wish to keep the windows and linux examples separate if we can.

One thing that's still not-LGTM is the dependencies you refer to in CustomScriptExtension should probably be checked in to this repo or another docker repo, otherwise when `azure-quickstart-templates` repo changes something (which is something it does quite often) this will be immediately broken. Please do not rely that repo and find a more persistent place (preferably outside github) to store these scripts/binaries.

ppadala:
> Azure driver changes LGTM. However when it is time to update the docs, I wish to keep the windows and linux examples separate if we can.

Thanks @ahmetalpbalkan. I'll add docs, when we are ready to merge and agree with separating Windows examples to reduce confusion.

> One thing that's still not-LGTM is the dependencies you refer to in CustomScriptExtension should probably be checked in to this repo or another docker repo, otherwise when azure-quickstart-templates repo changes something (which is something it does quite often) this will be immediately broken. Please do not rely that repo and find a more persistent place (preferably outside github) to store these scripts/binaries.

Agree, will bring them back in.

nathanleclaire:
Something to be thinking about, how about tests for this?  (Unit and integration)

Why not put the empty `WinRM`-related methods inside of the `BaseDriver` struct which gets embedded by most drivers?

ppadala:
> Something to be thinking about, how about tests for this? (Unit and integration)

I had a unit test that calls create and delete, but I have to manually enter Azure credentials. How are the tests done for other drivers? @ahmetalpbalkan can you point me to existing tests for Azure Linux driver?

> Why not put the empty WinRM-related methods inside of the BaseDriver struct which gets embedded by most drivers?

They are already in BaseDriver.

nathanleclaire:
> They are already in BaseDriver.

Ah, OK, I was not realizing that the Drivers they have been explicitly set for are mostly the testing-related drivers.

ahmetalpbalkan:
@ppadala we don't have tests. 😳

ppadala:
@ahmetalpbalkan :-)

@nathanleclaire there's no easy way to do automated tests without Azure credentials. If you have suggestion here, let me know.

StefanScherer:
Has anyone tried this with TP5? I can't find any TP5+Docker image, so probably the provisioning of a vanilla TP5 with `--azure-image MicrosoftWindowsServer:WindowsServer:Windows-Server-Technical-Preview:5.0.20160420` won't work. 

Can we add provisioning like in "https://get.docker.com/" for bash, but for PowerShell to enable Containers feature if missing, to install Docker if it is missing, to download the base OS image as long as we have to install it w/o docker pull?

ppadala:
@StefanScherer we are looking into it. It looks like Windows Server Core image for TP5 is not yet available. Also, now the scripts (UpdateHost.PS1 for example) are broken on TP4. I am guessing it's because of the URLs pointing to new versions.

cc @enderb-ms

StefanScherer:
@ppadala I followed the updated docs at https://msdn.microsoft.com/virtualization/windowscontainers/deployment/deployment

```
Install-WindowsFeature containers
Install-PackageProvider ContainerImage -Force
Install-ContainerImage -Name WindowsServerCore
Invoke-WebRequest https://aka.ms/tp5/Update-Container-Host -OutFile update-containerhost.ps1
.\update-containerhost.ps1
```

I also try to automate all these new steps for my packer build.

ppadala:
@StefanScherer update-containerhost.ps1 is broken for TP4 now. I haven't tried it on TP5 yet.

ppadala:
Another problem we just saw on TP5. https://github.com/Microsoft/Virtualization-Documentation/issues/225. cc @enderb-ms and @taylorb-microsoft

enderb-ms:
@ppadala the main reason the new scripts are broken in TP4 is due to how much networking set up has changed between TP4 and TP5.
@StefanScherer in regards to the TP5+Docker image, our current plan is to not publish one, and just have the TP5 image. This is due to some people having issues in TP4 once the version of Docker in the images went stale, and now we could also have an issue with the base images going stale. But if you could give us feedback on whether you think this was a good or bad choice, I'd appreciate it.

StefanScherer:
@enderb-ms I have seen this updated azure template https://github.com/Azure/azure-quickstart-templates/tree/master/windows-server-containers-preview which I want to try out. If that template works it is a starting point to update the steps needed in docker-machine.

If it's a good or bad choice to install docker and the base image depends on the time it takes for a user to have the whole Docker engine up and running. Sure, as long as it's a tech preview it is good to have an up-to-date Docker version.

ppadala:
Agree with @StefanScherer. Installing and setting up docker-engine takes considerably long time, especially getting the core images etc.

@enderb-ms, if you want better user experience have the clean Windows server image, but also have an image with docker-engine installed and configured to run locally.

cc @selvik

StefanScherer:
I would prefer a prebuilt Azure VM with Docker engine and the windowsservercore base image to have a working VM after let's say **five minutes**.

I've created a VM with the Azure template linked above and it took **59 minutes** to have docker installed and to run the first docker commands. And that time would be needed for a docker-machine create. I think that's too long to be useful.

I don't know the process of providing new Azure VMs in the catalog and how long it takes to test/proof/stage/publish it. Having a prebuilt TP5+Docker VM in a regular basis (eg. every 1-2 weeks) would be new enough. And as a user I still can run the update-containerhost to update docker engine to the latest version and so on.

Behind the scenes there should be an automated way to create a new TP5+Docker Azure VM to update the catalog that can be run after a major change (base image, docker engine) or every weekend. Don't know if [packer-azure](https://github.com/Azure/packer-azure) may help here or if you have other automation tools at Azure.

StefanScherer:
Any updates for TP5?

Microsoft switched the [installation guide](https://msdn.microsoft.com/en-us/virtualization/windowscontainers/quick_start/quick_start_windows_server) from Install-ContainerHost.ps1 to use dockerd.exe as native Windows service to remove a dependency to NSSM.

But this means that docker-machine should be able to modify the `C:\ProgramData\docker\config\daemon.json` instead of adding command line arguments to the daemon. -> https://github.com/docker/machine/issues/3062

StefanScherer:
I just have seen that there is now a "Windows Server 2016 Core with Containers Tech Preview 5" virtual machine in Azure. This might help to update this PR.

StefanScherer:
Any plans to get this running with official Windows Server 2016 VM now available in Azure?

davidarcher:
Just leaving a note here that I am planning to resurrect this PR as I am working on a project where I need to spin up Windows docker hosts via docker-machine in OpenStack and having this WinRM support would be great.

dgageot:
@davidarcher Please do refresh this PR! I'll be here to review it

mitchellmaler:
Is there any status on this PR? I have a use case for windows containers.

PR #3330, opened 2016-04-17
[Generic] Copy the public key to machine directory (fixes #3268)

### LINKED ISSUE #3268: generic ssh key 'Too many retries waiting for SSH to be available.'
Hi, when I use docker-machine for provisioning a remote host and set a ssh key with `--generic-ssh-key ~/.ssh/id_rsa` I get the following error `Error creating machine: Error detecting OS: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded`. When I don't set the --generic-ssh-key flag it works. I even tried to place the ssh key on another path which didn't work aswell.

bamarni:
It should be possible to use the specified location directly. Copying
the private key can also have some drawbacks :

http://apple.stackexchange.com/questions/18458/password-dialog-appears-when-ssh-private-key-permissions-are-set-to-0600/26252#26252

### DIFF
--- a/drivers/generic/generic.go
+++ b/drivers/generic/generic.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"strconv"
 	"time"
 
@@ -87,13 +88,6 @@ func (d *Driver) GetSSHUsername() string {
 }
 
 func (d *Driver) GetSSHKeyPath() string {
-	if d.SSHKey == "" {
-		return ""
-	}
-
-	if d.SSHKeyPath == "" {
-		d.SSHKeyPath = d.ResolveStorePath("id_rsa")
-	}
 	return d.SSHKeyPath
 }
 
@@ -114,8 +108,10 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 func (d *Driver) PreCreateCheck() error {
 	if d.SSHKey != "" {
 		if _, err := os.Stat(d.SSHKey); os.IsNotExist(err) {
-			return fmt.Errorf("Ssh key does not exist: %q", d.SSHKey)
+			return fmt.Errorf("SSH key does not exist: %q", d.SSHKey)
 		}
+
+		// TODO: validate the key is a valid key
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Will this break existing drivers?  e.g. if you had one created with `SSHKey` set before it's now supposed to be `SSHKeyPath` but there's no way that a machine created before this change could know that.

nathanleclaire:
What's the desired state here?  The pub key should also be in the same directory as the private key?  If that's the case, can we copy that over to the Machine directory as well?

bamarni:
I've double-checked for BC breaks. Currently, `SSHKey` references the original key and is only used during machine creation, later on `SSHKeyPath` is always used. With this PR, previous machines would still be working :)

You're right about copying the public key, that would also work. The thing is that I'd rather solve this by removing code if possible. Do you remember if there were a specific reason to copy the key over in the first place? I can't think of any..

bamarni:
I've also rephrased some messages, as we force `IdentitiesOnly`.

nathanleclaire:
> I've double-checked for BC breaks. Currently, SSHKey references the original key and is only used during machine creation, later on SSHKeyPath is always used. With this PR, previous machines would still be working :)

👍 

> You're right about copying the public key, that would also work. The thing is that I'd rather solve this by removing code if possible. Do you remember if there were a specific reason to copy the key over in the first place? I can't think of any..

Our main motivation for supporting this historically (including copying certs, keys, etc.) is to leave the door open for sharing machines by bundling up just one directory where possible.  It's also why custom CA certs and so on are meant to be copied to the individual machine directory when needed.

bamarni:
Thanks for the explanation @nathanleclaire. So I've updated the PR to copy the pubkey too, I've made it fail with a log message only (to keep BC and as it's not a critical failure). I've moved the rest to another PR as it's not directly related.

drivers/generic/generic.go
```diff
@@ -127,13 +122,14 @@ func (d *Driver) Create() error {
 			" future will require the ssh agent to contain the appropriate key.")
 	} else {
 		log.Info("Importing SSH key...")
-		// TODO: validate the key is a valid key
-		if err := mcnutils.CopyFile(d.SSHKey, d.GetSSHKeyPath()); err != nil {
-			return fmt.Errorf("unable to copy ssh key: %s", err)
+
+		d.SSHKeyPath = d.ResolveStorePath("id_rsa")
```
nathanleclaire: Can you make this a top-level `const` please?

nathanleclaire: As a side note:  I think it's not strictly the case that an SSH key will be named `id_rsa` (the key might use another algo than RSA), but RSA is so massively popular vs. the others that we might be OK to just assume it.

libmachine/mcnutils/utils.go
```diff
@@ -63,11 +63,7 @@ func CopyFile(src, dst string) error {
 		return err
 	}
 
-	if err := os.Chmod(dst, fi.Mode()); err != nil {
-		return err
-	}
-
-	return nil
+	return os.Chmod(dst, fi.Mode())
```
nathanleclaire: Nice

drivers/generic/generic.go
bamarni: Good point. Just to clarify, this is the key we're saving so it'd still work for other algos. In that case the hardcoded `id_rsa` filename might be confusing indeed, I've changed it to be the basename of `SSHKey`, hope that's fine.

nathanleclaire:
LGTM

PR #3342, opened 2016-04-20
[Generic] clarified doc / messages about ssh-agent

bamarni:
As the IdentitiesOnly option is set with the SSH external client,
existing identities in ssh-agent won't be used unless it's the default
location.

### DIFF
--- a/docs/drivers/generic.md
+++ b/docs/drivers/generic.md
@@ -33,24 +33,8 @@ to the host.
     $ docker-machine create \
       --driver generic \
       --generic-ip-address=203.0.113.81 \
-      --generic-ssh-key=~/.ssh/id_rsa \
+      --generic-ssh-key ~/.ssh/id_rsa \
       vm
-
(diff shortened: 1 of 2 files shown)


docs/drivers/generic.md
```diff
@@ -33,25 +33,9 @@ to the host.
     $ docker-machine create \
       --driver generic \
       --generic-ip-address=203.0.113.81 \
-      --generic-ssh-key=~/.ssh/id_rsa \
+      --generic-ssh-key ~/.ssh/id_rsa \
       vm
 
-### Password-protected SSH keys
```
bamarni: Here I've removed the whole section as there is nothing special with password-protected keys. It had probably been added because of the bug described in  #3306.

nathanleclaire: Nice, good catch.

nathanleclaire:
LGTM.  Thanks for putting it in Bilal!

PR #3347, opened 2016-04-20
exoscale: fix cloud-init file to be compatible with CoreOS

vincentbernat:
The cloud-init implementation in CoreOS is quite limited and will choke
on `manage_etc_hosts: true`. Use `manage_etc_hosts: localhost` instead
which is supported. This is also supported by the classic Python
cloud-init implementation and the result is quite similar.


nathanleclaire:
LGTM

PR #3350, opened 2016-04-20
Fix Default Address Space for Azure VNets

xyziemba:
RFC 1918 specifies 10.0.0.0/8, not 10.0.0.0/6, as part of the private network address space. This fixes what appears to be a typo.


nathanleclaire:
@ahmetalpbalkan  ?

ahmetalpbalkan:
Yes it looks like a typo. So if we merge this only the newly created vnets will get the fix as we check existence of vnet and skip creating/updating it if exists. I don't see a problem merging this.

nathanleclaire:
So I think what you're trying to say is...

nathanleclaire:
LGTM

PR #3362, opened 2016-04-22
Ensure that attempt to remove instance and SSH key both go through

nathanleclaire:

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -748,16 +748,23 @@ func (d *Driver) Kill() error {
 }
 
 func (d *Driver) Remove() error {
+	multierr := mcnutils.MultiError{
+		Errs: []error{},
+	}
+
 	if err := d.terminate(); err != nil {
-		return fmt.Errorf("unable to terminate instance: %s", err)
+		multierr.Errs = append(multierr.Errs, err)
 	}
 
-	// remove keypair
 	if err := d.deleteKeyPair(); err != nil {
-		return fmt.Errorf("unable to remove key pair: %s", err)
+		multierr.Errs = append(multierr.Errs, err)
 	}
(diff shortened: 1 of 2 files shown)


nathanleclaire:
@tianon @dgageot  What do you think about this? I tried to test it by deleting the instance but not the SSH key pair before attempting `docker-machine rm` and it seemed to work, but it also didn't spit any error message back at me, so I suspect the conditions resulting in dangling SSH keys might be slightly more difficult to track down.

tianon:
If the terminate fails for some reason (lack of network connection, etc), wouldn't this mean that our SSH key is now gone and we have no way of accessing the VM anymore?

nathanleclaire:
@tianon Indeed, but if you're requesting `docker-machine rm`, that instance is as good as gone to the user anyway, right?

nathanleclaire:
So, to me it sort of seems like expecting to be able to salvage deleted files if UNIX `rm` were to fail for some reason.

tianon:
Yeah, that's fair -- that makes me wonder what `rm`'s behavior is when it comes across a failure halfway through a recursive `rm`. :smile:

tianon:
Well, that answers that:
```console
$ mkdir -p a/b/c
$ mkdir -p e/f/g
$ mkdir -p h/i/j
$ sudo chown -R root:root e
$ rm -rf a e h
rm: cannot remove ‘e/f/g’: Permission denied
$ find
.
./e
./e/f
./e/f/g
$ 
```

LGTM :+1:

nathanleclaire:
Nice

nathanleclaire:
> Yeah, that's fair -- that makes me wonder what rm's behavior is when it comes across a failure halfway through a recursive rm. :smile:

It doesn't just print "Oops" to STDERR and exit 1?

PR #3365, opened 2016-04-25
Fixes #2302 Possibility to create master with manager, but without agent

### LINKED ISSUE #2302: Possibility to create master with manager, but without agent
Originally reported as https://github.com/docker/swarm/issues/771, but seems to be Machine-related.
I'd like to also have master with just `swarm-agent-master` without `swarm-agent` on it.
I'm wondering that this is not possible:(
Especially having such message in docs:
> The master is running both the swarm manager and a swarm agent container. This isn’t recommended in a production environment because it can cause problems with agent failover. However, it is perfectly fine to do this in a learning environment like this one.

It would be helpful to run only manager, while currently it doesn't seem to be possible to do out of the box - we need to dig into machine and remove agent manually.

ndeloof:
Signed-off-by: Nicolas De Loof <nicolas.deloof@gmail.com>

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -81,7 +81,7 @@ var (
 		},
 		cli.BoolFlag{
 			Name:  "swarm",
-			Usage: "Configure Machine with Swarm",
+			Usage: "Configure Machine to join a Swarm cluster",
 		},
 		cli.StringFlag{
 			Name:   "swarm-image",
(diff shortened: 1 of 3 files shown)


nathanleclaire:
Nice

Usually changing the swarm options would require a config migration, but in this case it really is a very small change (and also around some code that already has some issues, so we're not breaking anything existing).  Still, it makes me slightly nervous.

nathanleclaire:
FYI, just merged https://github.com/docker/machine/pull/3287 which will affect this.

ndeloof:
rebased

nathanleclaire:
@ndeloof Should I continue evaluating this for merge or wait until you've submitted the changes you suggest in the other thread?

ndeloof:
@nathanleclaire rebased after #3387 has been merged

nathanleclaire:
Nice.  Thanks for your persistence and good work @ndeloof.

LGTM

PR #3368, opened 2016-04-25
[WIP] Add AWS iam role

kencochrane:
AWS has the ability to assign an IAM role to your EC2 instance [1], and this will allow you to make API calls from the EC2 instance, without needing to provide credentials. This works by pinging the metadata service and pulling down some temp credentials and using those until they expire. Once expired, they will get new ones from the same meta data service.

The Go AWS sdk offers the ability to use this out of the box [2], if you use the correct credentials. The goal of this PR is to use those credentials for creating other EC2 instances, and thus, not having to provide your api key and secret in order to do so.

This is a work in progress, when I compile and run, I get an error, but not sure if the error is related to this change, or something else.

There also needs to have a better way to validate the missing AWS key/secret when not using an IAM role. The current way prevents you from going forward without those items, and thus the validation is removed from this PR for now, but a better way needs to get added before we can merge.

/cc @nathanleclaire 

[1] http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html
[2] http://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/ec2rolecreds/EC2RoleProvider.html

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -15,6 +15,9 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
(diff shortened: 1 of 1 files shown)


kencochrane:
@nathanleclaire are we no longer going to add this feature?

PR #3378, opened 2016-04-26
azure: create temporary file next to destination

### LINKED ISSUE #3313: Azure driver: cannot login on ubuntu 14.04
I appear to be running into a couple of problems trying to do a "docker-machine create -d azure ..." from my Ubuntu 14.04 box with the new 0.7.0 driver:

- `Error creating Azure client: Failed to save tenant ID: Failed to rename file: rename /tmp/tenantidxxxxxxxxxx /home/<username>/.docker/machine/credentials/azure/<GUID>.tenantid: invalid cross-device link` (cp'ing appears to work, but rename runs into script errors if I run it manually)
- Even if I copy the tenantid file over manually, the token retrieval process that follows times out before I can go back and forth to the web browser to finish it.

I have the azure-cli installed on this box and it logs in just fine. This appears to be specific to the docker-machine azure driver.

ahmetalpbalkan:
We were doing atomic file writes by writing to /tmp
and moving to the actual destination, however some filesystems
do not support copying from /tmpfs.

Bumping the `go-autorest` dependency version to latest which fixes this.

Fixes #3313.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -96,23 +96,23 @@
 		},
 		{
 			"ImportPath": "github.com/Azure/go-autorest/autorest",
-			"Comment": "v7.0.4",
-			"Rev": "b01ec2b60f95678fa759f796bac3c6b9bceaead4"
+			"Comment": "v7.0.4-6-gec603a8",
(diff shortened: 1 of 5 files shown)


nathanleclaire:
LGTM 👍  Thanks @ahmetalpbalkan

PR #3381, opened 2016-04-27
[SSH] Allow private keys with 0400 permissions (fixes #3359)

### LINKED ISSUE #3359: docker-machine ssh in v1.11.0 broken when talking to 1.10.3 instances
1. docker-machine ssh in v1.11.0 broken when talking to 1.10.3 engines 
2. However it works with --native-ssh flag.  
3. Not broken when talking to 1.11.0 engines
4. In the first example below, notice how it complains about permissions, but the permissions are what they should be

$ docker-machine ssh aws-demo2
Permissions 0400 for '/Users/olddfiliatrault/.docker/machine/machines/aws-demo2/id_rsa' are too open.

$ ls -al /Users/olddfiliatrault/.docker/machine/machines/aws-demo2/id_rsa
-r--------  1 olddfiliatrault  staff  1696 Mar  6 15:16 /Users/olddfiliatrault/.docker/machine/machines/aws-demo2/id_rsa

$ docker-machine --native-ssh ssh aws-demo2
Welcome to Ubuntu 15.10 (GNU/Linux 4.2.0-18-generic x86_64)
Last login: Fri Apr 22 14:29:16 2016 from 10.0.0.144
ubuntu@aws-demo2:~$ 

$ docker-machine ls
NAME        ACTIVE   DRIVER       STATE     URL                     SWARM   DOCKER    ERRORS
aws-demo2   *        amazonec2    Running   tcp://10.0.6.20:2376            v1.10.3   
default     -        virtualbox   Stopped                                   Unknown   
qa3         -        amazonec2    Running   tcp://10.0.4.222:2376           v1.10.3   
uat         -        amazonec2    Running   tcp://10.0.4.23:2376            v1.11.0   
$ docker-machine --version
docker-machine version 0.7.0, build a650a40
$ docker-machine ssh uat
Welcome to Ubuntu 15.10 (GNU/Linux 4.2.0-35-generic x86_64)
ubuntu@uat:~$

bamarni:

I'm wondering if https://github.com/docker/machine/commit/be8f469a828ffd6ad8ca0baaef017c49e3bab713 shouldn't be reverted instead of merging my PR. It caused a few false positives and kind of "duplicates" the ssh binary behaviour. If the original intention was to have more helpful error messages, maybe the default could be to print the output when an SSH command is failing?

### DIFF
--- a/libmachine/ssh/client.go
+++ b/libmachine/ssh/client.go
@@ -332,9 +332,12 @@ func NewExternalClient(sshBinaryPath, user, host string, port int, auth *Auth) (
 				mode := fi.Mode()
 				log.Debugf("Using SSH private key: %s (%s)", privateKeyPath, mode)
 				// Private key file should have strict permissions
-				if mode != 0600 {
-					// Abort with correct message
-					return nil, fmt.Errorf("Permissions %#o for '%s' are too open.", mode, privateKeyPath)
+				perm := mode.Perm()
+				if perm&0400 == 0 {
+					return nil, fmt.Errorf("'%s' is not readable", privateKeyPath)
+				}
+				if perm&0077 != 0 {
+					return nil, fmt.Errorf("permissions %#o for '%s' are too open", perm, privateKeyPath)
 				}
 			}
 			args = append(args, "-i", privateKeyPath)
--- a/libmachine/ssh/client_test.go
+++ b/libmachine/ssh/client_test.go
@@ -1,6 +1,9 @@
 package ssh
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
 	"runtime"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


nathanleclaire:
The PR has definitely been causing nothing but headaches so far.  I think consideration should be given to a revert.

bamarni:
Me too 👍

I'll try to revert it and see if we still get meaningful messages for such errors.

bamarni:
Reverting it can also cause a headache or two 😜  

Once this is reverted, trying to create a machine with an invalid SSH setup leaves the command hanging for a few minutes, before it eventually fails with the following output :

```sh
Running pre-create checks...
Creating machine...
(docknic) No SSH key specified. Assuming an existing key at the default location.
```

We only get a generic retry limit error, not sure why exactly. But in any case the ssh client wouldn't give a useful message here because it's set to `quiet` (https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L71), we probably should set it to `error`.

Even though we fix it, it'd still hang for a few minutes before showing that message. I think there should be different retry timeouts depending on the driver action (waiting a few minutes for machine creation makes sense, but for SSH not that much). Let me know if you have other ideas on how to tackle this.

bamarni:
@nathanleclaire : did you make up your mind about what we should do here? Unfortunately I don't see any quick fix.. 😔

nathanleclaire:
This PR should basically resolve the issue, no?

I don't really recall from my reading of the above what the hangup about merging this might be (and it may have just dropped off my radar, so thanks for the ping).

Can you add a new test for the failing case to make sure the bug doesn't appear again please?

Assuming those are true, I'm game to merge.

bamarni:
Our previous discussion was more about how to fix it while dropping some code, but turned out to be trickier than expected... This PR would also solve the issue, I've added a a few test cases with different permission modes.

nathanleclaire:
LGTM

nathanleclaire:
Thanks @bamarni :)

Issue #3391, opened 2016-04-28
When provisioner needs to inject strings, it should use stdin


vincentbernat:
Hey!

The provisioner does stuff like that:

```
p.SSHCommand(fmt.Sprintf("printf %%s \"%s\" | sudo tee %s", dkrcfg.EngineOptions, dkrcfg.EngineOptionsPath))
```

This is not robust as if `dkrcfg.EngineOptions` contains a quote, it will just break horribly or may even execute dangerous stuff. Instead, we should do this:

```
p.SSHCommandWithStdin(fmt.Sprintf("sudo tee %s", dkrcfg.EngineOptionsPath, dkrcfg.EngineOptions)
```

I have tried to look at what should be done to do that, but accessing stdin would be quite invasive as it needs to be integrated in `libmachine/ssh/client.go` in the `Start` command.

Is there an interest in fixing such issues? If yes, should `Start` takes an additional argument or should an additional method, `StartWithStdin` be used?

nathanleclaire:
I think you would have to connect to `cmd.Stdin` and, as you noted, likely require a new way of invoking the command.  (Perhaps using something like [Cmd.StdinPipe](https://golang.org/pkg/os/exec/#Cmd.StdinPipe))

Yes, there is interest in fixing this, as you noted it will probably be a decent sized chunk, if you're feeling motivated, go for it!

vincentbernat:
I'll do it at some point, unfortunately, not right now. Next week maybe.

PR #3392, opened 2016-04-28
coreos: better docker integration with existing systemd unit

vincentbernat:
CoreOS is shipping a systemd unit starting Docker through socket
activation. The content of the file is the following (CoreOS 899):

```
[Unit]
Description=Docker Socket for the API
PartOf=docker.service

[Socket]
ListenStream=/var/run/docker.sock
SocketMode=0660
SocketUser=docker
SocketGroup=docker

[Install]
WantedBy=sockets.target
```

The service file provided by CoreOS is the following:

```
[Unit]
Description=Docker Application Container Engine
Documentation=http://docs.docker.com
After=docker.socket early-docker.target network.target
Requires=docker.socket early-docker.target

[Service]
EnvironmentFile=-/run/flannel_docker_opts.env
MountFlags=slave
LimitNOFILE=1048576
LimitNPROC=1048576
ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ

[Install]
WantedBy=multi-user.target
```

Docker Machine is setting up a regular systemd unit service file. The
content is the following:

```
[Unit]
Description=Docker Socket for the API
After=docker.socket early-docker.target network.target
Requires=docker.socket early-docker.target

[Service]
Environment=TMPDIR=/var/tmp
EnvironmentFile=-/run/flannel_docker_opts.env
MountFlags=slave
LimitNOFILE=1048576
LimitNPROC=1048576
ExecStart=/usr/lib/coreos/dockerd daemon --host=unix:///var/run/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=exoscale $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
Environment=

[Install]
WantedBy=multi-user.target
```

It's mostly a copy/paste of the original service. There are multiple problems:
1. We have two processes listening to /var/run/docker.sock: systemd
   through socket activation and Docker itself. A few tests show that
   Docker gets all the requests but it could be luck and break at some
   point or under some conditions.
2. The unit setup by Docker Machine is not enabled on start. After a
   reboot, Docker is not started anymore. If we trigger it locally, it
   starts but blocks because it doesn't use socket activation. Further
   invocations will then work.
3. The environment is in fact empty because of the empty `Environment=`
   at the end.

Those commits integrates Docker a bit better with CoreOS by solving the
first and last point (first commit) and the middle one (last commit).
We cannot do socket activation with TLS because in
stable channels, CoreOS ships only Docker 1.9. Therefore, we keep socket
activation for the Unix socket but we use a regular socket for the TLS
one. We override the existing service to avoid repeat what's in it. We
add /etc/systemd/system/docker.service.d/machine.conf:

```
[Service]
Environment="DOCKER_OPTS=--host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=exoscale"
```

In the future (Docker 1.10+), we could add
/etc/systemd/system/docker.socket.d/machine.conf:

```
[Socket]
ListenStream=0.0.0.0:2376
```

And only append `Environment` to the `.service` file.

Note that the way the systemd unit is passed over SSH is quite
fragile. Notably, characters need to be escaped properly and a double
quotes should not be present. This can be problematic for the
environment variables. See #3391.

### DIFF
--- a/libmachine/provision/systemd.go
+++ b/libmachine/provision/systemd.go
@@ -22,7 +22,7 @@ func NewSystemdProvisioner(osReleaseID string, d drivers.Driver) SystemdProvisio
 		GenericProvisioner{
 			SSHCommander:      GenericSSHCommander{Driver: d},
 			DockerOptionsDir:  "/etc/docker",
-			DaemonOptionsFile: "/etc/systemd/system/docker.service",
+			DaemonOptionsFile: "/etc/systemd/system/docker.service.d/machine.conf",
 			OsReleaseID:       osReleaseID,
 			Packages: []string{
 				"curl",
@@ -41,15 +41,11 @@ func (p *SystemdProvisioner) GenerateDockerOptions(dockerPort int) (*DockerOptio
 	p.EngineOptions.Labels = append(p.EngineOptions.Labels, driverNameLabel)
 
 	engineConfigTmpl := `[Service]
-ExecStart=/usr/bin/docker daemon -H tcp://0.0.0.0:{{.DockerPort}} -H unix:///var/run/docker.sock --storage-driver {{.EngineOptions.StorageDriver}} --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
-MountFlags=slave
-LimitNOFILE=1048576
-LimitNPROC=1048576
-LimitCORE=infinity
-Environment={{range .EngineOptions.Env}}{{ printf "%q" . }} {{end}}
-
-[Install]
-WantedBy=multi-user.target
+ExecStart=
+ExecStart=/usr/bin/docker daemon -H fd:// -H tcp://0.0.0.0:{{.DockerPort}} --storage-driver {{.EngineOptions.StorageDriver}} --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
+{{range .EngineOptions.Env}}
+{{ printf "Environment=%q" . }}
+{{end}}
 `
 	t, err := template.New("engineConfig").Parse(engineConfigTmpl)
 	if err != nil {
--- a/libmachine/provision/coreos.go
+++ b/libmachine/provision/coreos.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/machine/libmachine/engine"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/provision/pkgaction"
+	"github.com/docker/machine/libmachine/provision/serviceaction"
 	"github.com/docker/machine/libmachine/swarm"
 )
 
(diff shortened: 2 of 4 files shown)


vincentbernat:
Note that the same thing can be done for Ubuntu. The installed package now comes with a systemd unit and it seems a bit different (`Delegate=yes`). I'll add a commit similar to the first one but using only socket activation. Currently, my change breaks Ubuntu.

vincentbernat:
So, I did the modification for Ubuntu too. It's similar. Not easy to use socket activation with Ubuntu due to the need to modify two files.

bjaglin:
Nice, I was working on the exact same thing in https://github.com/docker/machine/commit/42254e5bf3e91698f822580342b35cc5c50f4af1.

My own motivation was that things like https://github.com/coreos/coreos-overlay/pull/1912 would need to be replicated in the current copy/paste approach - until that CoreOS Alpha is pretty much unusable with machine.

Since this PR is already open, I'll comment here my findings so far.

libmachine/provision/systemd.go
```diff
@@ -22,7 +22,7 @@ func NewSystemdProvisioner(osReleaseID string, d drivers.Driver) SystemdProvisio
 		GenericProvisioner{
 			SSHCommander:      GenericSSHCommander{Driver: d},
 			DockerOptionsDir:  "/etc/docker",
-			DaemonOptionsFile: "/etc/systemd/system/docker.service",
+			DaemonOptionsFile: "/etc/systemd/system/docker.service.d/machine.conf",
```
bjaglin: This has a side effect on the `redhat` provisioner as it currently verrides `GenerateDockerOptions` to override the system unit (rather than a drop-in) but references this var. I haven't tested yet https://github.com/bjaglin/machine/commit/ebe05412a09778f5f5e396c8c40fada29b556235, but simply removing `GenerateDockerOptions` looked promising.

vincentbernat: Yes, I did miss the redhat provisioner. Can you test it? I can push an additional commit to do as you say.

bjaglin: I can probably run the tests on a Redhat AMI with the EC2 driver, but maybe this can/will be triggered automatically on the PR?

vincentbernat:
I did test with CoreOS and the provisioning works fine too.

vincentbernat:
@nathanleclaire This PR is ready to be reviewed.

libmachine/provision/systemd.go
nathanleclaire: It will not be triggered automatically - we would appreciate the help testing!

nathanleclaire:
Code looks pretty good, curious to hear back from @bjaglin, did you test on Ubuntu as well?

libmachine/provision/systemd.go
vincentbernat: I did test with CoreOS 899 stable and with CentOS 7.

vincentbernat:
Works fine with Ubuntu 16.04 as well.

vincentbernat:
In the meantime, #3870 has been accepted while this PR could have solved the problem in a more robust way. I can work again on this PR if there is some consensus this is the right way forward.

nathanleclaire:
@vincentbernat Sure, open to ideas, and certainly not an expert on CoreOS / systemd, so I appreciate the help. Sorry to have this one lag if you feel it overlaps with the other PR, I know that feels bad, the other was more on my periphery since I was getting pinged on it and didn't realize that they may overlap.  How do you feel this one should move forward?

vincentbernat:
This PR makes only sense if you agree to let the "distributor" (CoreOS for CoreOS, Docker for Redhat/CentOS/Ubuntu) decide how to properly exec the docker daemon. Docker Machine would only "insert" the additional parameters it needs for authentication and external access.

If yes, I would modify this PR by reverting #3870, resolve conflcts, test with a CoreOS with Docker 1.10, CoreOS with Docker 1.12 (just to validate this works for both CoreOS versions, like with #3870), Ubuntu 16.04 and CentOS 7.

Letting the distributor decides the best way to execute the docker daemon has the following advantages:

 - the distributor usually knows how to run the docker daemon it provides (if this is Docker 1.10, it does that, if this is Docker 1.12, it does that)
 - the distributor knows better how to best integrate into the system (user settings, security, paths, ...)
 - less code in Docker Machine

The only drawback I could see is that it may break if the distributor decides to remove the environment variable Docker Machine relies on for its integration. This is quite possible, but since the main distributor (Docker) is near you, it should be easy to sort that out.

vincentbernat:
Well, #3966 implements something similar, in a less robust way since the kludge introduced in #3870 has to be kept. There is no point keeping this PR opened as there is clearly no path forward.

PR #3395, opened 2016-04-28
azure: add link to Azure.com for Azure Examples

ahmetalpbalkan:
Linking to our docs website for Azure examples.

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -15,3 +15,6 @@ weight="-50"
 
 -   [Digital Ocean Example](ocean.md)
 -   [AWS Example](aws.md)
+-   [Microsoft Azure Example][azure-tutorial]
+
+[azure-tutorial]: https://azure.microsoft.com/documentation/articles/virtual-machines-linux-docker-machine/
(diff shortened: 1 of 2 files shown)


squillace:
+1: that's a good idea, since we have that as a straightforward walkthrough already, and it points back to more docker goodness.

ahmetalpbalkan:
and our website is more pretty 💁

nathanleclaire:
I think we might have a policy on this to only link to docs.docker.com links.

@moxiegirl @londoncalling what do you think?

londoncalling:
@nathanleclaire @moxiegirl we did get a specific request from @shykes to link to our own examples and tutorials in our docs. We spent some time in recent releases adding new examples, including the AWS and Digital Ocean ones in these docs. If you guys, together, come to a different decision, I'm fine with it, but right now the Azure example does break the current model, given the re-write of the Machine docs.

(also looping in @dmp42 )

nathanleclaire:
Mm, for consistency, I think we should only link to the Docker website (we can actually take responsibility for it, for one -- we've had external links die on us before).  Feel free to write a guide for our docs though.  Thanks

londoncalling:
Hey Nathan, I can write one too if you like. I can base it on this one or
some other , and I'll send it to you guys for review .--Vicky

PR #3406, opened 2016-05-04
Align machine storage path flag name

ronakbanka:
Signed-off-by: Ronak Banka <ronak.banka@rakuten.com>

Short form for storage path was added in https://github.com/docker/machine/commit/4e9fc9c248216496a0b4ef8d857d6144878f540d but is not aligned with other flag names.


nathanleclaire:
LGTM

PR #3410, opened 2016-05-05
Machine-specific persistent option to specify SSH config

brettdh:
@nathanleclaire

This adds two flags to `docker-machine create`:

    --generic-ssh-config-file <path/to/config>

        Use the provided config file with the external SSH client (-F)

    --generic-use-user-ssh-config

        Use ~/.ssh/config as the config file

(The flags are mutually exclusive in order to avoid ambiguity.)

Tried this out in my environment, which requires a `ProxyCommand` to
access a remote site via SSH. Before, the issue described in #3310
occurred, preventing me from using `docker-machine` to configure this
host. With this, the `--generic-use-user-ssh-config` option adds
the config file in place of the previous-default `/dev/null` and all
seems well.

Also added a couple unit test cases.

Depends on an updated version of #3209.
Note that this PR includes brettdh/machine@d39de4f
(which is just #3209 rebased on master).

The merge will be a bit messy, since this will get rebased on whatever #3209
becomes, but at least now it can be reviewed in relative isolation.
I considered setting the base for this PR on brettdh/machine@d39de4f,
but I'm pretty sure that would make a PR on my fork, rather than on docker/machine.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,6 +1,7 @@
 {
 	"ImportPath": "github.com/docker/machine",
 	"GoVersion": "go1.6",
+	"GodepVersion": "v62",
 	"Packages": [
 		"github.com/docker/machine",
 		"github.com/docker/machine/cmd",
@@ -48,6 +49,7 @@
 		"github.com/docker/machine/libmachine/mcnerror",
 		"github.com/docker/machine/libmachine/mcnflag",
 		"github.com/docker/machine/libmachine/mcnutils",
+		"github.com/docker/machine/libmachine/opt",
 		"github.com/docker/machine/libmachine/persist",
 		"github.com/docker/machine/libmachine/persist/persisttest",
 		"github.com/docker/machine/libmachine/provider",
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -158,6 +158,17 @@ func main() {
 			Usage:  "BugSnag API token for crash reporting",
 			Value:  "",
 		},
+		cli.StringFlag{
+			EnvVar: "MACHINE_SSH_CONFIG_FILE",
+			Name:   "ssh-config-file",
+			Usage:  "Use the specified SSH config file",
+			Value:  "",
+		},
+		cli.BoolFlag{
+			EnvVar: "MACHINE_USE_USER_SSH_CONFIG",
+			Name:   "use-user-ssh-config",
+			Usage:  "Use the user's SSH config file (~/.ssh/config)",
+		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -158,17 +158,6 @@ func main() {
 			Usage:  "BugSnag API token for crash reporting",
 			Value:  "",
 		},
-		cli.StringFlag{
-			EnvVar: "MACHINE_SSH_CONFIG_FILE",
-			Name:   "ssh-config-file",
-			Usage:  "Use the specified SSH config file",
-			Value:  "",
-		},
-		cli.BoolFlag{
-			EnvVar: "MACHINE_USE_USER_SSH_CONFIG",
-			Name:   "use-user-ssh-config",
-			Usage:  "Use the user's SSH config file (~/.ssh/config)",
-		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/codegangsta/cli"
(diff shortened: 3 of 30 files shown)


brettdh:
Gordon is a little excitable about a missing sign-off, it seems.

brettdh:
@nathanleclaire any further thoughts on this and/or #3209?

bamarni:
Is `--generic-use-user-ssh-config` really needed? As it's seems equivalent to `--generic-ssh-config-file ~/.ssh/config`.

brettdh:
@bamarni yep, that's correct. Just a shortcut (barely).

lox:
Any progress on this? This will block anyone with a slightly complicated SSH setup.

lox:
@brettdh I'd happily take this over if you aren't inclined.

brettdh:
@lox rebased and resolved conflicts. I'd say this is still in @nathanleclaire 's court, though - I don't know what he wanted to do about #3209, so that conversation was blocking this, IIRC.

lox:
Yup fair enough. Merging this would at least let the majority of users who need ssh proxying get on with things though, as it stands machine just doesn't work for us.

brettdh:
Good news is that docker-machine is tremendously easy to build, so if you
need this today you can always go that route, as I've done at my company.

lox:
I'm fairly allergic to distributing custom builds of docker things internally, but yup, fair point.

nathanleclaire:
Here's some thinking on this.  Sorry it's been slow going but it's a bit riskier change than most.

- it's good feature and i'd like to support it if possible, additionally fixing the long-standing propagating of options bug would be nice
- we could probably avoid bumping the API version and just print a warning like we've done here, since we are just adding methods, not changing or removing existing ones, but this should be tested to make sure it behaves as expected.  maybe we could treat this as experiment since api version bump / modification hasn't really been done before
- @brettdh  you can't [edit vendored code directly](https://github.com/docker/machine/pull/3410/files#diff-d93d18405022dacc9ebd59909ac5a9ba) but it seems you could just set `HTTPClient.Proxy` on the returned client where we generate the code.
- what's the utility of `GetMachineOptions` in general?  I would expect to just call `SetMachineOptions` for each server driver... It's a one-way operation dictating what those options should be from the main binary/flags, right?
- isn't [this](https://github.com/docker/machine/pull/3410/files#diff-c40a162070d4aff6c77a9a75563701a7R55) a bit risky?  have you tested invoking this code on hosts originally created with 0.8.2?  i think it will need either a proper host migration or some other way to assure safety, the nil checking seems a bit off to me... i guess it's so you don't have to specify the ssh config file each invocation? (i agree our lack of a default config generally, which would help to avoid having to set the flags each time, is annoying).
- why `--use-user-ssh-config` when they could just specify that themselves with the other new flag?

thanks for your persistence and interest in this.

lox:
Agree with dropping `--use-user-ssh-config`, seems like unneeded syntactic sugar.

brettdh:
> @brettdh you can't edit vendored code directly but it seems you could just set HTTPClient.Proxy on the returned client where we generate the code.

I'm taking a crack at this, but my golang is rusty, and I feel like I'm doing something rather dumb - or just not fully understanding the details of `http.Transport` and related bits.

> what's the utility of GetMachineOptions in general? I would expect to just call SetMachineOptions for each server driver... It's a one-way operation dictating what those options should be from the main binary/flags, right?

I think you answered your own question in the next bit; we need to be able to retrieve the saved options because the user isn't providing this option on every invocation.

...at least, that would have been my answer a few minutes ago. Looking more carefully, it looks like `GetMachineOptions` is only called during `create`, and it only ever returns the `defaultOptions` from `libmachine/opt/options.go`. So that's pretty silly. I'll rip it out and just call `mcnopt.Opts()` directly.

> isn't this a bit risky? have you tested invoking this code on hosts originally created with 0.8.2? i think it will need either a proper host migration or some other way to assure safety, the nil checking seems a bit off to me... i guess it's so you don't have to specify the ssh config file each invocation? (i agree our lack of a default config generally, which would help to avoid having to set the flags each time, is annoying).

Maybe? It's been a while since I tested it, but I thought that the missing value would end up with a value of `nil`, since it's just saved as JSON.

Giving this a quick test: with a machine created with 0.8.1, and using this PR's build to talk to it, `docker-machine inspect` shows `SSHOptions: null`, and `docker-machine ssh` works fine. (Interestingly, using 0.8.1  to `docker-machine inspect` a machine created with this PR's build also works fine; it just doesn't show `SSHOptions`.)

> why --use-user-ssh-config when they could just specify that themselves with the other new flag?

This has now been brought up by three different people, so I give. 😅 
I might like to try just having `--generic-ssh-config-file` having a default value of `$HOME/.ssh/config` if it's given without a value, but it appears that's tricky, because (I think) you get `""` for its value whether it's provided without a value or .

libmachine/mcndockerclient/docker_client.go
```diff
@@ -19,7 +21,13 @@ func DockerClient(dockerHost DockerHost) (*dockerclient.DockerClient, error) {
 		return nil, fmt.Errorf("Unable to read TLS config: %s", err)
 	}
 
-	return dockerclient.NewDockerClient(url, tlsConfig)
+	client, err := dockerclient.NewDockerClient(url, tlsConfig)
+	if err != nil {
+		transport := (client.HTTPClient.Transport).(*http.Transport)
+		transport.Proxy = http.ProxyFromEnvironment
```
brettdh: This is the part I'm uncertain about - whether it is sane to update the proxy setting in this way.

brettdh:
@nathanleclaire any further thoughts on this? I think I've addressed all your concerns stated above, except perhaps the API version non-bump testing that you mentioned, as I'm not sure how to go about that.

brettdh:
@nathanleclaire had another coworker get bit by the lack of this feature today. Any chance we could take another look at this soon?

bilts:
Any status on this? I can't use docker-machine because of it.

nathanleclaire:
I am not maintainer any more -- ping @shin-

brettdh:
And I'm not using docker-machine these days, so @lox if you still want to take it over, be my guest :)

davidneudorfer:
@shin- Any thoughts on this? Thanks!

EvgeniiProkofevItomych:
@shin this makes impossible to use GitLab CI autoscale approach within isolated network

GuSuku:
What is the status of this? This is currently a big impediment for connecting to remotes beyond bastions.

letalvoj:
Could please any _maintainer_ summarize the changes needed to get this to master? Thanks 👍

josephwibowo:
Is there an update on this? 
How can I use docker-machine if I'm using Bastion and not able to change the ssh config?

PR #3423, opened 2016-05-11
Fix port binding of swarm master

npcode:
The port is 3376 by default but could be modified by --swarm-host
option.

Signed-off-by: Yi EungJun <eungjun.yi@navercorp.com>


npcode:
I am also worry about advertise port which is fixed to 3376...

nathanleclaire:
LGTM

PR #3444, opened 2016-05-18
Add HPE Linux provisioner, based on Debian provisioner

tpot:
Signed-off-by: Tim Potter <tpot@hpe.com>

### DIFF
--- a/libmachine/provision/hpelinux.go
+++ b/libmachine/provision/hpelinux.go
@@ -1,0 +1,143 @@
+package provision
+
+import (
+	"fmt"
+
+	"github.com/docker/machine/libmachine/auth"
+	"github.com/docker/machine/libmachine/drivers"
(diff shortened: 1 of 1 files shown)


tpot:
Hi there.  I'm still messing with this, but am just wondering whether any extra tests would be required as HPE Linux is a Debian derivative.  In a similar vein, I was thinking that maybe there should be some inheritance here rather than cut&pasting the content of libmachine/provision/debian.go. 

Thanks!

nathanleclaire:
Think it can just embed the default Debian provisioner?  What exactly has changed in this one vs. the default Debian provisioner?  Could it follow the pattern of the existing `centos` provisioner which simply calls RH provisioner but identifies by a different ID string?

nathanleclaire:
Hi @tpot, going to close this due to lack of response, please feel free to ping for re-open or open a new PR if you'd like to continue

PR #3463, opened 2016-05-28
Adding an optional VM's UI type for the virtualbox driver

malex984:
With this option one can specify the VirtualBox UI Type: "gui"/"sdl"/"separate" instead of the default "headless"

NOTE: running X11 server is required if UI Type is not "headless"

### DIFF
--- a/drivers/virtualbox/virtualbox.go
+++ b/drivers/virtualbox/virtualbox.go
@@ -28,6 +28,7 @@ const (
 	defaultHostOnlyCIDR        = "192.168.99.1/24"
 	defaultHostOnlyNictype     = "82540EM"
 	defaultHostOnlyPromiscMode = "deny"
+	defaultUIType              = "headless"
 	defaultDiskSize            = 20000
 	defaultDNSProxy            = true
 	defaultDNSResolver         = false
@@ -62,6 +63,7 @@ type Driver struct {
 	HostOnlyCIDR        string
 	HostOnlyNicType     string
 	HostOnlyPromiscMode string
+	UIType              string
 	NoShare             bool
 	DNSProxy            bool
 	NoVTXCheck          bool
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Hi @malex984, thanks for the PR.  What's the motivation?  We are strongly inclined to be prescriptive about using headless VMs here.

malex984:
Hi @nathanleclaire, our main purpose is to test containerized GUI applications (e.g. a-la https://blog.jessfraz.com/post/docker-containers-on-the-desktop/) on minimal linux systems (only VirtualBox Guest Additions are necessary on the guest system while X11 Server may run in a container). Such a testing should be possible on OSX and Linux desktop systems.
Previously we did that with vagrant but docker-machine should be able to do that as well after a few minor improvements.

Since even in the headless mode there is a VM Display Preview available from inside VirtualBox Manager it seems that headless mode simply disables a separate VM Display Window...

On the other hand having that Display Window provides user with a BOOT loader interactive menu, boot/provision output log as well as an extra interactive console and an option to have GUI applications running within such a VM.

Do you see any problems with such a usage of Docker Machine on desktop systems?

nathanleclaire:
It is a bit different than the usual prescriptive workflow but since it's optional and the default doesn't change it doesn't seem likely to be problematic.

Do you have any little sample app I could try it out?

malex984:
Yes, this is an optional thing which changes nothing for people that do not use it.

The following will starts X11 server and xclock in the created VirtualBox Display Window for me under Linux:
```
dm create -d virtualbox --virtualbox-ui-type gui VBGUI
eval $(dm env VBGUI) # or `dm ssh VBGUI`

docker run -d --privileged alpine:3.4 /bin/sh -c \
'apk update && apk add xorg-server xf86-video-vesa xf86-input-evdev xf86-input-mouse xf86-input-keyboard udev && echo "exec xclock">>~/.xinitrc && apk add xclock  --update-cache --repository "http://dl-cdn.alpinelinux.org/alpine/edge/testing" && startx'
```

where dm is my custom built docker-machine supporting `--virtualbox-ui-type`. 

ps: unfortunately it seems that mouse support has been disabled :-( i suspect VirtualBox Guest Additions, Linux Kernel or missing kernel modules...

malex984:
Hi! Is something else wrong with this PR?

nathanleclaire:
@malex984 Nah I think it's OK, just got lost in the shuffle.

LGTM

malex984:
thanks!

PR #3494, opened 2016-06-10
drivers: vmwarefusion: fix CPU count

philips:
The current default for CPU count is 0. At some point it looks like
someone started using CPU instead of CPUS field
(13c1006f3bc19d8db3468ffaa257c08cd25e3ace) but things got messed up
later.


philips:
I don't understand what the problem is with the DCO. This is a trivial fix if someone else wants to pick it up.

philips:
Closing in favor of #3503.

PR #3509, opened 2016-06-14
Extends 60 attemps to config

sebadiaz:
Extends 60 attemps to config


bamarni:
What is the motivation for this? Do you think it's not high enough?

sebadiaz:
In fact during some test with intensive AWS instance demand, the number of "connection reset by peer" are very frequent. Causing invalid connection and a non stable docker machine status where machine removing become not good (non clean of keys,...) .  The target is to have the best smoothly machine deployment instead the time it could take.

bamarni:
I'm not really familiar with AWS but changing this value is unlikely to help, it is already something very high. Maybe you'd be more interested by this https://github.com/docker/machine/commit/770d5bbb6f493eff979f24bb0c71e72c17e05b31.

sebadiaz:
@bamarni this bug is absolutely not the same thing
In fact and If you have a look on the code the time out will 60 * 3s =180 s --> It is just 3 minutes of delay that is really short when you demand for a big virtual machine on a big cloud. Searching make more time due to the scarcity of this kind of VM.

bamarni:
Ah ok, I thought you were talking about an issue during machine removal.

libmachine/mcnutils/utils.go
```diff
@@ -99,7 +99,13 @@ func WaitForSpecific(f func() bool, maxAttempts int, waitInterval time.Duration)
 }
 
 func WaitFor(f func() bool) error {
-	return WaitForSpecific(f, 60, 3*time.Second)
+	i, err := strconv.Atoi(os.Getenv("MAX_ATTEMPTS"))
+    	if err != nil {
+        	return WaitForSpecific(f, 60, 3*time.Second)
```
bamarni: this could be replaced by `i = 60` to avoid code duplication.

nathanleclaire:
Hi @sebadiaz, thanks for the PR.  If you would like to implement such functionality it will need to be done with a proper flag and propagated through the system via CLI though.

sebadiaz:
@nathanleclaire I make a parameter flag

sebadiaz:
@nathanleclaire Could you add in gordonTheturtle inspection for signature the special git option for merge during the rebase
git rebase -i HEAD~9 --preserve-merges

Thanks

docs/reference/create.md
```diff
@@ -51,6 +51,7 @@ customize.
        --engine-registry-mirror [--engine-registry-mirror option --engine-registry-mirror option]           Specify registry mirrors to use [$ENGINE_REGISTRY_MIRROR]
        --engine-label [--engine-label option --engine-label option]                                         Specify labels for the created engine
        --engine-storage-driver                                                                              Specify a storage driver to use with the engine
+       --engine-max-attempts                                                                                Specify the number of attempts allowed on waiting period
```
bamarni: I don't think the flag should have the engine prefix (as it's not related to Docker engine's installation / configuration).

libmachine/drivers/base.go
```diff
@@ -9,6 +9,7 @@ const (
 	DefaultSSHUser          = "root"
 	DefaultSSHPort          = 22
 	DefaultEngineInstallURL = "https://get.docker.com"
+	MaxAttempt              = 60
```
bamarni: `DefaultMaxAttempt`?

libmachine/drivers/base.go
```diff
@@ -35,6 +37,14 @@ func (d *BaseDriver) GetMachineName() string {
 	return d.MachineName
 }
 
+// GetMaxAttempt returns the maximum attempt
+func (d *BaseDriver) GetMaxAttempt() int {
+	if d.MaxAttempt == 0 {
+		d.MaxAttempt = DefaultSSHPort
```
bamarni: `d.MaxAttempt = DefaultMaxAttempt` (assuming you rename the constant)

libmachine/drivers/serial_test.go
```diff
@@ -62,6 +63,14 @@ func (d *MockDriver) GetIP() (string, error) {
 	return d.ip, nil
 }
 
+// GetMachineName returns the machine name
+func (d *MockDriver) GetMaxAttempt() int {
+	if d.maxAttempt == 0 {
+		d.maxAttempt = DefaultSSHPort
```
bamarni: same here

libmachine/ssh/client.go
```diff
@@ -89,18 +90,18 @@ func SetDefaultClient(clientType ClientType) {
 	}
 }
 
-func NewClient(user string, host string, port int, auth *Auth) (Client, error) {
+func NewClient(user string, host string, port int, maxAttempt int, auth *Auth) (Client, error) {
```
bamarni: Not sure if there is an easy way around it, but I find it strange to have this setting at the ssh client level, feels more like external logic.

bamarni:
@nathanleclaire : as there would be an explicitly documented flag, should we also mention that except for rare cases it shouldn't be changed? It feels like some docker machine internals and might be confusing to some.

sebadiaz:
@bamarni corrected .

Thank you. you awake me

sebadiaz:
Just a workaround
while test "`docker-machine ls --format "{{.Name}}:{{.Error}}" |grep osnode0|cut -d':' -f 2|wc -w`" != "0" ;do docker-machine rm osnode0;docker-machine create  --openstack-ssh-user centos --openstack-active-timeout 200000 --openstack-sec-groups default --openstack-password XXX --driver openstack --openstack-tenant-name fabric --openstack-image-name centos7 --openstack-auth-url http://xxxxxx:5000/v2.0 --openstack-net-name fabric --openstack-floatingip-pool public --openstack-flavor-name m1.2xlarge --openstack-username fabric --engine-storage-driver overlay osnode0;done

sebadiaz:
My last test on overloaded Amazon

XXXXXX --amazonec2-region us-west-2 --amazonec2-root-size 20 --amazonec2-zone b awsnode0
Running pre-create checks...
Creating machine...
(awsnode0) Launching instance...

lavalamp-:
<3 you guys

nathanleclaire:
Thanks for the contribution on this @sebadiaz -- I think it's too wide-ranging of a change to merge in this form, but how would you feel about kicking up the defaults.  Do you have a comfortable level in mind?

sebadiaz:
@nathanleclaire . 

The issue that this level is really to short if you are a AWS consumer. Reserving a virtual machine (not a litlle) could make more than 10 minutes(depending the region, the image, the disk space,...). 
Actualy the timeout will occur in waiting a less 3 minutes for the SSH connection that is the first step to install the docker host. if we multiply by 5 , I expect to have enought time to have. 60->300

Do you think that it will be acceptable ?

sebadiaz:
I can make this simple change on a pull request.

bamarni:
@sebadiaz @nathanleclaire : If I may suggest, I think this should be tackled differently. The main problem with the current ssh retries is that we don't know what we're waiting for, is the machine still getting created or did the user provided invalid flags? If we raise max attempts to 300, people in the latter situation will have to wait 12 more minutes before it eventually fails and they know they should investigate, which can become pretty frustrating. With coreos for example it's a common mistake to forget to specify the "coreos" ssh user. Personally each time I reached the timeout it was also because of a mistake from my end.

An alternative would be to poll for machine status before returning from the driver create function. Just looking at amazon and digitalocean drivers it seems fairly easy to do, as there are already some sort of polling (to get the IP address). In the case of digitalocean for instance, it's just a one liner change [here](https://github.com/docker/machine/blob/6c39ec9e3fabf93e82162f2d9574478dadda2768/drivers/digitalocean/digitalocean.go#L227). Then we could get rid of ssh retries, or at least keep a very low value.

Issue #3506, opened 2016-06-14
Docker Compose and Overlay Networks require port 4789


vvhof:
When trying to deploy multiple containers via docker-compose spread across multiple AWS EC2 instances, an overlay network is created to enable cross-host inner-container communication.

This requires TCP port 4789 to be opened.

Proposition:
Include a rule for TCP 4789 0.0.0.0/0 in the created Security Group.
https://github.com/docker/machine/blob/master/drivers/amazonec2/amazonec2.go

kacole2:
@vvhof would you also want port 7946 opened as well? https://docs.docker.com/engine/userguide/networking/dockernetworks/

vvhof:
@kacole2 Interestingly for the key-store  Consul that we utilized in our experiments, opening the TCP port 4789 for the **public** network proofed adequate. I may add, that we had allowed all traffic on the private IP network, which may explain while this was sufficient.

It would probably be a good course of action to abstract from specific key-value store implementations and instead just open the ports Docker, Inc. specifies? Maybe implementation wise a flag for the ec2 driver would be the solution. But i'll leave this up to debate.

kacole2:
@vvhof things are going to change slightly with 1.12 https://github.com/docker/docker/pull/23606 so the 7946 probably isn't necessary. not sure sure what ports are required for swarmKit

kacole2:
@vvhof I think a flag for `--overlay` could be done but would have to be at a global level and not for this specific driver. It would need to be the same way `--swarm-master` is done. Also, `--overlay` as a global flag should mean more than just opening a port. So not sure what should be done. I made a build for it anyway.

PR #3515, opened 2016-06-16
Added overlay networking port for AWS security profile

kacole2:
This is based on a request for #3506. Added in port 4789/udp for the security profile on AWS and fixed the tests so they all pass.

Signed-off-by: Kendrick Coleman <kendrickcoleman@gmail.com>

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -48,6 +48,7 @@ const (
 var (
 	dockerPort                  = 2376
 	swarmPort                   = 3376
+	overlayNetworkPort          = 4789
 	errorMissingAccessKeyOption = errors.New("amazonec2 driver requires the --amazonec2-access-key option or proper credentials in ~/.aws/credentials")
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Nice @kacole2 .  I think it needs `7946` for Serf as well though.

How will this be enabled?

Additionally, I think it should only be open to instances in same VPC, not publicly to the world, right?

kacole2:
@nathanleclaire adding in 7946 can be done on the security profile pretty easily. 

As a counter to not be public... what about the idea of cross-cloud Swarms? I might be mistaken, but I thought you could use overlay networking to have hosts on AWS & Digital Ocean all in the same Swarm. No?

nathanleclaire:
> As a counter to not be public... what about the idea of cross-cloud Swarms? I might be mistaken, but I thought you could use overlay networking to have hosts on AWS & Digital Ocean all in the same Swarm. No?

Well, it's security issue.  Not really safe to allow inbound VXLan from basically anywhere.

kacole2:
Security puts the "no" in innovation. 

I see your point. I'll close this out because it's not going to go anywhere. 

This convo can continue in #3506

nathanleclaire:
Thanks @kacole2

PR #3535, opened 2016-06-23
bumped version up to fedora 23 (should'nt it use a matching versi…

tronicum:
…on and only default to something else otherwise?)

Signed-off-by: tronicum <tronicum@gmail.com>


nathanleclaire:
23 is latest I think.

LGTM

cdrage:
@nathanleclaire 24 is the latest now (https://getfedora.org/en/server/download/) should we bump it up again?

tronicum:
It refers to, https://yum.dockerproject.org/repo/main/fedora/ where 23 is the latest build.

PR #3549, opened 2016-06-27
Use machine name from environment variable if available

denisvm:
If available, use `MACHINE_NAME` environment variable to make it easier to run docker-machine commands when not using the default machine name `default`.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -79,13 +79,18 @@ func (c *contextCommandLine) Application() *cli.App {
 // arg, or the default host name if no host is specified.
 func targetHost(c CommandLine, api libmachine.API) (string, error) {
 	if len(c.Args()) == 0 {
-		defaultExists, err := api.Exists(defaultMachineName)
+		machineName := os.Getenv("MACHINE_NAME")
+		if machineName == "" {
+			machineName = defaultMachineName
+		}
+
+		defaultExists, err := api.Exists(machineName)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


denisvm:
Would it be necessary/interesting to add a test for this small change? Thanks!

nathanleclaire:
Thanks for the contribution @denisvm .  I think it would be too confusing to have the short form commands resolve to a host other than `default` though.  We had a situation like this before with machine and it was very difficult to tell what was supposed to be happening in output that folks were posting.

denisvm:
Thanks for your feedback, @nathanleclaire! I thought this patch would be interesting, for instance, when used with vmwarefusion driver, because the machine is also displayed in the VMware Fusion Library, and the name 'default' could be confusing, and if used other name than that, you would loose the short form commands, that are very handy.
For users who would like to use another default name, it would be easier to keep it organised and make the short form available for them as an option. In those special cases the default debug option could be enough to investigate more, or just make a verbose mode with less detailed information. What you think?

nathanleclaire:
Thanks for the contribution @denisvm .  Unfortunately I do not think it is in the best interests of the project to merge this feature.  We really need simplicity and predictable behavior first and foremost and that can't be guaranteed if whenever someone copy-pastes `docker-machine` commands to demonstrate something to me I have a guessing game about which machine they're targeting based on environment variables.  Thanks for your contribution.

PR #3552, opened 2016-06-28
Change VirtualBox on Windows base paths to use long path prefix

juanjose49:
*The current implementation for the Windows VirtualBox driver maps
directly to the 'c:\\Users' directory. This regularly causes exceptions
when a file is written to a volume on a Windows machine where there is a
limitation on the number of characters allowed in a file path.
*By instead using the proposed mapping of '\\\\?\\c:\\Users' the Windows
API will allow the file path to exceed the character limit and should
resolve issues and cryptic errors for many users that are using Docker
Machine to build artifacts with lengthy file paths.

### DIFF
--- a/drivers/virtualbox/virtualbox_windows.go
+++ b/drivers/virtualbox/virtualbox_windows.go
@@ -91,7 +91,7 @@ func findVBoxInstallDirInRegistry() (string, error) {
 }
 
 func getShareDriveAndName() (string, string) {
-	return "c/Users", "c:\\Users"
+	return "c/Users", "\\?\c:\\Users"
 }
 
 func isHyperVInstalled() bool {
--- a/drivers/virtualbox/virtualbox_windows.go
+++ b/drivers/virtualbox/virtualbox_windows.go
@@ -91,7 +91,7 @@ func findVBoxInstallDirInRegistry() (string, error) {
 }
 
 func getShareDriveAndName() (string, string) {
-	return "c/Users", "\\?\c:\\Users"
+	return "c/Users", "\\\\?\\c:\\Users"
 }
 
 func isHyperVInstalled() bool {


nathanleclaire:
Can you please reference some information / documentation about this alternative path?

juanjose49:
[Support Ticket at VirtualBox](https://www.virtualbox.org/ticket/11976)
- This support ticket describes the exact issue and the solution I've been using on my local machine to workaround the issue.

[Microsoft Documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx)
- This article describes the general limitations of the file paths in Windows and describes the workaround that is proposed in the VirtualBox support ticket and in this Pull Request. Specifically, review the section called "Maximum Path Length Limitation".

nathanleclaire:
thanks -- will take a look

nathanleclaire:
Should this be done in all cases though?  Is there a way we can check the total path length and only enable this when needed?

juanjose49:
I think it has to be done for all cases due to the fact that docker-machine
is relying on VirtualBox's shared folder capabilities for I/O on the
Windows machine. Do you have any specific reservations? I've personally
been using this solution for several weeks now and there is absolutely no
noticeable speed difference or general drawbacks as far as I'm concerned.

nathanleclaire:
It's not performance I'm concerned about, it's breakage.  To me this:

> Basically, the software (VirtualBox in that case) would have to use special paths starting with "
?\" to trigger the Windows APIs allowing longer paths. One of the caveats described in that page is that relative paths are not supported.

Is magic that can be used to hack around the fact that Windows is disallowing long file paths.  Hacks are OK if absolutely needed but since we know that `C:\\` works (and is the "right" way to do it) it'd be preferable to enable only when needed vs. by default for _everyone_.

nathanleclaire:
OK, the Windows docs are much more informative.  Still, I'm not 100% sure if there are tradeoffs involved.

> The Windows API has many functions that also have Unicode versions to permit an extended-length path for a maximum total path length of 32,767 characters. This type of path is composed of components separated by backslashes, each up to the value returned in the lpMaximumComponentLength parameter of the GetVolumeInformation function (this value is commonly 255 characters). To specify an extended-length path, use the "\\?\" prefix. For example, "\\?\D:\very long path".

nathanleclaire:
I think it's safe based on my read of the documentation -- I'll go ahead and merge and test it in the next RC, we can always revert if something seems wrong.  Thanks for your contribution @juanjose49 !

nathanleclaire:
LGTM

PR #3557, opened 2016-06-29
Add Network options to Google Driver, shorten SourceImage URI

tydavis:
* Permits specifying `--google-network=SomethingNotDefault` for a network not "default"
* Only tested with [legacy network](https://cloud.google.com/compute/docs/subnetworks#legacy_non-subnet_network)
* Shortens source-image requirements e.g. `--google-machine-image=debian-cloud/global/images/debian-8-jessie-v20160606`  (As far as I can determine) Google does not permit source images to be loaded outside its googleapis.com domain.

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -26,6 +26,7 @@ type ComputeUtil struct {
 	project           string
 	diskTypeURL       string
 	address           string
+	network           string
 	preemptible       bool
 	useInternalIP     bool
 	useInternalIPOnly bool
@@ -64,6 +65,7 @@ func newComputeUtil(driver *Driver) (*ComputeUtil, error) {
 		project:           driver.Project,
 		diskTypeURL:       driver.DiskType,
 		address:           driver.Address,
+		network:           driver.Network,
 		preemptible:       driver.Preemptible,
 		useInternalIP:     driver.UseInternalIP,
 		useInternalIPOnly: driver.UseInternalIPOnly,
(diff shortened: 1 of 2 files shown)


drivers/google/compute_util.go
```diff
@@ -274,7 +277,7 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 	if disk == nil || err != nil {
 		instance.Disks[0].InitializeParams = &raw.AttachedDiskInitializeParams{
 			DiskName:    c.diskName(),
-			SourceImage: d.MachineImage,
+			SourceImage: "https://www.googleapis.com/compute/v1/projects/" + d.MachineImage,
```
nathanleclaire: Why change this just to add it back in again later?

drivers/google/google.go
```diff
@@ -34,10 +35,11 @@ const (
 	defaultZone        = "us-central1-a"
 	defaultUser        = "docker-user"
 	defaultMachineType = "n1-standard-1"
-	defaultImageName   = "https://www.googleapis.com/compute/v1/projects/ubuntu-os-cloud/global/images/ubuntu-1510-wily-v20151114"
+	defaultImageName   = "ubuntu-os-cloud/global/images/ubuntu-1510-wily-v20160627"
```
nathanleclaire: It was already present here

nathanleclaire:
I think it looks OK after code nitpick, it will have to wait until after 0.8.0 though.

tydavis:
@nathanleclaire Removing the absolute URL was due my understanding that Google wouldn't accept URIs outside its own "https://www.googleapis.com/compute/v1/project/" path, and resulted in a shorter command line argument for an otherwise unchanging parameter.

tydavis:
Just updated my copy of Docker with the latest commits. All tests should pass.

nathanleclaire:
LGTM

PR #3560, opened 2016-06-30
Remove "open : no such file or directory" message during crash report

### LINKED ISSUE #3559: Confusing "open : no such file or directory" message when creating a machine with a driver other than virtualbox
If an error occurs when creating a machine with a driver other than virtualbox, "open : no such file or directory" message will be reported in the debug log regardless of the root cause of the error. Here is an example output using the openstack driver with "--openstack-active-timeout 1" to cause an error.

```
(default) Creating machine...
(default) DBG | Waiting for the OpenStack instance to be ACTIVE... map[MachineId:da744b65-383c-401a-bd13-c7d12112f223]
Error creating machine: Error in driver during machine creation: Maximum number of retries (0) exceeded
open : no such file or directory
notifying bugsnag: [Error creating machine: Error in driver during machine creation: Maximum number of retries (0) exceeded]
```

Although the error of open itself is not harmful, this message misleads many people into thinking that this is the root cause of the problem they are encountering.

Please note that "no such file or directory" message might be different on executing environment.

s-koba:
Add an empty string check to the path argument of addFile(). Send()
method of BugsnagCrashReporter calls addFile() passing err.LogFilePath
as the path argument, which is an empty string if the driver is not
virtualbox.

Fixes #3559

Signed-off-by: KOBAYASHI Shinji <koba@jp.fujitsu.com>

### DIFF
--- a/libmachine/crashreport/crash_report.go
+++ b/libmachine/crashreport/crash_report.go
@@ -112,6 +112,9 @@ func (r *BugsnagCrashReporter) noReportFileExist() bool {
 }
 
 func addFile(path string, metaData *bugsnag.MetaData) {
+	if path == "" {
+		return
+	}
 	file, err := os.Open(path)
 	if err != nil {
 		log.Debug(err)


nathanleclaire:
LGTM -- thanks Shinji!

Although I am not sure why we are sending empty string to this method in the first place?

s-koba:
Thanks @nathanleclaire for merging the PR quickly.

The empty string comes from vBoxLog in the function cmdCreateInner() in commands/create.go. It is initialized to an empty string and is set to "VBox.log" when the driver is virtualbox.

```go
                vBoxLog := ""
                if h.DriverName == "virtualbox" {
                        vBoxLog = filepath.Join(api.GetMachinesDir(), h.Name, h.Name, "Logs", "VBox.log")
                }

                return crashreport.CrashError{
                        Cause:       err,
                        Command:     "Create",
                        Context:     "api.performCreate",
                        DriverName:  h.DriverName,
                        LogFilePath: vBoxLog,
                }
```

So my guess is that there is no appropriate log file to attach to the crash report for the drivers other than virtualbox.

PR #3568, opened 2016-07-05
Add an option to use an EC2 instance's IAM role

### LINKED ISSUE #3500: Ability to provision AWS EC2 with IAM Role instead of keys
Right now AWS access keys are required. If a host server has an IAM Role with the appropriate policy to provision an EC2 server access keys should not be required.

sgsabbage:
Uses the AWS Go SDK's role provider to provide credentials with use of the `--amazonec2-use-iam-role` flag.

I wasn't quite sure about the flag name, so more than happy to change it if needed!

Fixes #3500 

Signed-off-by: Sean Sabbage s.sabbage@siphonnetworks.com

### DIFF
--- a/docs/drivers/aws.md
+++ b/docs/drivers/aws.md
@@ -46,6 +46,12 @@ You can use environment variables:
     $ export AWS_SECRET_ACCESS_KEY=MY-SECRET-KEY
     $ docker-machine create --driver amazonec2 aws01
 
+### Instance IAM role
+
+If run on an EC2 instance with its own IAM role, you can use this by using the `--amazonec2-use-iam-role` flag on the command line:
+
+    $ docker-machine create --driver amazonec2 --amazonec2-use-iam-role aws01
+
 ## Options
 
 -   `--amazonec2-access-key`: Your access key id for the Amazon Web Services API.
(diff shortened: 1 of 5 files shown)


nathanleclaire:
Thanks for the PR.

Is it possible to just do so automatically?  Like, if an IAM profile for the instance exists, then use that, otherwise fall back to trying to read them from file, and otherwise fall back on the Machine flags / env vars.  Seems like we should just use the profile automatically if possible.

sgsabbage:
@nathanleclaire My concern with doing that is that it would potentially break existing installs - where docker-machine might be on an instance with a profile, but using credentials for a different role. It would then be impossible to override the instance's profile.

I could also make it work without the flag as a fallback - i.e. try taking the credentials from the command line, then file, then if none specified use the profile, but this would cause delays when run on a non EC2 box as it would try and contact the metadata server every time before throwing the error - I'm not sure if there's an easy way to check if run on an EC2 instance without trying to contact the metadata server. 

Also, the same problem exists, if you have the credentials file in place for a different use but want to use the instance role then there's no way to override and say 'use the role'.

Very happy to take direction here as to what would be best.

nathanleclaire:
I think this is fixed with https://github.com/docker/machine/pull/3799 ?

nathanleclaire:
Please re-open if I'm mistaken!  Thanks

PR #3569, opened 2016-07-05
Make the plugin driver prefix a variable.

dlorenc:
This allows other clients of libmachine to configure this prefix, so they
can avoid conflicting with drivers installed for docker-machine itself.


nathanleclaire:
Can you elaborate a bit more on the motivations for this change?

dlorenc:
Hey,

We're using 'libmachine' directly as a vendored dependency, and we'd like to enable the usage of some non-core drivers. If we can configure the driver prefix, we'll be able to bundle these drivers without having to name them 'docker-machine-driver-SOMETHING', and can instead use our own naming prefix.

nathanleclaire:
Thanks for the submission @dlorenc, however it is not a use case we are interested in supporting at this time.  If folks want to use `libmachine`, compatibility with the `docker-machine` upstream is expected.  This includes the convention supported for naming and detecting plugins.  I suggest following the existing naming convention if you wish to use plugins.  Thanks again.

- N

Issue #3596, opened 2016-07-18
Create virtualbox machine fails with ip collision error


mjstevens777:
When I try to run docker-machine create on a fresh install, I get an ip conflict error.

I am running arch linux with docker-machine 0.7.0 and virtualbox 5.1.0.

```sh
> docker-machine create -d virtualbox default
Running pre-create checks...
Creating machine...
(default) Copying /home/matt/.docker/machine/cache/boot2docker.iso to /home/matt/.docker/machine/machines/default/boot2docker.iso...
(default) Creating VirtualBox VM...
(default) Creating SSH key...
(default) Starting the VM...
(default) Check network to re-create if needed...
(default) Found a new host-only adapter: "vboxnet0"
(default) Waiting for an IP...
Error creating machine: Error in driver during machine creation: host-only cidr conflicts with the network address of a host interface
```

After some debugging, it appears that the conflict is between the dhcp address and the main address of the virtualbox host only interface.

PR #3597, opened 2016-07-18
Exclude all addresses from host-only interfaces

mjstevens777:
This PR addresses issue #3596 

It excludes all addresses from a host-only interface from being included in the list of host interfaces. The previous behavior was to exclude only addresses listed in excludeNets.


mjstevens777:
The tests for this file are set up around network interfaces with a single address, which is not guaranteed by the go api. To test behavior with multiple interfaces, I would have to refactor the entire test file. Since I am not very familiar with go, I don't really think i should do this.

nathanleclaire:
What issue does this fix?

nathanleclaire:
LGTM

nathanleclaire:
Thanks!

PR #3602, opened 2016-07-21
fix repo install on redhat provisioner

tronicum:
redhat provisioner
* added repo config (found in older version)
* try to install redhat/fedora rpm from docker
before generic get.docker.com tries to config & install (and fails at the moment)

Signed-off-by: Stefan Sels <tronicum@gmail.com>

### DIFF
--- a/libmachine/provision/redhat.go
+++ b/libmachine/provision/redhat.go
@@ -170,9 +170,20 @@ func (provisioner *RedHatProvisioner) Provision(swarmOptions swarm.Options, auth
 		}
 	}
 
+	// configure dockers rpm repo
+        if err := provisioner.ConfigurePackageList(); err != nil {
+                return err
+        }
+
 	// update OS -- this is needed for libdevicemapper and the docker install
 	if _, err := provisioner.SSHCommand("sudo -E yum -y update"); err != nil {
 		return err
+	}
+
+        // try to install docker rpms via repo that provisioner.ConfigurePackageList created 
+        // if that does not succed, installDocker will install generic or skip install if docker exists
+        if _, err := provisioner.SSHCommand("sudo -E yum -y -q install docker-engine"); err != nil {
+                return err
 	}
 
 	// install docker
--- a/libmachine/provision/redhat.go
+++ b/libmachine/provision/redhat.go
@@ -171,19 +171,19 @@ func (provisioner *RedHatProvisioner) Provision(swarmOptions swarm.Options, auth
 	}
 
 	// configure dockers rpm repo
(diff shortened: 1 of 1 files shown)


nathanleclaire:
I think upstream `get.docker.com` is supposed to work with RHEL now?  We should probably switch to that instead.

tronicum:
This fixes a setup where fedora 24 will result into get.docker.com to point to a non existing repo instead of using 23 which works fine with 24. 
I can not look at the code right now for RHEL. 

If we point to get.docker.com always, everytime fedora is newer than dockers repo it fails or installs older fedora packages (without notice).

nathanleclaire:
> If we point to get.docker.com always, everytime fedora is newer than dockers repo it fails or installs older fedora packages (without notice).

Then that issue should be fixed in docker/docker upstream, no?  Won't Machine attempting to work around this be counter-productive?

tronicum:
> > If we point to get.docker.com always, everytime fedora is newer than dockers repo it fails or installs older fedora packages (without notice). 

> Then that issue should be fixed in docker/docker upstream, no? Won't Machine attempting to work around this be counter-productive?

Yes, it should be covered by docker/docker. But I guess it wont. The official policy at docker/docker is, if its not released at time of a docker release, it is not supported.

Also, if you don't even try to install via a docker-machine configured repo, close #3627 too and we can remove all the repository setup code as if we always use get.docker.com to setup those, we do not need them.

nathanleclaire:
Hey @tronicum, what's the status on this?  Do you think it's still needed?

tronicum:
> Hey @tronicum, what's the status on this? Do you think it's still needed?
> 
If get.docker.com is the choice to setup repos it's not needed. 

But then I would suggest to remove all that platform specific code to setup repos because it would be never be used.

nathanleclaire:
> If get.docker.com is the choice to setup repos it's not needed. 

OK. It is so I will close this issue

> But then I would suggest to remove all that platform specific code to setup repos because it would be never be used.

Which code do you refer to?

Thanks @tronicum

tronicum:
ConfigurePackageList, which sets the official yum repositories. This task is performed by get.docker.com ( https://github.com/docker/docker/blob/master/hack/install.sh around line 458). 
I opened this PR initially, because to my surprise, ConfigurePackageList was never used in latest code.

In my opinion it can be removed because it is never called. If you want I could write a PR to remove it and its tests.

nathanleclaire:
> In my opinion it can be removed because it is never called. If you want I could write a PR to remove it and its tests.

Yeah that'd be great, thanks!

PR #3612, opened 2016-07-27
Update digitalocean driver image defaults

GenPage:
Ubuntu 15.10 image is no longer available on DigitalOcean since it is EOL, We recommend utilizing 16.04.1 LTS as the new default image.

### DIFF
--- a/drivers/digitalocean/digitalocean.go
+++ b/drivers/digitalocean/digitalocean.go
@@ -35,7 +35,7 @@ type Driver struct {
 const (
 	defaultSSHPort = 22
 	defaultSSHUser = "root"
-	defaultImage   = "15621816" // Ubuntu 15.10 x86_64
+	defaultImage   = "ubuntu-16-04-x64"
 	defaultRegion  = "nyc3"
 	defaultSize    = "512mb"
 )


nathanleclaire:
I still don't understand why it is possible to spin up an image for this by ID and not by slug if DigitalOcean is dropping support for it.  Can you explain?

And why was there absolutely no deprecation notice or warning on this?  Canonical _still_ maintains Amazon machine images even for versions that have long been EOLed like 12.04.  Is it so unreasonable to expect that something which was only released nine months ago will continue to work?

GenPage:
@nathanleclaire I am a bit confused as it should not be possible to spin up a 15.10 image using the ID or slug. The image is no longer available via control panel either. 

There was no deprecation or warning deemed necessary since the image itself was EOL'd as it has been 9 months since its release and internal stats showed no usage of the image.

We still maintain 12.04 and 14.04 images just like Canonical does because they are LTS, not EOL'd. AMIs that that are labeled EOL are only provided as reference.

Note: I am also going to be updating the PR to use the slug instead so that any 16.04 LTS point releases do not break future functionality.

nathanleclaire:
Image ID 15621816, which referenced Ubuntu 15.10 is now gone completely from the API, was present _yesterday_ when I checked and made a VM using that image (which is why I turned in [this fix](https://github.com/docker/machine/pull/3611)).

```console
$ curl -s -X GET "https://api.digitalocean.com/v2/images/15621816" -H "Authorization: Bearer $DIGITALOCEAN_ACCESS_TOKEN" | jq .
{
  "id": "not_found",
  "message": "The resource you were accessing could not be found."
}
```

You'll note that the 32 bit version is still around with no slug.  This is extremely bizarre behavior.  I'm really quite disappointed. Bumping OS versions is not something that we can take lightly as it can result in backwards compatibility issues and should be done for all drivers.

```console
$ curl -s -X GET "https://api.digitalocean.com/v2/images/15621817" -H "Authorization: Bearer $DIGITALOCEAN_ACCESS_TOKEN" | jq .
{
  "image": {
    "id": 15621817,
    "name": "15.10 x32",
    "distribution": "Ubuntu",
    "slug": null,
    "public": true,
    "regions": [
      "nyc1",
      "sfo1",
      "nyc2",
      "ams2",
      "sgp1",
      "lon1",
      "nyc3",
      "ams3",
      "fra1",
      "tor1",
      "sfo2",
      "blr1"
    ],
    "created_at": "2016-02-03T16:06:25Z",
    "min_disk_size": 20,
    "type": "snapshot",
    "size_gigabytes": 1.43
  }
}
```

Thanks for the PR and please communicate to the team at DigitalOcean that yanking support for images so suddenly without any advance warning, notice, or existing deprecation policy is extremely disheartening, and resulted in the DigitalOcean driver defaults for Machine being broken with no chance to prepare.  Likewise having the image I was referencing vanish completely from the API after I file a support ticket indicating that the slug is missing is really quite strange.

 In this case since the 0.8.0 release is so soon I am going to go ahead and merge this and work on updating the other drivers to use this Ubuntu version over time.  Thanks.

nathanleclaire:
LGTM

tassels:
@nathanleclaire Thanks for merging, and I apologize for the lack of notice. We follow a pretty standard EOL policy. When an OS is EOL'd by the OS's maintainers, whether it be CentOS, Debian, Ubuntu, Fedora, or CoreOS, we drop support and remove the image from rotation. In this case, we made an oversight and while we removed it from the UI, we didn't disable it for public use. We'll pull the 32-bit version as well, in the morning. In the future you should be able to expect that both the slug and the image-ID will be pulled upon EOL, as determined by the OS maintainers.

By using an LTS release, you can expect much more stable support, but for a release like 15.10 where support is dropped by Canonical only 9 months after release, you can expect us to move quickly to remove it to prevent further Droplets being created with an out of date, potentially insecure OS.

Bumping versions **absolutely** shouldn't be done lightly, but we also can't promote the use of old, insecure images by keeping them the default. In an ideal situation, support for 15.10 should have been removed, across all drivers, shortly before EOL -- not after the fact. 

On my end, I'll see what we can do to be more transparent about our image updates and removals. And, of course, this policy which should be public somewhere.

nathanleclaire:
Yeah I think partially the moral of the story here is that Machine should probably not use non-LTS releases.  At any rate thanks for the PR and followup.

PR #3620, opened 2016-07-28
Support building a Linux ARMv6 binary

DieterReuter:
In order to support building for Linux on ARM (32bit) we could easily cross compile Docker Machine. As Go defaults to GOARM=6 the compiled binary supports running on ARMv6 and ARMv7 machines which are typically called "armhf" for the Debian distro.

Testing a build locally within Docker and only compile an ARM binary is easy with this command:
```
USE_CONTAINER=true TARGET_OS=linux TARGET_ARCH=arm make build-x
```
It builds the binary `./bin/docker-machine-Linux-armhf`.

Signed-off-by: Dieter Reuter <dieter.reuter@me.com>


nathanleclaire:
LGTM thanks

PR #3547, opened 2016-06-27
Add affinity group support in exoscale driver

runseb:
This improves the exoscale drivers and adds support for affinity groups.

Signed-off-by: Sebastien Goasguen <runseb@gmail.com>

PR #3629, opened 2016-08-02
Add affinity group support in exoscale driver

cab105:
Based on the suggestion from #3547 since I don't have write access to @runseb's repo, I'm creating new PR.

### DIFF
--- a/docs/drivers/exoscale.md
+++ b/docs/drivers/exoscale.md
@@ -32,6 +32,8 @@ Options:
 
 If a custom security group is provided, you need to ensure that you allow TCP ports 22 and 2376 in an ingress rule. Moreover, if you want to use Swarm, also add TCP port 3376.
 
+There is a limit to the number of docker machines that an anti-affinity group can have.  This can be worked around by specifying an additional anti-affinity group using `--exoscale-affinity-group=docker-machineX`
+
 Environment variables and default values:
 
 | CLI option                      | Environment variable         | Default                           |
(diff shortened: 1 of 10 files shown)


nathanleclaire:
@vincentbernat Does it LGTY?

vincentbernat:
Codewise, it's OK. However, by putting everyone in the docker-machine anti-affinity group by default, it means that a user spawning more than 10 docker machines will get an error about reaching the limit of the anti-affinity group size. I think that's a good default to have a default anti-affinity group, we just need to add a tidbit in the documentation to mention the case and the solution would be to override the default anti-affinity group. Something like:

> By default, machines are spawned in the `docker-machine` anti-affinity group. There is a limit in the number of machines that can be part of the same anti-affinity group. If you reach the limit, just specify another anti-affinity group with `--exoscale-affinity-group=docker-machine2`.

docs/drivers/exoscale.md
```diff
@@ -32,6 +32,8 @@ Options:
 
 If a custom security group is provided, you need to ensure that you allow TCP ports 22 and 2376 in an ingress rule. Moreover, if you want to use Swarm, also add TCP port 3376.
 
+There is a limit to the number of anti-affinity groups a machine can belong to.  You can work around it by specifying a second anti-affinity group using `--exoscale-affinity-group=docker-machineX`
```
vincentbernat: It's the reverse. There is a limit to the number of machines an anti-affinity group can contain.

vincentbernat:
IMO, the wording is still incorrect. Sorry for the nitpick. You say "There is a limit to the number of machines that an anti-affinity group can belong to", you should say "There is a limit to the number of machines that an anti-affinity group can have". You'll also need to squash the commits.

cab105:
Its okay.  Wouldn't want to make a substandard commit.

nathanleclaire:
Is this good to go @vincentbernat @cab105 ?

vincentbernat:
LGTM, sorry for the delay!

nathanleclaire:
cool -- LGTM

Issue #3639, opened 2016-08-03
Please add more dangerous prompt when executing 'docker-machine rm'


zhanghan177:
I know that it was on the docker-machine documentation that **rm** option
>  This will remove the local reference as well as delete it on the cloud provider or virtualization management platform. 

However, it was not specified that clear in the CLI.
> $ docker-machine help
...
Commands:
rm   Remove a machine
...

 I thought it will only remove the local reference(*dumb*), and in the command prompt it only asks
> $ docker-machine rm \<machine\>
About to remove \<machine\>
Are you sure? (y/n): y

It would be extremely helpful for *newbies* like me if there was a warning saying that 
> $ docker-machine rm \<machine\>
About to remove \<machine\>
WARNING: This will remove the remote instance as well.
Are you sure? (y/n): y

I would definitely not do the same thing as I just did and it was disastrous...

nathanleclaire:
Would be happy to accept a PR for this.

PR #3640, opened 2016-08-03
Add specified Warnings in rm option.

zhanghan177:
Signed-off-by: Han Zhang <han@zhanghan177.me>

#3639 

Now in `docker-machine rm <machine>`, it will prompt additional warning about the fact that action will delete instance in cloud provider as well as the local reference, as discussed in the closed issue.


zhanghan177:
I really don't think my change will break the build on Travis.

nathanleclaire:
LGTM -- thanks

PR #3645, opened 2016-08-05
Support for Docker hosts on GoDaddy Cloud Servers

aka-bo:
Initial support for new godaddy driver.

Signed-off-by: Bo Thompson <bo.thompson@gmail.com>


nathanleclaire:
Thanks for the contribution -- we don't accept any more drivers into core but please feel free to distribute it as a separate plugin binary

PR #3664, opened 2016-08-15
Update certificates error message

bamarni:
Which is not necessarily true since live-restore was introduced
(https://docs.docker.com/engine/admin/live-restore/).


nathanleclaire:
Heh... Funny how the meaning of words changes over time

nathanleclaire:
Seems OK. LGTM

PR #3669, opened 2016-08-16
Bump Go version to 1.7.0

DieterReuter:
Signed-off-by: Dieter Reuter <dieter.reuter@me.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.6.3
+FROM golang:1.7.0
 
 RUN go get  github.com/golang/lint/golint \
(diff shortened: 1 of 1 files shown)


nathanleclaire:
LGTM!

Might even consider switching this to `golang:1.7.0-alpine` since we're not doing anything Debian-specific

DieterReuter:
I already tested it with `golang:1.7.0-alpine`, but this causes some errors. 😇

nathanleclaire:
just need to `apk add --update --no-cache git` in the `Dockerfile`.  I'm surprised the Golang alpine distributions don't include it actually.

Issue #3642, opened 2016-08-04
Proposal/Drivers: Support for additional security permissions


andrestc:
It would be useful if the core drivers provided a way for the user to add additional security permissions(opened ports on the host). 

My use case: I'm currently using docker machine drivers to automate the installation of multiple pre-determined services with known exposed ports that should be accessible from the outside. In order to have my workflow fully automated I need to create the security group manually (before starting the provision process) and then feeding it to the ec2 driver with `amazonec2-security-group`. This can be error prone and boring.

My proposal is to add a new, optional parameter: `amazonec2-security-rule` where one could configure additional rules to be appended to the default security group.

I believe we could support this kind of feature on every core driver that has some kind of security permission and would be happy to submit a PR to each of them.

andrestc:
Hi @nathanleclaire, do you think a PR adding this kind of support would be merged into docker-machine? We are in need of this feature and can solve it either by contributing it docker-machine or handling this feature for each driver separately in our tool.

nathanleclaire:
`--amazonec2-security-group` doesn't work for your needs ?

```
   --amazonec2-security-group [--amazonec2-security-group option --amazonec2-security-group option]	AWS VPC security group [$AWS_SECURITY_GROUP]
```

andrestc:
Actually, no. I'm building a tool that uses docker-machine and its drivers to bootstrap docker hosts and start some services on them. I can't rely on having my end user create the security group needed before running my tool. So I need to provide a way to automatically add this security permissions.

andrestc:
So I think I have two options: (1) implement this logic after the machine is provisioned (using the aws client(and the azure client etc etc) directly in my tool) or (2) adding this kind of feature to every driver it fits.

andrestc:
Any thoughts on this @nathanleclaire?

ahmetalpbalkan:
@andrestc for azure we create the Network Security Group for each machine and open Docker/SSH ports by default. We already get `--azure-open-port` user input to open extra ports (I believe this part is missing in other drivers). Would having such a generic create flag and having drivers to implement it solve your problem?

andrestc:
@ahmetalpbalkan, yes, that would definitely solve my problem! I could get these PR's going if everyone is cool with it.

ahmetalpbalkan:
[...]

PR #3671, opened 2016-08-16
amazonec2: add support for additional open ports

andrestc:
This PR adds a flag to the `amazonec2` driver to enable opening custom ports on the host.
This is a step to add this support for every core driver(where applied) as discussed in #3642(the azure driver already support this flag). 

I kept the same behavior as the azure driver, but I was thinking about supporting the format `port/protocol`(eg `8888/tcp`, `9200/udp`), defaulting the protocol to `tcp` if not set by the user. I decided to open the PR like this because I think the behavior should be the same across all core drivers. If we go for this behavior, I can update this PR and them move on to the other drivers.

cc @nathanleclaire, @ahmetalpbalkan 

### DIFF
--- a/docs/drivers/aws.md
+++ b/docs/drivers/aws.md
@@ -57,6 +57,7 @@ You can use environment variables:
 -   `--amazonec2-zone`: The AWS zone to launch the instance in (i.e. one of a,b,c,d,e).
 -   `--amazonec2-subnet-id`: AWS VPC subnet id.
 -   `--amazonec2-security-group`: AWS VPC security group name.
+-   `--amazonec2-open-port`: Make additional port number(s) accessible from the Internet.
 -   `--amazonec2-tags`: AWS extra tag key-value pairs (comma-separated, e.g. key1,value1,key2,value2).
 -   `--amazonec2-instance-type`: The instance type to run.
 -   `--amazonec2-keypair-name`: AWS keypair to use; requires --amazonec2-ssh-keypath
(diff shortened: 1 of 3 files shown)


nathanleclaire:
> I was thinking about supporting the format port/protocol(eg 8888/tcp, 9200/udp), defaulting the protocol to tcp if not set by the user.

This is how Docker works and I think how I'd like to do it here too.  @ahmetalpbalkan what do you think.  I'd prefer not to just hardcode to TCP.  There are other protos user might want as well BTW.

ahmetalpbalkan:
@nathanleclaire hmm in Azure case I guess we just open TCP. It makes sense to add `/tcp`/`/udp` distinction probably in a separate PR.

andrestc:
@nathanleclaire, @ahmetalpbalkan I've updated the PR adding support for the port/protocol format, defaulting the protocol to "tcp". 

After this is good to go, I can change the azure driver to accept the same format. I'll, then, move the function splitPortProtocol to some utility package where it can be used by the other drivers (any suggestion where to place it)? Then, I'll move to add the analogous flag to other core drivers that it fits.

ahmetalpbalkan:
Sounds good. :+1:

nathanleclaire:
@andrestc It's ready for review?

andrestc:
@nathanleclaire yes!

nathanleclaire:
LGTM

PR #2535, opened 2015-12-09
Fixes for on-premise EC2 endpoints

bgokden:
Signed-off-by: Berk Gokden <berk.gokden@gmail.com>

PR #3675, opened 2016-08-18
Fix for on-premise EC2 endpoints

bgokden:
Added option to select amazonec2-endpoint and amazonec2-insecure-endpoint.

- Disabled region checking when custom endpoint used.

Signed-off-by: Berk Gokden <berkgokden@gmail.com>

This is re based version of pull request 2535 : 
Fixes for on-premise EC2 endpoints #2535 https://github.com/docker/machine/pull/2535

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -46,12 +46,13 @@ const (
 )
 
 var (
-	dockerPort                  = 2376
-	swarmPort                   = 3376
-	errorNoPrivateSSHKey        = errors.New("using --amazonec2-keypair-name also requires --amazonec2-ssh-keypath")
-	errorMissingAccessKeyOption = errors.New("amazonec2 driver requires the --amazonec2-access-key option or proper credentials in ~/.aws/credentials")
-	errorMissingSecretKeyOption = errors.New("amazonec2 driver requires the --amazonec2-secret-key option or proper credentials in ~/.aws/credentials")
-	errorNoVPCIdFound           = errors.New("amazonec2 driver requires either the --amazonec2-subnet-id or --amazonec2-vpc-id option or an AWS Account with a default vpc-id")
+	dockerPort                           = 2376
+	swarmPort                            = 3376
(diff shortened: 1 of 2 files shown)


nathanleclaire:
Insecure endpoint should never be set without the other option to specify the endpoint, right?  Can you please add a check and error out if user attempts to set the insecure endpoint flag without also specifying an alternative endpoint to connect o?

jgallucci32:
A few recommendations:
* Rename the `--amazonec2-insecure-endpoint` flag to `--amazonec2-disable-ssl` to match the aws-sdk-go function WithDisableSSL. It makes it more obvious that it is a bool and not a string (endpoint implies string)
* Improve list of valid regions by adding a "custom-endpoint" region in the list with a blank ami ID rather than return any value
 * Some third-party apps like Rancher use the list to populate the region field
 * If region and endpoint are both defined, region is ignored by aws-sdk-go

nathanleclaire:
> Rename the --amazonec2-insecure-endpoint flag to --amazonec2-disable-ssl to match the aws-sdk-go function WithDisableSSL. It makes it more obvious that it is a bool and not a string (endpoint implies string)

If the concern is around conflating bool and string, maybe we can call it something like `--amazonec2-insecure-transport`, but I feel strongly that since it is insecure it should be called out as such

bgokden:
@nathanleclaire I have chosen amazonec2-disable-ssl. Also made some adjustments since region is ignored by aws-sdk-go when endpoint is used. I also realized there is a regionzone variable which is actually just available  zone when custom endpoint is used, so I set region to "" when endpoint is set.

nathanleclaire:
Let's please go with `--amazon-insecure-transport`

bgokden:
@nathanleclaire I changed it to --amazonec2-insecure-transport

nathanleclaire:
LGTM 👍

nathanleclaire:
Thanks for your contribution @bgokden

bgokden:
@nathanleclaire Thank you.

PR #3681, opened 2016-08-19
[scp] allow ssh-agent when no key is provided

### LINKED ISSUE #3663: docker-machine scp passes "-o IdentitiesOnly=yes" with no ssh keys specified
This causes the scp to always fail since there are no SSH keys specified in the command either.

bamarni:
Fixes #3663


nathanleclaire:
LGTM

Thanks @bamarni !

PR #3682, opened 2016-08-19
[proposal] poll for machine status during creation

bamarni:
This is a follow-up from #3509, I think it'd be easier to also poll for machine status here, instead of making SSH attempts later on.

I don't know if it's easily doable with all drivers? I've just checked aws and digitalocean and there are already some polling.

Here is the output after this patch when creating a droplet on digitalocean : 
```
Creating machine...
(toto) Calling .Create
(toto) Creating SSH key...
(toto) Creating Digital Ocean droplet...
(toto) Waiting for IP address to be assigned to the Droplet...
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Waiting for droplet to be created
(toto) DBG | Created droplet ID 23040331, IP address 104.131.15.84
(toto) Calling .GetConfigRaw
(toto) Calling .DriverName
(toto) Calling .DriverName
(toto) Calling .GetState
Getting to WaitForSSH function...
(toto) Calling .GetSSHHostname
(toto) Calling .GetSSHPort
(toto) Calling .GetSSHKeyPath
(toto) Calling .GetSSHKeyPath
(toto) Calling .GetSSHUsername
Using SSH private key: /Users/bamarni/.docker/machine/machines/toto/id_rsa (-rw-------)
&{[-F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=q
uiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none root@104.131.15.84 -o Identi
tiesOnly=yes -i /Users/bamarni/.docker/machine/machines/toto/id_rsa -p 22] /usr/bin/ssh <nil>}
About to run SSH command:
exit 0
SSH cmd err, output: <nil>:
Detecting the provisioner...
```

As you can see, there is no SSH retries anymore. If we had similar checks for all drivers we could get rid of ssh attempts, what do you think?


nathanleclaire:
Once upon a time we waited for `state.Running` before attempting the SSH ... I wonder what happened to that (sometimes things get lost in the shuffle).  Why not use `GetState` to do this in a uniform way for all drivers in the main `create` code path?

nathanleclaire:
Ah yeah, here we go: https://github.com/docker/machine/blob/8141874120abdc5c9056b6dbeb09bfb76a3eb290/libmachine/libmachine.go#L161

What is this PR improving upon that?  SSH should not be attempted until the machine reports state as `Running`, and personally I'm a bit confused why an EC2 instance etc. would report running but then still take >5 minutes for SSH to be available.  Not saying it's impossible, but usual Linux boot is a fair bit faster than that.

bamarni:
Ah yep :) there's definitely a wait for `state.Running`, I thought it was trying SSH right after creation.

My main concern about the mentioned issue is that if we raise max attempts it will make very long waiting timeouts in case of failures (eg. wrong flags), for instance with the generic driver it might hang 15 minutes while actually we're not waiting for any machine creation, so failing very early should be possible in that case. Should there be an exception for the generic driver, and don't wait for the timeout (because either it works or it doesn't)?

(In any case this PR is not needed as you pointed out)

nathanleclaire:
> Should there be an exception for the generic driver, and don't wait for the timeout (because either it works or it doesn't)?

That's certainly an interesting idea -- I would argue that SSH should be live for machines you expect to work with generic driver, or at least the timeout should be much less forgiving...

> (In any case this PR is not needed as you pointed out)

I'd be really curious to know _why_ there's such a big gap between `Running` state and SSH availability though.

bamarni:
> I'd be really curious to know why there's such a big gap between Running state and SSH availability though.

No idea :/ According to @sebadiaz, it can happen when creating big machines on AWS.

So I've updated the PR with another proposal, what about having a separated `WaitFor` function in drivers util? It'd be more flexible than the one in mcnutils and would be used in places with potentially some delta depending on use cases and setups (mainly wait for `state.Running` and wait for SSH), with this priority :
- a `MACHINE_MAX_ATTEMPTS` env var
- driver type (eg. reduced value for generic driver)
- fallback to the default `60`

nathanleclaire:
Seems reasonable to me although I think if we're going to rely on an environment variable like so we might as well make it a proper flag associated to an environment variable.  [Like these](https://github.com/docker/machine/blob/52d0020542afa76242fce663b394f749d4fb05fe/commands/commands.go#L151-L152), although they have their own associated issues

bamarni:
@nathanleclaire I've added a second commit, PTAL.

Personally I'm not very conviced by the global flag approach, as tweaking this value is basically only useful for the create command. Exposing this in the cli might also be confusing because it is not something one would usually have to worry about.

Let me know if you still want to go this way and I'll squash (and I'll sign too @GordonTheTurtle !) :)

nathanleclaire:
> Personally I'm not very conviced by the global flag approach, as tweaking this value is basically only useful for the create command. Exposing this in the cli might also be confusing because it is not something one would usually have to worry about.

Yeah, I agree it's not ideal, but there are other commands such as `provision` which might need this.

What about the other `WaitFor` function in `mcnutils`?  Does this replace it completely and if so why not remove it?  And/or, why not make the change to the function there?

Why not set the flag default to `60` in the flag definition instead of the `else` block?

bamarni:
> What about the other WaitFor function in mcnutils? Does this replace it completely and if so why not remove it? And/or, why not make the change to the function there?

I made a separate one because this PR is only about machine creation actions (wait for state.Running, and then for ssh as required with the original issue), which can vary a lot depending on use case / driver. I think other places using `mcnutils.WaitFor` are pretty even and low in terms of possible duration, we could probably switch a few of them to drivers.WaitFor (the ones inside some drivers create function which are doing some sort of polling like wait for ip, etc.) and then be able to reduce the default attempts in `mcnutils.WaitFor`?

> Why not set the flag default to 60 in the flag definition instead of the else block?

I thought about this priority : use the provided value, or a low value if it's the generic driver, or 60. If it's set as 60 as a flag's default we don't know anymore if the value was provided explicitly or not, so there would be no way to put the low value for the generic driver in the middle priority. If you're fine with forcing the low value with the generic driver then I can put 60 as a default.

nathanleclaire:
Hm, I originally added `mcnutils` deliberately to contain little functions like `WaitFor` that don't really make sense scattered across packages like `driver` (what does a "wait" function have to do with drivers specifically? also causes problems with circular imports etc.)   Could we just change the one in `mcnutils`?

nathanleclaire:
> I thought about this priority : use the provided value, or a low value if it's the generic driver, or 60. If it's set as 60 as a flag's default we don't know anymore if the value was provided explicitly or not, so there would be no way to put the low value for the generic driver in the middle priority. If you're fine with forcing the low value with the generic driver then I can put 60 as a default.

Could you define a constant for this somewhere and check against it?

bamarni:
I put this one in the drivers package because it gets the driver as argument, it is only used at two places. Do you mean I should keep one function and change all the places to use it? The thing is that other places shouldn't have a reduced timeout in case it's "generic" imo.

An alternative is also to keep the original function, and in those two places I'm using a different timeout from the outside depending of it's the generic driver or not.

Sorry for all the back and forth it's not really clear to me :)

alambike:
Any chance to merge this soon? We are having random "Maximum number of retries exceeded" issues with Azure:
```
Expected process to exit with 0, but received '1'. Details: Running pre-create checks...
Creating machine...
(cargoXXX-prefapp-in) Creating Azure machine...
The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.
, Error creating machine: Error waiting for machine to be running: Maximum number of retries (60) exceeded
```

PR #3605, opened 2016-07-21
Add support to install Docker on raspbian/jessie

DieterReuter:
**\- What I did**
Adding support to provision the Docker Engine on a standard Raspberry Pi running a raspbian/jessie OS. Refers to https://github.com/docker/docker/issues/24561 and https://github.com/docker/docker/issues/24815.

**\- How I did it**
As Raspbian is just a fork of Debian and is compiled for ARMv6, I just cloned the Debian provisioner and modified it to use Raspbian and use "overlay" as the default storage driver.

**\- How to verify it**
Flash a standard SD card image with latest RASPBIAN JESSIE or RASPBIAN JESSIE LITE, boot the Raspberry Pi and configure the SSH access via SSH keys. A `ssh pi@<ipaddress>` should work correctly without requiring an interactive password.

Next, use `docker-machine` to provision the Docker Engine to the Raspberry Pi with the following commands or shell script:

```
#!/bin/sh
set -x

# access the Raspberry Pi running Raspbian/Jessie
IPADDRESS=192.168.2.115
PI_USERNAME=pi
PI_PASSWORD=raspberry

# deploy a Docker 1.12.0-rc4 on ARMv6
docker-machine --debug create \
  --driver=generic \
  --generic-ip-address=$IPADDRESS \
  --generic-ssh-user=$PI_USERNAME \
  --engine-install-url=https://github.com/DieterReuter/docker/raw/430bf992d933a1a41d7e61e23ca8fa051e98659e/hack/install.sh \
  pi
```

As soon as the PR https://github.com/docker/docker/issues/24815 is merged and a docker-engine .deb package for raspbian/jessie is in the official Docker APT repository, we can use `--engine-install-url=https://test.docker.com`.

Signed-off-by: Dieter Reuter dieter.reuter@me.com

PR #3692, opened 2016-08-22
[provision] fix replace pattern in /etc/hosts

bamarni:
Require a space before the hostname.

Since #2486, as we don't check for a space before the hostname, an existing one finishing by the new one wouldn't be replaced.

### DIFF
--- a/libmachine/provision/generic.go
+++ b/libmachine/provision/generic.go
@@ -47,9 +47,9 @@ func (provisioner *GenericProvisioner) SetHostname(hostname string) error {
 
 	// ubuntu/debian use 127.0.1.1 for non "localhost" loopback hostnames: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution
 	if _, err := provisioner.SSHCommand(fmt.Sprintf(`
-		if ! grep -xq .*%s /etc/hosts; then
-			if grep -xq 127.0.1.1.* /etc/hosts; then 
-				sudo sed -i 's/^127.0.1.1.*/127.0.1.1 %s/g' /etc/hosts; 
+		if ! grep -xq '.*\s%s' /etc/hosts; then
+			if grep -xq '127.0.1.1\s.*' /etc/hosts; then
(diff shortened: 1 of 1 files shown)


bamarni:
Some other concerns not addressed in this PR :

- the comment states that this change in /etc/hosts is specific to debian-based distros while the code is in the generic provisioner, so it might be done for other OS too. Should we check for the os release id here and only do this for debian / ubuntu / raspbian?

- according to https://wiki.debian.org/HowTo/ChangeHostname#Rename_a_computer, there are more steps needed to properly rename a hostname on a debian system. This is notably problematic for the raspbian integration (#3605), because on a raspberry pi it is very common to rely on the avahi daemon for its mDNS capabilities (to access it through `hostname.local`). Can we do something about this? Or is it up to the user to reload those services or reboot the rpi after the create command?

cc @DieterReuter

nathanleclaire:
> the comment states that this change in /etc/hosts is specific to debian-based distros while the code is in the generic provisioner, so it might be done for other OS too. Should we check for the os release id here and only do this for debian / ubuntu / raspbian?

Yes, it's been existing issue for a while, that this is done on non-Debian based distros when it should not be AFAIK.

> according to https://wiki.debian.org/HowTo/ChangeHostname#Rename_a_computer, there are more steps needed to properly rename a hostname on a debian system. This is notably problematic for the raspbian integration (#3605), because on a raspberry pi it is very common to rely on the avahi daemon for its mDNS capabilities (to access it through hostname.local). Can we do something about this? Or is it up to the user to reload those services or reboot the rpi after the create command?

Probably at least run the proper `hostname` commands etc. if possible.  As for avahi, I'm not sure.  If there were already a hostname set by avahi I guess Machine could just over-write it.

nathanleclaire:
LGTM

bamarni:
> Probably at least run the proper hostname commands etc. if possible. As for avahi, I'm not sure. If there were already a hostname set by avahi I guess Machine could just over-write it.

It is already overwritten (by default avahi would use the machine hostname), it's just not reloaded so after provisioning it'd still use the old one. I'll leave a comment on the raspberry pi PR to keep track of this.

PR #3695, opened 2016-08-22
Fix for  amazonec2-insecure-transport usage  and region checking for custom endpoints.

bgokden:
Fix for  amazonec2-insecure-transport usage  and region checking for custom endpoints.

- When insecure-transport set custom endpoint should also be set.
- Aws client requires non-empty  region, so added a method to properly set RegionZone.
- Proper tests added to check these behaviors.

Unfortunately, I have introduced this bug and realized it today.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -307,10 +307,6 @@ func (d *Driver) SetConfigFromFlags(flags drivers.DriverOptions) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


bgokden:
@nathanleclaire  hi, could you check this one. Today, while testing it, I realized I accidentally introduced two bugs, so I wrote fix and tests to be sure.

nathanleclaire:
LGTM

Thanks for the followup @bgokden

PR #3698, opened 2016-08-23
Added check if subnet.DefaultForAz is not nil

bgokden:
Added check if subnet.DefaultForAz is not nil

- subnet.DefaultForAz is an optional parameter and can be empty.
- if subnet.DefaultForAz is empty, it is considered as false.

Reference: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Subnet.html


bgokden:
@nathanleclaire I would be glad if you could look at this. I have another PR depending on this. Thank you.

nathanleclaire:
It causes panic if trying to dereference and not instantiated I suppose?

nathanleclaire:
LGTM

bgokden:
@nathanleclaire yes, exactly. Thank you.

PR #3701, opened 2016-08-25
[proposal] Added amazonec2-use-default-network-interface option

bgokden:
Added amazonec2-use-default-network-interface option to skip NetworkInterfaces parameter.

- With use default network option, it is possible to use automatic network offered by the provider.

Signed-off-by: Berk Gokden <berkgokden@gmail.com>

This is required for our on-promise cluster to work. I am not sure if it helps anyone other than us. So it would be nice if someone who is using aws could review it.


bgokden:
@nathanleclaire hi, can you look at it when you have time?

drivers/amazonec2/amazonec2.go
```diff
@@ -566,12 +573,21 @@ func (d *Driver) Create() error {
...
 	}
-	netSpecs := []*ec2.InstanceNetworkInterfaceSpecification{{
-		DeviceIndex:              aws.Int64(0), // eth0
-		Groups:                   makePointerSlice(d.securityGroupIds()),
-		SubnetId:                 &d.SubnetId,
-		AssociatePublicIpAddress: aws.Bool(!d.PrivateIPOnly),
-	}}
+
+	var netSpecs []*ec2.InstanceNetworkInterfaceSpecification = nil
+	var securityGroupIds []*string = nil
```
nathanleclaire: Why are these `*string` instead of `string`?  Seems a potential nil deference panic

drivers/amazonec2/amazonec2.go
```diff
@@ -566,12 +573,21 @@ func (d *Driver) Create() error {
...
+	var subnetId *string = nil
+	if !d.UseDefaultNetworkInterface {
+		netSpecs = []*ec2.InstanceNetworkInterfaceSpecification{{
```
nathanleclaire: Shouldn't this be set to _something_ even if `UseDefaultNetworkInterface` is true?  Passing `nil` for this doesn't seem quite right to me

drivers/amazonec2/amazonec2.go
bgokden: I have used `*string` since `func makePointerSlice(stackSlice []string) []*string` returns `[]*string`

drivers/amazonec2/amazonec2.go
bgokden: While creating the REST request, aws-sdk-go ignores nil values. If they have any other value, they are added to the request as parameters.

vtintillier:
@nathanleclaire is there something you need before we can merge this?

@bgokden left our company so he may not need this anymore, but it is still necessary for our ec2 compliant private cloud.

I can help rebasing on master (or merge master into this branch?) to resolve conflicts. I did that already on my local machine, to validate this still works with latest version.

PR #3708, opened 2016-08-31
Filter out invalid IPs from the vSphere driver

mbhinder:
On my local machine, the docker machine env command would return the
Docker0 bridge IP address which is not reachable from the outside. Added
a check for the default bridge IP as well as another check that filters
out link local IP addresses.

Signed-off-by: Preet Bhinder <preet@apcera.com>

@nathanleclaire

### DIFF
--- a/drivers/vmwarevsphere/vsphere.go
+++ b/drivers/vmwarevsphere/vsphere.go
@@ -34,7 +34,9 @@ import (
 )
 
 const (
-	isoFilename = "boot2docker.iso"
+	// dockerBridgeIP is the default IP address of the docker0 bridge.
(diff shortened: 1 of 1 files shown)


nathanleclaire:
cc @frapposelli

nathanleclaire:
Guess Fabio's not available for review right now -- but at any rate, LGTM since returning `docker0` IP definitely seems wrong.

frapposelli:
Sorry @nathanleclaire, VMworld week is crazy.

Definitely LGTM.

nathanleclaire:
Cool thanks @frapposelli !

PR #3720, opened 2016-09-02
libmachine: implements save and remove on FakeStore

andrestc:
This makes the `Save` and `Remove` methods of the `FakeStore` manipulate the Hosts slice accordingly.

The `Remove` method will only return an error if `fs.RemoveErr` is set (even when trying to remove an invalid host).

If we want to make sure that the `Hosts` slice stay the same, another possible approach would be to create two separated slices and just append to them on each operation, so the user can assert that the methods were called.


nathanleclaire:
Cool.  Did you have writing some tests in mind?

andrestc:
Sure, @nathanleclaire. There it is. I've also renamed `fake_store.go` to `fakestore.go` to match the `filestore.go` one.

nathanleclaire:
LGTM

PR #3654, opened 2016-08-08
Enhance softlayer driver

wzymaster:
To enhance the driver, I added two new parameters to address two issues here :

- SOFTLAYER_DOCKER_ON_PRIVATE_IP, enforce docker daemon to run on the private ip if needed. Especially for some use cases in private cloud.

- SOFTLAYER_NETWORK_MAX_SPEED, set up network speed and the default is set to 100 Mbps. The SoftLayer driver in the release version only provision VM with 10 Mbps by the default, which is not good in the most cases for users.

PR #3728, opened 2016-09-06
Regression bug fix for pull request #3654

wzymaster:
To enable Docker Daemon running on private IP if environment variable - SOFTLAYER_DOCKER_ON_PRIVATE_IP is set to true.


wzymaster:
@nathanleclaire  all checks are passed and please feel free to review it.

nathanleclaire:
Which issue does this fix?  Seems a bit of a wide-ranging change for just a regression fix, no?

wzymaster:
@nathanleclaire probably regression fix isn't a good description. I found that the final goal is to enable Docker Daemon running on Private IP and previous pull request #3654 only does displaying part but doesn't do the real job.

nathanleclaire:
@wzymaster OK -- couple things:

1. [This](https://github.com/docker/machine/pull/3654/files#diff-50b71984e556cb76a3e64ec0009d43f5R366) (checking environment variable) will inevitably fail when it is not set (i.e., operations invoked which aren't `create`), you should be checking the value in the `Driver` struct which will be unmarshaled instead -- sorry, should have caught that in initial code review
2. Why do a driver-specific check in the provisioner here?  I don't like that -- ideally provisioner should be driver-agnostic and only require basics like SSH connection.  `Driver.GetURL` is intended to return the "Docker URL" (e.g., `<private-ip>:2376`) so consider switching that out instead.  However it's a bit tricky with all the `dockerPort` stuff.

wzymaster:
@nathanleclaire
1) In fact,  even environment variable isn't set up, my existing codes will handle it well. 
2) I want Docker Daemon to bind on private IP instead of '0.0.0.0' which is hard coded.  I don't quite understand what you mean `Driver.GetURL` is related to this ?

wzymaster:
@nathanleclaire Can you tell me where  this pull request will go ?

nathanleclaire:
I just don't like the reference to softlayer driver specifically in the driver.  I think it might be case that if we just always set it to `GetIP` output the bind will still work.  Since most times it's public IP and service should be able to advertise on that IP.  Do you have another driver you can try with?  especially a cloudy one such as `digitalocean`

Issue #533, opened 2015-02-12
EC2 credentials could expire


sthulb:
Under certain situations, credentials could expire for the EC2 driver, i.e. when using roles.

We could perhaps mitigate it by checking credentials and then either asking for more or using env vars.

nathanleclaire:
Good call out - this seems like something which could potentially be mitigated using config (as in `docker-machine config Drivers.AmazonEc2.AmazonSecretAccessKey mykey` to update the current key that you are using).  It's good practice to rotate keys regularly so I'm +1 on thinking about and covering this case well.

sthulb:
What do we do in the event of multiple accounts?

Keep track of the account ID the instances are in and then match them against the config so we end up with a config key of: `Drivers.EC2.[12 digit account ID].[KeyType]`?

ehazlett:
Why wouldn't we just throw an error and let the user present new
credentials?

If we want to add a grouping construct we would need additional options at
create time as well.

sthulb:
If we just throw errors things like the ls command break. I'd rather just try to detect them again.

ehazlett:
Isn't that going to do the same thing if you cannot detect the new ones?

PavelVanecek:
I just ran into this issue. In our configuration, AWS tokens expire after ~30 minutes. I can however see the old credentials cached in `~/.docker/machine/machines/mymachinename/config.json`. This is effectively prohibiting us from using amazon EC2 driver.

If I delete the credentials from JSON, I get `error EmptyStaticCreds: static credentials are empty`.

Relates to https://github.com/docker/machine/issues/2739 and https://github.com/docker/machine/issues/1840

nathanleclaire:
I wonder if we could modify the AWS driver to support temporary tokens.  I might be mis-remembering, but I think @jeffellin may have done some research in this regard.

jeffellin:
So this is a little nuanced depending on if you want to break backwards compatibility.  The truth is now that you are using the aws sdk you don't need to store credentials at all.  The sdk has a pre-determined notion of where they should be,  env variables,  .aws dir, etc.  So my recommendation is to stop reading the credentials from the config file and just use whats in the environment.  This should even enable STS tokens to work without issues since the sdk knows already how to work with those.

The issue with making this change is that users who don't have their amazon environment configured will need set them up otherwise an upgrade will break them. 

[...]

Issue #3748, opened 2016-09-13
AWS Driver: Does not retrieve credentials from metadata service


pasky:
docker-machine with AWS driver does not work well when actually run _from_ an ec2 instance with an IAM role assigned - in that case, even if credentials are not present in ~/.aws/credentials, the proper behavior is to retrieve them from `http://169.254.169.254/latest/meta-data/iam/security-credentials/...` which is what e.g. aws cli or the sdks do but for some reason docker-machine doesn't do that.

mrburrito:
+1

I'm trying to configure GitLab-CI runners that use docker-machine to spin up instances and the lack of instance profile support is making this both incredibly difficult and/or incredibly insecure. Need a fix ASAP.

PR #3763, opened 2016-09-19
Update AWS Credentials lookup to support use of assigned instance roles.

mrburrito:
This pull request addresses #3748, adding support for automatic configuration of AWS credentials from Instance Roles when running `docker-machine` from an EC2 server or ECS container. It replaces the original credential lookup code with an Amazon ChainProvider that retrieves credentials from the first valid source in this list:

1. Static credentials from the `--amazonec2-access-key`, `--amazonec2-secret-key`, and `--amazonec2-session-token`command line options

2. Credentials found in a specific credentials file or a non-default profile (not currently supported by CLI options, but hooks exist)

3. Amazon's default credential chain: (at time of submission)
    a.  Environment Variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`)
    b.  Default profile in `~/.aws/credentials`
    b.  IAM Role (on supported AWS infrastructure)


nathanleclaire:
Cool.

Can you update the documentation for the AWS driver to clearly outline the hierarchy / mechanics of how this works?  Otherwise I worry it will be quite opaque / unpredictable.

mrburrito:
Will do. I'll be pushing updates to the AWS SDK as well to support ECS task roles.

nathanleclaire:
Whoa, if you need to bump the vendored lib, please do so in a separate PR, otherwise this one becomes fairly difficult to review

mrburrito:
Will do. I'll separate them out tomorrow.

drivers/amazonec2/awscredentials.go
```diff
@@ -1,19 +1,79 @@
...
+	return &creds
+}
+
+func (c *defaultAWSCredentials) Credentials() *credentials.Credentials {
+	providers := []credentials.Provider{}
+	if c.AccessKey != "" && c.SecretKey != "" {
+		providers = append(providers, c.providerFactory.NewStaticProvider(c.AccessKey, c.SecretKey, c.SessionToken))
+	}
+	if c.Filename != "" || c.Profile != "" {
+		providers = append(providers, c.providerFactory.NewSharedProvider(c.Filename, c.Profile))
```
mjiderhamn: In the [SDK](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L91), `EnvProvider` has precedence over `SharedCredentialsProvider`. If there is a reason for doing things differently here, I suggest it is properly explained in comments.

drivers/amazonec2/awscredentials.go
```diff
@@ -1,19 +1,79 @@
...
+	AccessKey        string
+	SecretKey        string
+	SessionToken     string
+	Filename         string
+	Profile          string
+	providerFactory  ProviderFactory
+	fallbackProvider awsCredentials
+}
+
+func NewAWSCredentials(id, secret, token, filename, profile string) *defaultAWSCredentials {
```
mjiderhamn: I cannot see `filename` or `profile` being anything other than `""` except in test cases. Could these be removed altogether...?

mjiderhamn:
Fixes #533

drivers/amazonec2/awscredentials.go
mrburrito: Easily. I kept them in there to maintain the previous behavior. Since there is no external way for those to be updated, that code can be killed. I'll do that before I resubmit on top of pull #3772.

mrburrito:
Closing. This PR is superceded by the combination of #3772 and #3790

PR #3768, opened 2016-09-20
Remove incorrect hard-coded "fish" shell identification.

### LINKED ISSUE #3496: docker-machine env incorrectly identifies fish shell on master
If you run the following within fish it produces fish syntax instead of bash syntax, which is unexpected and breaks some of our automated scripts:
```
$ bash -c 'docker-machine env name'
set -gx DOCKER_TLS_VERIFY "1";
set -gx DOCKER_HOST "tcp://52.36.94.168:2376";
set -gx DOCKER_CERT_PATH "/home/v/.docker/machine/machines/name";
set -gx DOCKER_MACHINE_NAME "name";
# eval (docker-machine env name)
```
I tested this on master.

It's clearly doing the wrong thing. It should at least only check the value of $SHELL. It seems to be checking some random variable in addition to $SHELL: https://github.com/docker/machine/blob/8141874120abdc5c9056b6dbeb09bfb76a3eb290/libmachine/shell/shell.go#L25

Better yet, I would like a flag that lets me control the output format. I would also like to have some more machine readable formats available.

sakoht:
Fixes #3496 

The lines I removed cause the shell to be identified as `fish` if the `__fish_bin_dir` environment variable is set, even when `$SHELL` is explicitly set to something else.

This causes a fish user to get the `fish` env when running a non-fish sub-shell.  Until fixed all bash scripts that use `docker-machine env` will fail for `fish` users.

### DIFF
--- a/libmachine/shell/shell.go
+++ b/libmachine/shell/shell.go
@@ -22,9 +22,5 @@ func Detect() (string, error) {
 		return "", ErrUnknownShell
 	}
 
-	if os.Getenv("__fish_bin_dir") != "" {
-		return "fish", nil
-	}
-
 	return filepath.Base(shell), nil
 }


vikstrous:
:+1:

nathanleclaire:
Cool, thanks @sakoht -- you'll need to update the test as well please:

```
ok  	github.com/docker/machine/libmachine/provision/serviceaction	1.020s
--- FAIL: TestDetectFish (0.00s)
	Error Trace:	shell_test.go:29
	Error:		Not equal: "fish" (expected)
			        != "bash" (actual)
```

sakoht:
This PR is flawed.  

The real problem is that `$SHELL`  contains the login shell, not the current shell, which is what `docker-machine env` requires.

The real solution would have to look through the stack of parent processes until it hits one that is a shell.  The only reasonable way to know a shell is to check /etc/shells, though this is technically just the shells that can be used for login.

It would still probably be better to remove these lines, because it will force things to think they are fish that are not.  But this is undoubtedly helping someone who put it there.  I'm going to revisit this for a more thorough replacement....

sakoht:
Unless you want to leave it to the user to always use `--shell` if not running in your login shell.

nathanleclaire:
Thanks @sakoht , I'm going to close this for now, if you have a change of heart or file a new PR to help with this feel free to ping me

vikstrous:
This PR is still an improvement over what existed.

nathanleclaire:
I'm OK with whatever -- not being a fish user, it's hard for me to say what's the best course, but if you think it'd be good to merge I'm in favor @vikstrous .  It certainly is a little odd to try and auto-detect this somewhat arbitrary environment variable.

nathanleclaire:
Going to merge this and see how it turns out for us...

nathanleclaire:
LGTM

vikstrous:
Thanks

nathanleclaire:
👍

PR #3772, opened 2016-09-21
Updated aws-sdk-go to 1.4.10 to support ECS credential loading.

mrburrito:
This PR updates the AWS SDK to the latest version to support loading IAM credentials from ECS task roles. This is in support of the submitted fix for #3748, which will be submitted as an additional pull request once this is accepted.

Reference PR #3763 for initial conversations.


Godeps/Godeps.json
```diff
@@ -435,82 +462,82 @@
 		},
 		{
 			"ImportPath": "github.com/vmware/govmomi",
-			"Comment": "v0.1.0-454-g9051bd6",
+			"Comment": "v0.6.2",
```
mrburrito: Not entirely sure. I just ran the steps in CONTRIBUTING.md to update a library:

```
make dep-restore
go get -u github.com/aws/aws-sdk-go
make dep-save
```

I assumed any version changes were transitive dependencies.

mjiderhamn: AFAI can see this is only used for VMWare driver, and should supposedly be left untouched when making AWS changes?

nathanleclaire: Yeah, exactly.  `godep` has some very erratic behavior sometimes so it seems kinda wrong to me to modify these vmware lib parameters when we just want to bump the aws driver.

mrburrito: It looks like vmware changed the tag to [0.6.2](https://github.com/vmware/govmomi/tree/v0.6.2). The actual commit is the same as it was previously but the `v0.1.0-454.g9051bd6` tag no longer exists in their repository.

nathanleclaire: Ah, weird... Ok thanks.  I don't think there's anything we can do about that WRT this change so I guess we'll just have to accept that weirdness for the version bump.

nathanleclaire:
LGTM

PR #3779, opened 2016-09-22
Add an option for configuring custom userdata script for the amazonec2 driver.

### LINKED ISSUE #3778: Support custom userdata in amazonec2 driver.
Add support for custom userdata when deploying instances in AWS. This will allow runtime configuration of machines across AMIs rather than requiring all configs be hard-coded in an AMI before launch.

I'll be submitting a PR for this ticket.

mrburrito:
Added --amazonec2-userdata flag that accepts the path to a file containing a custom User Data script provided to the instance when it is launched. Updated documentation

for the amazonec2 driver to reflect the new option.  Fixes #3778

Signed-off-by: Gordon Shankman gshankman@gmail.com

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -563,8 +563,8 @@ func (d *Driver) securityGroupIds() (ids []string) {
 
 func (d *Driver) Base64UserData() (userdata string, err error) {
 	if d.UserDataFile != "" {
-		buf, io_err := ioutil.ReadFile(d.UserDataFile)
-		if io_err != nil {
+		buf, ioerr := ioutil.ReadFile(d.UserDataFile)
+		if ioerr != nil {
 			err = errorReadingUserData
 			return
 		}
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -3,6 +3,7 @@ package amazonec2
 import (
 	"crypto/md5"
 	"crypto/rand"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


nathanleclaire:
The AWS SDK will not base 64 encode it for you automatically?  I seem to recall that their CLI supports that

nathanleclaire:
Yeah, I think you do not need to perform the base64 encoding https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#UserData

nathanleclaire:
You might be able to use `file://` and avoid reading the file yourself as well

mrburrito:
You would think by reading their docs that the Go SDK qualifies as "using an Amazon SDK" and would be automatically encoded, but it failed when I didn't encode it myself.  :angry: I should probably submit a bug to Amazon about that. I tested it on both the current version of the Go SDK and my updated version in #3772 and I had to manually encode it for both.

I'll try the `file://` URI and see if that works.

nathanleclaire:
hmmm... yeah try `file://` since they might not have a way to check whether a raw string should be encoded or not if the "protocol" is not set

mrburrito:
I tried passing a file URI and an open File pointer (`os.Open`) to the `UserData` attribute and both failed with the non-Base64 encoding error. Not sure if its a bug in Amazon's code or I'm just not finding the right combination/type/format of value to provide to the SDK.

ifourmanov:
So user data functionality made it's way to official documentation (https://docs.docker.com/machine/drivers/aws/) but is not yet available in docker-machine

Shame, really need to use this.

+1

mrburrito:
@londoncalling I think this docs change (https://github.com/docker/docker.github.io/issues/37) didn't get rolled back from the master branch of docker.github.io and the docs are currently live.

nathanleclaire:
Just one nitpick to please take care of (the `snake_case` nit) and then I'm good to go on this.

mrburrito:
That's what I get for bouncing between 16 different languages in a day! Fixed.

nathanleclaire:
LGTM

Thanks for the hard work and persistent followups @mrburrito !!! 🌯 🌯 🌯

mrburrito:
Anytime. Is there a tentative release date for 0.9?

nathanleclaire:
Alongside Docker 1.13.0: https://github.com/docker/docker/milestone/56 (no fixed release date, but you can see the progress in the bar there)

PR #3783, opened 2016-09-23
Use net.JoinHostPort in ssh.Dial commands to allow for ipv6 addresses to work properly.

aaron-prindle:
There was an issue encountered in Windows where when using Powershell/cmd (which did not have an ssh command by default), calls using ssh.Dial would use the native go ssh binary which had an error handling ipv6 addresses (which Hyper V uses by default).  This fix puts the hostname:port in a format that allows it to work properly.


nathanleclaire:
LGTM

Thanks!

PR #3784, opened 2016-09-24
azure: add flag for static dns label

ahmetalpbalkan:
Introducing `--azure-dns` argument which creates a FQDN to the
created public IP resource for the VM. This way, users don't have to
pay for a static IP address.

e.g. a value like `foo` will be later used as IP `foo.westus.azure.cloudapp.net`.

Also modifying `ip` subcommand to operate on FQDN, where possible so that
the generated certs etc do not require re-generation when IP of the VM changes.


ahmetalpbalkan:
cc: @nathanleclaire

nathanleclaire:
LGTM 👍

Issue #3777, opened 2016-09-22
Azure driver azure-open-port option fails to process protocol


fjollberg:
Creating a machine with --azure-open-port 123/udp or --azure-open-port 123/tcp to specify protocol fails due to illegal naming of firewall rule.

E.g:
Error creating machine: Error in driver during machine creation: network.SecurityGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Long running operation terminated with status 'Failed': Code="InvalidResourceName" Message="Resource name Port7946/tcpAllowAny is invalid. The name can be up to 80 characters long. It must begin with a word character, and it must end with a word character or with '_'. The name may contain word characters or '.', '-', '_'."

docker-machine version 0.8.1, build 41b3b25

ahmetalpbalkan:
@andrestc it appears like the name for the azure firewall rule has `/` character which is not allowed, do you mind sending a fix, perhaps replacing `/` with `-` or removing it?

andrestc:
Hi @ahmetalpbalkan, I created a PR with a test trying to reproduce this issue(but failed to). I'll be happy to fix this once Im able to reproduce it somehow.

ahmetalpbalkan:
@andrestc here's an easy repro, you need to explicitly specify `/proto` part.

Also let's make sure that even when `/proto` is not specified, it's included in the name.

```
~ dm create -d azure --azure-open-port 5000/udp --azure-image 'coreos:coreos:stable:latest'  azure$RANDOM
Running pre-create checks...
(azure15930) Completed machine pre-create checks.
Creating machine...
(azure15930) Querying existing resource group.  name="docker-machine"
(azure15930) Resource group "docker-machine" already exists.
(azure15930) Configuring availability set.  name="docker-machine"
(azure15930) Configuring network security group.  name="azure15930-firewall" location="westus"
Error creating machine: Error in driver during machine creation: network.SecurityGroupsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Long running operation terminated with status 'Failed': Code="InvalidResourceName" Message="Resource name Port5000/udpAllowAny is invalid. The name can be up to 80 characters long. It must begin with a word character, and it must end with a word character or with '_'. The name may contain word characters or '.', '-', '_'."
```

PR #3785, opened 2016-09-24
drivers/azure: adds test to getSecurityRules

andrestc:
This adds a test to the getSecurityRules driver method, and is related to #3777.

@ahmetalpbalkan, I was not able to reproduce the issue reported on #3777, I tried with the test on this PR, the rules are not being created with `/` in their name. 

I will try to dig in a little more, but since I don't have an Azure account, I can't try to reproduce it with a manual test to get more insight

### DIFF
--- a/drivers/azure/util_test.go
+++ b/drivers/azure/util_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/arm/network"
+	"github.com/docker/machine/libmachine/drivers"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
@andrestc tests are always good. but can't say this one is particularly useful. What I can suggest instead is, extract `mkRule` inner function into an actual method and test it instead. Because it seems like you're actually trying to test the `mkRule`'s functionality.

andrestc:
@ahmetalpbalkan i was actually trying to reproduce the issue by testing the naming, which is actually done on the GetSecurityRules func. I was able to reproduce the bug testing on azure, will send a pr with tests for more methods afterwards (thats why i closed this one).

PR #3791, opened 2016-09-27
Update fish test now that magic env var is not used

nathanleclaire:
Not an amazing test, but better than nothing / failing


nathanleclaire:
LGTM

PR #3800, opened 2016-10-04
Add build support for FreeBSD

tolbrino:
I've verified that this builds a working binary on FreeBSD `11.0-RELEASE`.


nathanleclaire:
Oh, that's cool!  I don't think we'll distribute binaries directly but the option to compile from source is quite nice.

nathanleclaire:
LGTM!

Issue #3815, opened 2016-10-12
Support AzureGermanCloud


arafato:
Hi--,

could you please update your reference to the [Azure Go Autorest library](https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go)? The most current version provides support for all four Azure cloud environments including the Microsoft German Cloud. Thanks!

ahmetalpbalkan:
@arafato can you change title to `Support AzureGermanCloud`? It's unfortunately not as simple as updating go-autorest with right URLs. We need to create a subscription there to host the AAD application for docker-machine in Germany AAD PROD in the long term.

ahmetalpbalkan:
@arafato I have created a series of patches that updates dependencies and adds Germany support. It's on my branch here: https://github.com/ahmetalpbalkan/machine/tree/germany

You can compile from source (`make build-x`) or alternatively you can grab the binaries from this ZIP bundle if you trust me enough: https://cl.ly/2Y0s3J1M2W1k/docker-machine-germany.zip 

I do not have a Germany subscription for dev/test but I confirmed that when you tried to create a machine, it prompts you to log in with AAD endpoint in Germany, so things should work...

arafato:
@ahmetalpbalkan thank you so much! We will give it a try asap and let you know.

PR #3818, opened 2016-10-12
Update azure driver dependencies

ahmetalpbalkan:
In order to address #3815, bringing azure dependencies to
more up-to-date versions.

Also fixing some code caused by breaking changes introduced in Azure SDK.

@nathanleclaire PTAL


nathanleclaire:
LGTM

PR #3820, opened 2016-10-13
Close properly SSH clients created by NativeClient

PouuleT:
SSH clients created by ssh.Dial in case of a SSH NativeClient are not properly closed

Signed-off-by: Lucas BEE <pouulet@gmail.com>

### DIFF
--- a/libmachine/ssh/client.go
+++ b/libmachine/ssh/client.go
@@ -46,6 +46,7 @@ type NativeClient struct {
 	Hostname    string
 	Port        int
 	openSession *ssh.Session
+	openClient  *ssh.Client
 }
 
 type Auth struct {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Cool.  Thanks for this.  I wonder why error [here](https://github.com/docker/machine/pull/3820/files#diff-b19d47d6d93137e81fd31b886ef14559L246) is swallowed though.  Maybe it should be checked as well as the other one

PouuleT:
I was wondering why it was explicitly swallowed too, so I just copied the
way it was done for the session

No problem to check the error, but I have no idea how to handle properly a
Close() error here. They are usually discarded in the defers

What should we do if we have an error on the close? Log it (do we have a
way to log it properly from here?)? Panic (maybe not a good idea)?
If we do decide to do something here, does that mean we'll have to check
for all the other discarded error on Close() ( session or conn ) calls in
the file?

Le sam. 15 oct. 2016 à 01:45, Nathan LeClaire <notifications@github.com> a
écrit :

nathanleclaire:
Yeah, I think that since close errors are relatively uncommon, it's tended to be unfortunately common practice in Go programs to not check them on `defer`-ed statements, but here we don't need to `defer`, so might as well check it.

> What should we do if we have an error on the close?

Since the function in question returns `error`, let's return it if it's not nil.  Otherwise I would say `log.Debug` -- panic is overkill, sometimes SSH operations and things like that get re-tried.

nathanleclaire:
Looks good, there's an error with build but I'm 99% sure it's not your fault

PouuleT:
Yep, I added the log on the client's Close
For the session's Close(), I've made some tests to log when there's an error too, but it seems like there's one EOF error on all my tests
I've tried digging deeper, but even in a simple SSH test as in [this](https://godoc.org/golang.org/x/crypto/ssh#example-Dial), the session.Close() returned the EOF error

nathanleclaire:
> For the session's Close(), I've made some tests to log when there's an error too, but it seems like > there's one EOF error on all my tests
> I've tried digging deeper, but even in a simple SSH test as in this, the session.Close() returned the EOF error

Hmm... Yeah, no need to stress about it / fix it yourself if there was already an error there before which simply wasn't caught, but it is a bit troubling.

nathanleclaire:
LGTM

Thanks!

PR #3823, opened 2016-10-13
Dialer TLS timeout is too short, 2 -> 20 seconds

### LINKED ISSUE #3479: "tls: DialWithDialer timed out " error during docker-machine create
when I doing docker-machine create, it failed due to the following error for many times:
**[Error creating machine: Error checking the host: Error check
ing and/or regenerating the certs: There was an error validating certificates fo
r host "10.200.39.81:2376": tls: DialWithDialer timed out
You can attempt to regenerate them using 'docker-machine regenerate-certs [name]'.
Be advised that this will trigger a Docker daemon restart which will stop running containers.]**

Actually, the machine has been created successfully, the machine  can be list by 'docker-machine ls' command, and can be managed by docker-machine.
After review the source code, it seems this failure is caused by **Certificate Validation Timeout** , in machine/libmachine/cert/cert.go
// ValidateCertificate validate the certificate installed on the vm.
func (xcg *X509CertGenerator) ValidateCertificate(addr string, authOptions *auth.Options) (bool, error) {
	tlsConfig, err := xcg.ReadTLSConfig(addr, authOptions)
	if err != nil {
		return false, err
	}

	dialer := &net.Dialer{
		Timeout: time.Second * 2,
	}

	_, err = tls.DialWithDialer(dialer, "tcp", addr, tlsConfig)
	if err != nil {
		return false, err
	}

	return true, nil
}

There is a timeout variable, and it's value is 2 second, it's not enough!   in some cases, this operation needs more time than 2 second. 
There should be a recursive check with specific time in above function, at least, the timeout value should be assigned to a bigger value. (10 or 20 seconds )

sblackstone:
Fixes #3479

### DIFF
--- a/libmachine/cert/cert.go
+++ b/libmachine/cert/cert.go
@@ -257,7 +257,7 @@ func (xcg *X509CertGenerator) ValidateCertificate(addr string, authOptions *auth
 	}
 
 	dialer := &net.Dialer{
-		Timeout: time.Second * 2,
+		Timeout: time.Second * 20,
 	}
 
 	_, err = tls.DialWithDialer(dialer, "tcp", addr, tlsConfig)


nathanleclaire:
Seems reasonable!

Thanks for this!!

LGTM

nathanleclaire:
(will merge when CI is done, please ping me if I've forgotten)

nathanleclaire:
LGTM

PR #3830, opened 2016-10-17
Add COMPOSE_CONVERT_WINDOWS_PATHS to `docker-machine env` output

shin-:
See related Compose change for context: https://github.com/docker/compose/pull/4026

This is my first contribution to machine (the shame!), so this might be a little rough around the edges.

Also, I tried to add some tests for this, but with the dependency on `runtime.GOOS` I'm not quite sure how to do that. Appreciate any guidance!

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


nathanleclaire:
`runtime.GOOS` is public (exported) so you can just set it and set it back in the test probably

nathanleclaire:
this might break the existing tests when run on windows anyway though...

shin-:
Added test and updated the existing tests to make sure they don't break on Windows.

nathanleclaire:
Oh yeah, cool, this solution seems pretty good.  Joffrey on beast mode with these tests

nathanleclaire:
LGTM

PR #3839, opened 2016-10-20
Add DOCKER_API_VERSION to env

### LINKED ISSUE #2046: managing different versions of docker difficult with docker-machine
For presumably relatively obvious reasons, I may not have all my versions of docker running on the same version. This causes issues in that the docker client isn't smart enough to fall back on the api version if it's newer than the daemon it's trying to connect to.

It would be nice if `docker-machine env` could provide a hint that I could use to adjust my path.

e.g.
```
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.100:2376"
export DOCKER_CERT_PATH="/Users/xxx/.docker/machine/machines/dev"
export DOCKER_MACHINE_NAME="dev"
export DOCKER_VERSION="1.8.3"
export DOCKER_API_VERSION="v1.20"
# eval "$(docker-machine env dev)"
```

Obviously the best would be if the client was smarter and could fall back, or specify an api version, but I feel like this has a better chance of making it in "soon" than that would.

s-koba:
Fixes #2046

Signed-off-by: KOBAYASHI Shinji <koba@jp.fujitsu.com>

### DIFF
--- a/commands/env.go
+++ b/commands/env.go
@@ -13,11 +13,12 @@ import (
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/check"
 	"github.com/docker/machine/libmachine/log"
+	"github.com/docker/machine/libmachine/mcndockerclient"
 	"github.com/docker/machine/libmachine/shell"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


nathanleclaire:
Thanks!  I'm pretty interested in this but just merged another change to the similar code so can you  please rebase?

s-koba:
Sure. I just rebased it.

nathanleclaire:
Why have `DockerAPIVersionVar`?  It should pretty much just always print it right?

s-koba:
At first, I thought that `mcndockerclient.DockerAPIVersion()` may return an error when the remote docker engine is stopped for example. That's why I introduced `DockerAPIVersionVar`. But in fact, `docker-machine env` itself exits with an error in such cases. So it may be better to return the error if `mcndockerclient.DockerAPIVersion()` returns an error by any chance. How about the updated commit?

nathanleclaire:
> At first, I thought that mcndockerclient.DockerAPIVersion() may return an error when the remote docker engine is stopped for example. That's why I introduced DockerAPIVersionVar. But in fact, docker-machine env itself exits with an error in such cases. So it may be better to return the error if mcndockerclient.DockerAPIVersion() returns an error by any chance. How about the updated commit?

Yeah, `env` is not useful to user in this case (stopped/unreachable machine), so returning error makes sense.  I'd rather know about error as it happens than think everything is OK and find out it doesn't work (confusing).

This LGTM.

Thanks @s-koba !

PR #3841, opened 2016-10-20
Refactoring of the SUSE/openSUSE provisioner

flavio:
This provisioner is not relying on the bash script hosted at get.docker.com.

Helps fixing issue #23480


nathanleclaire:
LGTM

nice integration w/ the systemd provisioner as well

PR #3847, opened 2016-10-25
Fix ssh keyfile permissions on FreeBSD

tolbrino:
Previously, the permissions would be left open which results in a broken SSH setup between the client and the docker machine host.


tolbrino:
With this patch I was able to get a working setup of docker-machine, docker-compose and docker (client-only) running on FreeBSD 11.

nathanleclaire:
LGTM

Thanks @tolbrino ... It's not officially supported but I see no harm

What's the hypervisor/plugin you're using?

tolbrino:
Thanks. Using this with VirtualBox.

PR #3852, opened 2016-10-28
Improve error message if driver binary not available

kad:
This change provide more information which exact binary for particular driver is missing in PATH


nathanleclaire:
LGTM

nathanleclaire:
Thanks @kad !

PR #3870, opened 2016-11-14
provision/coreos: start docker 1.12 correctly

pschultz:
Originally reported as coreos/bugs#1661.

Docker >= 1.12 does not accept the "daemon" argument anymore, making CoreOS machines in the beta and alpha branches fail to start properly.

This makes the provisioner check for the presence of `dockerd`. If it's there, don't add the argument. That's the simplest test I could think of. Other options are checking the Docker or CoreOS versions, but it wasn't immediately obvious to me how to do that.


nathanleclaire:
I'm confused on this one.  If `type dockerd` fails, then it shouldn't be `dockerd` in the init script at all like you have here, right?

nathanleclaire:
Unless `dockerd` was just what `docker` was called before... let me knwo

pschultz:
The image starts docker as `/usr/lib/coreos/dockerd`, which is some kind of wrapper and ends in `exec docker "$@"`. I couldn't figure out where this script is coming from (it doesn't seem to be part of docker-machine). But since it just passes its arguments on to docker, calling it with "daemon" is certainly wrong with 1.12 or newer.

I can paste the script here if that's useful.

pschultz:
By the way, the full command is in coreos/bugs#1661, in the "Actual Behavior" section.

pschultz:
After thinking about it a bit more, making decisions based on the presence (or absence) of a programm named dockerd in the path is very short sighted. There are many reasons dockerd may or may not exist.

I'll see if I can do something useful with `OsReleaseID`, and if that doesn't work parse `docker version` instead.

Definitely don't merge yet, please.

pschultz:
Updated with parsing the output of `docker --version`. Please review again.

nathanleclaire:
Thanks, seems like a good direction, I have some suggestions:

- Please name the package to `versioncmp` to make its purpose more explicit
- Maybe the comparision function(s) might be better off returning an error?  If I pass some invalid string to this function, it should probably return `error` right?
- Can you pull the string parsing of `Docker version 1.13.0-rc1, build 75fd88b` into its own little function with its own test?  e.g., it takes in that string and returns `1.13.0-rc1`
- What happens if the version is release candidate, e.g., `1.13.0-rc1`?

Quite creative solution.  Thanks!

nathanleclaire:
Don't mind the travis results for now, I think go-lint upstream changed on us.

pschultz:
I haven't considered the rc versions. I just stole the version functions from github.com/docker/docker/api/types/versions. I suppose the input is more controlled there.

There is no obvious way to compare versions with arbitrary suffixes, e.g. "1.13.0-rc1" with "1.13.0-beta2". I don't want to special case rc-$number, so I lean towards ignoring the suffix, assuming there are no API differences within a pre-release phase as far as provisioning is concerned. Objections?

libmachine/provision/coreos.go
```diff
@@ -61,8 +62,15 @@ func (provisioner *CoreOSProvisioner) GenerateDockerOptions(dockerPort int) (*Do
 		engineCfg bytes.Buffer
 	)
 
-	driverNameLabel := fmt.Sprintf("provider=%s", provisioner.Driver.DriverName())
-	provisioner.EngineOptions.Labels = append(provisioner.EngineOptions.Labels, driverNameLabel)
```
pschultz: That was an accident. Will fix soon.

nathanleclaire:
> I haven't considered the rc versions. I just stole the version functions from github.com/docker/docker/api/types/versions. I suppose the input is more controlled there.

Mmm, maybe we can do a proper vendor of those instead then.  I know godep can be a pain, but I'd much rather have `import versioncmp "github.com/docker/docker/api/types/versions"` than just copy-paste the code.

nathanleclaire:
We also just usually focus on making provisioning work for the latest version of Docker, so that's another possibility (to just go straight to `dockerd`), although I understand the CoreOS sometimes ships versions of Docker which are a few behind the latest.

pschultz:
github.com/docker/docker/api/types/versions deals with API versions, not the Docker binaries. That's probably why they don't have to deal with pre-releases and omit some error checks.

Right now, `versioncmp.Equal("1.13.0-rc1", "1.13.1-rc2")` returns true, which is wrong. I want to change `compare` here to ignore any dash-suffixes and not vendor the package in docker/docker.

pschultz:
I fixed the `versioncmp.Equal("1.13.0-rc1", "1.13.1-rc2")` case.

I don't think making versioncmp return errors is worth the trouble. It stops as soon as its clear which version is newer, e.g. given "1.a" and "2.b" it stops right after comparing 1 and 2. That comparison is even correct in a sense. 

If you insist I'll make compare check for well-formed versions, but I'd argue it's robust enough right now.

I've added the test for extracting the version string from `docker --version`.

nathanleclaire:
> github.com/docker/docker/api/types/versions deals with API versions, not the Docker binaries. That's probably why they don't have to deal with pre-releases and omit some error checks.

gotcha... yeah duplicating the code here seems OK since it's pretty small

> I don't think making versioncmp return errors is worth the trouble. It stops as soon as its clear which version is newer, e.g. given "1.a" and "2.b" it stops right after comparing 1 and 2. That comparison is even correct in a sense.

Yeah, you have some precedent there in your favor since the stdlib has many string comparison functions without error return, my concern was around "what if someone passes something totally bogus"... We can merge and see how things go though.

LGTM!

Thanks for the persistence and especially the extra elbow grease around testing, I find it can really pay off !

klausenbusk:
Should this be cherry-picked for v0.8.x?

Docker 1.12 just went into CoreOS stable yesterday, so now docker-machine stable isn't working with any CoreOS images..

It broke Gitlab autorunner platform yesterday: https://gitlab.com/gitlab-com/infrastructure/issues/957

Issue #3868, opened 2016-11-12
Modification of vmwarefusion driver to support


cab105:
I'd like to propose augmenting the vmwarefusion driver to support VMware Workstation at least for the Linux case.  The driver itself required slight modification to find the dhcpd leases, and an omission of exporting /home to the VM to prevent a conflict with setting up the docker user on the parent system.

The primary driver for this is to add VMware Fusion/Workstation support to minikube which also makes use of the docker-machine drivers.

The commit itself can be found at:
https://github.com/cab105/machine-1/commit/f85533b92ff1e7282ecd0e7a1e721a929eb09ee0

cab105:
Closing as the PR resolving this issue has been closed as well.

PR #3885, opened 2016-11-21
Add support for VMware Workstation to Fusion driver

cab105:
Signed-off-by: Christopher A. Baumbauer <cab@cabnetworks.net>


cab105:
This is to resolve issue #3868.

StefanScherer:
I think it would be better to have a `vmwareworkstation` driver instead of using the `vmwarefusion` driver for VMware Workstation. For Windows VMware Workstation there is an external driver plugin at https://github.com/pecigonzalo/docker-machine-vmwareworkstation/, maybe you should start something similar for Linux. An external plugin has the advantage that you don't have to wait until the driver plugin got merged into core codebase.

cab105:
Perhaps, but this would result in a lot of duplicate code between the Fusion
and Linux workstation driver which is why I moved the files back away from the
arch specific naming convention.

The modification itself is mainly adding arch switches to tell the driver to
look in /etc instead of /Library for the vmware specific network configs, and
/usr/bin instead of /Applications/VMWare\ Fusion.app for vmrun.

Windows Workstation is a completely different beast, and is one I'm not too
familiar with overal.

nathanleclaire:
@cab105 

> Perhaps, but this would result in a lot of duplicate code between the Fusion
and Linux workstation driver which is why I moved the files back away from the
arch specific naming convention.

Did you consider writing a driver plugin that embeds the existing workstation driver? e.g. (totally spitballing this code),

```go
package vmwareworkstation

import "github.com/docker/machine/drivers/vmwarefusion"

type Driver struct {
    vmwareFusion.Driver
}
```

You could re-use a lot of the methods, and just "over-ride" the methods you want to change.

nathanleclaire:
At any rate, we're unable to accept new plugins in the core right now.  Thanks for the contribution @cab105 , my suggestion would be to make and distribute a `docker-machine-driver` plugin

PR #3887, opened 2016-11-21
Close plugin binary even when driver server fails

andrestc:
When running `libmachine` inside a long lived process, I noticed some child processes being left in `<defunct>` state. This was happening whenever an operation like creating or removing a machine failed. Tracking this down, I figured sometimes closing the server driver would fail and the plugin binary would not be closed, which prevents `cmd.Wait()` from being called and properly handling the child process exit code.

With this commit the RPC client driver tries closing the plugin binary even when closing the driver server fails. This prevents the driver process from entering in `<defunct>` state when it failed before closing.

### DIFF
--- a/libmachine/drivers/rpc/client_driver.go
+++ b/libmachine/drivers/rpc/client_driver.go
@@ -213,10 +213,11 @@ func (c *RPCClientDriver) close() error {
 	log.Debug("Making call to close driver server")
 
 	if err := c.Client.Call(CloseMethod, struct{}{}, nil); err != nil {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Nice, thanks.  Looks pretty good just need to test the behavior of a hung SSH process as I think that's what made me introduce the early return in the first place.

andrestc:
Cool, I´ll do some tests with ssh.

nathanleclaire:
Tried it out and this LGTM

PR #3890, opened 2016-11-24
Add several EnvVars and a default region for Rackspace driver

twang2218:
Add several missing EnvVars in Rackspace driver, which exist in 'openstack' driver.
Add a default region, Northern Virginia (IAD), to the Rackspace driver, just like other cloud providers' driver.

Signed-off-by: Tao Wang <twang2218@gmail.com>


nathanleclaire:
Cool -- thanks! -- LGTM

PR #3891, opened 2016-11-25
libmachine/provision: wait for apt-get update lock

andrestc:
This PR adds a retry mechanism to the `apt-get update` calls. It will retry if a failure to obtain the lock is detected which may be caused by cloudinit scripts running `apt-get update` before the provisioner.

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -52,7 +52,9 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
 	}
 
 	if updateMetadata {
-		waitForLockAptGetUpdate(provisioner)
+		if err := waitForLockAptGetUpdate(provisioner); err != nil {
+			return err
+		}
 	}
 
 	command := fmt.Sprintf("DEBIAN_FRONTEND=noninteractive sudo -E apt-get %s -y  %s", packageAction, name)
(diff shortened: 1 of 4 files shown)


andrestc:
I would like some suggestions on how to properly test `waitForLockAptGetUpdate` as I'm finding it very hard to test a function that rely on the `WaitFor` func.

nathanleclaire:
Cool.  Really it probably should check before all `apt-get` calls, but this is a great start -- thanks.

> I would like some suggestions on how to properly test waitForLockAptGetUpdate as I'm finding it very hard to test a function that rely on the WaitFor func.

There's an existing `FakeSSHCommander` in `libmachine/provision/provisiontest/sshcommander.go`, or you could write your own since `SSHCommander` is an `interface`, so you could make a struct that mocks it out like so I think:

```go
type WaitTestSSHCommander struct {}

func (sshCmder *WaitTestSSHCommander) SSHCommand(args string) (string, error) {
    // ...does whatever you want
}
```

nathanleclaire:
LGTM -- feel free to followup with tests if you're feeling ambitious

andrestc:
thanks, will do it @nathanleclaire... really busy days past few weeks

nathanleclaire:
yup I understand :)

fazith27:
Hi, @andrestc, just checking when this will be available?

PR #3898, opened 2016-11-29
Add a `docker-machine reinstall`…

ayufan:
This commands allows to use a cloud instance API to schedule reinstall of machine, by completely wipping its content and then re-provisioning.

This PR is created as proof-of-concept supports only `DigitalOcean` reinstall API, but quite easily any other API can be supported.

This is the output of command execution on existing machine:
```
Stopping "kamil-gitlab-test-reinstall"...
Reinstalling "kamil-gitlab-test-reinstall"...
Waiting for "kamil-gitlab-test-reinstall" to become Running...
Machine "kamil-gitlab-test-reinstall" was started.
Detecting the provisioner...
Installing Docker...
Copying certs to the local machine directory...
Copying certs to the remote machine...
Setting Docker configuration on the remote daemon...
```

This feature is needed by [GitLab Runner docker-machine based auto-scaling](https://docs.gitlab.com/runner/configuration/autoscale.html) to make it possible to re-use existing machines.

What do you think of this? Is this something worth and mergeable (if improved)?

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -305,6 +305,17 @@ var Commands = []cli.Command{
 		Action: runCommand(cmdProvision),
 	},
 	{
+		Name:   "reinstall",
+		Usage:  "Re-install and re-provision existing machines",
+		Action: runCommand(cmdReinstall),
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "y",
+				Usage: "Assumes automatic yes to proceed with reinstall, without prompting further user confirmation",
+			},
+		},
+	},
+	{
 		Name:        "regenerate-certs",
 		Usage:       "Regenerate TLS Certificates for a machine",
 		Description: "Argument(s) are one or more machine names.",
@@ -420,6 +431,7 @@ func machineCommand(actionName string, host *host.Host, errorChan chan<- error) 
 		"upgrade":       host.Upgrade,
 		"ip":            printIP(host),
 		"provision":     host.Provision,
+		"reinstall":     host.Reinstall,
 	}
 
 	log.Debugf("command=%s machine=%s", actionName, host.Name)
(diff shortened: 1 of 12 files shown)


nathanleclaire:
Hmm... I dunno, I would think something like `docker-machine import --digitalocean-id ... existingbox`, `docker-machine provision existingbox` might be more apropos, no?

ahmetalpbalkan:
Agreed, `reinstall` is completely cryptic, especially when it's supported only by 1 out of 15 drivers.

> to make it possible to re-use existing machines.

In this case, does DigitalOcean actually wipe all the data in the machines (including docker certs) and burn the clean base OS image to the droplet's disk? If so, I would suggest calling the `RebuildByImageSlug` on your side and then using `docker-machine import` or `docker-machine create -d generic` as Nathan suggested.

nathanleclaire:
To be clear, `import` is not an existing feature, rather something I've thought might be nice for a while (bring your own existing machines based on instance ID), in this case I'm a bit confused why not just create a new machine if you're going to wipe out the whole OS anyway.

ayufan:
nathanleclaire: because of cost, if you destroy machine and re-create it often you are hit by full billing cycle.

Yes I saw that this is only supported by Digital Ocean.

Probably it doesn't make sense to introduce a feature just for one provider.

nathanleclaire:
What's the motivation for wanting to wipe the machine's state?  i.e., why is not `provision` alone sufficient?

ayufan:
nathanleclaire: we (at GitLab) use docker-machine to provision nodes on-demand for running CI jobs, since these nodes do run docker containers in privileged mode we have to wipe them before running another untrusted job

nathanleclaire:
Hm, yeah, I wonder if it might be a slightly better fit to run a little container which makes the DigitalOcean API call to wipe the machine on its own.

PR #3926, opened 2016-12-15
Disable update of existing docker installation on RedHat and its derivatives.

mrburrito:
This patch excludes the `docker-*` packages from the `yum update` command issued by the RedHat provisioner. When `docker-engine` was pre-installed on the image used by `docker-machine`, it was always updated to the "latest and greatest" during the provisioning process. This is contrary to the desired behavior of using an installed `docker` if it is found.

This issue came to light during discussions of a bug with interactions between Docker 1.12.4 and GitLab-CI. See discussions here: https://github.com/docker/machine/issues/1702

### DIFF
--- a/libmachine/provision/redhat.go
+++ b/libmachine/provision/redhat.go
@@ -177,7 +177,7 @@ func (provisioner *RedHatProvisioner) Provision(swarmOptions swarm.Options, auth
 	}
 
 	// update OS -- this is needed for libdevicemapper and the docker install
-	if _, err := provisioner.SSHCommand("sudo -E yum -y update"); err != nil {
+	if _, err := provisioner.SSHCommand("sudo -E yum -y update -x docker-*"); err != nil {
 		return err
 	}
 


mrburrito:
@nathanleclaire this is the PR for the RedHat docker exclusions we were discussing

mrburrito:
Tested this against a CentOS 7.1 image with Docker 1.12.3 pre-installed. Before this patch, Docker was upgraded to 1.12.4; after patching, 1.12.3 remained after provisioning.

nathanleclaire:
LGTM -- thanks @mrburrito .  It does seem a bit odd to update every package except Docker, but I don't know what else we can do without re-thinking the existing provisioning model a bit.

LinBoLen:
error occur

```
% docker-machine create \                                                                                                                                                   127 ↵
--driver generic \
--generic-ip-address *.*.*.* \
--generic-ssh-key ~/.ssh/aaaa_ee_rsa \
--generic-ssh-user root \
aaaa-ee
Running pre-create checks...
Creating machine...
(aliyun-ee) Importing SSH key...
Detecting the provisioner...
Provisioning with centos...
Error creating machine: Error running provisioning: something went wrong running an SSH command
command : sudo -E yum -y update -x docker-*
err     : exit status 1
output  : zsh:1: no matches found: docker-*
```

LinBoLen:
it's zsh problem

PR #3938, opened 2016-12-25
Extended the server key to Client auth usage

### LINKED ISSUE #3903: Failed to run a swarm as the server certs cannot be used as client cert.
Originally, I raised an issue, docker/docker#24982, as I could not follow the documentation, "Evaluate Swarm in a sandbox", <https://docs.docker.com/swarm/install-w-machine/>. 

I created a script to follow the document to create a swarm. And the script works before, but not anymore.

Here is the script:

```bash
#!/bin/bash

function create() {
    echo "Create 3 machines"
    docker-machine create -d virtualbox --engine-opt debug manager
    docker-machine create -d virtualbox --engine-opt debug agent1
    docker-machine create -d virtualbox --engine-opt debug agent2
    docker-machine ls

    echo "Create Discovery token"
    eval "$(docker-machine env manager)"
    Token=`docker run --rm swarm create`
    echo "Service Discovery Token is ${Token}"

    echo "Create Swarm Manager"
    docker run -dt \
        --name manager \
        -p 3376:3376 \
        -v /var/lib/boot2docker:/certs:ro \
        swarm \
            -l "debug" --debug \
            manage \
                -H 0.0.0.0:3376 \
                --tlsverify \
                --tlscacert=/certs/ca.pem \
                --tlscert=/certs/server.pem \
                --tlskey=/certs/server-key.pem \
                token://${Token}

    Size=2
    for i in $(seq 1 ${Size})
    do
        echo "Create Swarm Node ${i}"
        eval "$(docker-machine env agent${i})"
        docker run -d \
            --name agent \
            swarm \
                -l "debug" --debug \
                join \
                --addr="$(docker-machine ip agent${i}):2376" \
                token://${Token}
    done
}

function remove() {
    docker-machine rm -y manager agent1 agent2
}

function swarm_env() {
    echo "export DOCKER_TLS_VERIFY=1"
    echo "export DOCKER_CERT_PATH=~/.docker/machine/machines/manager"
    echo "export DOCKER_HOST=$(docker-machine ip manager):3376"
}

function main() {
    Command="$1"
    shift
    case "${Command}" in
        create) create ;;
        env)    swarm_env ;;
        remove) remove ;;
        *)      echo "Usage: $0 <create|env|remove>" ;;
    esac
}

main "$@"
```

My current `docker-machine` version is 0.8.2:

```bash
$ docker-machine -v
docker-machine version 0.8.2, build e18a919
```

After the swarm was created, the `docker info` showed me the swarm nodes are not working:

```bash
$ docker info
Containers: 0
 Running: 0
 Paused: 0
 Stopped: 0
Images: 0
Server Version: swarm/1.2.5
Role: primary
Strategy: spread
Filters: health, port, containerslots, dependency, affinity, constraint
Nodes: 2
 (unknown): 192.168.99.103:2376
  └ ID:
  └ Status: Pending
  └ Containers: 0
  └ Reserved CPUs: 0 / 0
  └ Reserved Memory: 0 B / 0 B
  └ Labels:
  └ UpdatedAt: 2016-12-03T01:17:41Z
  └ ServerVersion:
 (unknown): 192.168.99.102:2376
  └ ID:
  └ Status: Pending
  └ Containers: 0
  └ Reserved CPUs: 0 / 0
  └ Reserved Memory: 0 B / 0 B
  └ Labels:
  └ UpdatedAt: 2016-12-03T01:17:41Z
  └ ServerVersion:
Plugins:
 Volume:
 Network:
Swarm:
 NodeID:
 Is Manager: false
 Node Address:
Security Options:
Kernel Version: 4.4.27-boot2docker
Operating System: linux
Architecture: amd64
CPUs: 0
Total Memory: 0 B
Name: 5454869d197e
Docker Root Dir:
Debug Mode (client): false
Debug Mode (server): false
WARNING: No kernel memory limit support
```

And the logs of `swarm` container on the `manager` node shows following errors:

```bash
DEBU[1900] Failed to validate pending node: The server probably has client authentication (--tlsverify) enabled. Please check your TLS client certification settings: Get https://192.168.99.103:2376/info: remote error: bad certificate  Addr=192.168.99.103:2376
DEBU[1900] Failed to validate pending node: The server probably has client authentication (--tlsverify) enabled. Please check your TLS client certification settings: Get https://192.168.99.102:2376/info: remote error: bad certificate  Addr=192.168.99.102:2376
```

@dongluochen decided to change the swarm documentation instead (docker/swarm#2442), by using `docker-machine`'s `--swarm` option, so the issue was closed.

But I still don't think that is a fix for the original problem, the original document worked fine just months ago, and had more details of how the swarm was created without hiding too much details in `--swarm` option.

So, I'm not giving up, after I understand more about docker-machine, I gave it another try. then I found out the commit c0b721b in docker machine, which is "Disable extended key usage as a client for non-Swarm-master server certs", caused the problem.

```diff
-		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth}
 -		for _, h := range hosts {
 +		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}
 +		if opts.SwarmMaster {
 +			// Extend the Swarm master's server certificate
 +			// permissions to also be able to connect to downstream
 +			// nodes as a client.
 +			template.ExtKeyUsage = append(template.ExtKeyUsage, x509.ExtKeyUsageClientAuth)
 +		}
```

So, if I followed the documentation, create a docker host without using `--swarm-master` option, the generated `server.pem` will not have `TLS Web Client Authentication`.

Here is the difference in the `server.pem` between normal `docker-machine create` and `docker-machine create --swarm-master`:

```diff
         X509v3 extensions:
             X509v3 Key Usage: critical
                 Digital Signature, Key Encipherment, Key Agreement
             X509v3 Extended Key Usage:
-                TLS Web Server Authentication
+                TLS Web Server Authentication, TLS Web Client Authentication
             X509v3 Basic Constraints: critical
                 CA:FALSE
             X509v3 Subject Alternative Name:
```

That's the reason why it works before, as it used to generate both extended usage, but now it only generate `TLS Web Server Authentication` unless `--swarm-master` is specified.

Without having `TLS Web Client Authentication` extended usage, there is no easy way to create a working swarm, unless I regenerated the certs by myself.

So, please enable the `TLS Web Client Authentication` for `server.pem`.

twang2218:
Fixes #3903
This PR revert part of c0b721b, so the server key will have
Client usage. Otherwise, the server created without using
`--swarm-master` will not be run swarm master later. This
is important for create swarm manually, such as creating a
swarm by following the previous docs 'Evaluate Swarm in a sandbox'.

Signed-off-by: Tao Wang <twang2218@gmail.com>

### DIFF
--- a/libmachine/cert/cert.go
+++ b/libmachine/cert/cert.go
@@ -167,13 +167,9 @@ func (xcg *X509CertGenerator) GenerateCert(opts *Options) error {
 		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}
 		template.KeyUsage = x509.KeyUsageDigitalSignature
 	} else { // server
-		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}
-		if opts.SwarmMaster {
-			// Extend the Swarm master's server certificate
-			// permissions to also be able to connect to downstream
-			// nodes as a client.
-			template.ExtKeyUsage = append(template.ExtKeyUsage, x509.ExtKeyUsageClientAuth)
-		}
+		// Extend the server certificate permissions to be able to connect to downstream
+		// nodes as a client, this will allow the server can be act as a swarm master later.
+		template.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth}
 		for _, h := range opts.Hosts {
 			if ip := net.ParseIP(h); ip != nil {
 				template.IPAddresses = append(template.IPAddresses, ip)


tronicum:
Great! This is a real issue if you hot plug a Swarm Master after setting up a cluster.

nathanleclaire:
Thanks for the PR @twang2218 -- it is intended to be the way that it was before, where only nodes you have deliberately marked as masters should have master-level permissions.  Otherwise any node on the cluster could run commands on any other node by default, which is a privilege escalation concern (think about having only one node to protect vs. tens or hundreds).

Any other ideas?  e.g., revising the "legacy" Swarm documentation?

tronicum:
Can I provide a option/PR to have backup masters? I setup my cluster before the change so I am currently not affected by this change prevents a failover scenario. My kvstore(consul) allows multiple masters.

nathanleclaire:
> Can I provide a option/PR to have backup masters? I setup my cluster before the change so I am currently not affected by this change prevents a failover scenario. My kvstore(consul) allows multiple masters.

You're always welcome to submit code :) But in this case I'd suggest a more detailed proposal -- if you would like another flag for `create`, I'm guessing creating those nodes with `--swarm-master` as well wouldn't work?  I'm a bit iffy on understanding of multi-master support in https://github.com/docker/swarm

tronicum:
I am not an complete swarm expert, too maybe we CC @ehazlett ?
First I am a docker pre 1.12 user. 
On https://medium.com/on-docker/toward-a-production-ready-docker-swarm-cluster-with-consul-9ecd36533bb8#.ug1udf4ky and https://docs.docker.com/swarm/reference/manage/  there is a mysterious replication flag. 

If there is replication there should be at least master and slave. I dunno the native docker swarm code but it should at least support failover. 

Consul allows master election via the raft protocol. 

Greetings Stefan

nathanleclaire:
Going to close this one as it's intended to be setup the way it is already, however as always happy to take proposals / PRs for alternative solutions in the multi-manager idea scenario...

nathanleclaire:
Thanks as always for the contributions!

PR #3947, opened 2017-01-07
adds an example driver to --help

sanscontext:
Just adds a few words to the `create --help` text so you don't actually need to read the doc to remind yourself what some of the drivers are.

cc @nathanleclaire for 👀 

Signed-off-by: LRubin <lrubin@docker.com>

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -38,7 +38,7 @@ var (
 		cli.StringFlag{
 			Name: "driver, d",
 			Usage: fmt.Sprintf(
-				"Driver to create machine with.",
+				"Driver to create machine with, for example virtualbox",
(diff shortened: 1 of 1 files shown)


nathanleclaire:
How about `default: virtualbox` ?

I'm thinking about turning in a PR to change that like we discussed... Folks don't really use `none` much

sanscontext:
SGTM. Do you want me to edit my PR? Wait until your PR? Close and combine into your PR?

nathanleclaire:
hm, hold on for mine... it should change the default displayed and might be sufficient

nathanleclaire:
I changed the default!

PR #3965, opened 2017-01-25
driver: sadeem.io driver

Ahmah2009:
Add support for Sadeem.io, **[Sadeem.io](https://www.sadeem.io)** currently provide cloud in **Saudi Arabia**, **UAE**, and soon over **Turkey** and overall Middle East 

# Configuration

**The main configuration options are:**

**--sadeem-apikey**: Sadeem api key
**--sadeem_client_id**: client Id in Sadeem
**--sadeem_datacenter**: Sadeem Datacenter Name to create vm on
**--sadeem_service_offer**: Service offer name to use
**--sadeem_template**: Name of Template to use to create vm wih
	

SSH keypair will be generated for each new VM.

The default template used is Ubuntu Template and may be changed with the **--sadeem_template** option.

## Docker machine support

The following docker machine commands work as expected:

**create
ls
ssh
start, stop, restart, kill and rm
inspect**

### DIFF
--- a/cmd/machine.go
+++ b/cmd/machine.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/machine/drivers/none"
 	"github.com/docker/machine/drivers/openstack"
 	"github.com/docker/machine/drivers/rackspace"
+	"github.com/docker/machine/drivers/sadeem"
 	"github.com/docker/machine/drivers/softlayer"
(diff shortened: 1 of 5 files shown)


nathanleclaire:
Thanks for the PR, but we have a plugin model now and aren't accepting new drivers into the core, please see e.g. https://github.com/dhiltgen/docker-machine-kvm for an example of how to distribute your own.

Issue #3261, opened 2016-04-01
SSH keys already existing at create time should not be removed by 'rm'


nathanleclaire:
See Scott's article for an explanation: http://blog.scottlowe.org//2016/03/31/docker-machine-openstack-ssh-keys/

Essentially, drivers should not remove SSH keys which were not created as part of the `docker-machine create` process.

Ideally, we should vet all existing drivers with this functionality to ensure that they don't have such destructive operations.

OpenStack driver is known to have this bug -- cc @ggiamarchi

lowescott:
Thanks, @nathanleclaire. I searched the GitHub issues before posting that article and _thought_ I found at least one thread where this behavior was acknowledged as "expected" behavior, though it certainly wasn't how I expected it to work.

nathanleclaire:
Yeah, I think I remember the issue you mention, it's possible I may have misunderstood the problem.

jmyounker:
It's surprising behavior.  Horrifyingly surprising behavior.

nathanleclaire:
Thanks for your input @jmyounker.  I'm happy to accept a PR if you'd like to submit a fix.

adaikal:
@nathanleclaire This issue of docker machine deleting the key we pass to it, has really hurt our automation process. We prefer to have one keypair for all the hosts that we provision on openstack.

May I know whether this will be fixed at any point.

I'm a newbie to docker machine, could you please direct me to right path, so that I can give a try of fixing it, if any other is not working on it.

Thank you.

nathanleclaire:
Hi @adaikal, the code for the Openstack driver is here: https://github.com/docker/machine/tree/master/drivers/openstack

You most likely want to implement some kind of check that will not clean up the key if the user specified it explicitly.  (Implementation is up to you -- may need to add some sort of additional parameter to the `Driver` struct to track this information post-create)

@fsoppelsa Might have input on the Openstack-specific bits

fsoppelsa:
@adaikal removing the instance key when a machine is deleted is triggered automatically now, so to keep things in order, I couldn't even find a reason to save it: You can share the same Nova key since Machine 0.6.

adaikal:
@fsoppelsa  The plan was to setup a CaaS cloud and to have all the instances to share same SSH keypair so that sysadmin could use one key for all the instances that get launched.

Now because of machine deleting the keypair on instance creation failure, we have fall back to use machine's default behavior of  having different key for each instance.

fsoppelsa:
[...]

PR #3969, opened 2017-01-26
FIX : Openstack driver do not remove keypair by default

bruno-morel:
#3261  - proposed solution: flag to activate key removal only if asked

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -45,6 +45,7 @@ type Driver struct {
 	ComputeNetwork   bool
 	FloatingIpPoolId string
 	IpVersion        int
+	KeypairDelete    bool
 	client           Client
 }
 
@@ -52,6 +53,7 @@ const (
 	defaultSSHUser       = "root"
 	defaultSSHPort       = 22
 	defaultActiveTimeout = 200
+	defaultKeypairDelete = false
 )
 
 func (d *Driver) GetCreateFlags() []mcnflag.Flag {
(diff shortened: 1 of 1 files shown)


jrandall:
This seems like a reasonable solution to a massive problem with the docker-machine openstack driver - what is holding up getting this merged?

nathanleclaire:
Why have a flag to indicate not to delete the keypair? This is the desired behavior in the case of a machine which creates a key as part of the creation process, so shouldn't it check that `openstack-keypair-name` is `""` before deleting instead?

bruno-morel:
I'm not a big fan of non-explicit, magic behaviour, but if you think it's more on par with other driver's behaviour, I'll amend the commit to reflect this. Just let me know.

bruno-morel:
@nathanleclaire any feedback ? want me to change the logic ?
I'd really like to push that fix forward... :)

davidarcher:
I ran into this recently when using docker-machine against openstack the first time. For my use case, I'm actually fine with autogenerated keys but it was definitely surprising as a user when docker-machine deleted the keypair that i manually uploaded and was going to share between my launched instances.

+1 for just changing the logic so we only delete the key if it was autocreated by docker-machine.

davidarcher:
Since this PR hasn't moved forward, I've opened up one of my one that uses the same logic as the EC2 driver. See https://github.com/docker/machine/pull/4358

davidarcher:
This PR can be closed now that #4358 has gone in. It's finally fixed! 👍

bruno-morel:
good job David :-)

Issue #3908, opened 2016-12-07
--virtualbox-share-folder on windows


michzimny:
Hi,
I have installed docker-machine Version: 0.9.0-rc2, build 7b19591 on Windows 10 x64. I have VirtualBox 5.1.10 r112026. I tried to use the new --virtualbox-share-folder switch.

I try the following call:
```
docker-machine create -d virtualbox --virtualbox-no-share --virtualbox-share-folder c:\\workspace:workspace default
```

I expect it to mount the specified directory on /workspace. However, nothing is mounted and nothing is available in shared folders in VirtualBox GUI.

I have also tried the Windows directory path in different formats, e.g.:
- c:\workspace
- c:/workspace
- /c/workspace

How to use the switch in Windows?

nathanleclaire:
See if you can get it to work without `--virtualbox-no-share`.  That flag will [skip the parts of `Create` used to set up the share](https://github.com/docker/machine/pull/3798/files#diff-6a5763be0452879200e84388996162f8L451).

michzimny:
It does not work either. Tried with all the c:/workspace path formats as above.

If the flag -virtualbox-no-share skip setting up the share then I would like to propose to fix the docs - https://github.com/docker/docker.github.io/blob/master/machine/drivers/virtualbox.md - because I did earlier exactly as in the docs: I used -virtualbox-no-share to "disable the mount of your home directory" (as I wanted to mount another directory).

nathanleclaire:
> It does not work either. Tried with all the c:/workspace path formats as above

Can you please gist.github.com the output of at least one such attempt?

michzimny:
https://gist.github.com/michzimny/5a42a1fa1a4139b7be6b67bb3a81fdc1

nathanleclaire:
Can you show for:

```
$ docker-machine create -d virtualbox --virtualbox-no-share --virtualbox-share-folder "c:\\workspace:workspace" default
```

Note the quotes.

michzimny:
For the call given above:
https://gist.github.com/michzimny/ebef21474f3c061dd59358866d2ee743

And for the same but without --virtualbox-no-share:
https://gist.github.com/michzimny/2b7c6226b59d69e3cc4e60caa8f055ad

nathanleclaire:
What shell are you using? If you try it with Powershell or cmd.exe do you get different result?

btw `--virtualbox-no-share` should not ever be needed for this case.

michzimny:
[...]

PR #3975, opened 2017-01-30
Fix parsing of '--virtualbox-share-folder' on Windows

zigarn:
Allow use of ':' in the left part of the mapping to use Windows paths.
Assume the right part of the mapping does not contains ':'.
Should fix issue #3908

### DIFF
--- a/drivers/virtualbox/virtualbox_test.go
+++ b/drivers/virtualbox/virtualbox_test.go
@@ -63,6 +63,25 @@ func TestDefaultSSHUsername(t *testing.T) {
 	assert.Equal(t, "docker", username)
 }
 
+var parseShareFolderTestCases = []struct {
+	shareFolder       string
+	expectedShareDir  string
+	expectedShareName string
+}{
+	{"dir:name", "dir", "name"},
+	{"C:\\dir:name", "C:\\dir", "name"},
+	{"C:\\:name", "C:\\", "name"},
+}
+
+func TestParseShareFolder(t *testing.T) {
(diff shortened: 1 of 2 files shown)


nathanleclaire:
It's starting to get quite complicated.

Any ideas how to split it out and write some unit tests to indicate the desired results here?

zigarn:
I extracted the split in a function to be able to unit test it.
I don't know if there is any convention on the place of the function and the tests in the files, so feel free to correct it if it's wrong.

nathanleclaire:
Thanks! I'll have to give this a whirl.

drivers/virtualbox/virtualbox_test.go
```diff
@@ -63,6 +63,27 @@ func TestDefaultSSHUsername(t *testing.T) {
 	assert.Equal(t, "docker", username)
 }
 
+func TestParseShareFolder(t *testing.T) {
+	shareDir, shareName := parseShareFolder("dir:name")
```
nathanleclaire: Can you refactor this into one `TestParseShareFolder` function with a [test table](https://github.com/golang/go/wiki/TableDrivenTests)?

e.g., you'd do something like:

```go
testCases := []struct{
    input, expectedShareDir, expectedShareName string
}{
   {"dir:name", "dir", "name"}
   // ...
}
```

nathanleclaire: It makes the test cases easier to read in one go

zigarn: Changed it to use test table.

nathanleclaire:
LGTM 
image

PR #4003, opened 2017-02-23
Activate bash completion for Windows executable

albers:
On Unix shells running on Windows (cygwin, Docker Toolbox), command completion will expand the name of the Docker Machine binary to `docker-machine.exe`. If you manually installed bash completion, it will not trigger unless you remove the trailing `.exe`.
This PR makes bash completion work for binaries named `docker-machine` _and_ `docker-machine.exe`.


nathanleclaire:
LGTM

Thanks @albers !

PR #4004, opened 2017-02-23
Refine bash completion of machines in start and stop commands

albers:
Bash completion for `docker-machine start` and `docker-machine stop` always completed all machines, regardless of their state.
This PR refines this so that `docker-machine start` only completes stopped machines and `docker-machine stop` only completes running machines.

A new function `_docker_machine_machines` was factored out that lists machines and accepts additional parameters that get passed on to `docker-machine`. It can also be invoked directly, e.g.
```bash
_docker_machine_machines --filter state=Running
_docker_machine_machines --filter state=Running --filter driver=vmwarevsphere --format='{{.Name}}\t{{.State}}'
```
Such functions are widely used in Docker's bash completion.

### DIFF
--- a/contrib/completion/bash/docker-machine.bash
+++ b/contrib/completion/bash/docker-machine.bash
@@ -14,6 +14,16 @@
 #    . ~/.docker-machine-completion.sh
 #
 
+# --- helper functions -------------------------------------------------------
+
+_docker_machine_q() {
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Cool!

LGTM thanks @albers !

PR #4007, opened 2017-02-24
Add bash completion of machines to `docker-machine rm`

albers:
Bash completion of `docker-machine rm` intentionally does not complete machine names.
In practice, this is annoying. It feels as if completion were broken in this place.

There are use cases where machines are created and destroyed quite frequently. As examples see Viktor Farcic's book _The Devops 2.1 Toolkit_, where at the end of each chapter alle machines are removed, and his tutorials, e.g. http://proxy.dockerflow.com/swarm-mode-auto/.

Also added the missing `-f` option.


nathanleclaire:
Sure, seems reasonable.

nathanleclaire:
Thanks @albers  -- LGTM

PR #4011, opened 2017-02-28
Use default driver in bash completion for `docker-machine create`

albers:
Just found in the release notes that Docker Machine now has a default driver.

This PR reflects this change in bash completion: the virtualbox options are completed if no `--driver` is given.

As I already touched that line, I also tightened the command for harvesting driver options.


albers:
This change is not needed.
The virtualbox options are now included in `--help` and therefore detected by bash completion as-is.

PR #4019, opened 2017-03-07
Add delta option to the scp command, using rsync

afbjorklund:
When doing repeated syncs, or copying large directories, it is more
efficient to use "rsync" than to use "scp". Both use "ssh" to do it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
 FROM golang:1.7.4
 
+RUN apt-get update && apt-get install -y --no-install-recommends \
+                openssh-client \
+                rsync \
+        && rm -rf /var/lib/apt/lists/*
+
 RUN go get  github.com/golang/lint/golint \
             github.com/mattn/goveralls \
             golang.org/x/tools/cover
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -355,6 +355,10 @@ var Commands = []cli.Command{
 				Name:  "recursive, r",
 				Usage: "Copy files recursively (required to copy directories)",
 			},
+			cli.BoolFlag{
+				Name:  "delta, d",
+				Usage: "Reduce amount of data sent over network by sending only the differences (uses rsync)",
+			},
 		},
 	},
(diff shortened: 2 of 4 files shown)


afbjorklund:
Apparently rsync is not in the Travis default tool belt. \<sigh\>

nathanleclaire:
Ah cool, have long thought this would be a nice feature to have.

nathanleclaire:
I like it -- how about `--delta` or `--diff` for the flag name instead? Any objections, or other ideas? (`--update` is just so general, and `--rsync` short form of `-r` would already be taken.)

afbjorklund:
@nathanleclaire : OK, I changed the option to `--delta` and removed the `-u` flag...

It wasn't all that useful anyway, since rsync will still use the "quick check" on the files.
I avoided the name --diff, since it reminds me of diff(1). And --rsync was more a detail.

afbjorklund:
Aargh.

nathanleclaire:
Looks epic! Awesome, I've wanted this for quite a while

nathanleclaire:
Could you also turn in a PR to https://github.com/docker/docker.github.io updating the `scp` command documentation in `vnext-machine` branch?

LGTM

afbjorklund:
will do!

@nathanleclaire : as always with rsync, it's a bit peculiar when it comes to directories... (`-u`)
The "safe" option is to use trailing slash `'/'` on both source and destination to avoid surprises

i.e. /tmp/foo -> /tmp/foo, and you might end up with /tmp/foo/foo or something.

nathanleclaire:
> @nathanleclaire : as always with rsync, it's a bit peculiar when it comes to directories... (-r)
The "safe" option is to use trailing slash '/' on both source and destination to avoid surprises

Interesting, perhaps something to note in the documentation!

afbjorklund:
Yeah, that is what I thought. It should come as no surprise to rsync users.
But maybe it can trip someone up when switching from scp or something.

PR #4024, opened 2017-03-15
Update bash completion

albers:
Several improvements to bash completion
- add several missing options (mostly short options)
- improve completion for `env --shell`
- sort options alphabetically with the short form following the corresponding long form
- pull out nested treatment of `$cur` from `$prev` (readibility, more robust against future changes)
- do not set `$COMPREPLY` to its default value if no completion exists, use return for shortcutting instead (readibility, more robust against future changes)

This PR aligns this completion more closely to that of docker and docker-compose.

Signed-off-by: Harald Albers <github@albersweb.de>


nathanleclaire:
LGTM

Thanks @albers ! 🎉

PR #4047, opened 2017-04-04
Make the SUSE/openSUSE provisioner more resilient

flavio:
Do not let the provisioner fail when ran against an already initialized machine.


nathanleclaire:
LGTM

Thanks @flavio !

PR #4050, opened 2017-04-06
drivers/digitalocean: fix empty tags being added to droplets

cezarsa:
Currently an empty tag was always being added to the droplet. This causes digitalocean to reject the droplet creation request due to invalid tags.

This PR fixes this and adds a new test case to ensure the Tags attribute is nil if no tags were set.


nathanleclaire:
LGTM

Thanks @cezarsa !

Issue #4040, opened 2017-03-29
docker-machine ignores http_proxy settings


michalmedvecky:
I'm trying to provision a docker host using docker-machine using OpenStack module.

    docker-machine create --engine-env HTTP_PROXY=http://1.1.1.1:3128/ --engine-env HTTPS_PROXY=https://1.1.1.1:3128/ --driver openstack --openstack-auth-url=http://1.2.3.4:5000/v2.0/ --openstack-tenant-name=sec_tools --openstack-username=myuser --openstack-password=mypwd --openstack-image-name=66ae409c-f2ea-4a8c-a25c-3859d79f9f33 --openstack-net-name=mynet --openstack-keypair-name=mykey --openstack-private-key-file=/tmp/key.pem --openstack-flavor-name=m2.small bleh

Result:

    Error creating machine: Error in driver during machine creation: Post http://1.2.3.4:5000/v2.0/tokens: dial tcp 100.127.20.11:5000: getsockopt: connection timed out

My problem is that `OS_AUTH_URL` (1.2.3.4:5000) is reachable ony via http proxy and not directly.

I tried to set `http_proxy`, `https_proxy` (and capitalized as well), nothing works, always the same result.

PR #4055, opened 2017-04-12
Enable proxy for openstack

michalmedvecky:
This fix allows docker-machine to reach OS_AUTH_URL using proxy.


michalmedvecky:
Closes #4040

PR #4084, opened 2017-05-10
Add bash completion for `docker-machine provision`

albers:
I'm not sure whether this is correct. I assume that `docker-machine provision` should work on running and stopped machines. But on a quick test, `docker-machine provision` only worked with a running machine.

Is this how it is supposed to work? If so, I will have to update this PR to only complete running machines.
Currently all machines are completed.

### DIFF
--- a/contrib/completion/bash/docker-machine.bash
+++ b/contrib/completion/bash/docker-machine.bash
@@ -165,7 +165,7 @@ _docker_machine_regenerate_certs() {
     if [[ "${cur}" == -* ]]; then
         COMPREPLY=($(compgen -W "--force -f --help" -- "${cur}"))
     else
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Nice, thanks @albers. Only running machines, since the binary will log in via SSH and execute some commands to do the provisioning.

albers:
Is this also valid for `regenerate-certs`?

nathanleclaire:
Yes

albers:
@nathanleclaire The PR is updated so that only running machines are completed for `docker-machine provision`. Bash completion for `docker-machine regenerate-certs` was updated accordingly, too.
PTAL

nathanleclaire:
Sweet! Thanks @albers

nathanleclaire:
LGTM

PR #4087, opened 2017-05-11
Add bash completion for `docker-machine ls` filters

albers:
This adds bash completion of
- filters for the `--filter` option
- drivers for the `driver` filter
- states for the `state` filter

to bash completion for `docker-machine ls`

To achieve this, I had to factor out the list of drivers from `docker-machine create`.
I also needed helper functions which I took over from Docker's bash completion.
These functions also take care of the fact that MacOS ships with an ancient version of bash and an outdated version of bash-completion.


nathanleclaire:
LGTM

Thanks @albers !

PR #4116, opened 2017-06-01
Ignore "unknown instance" error when rm instance

### LINKED ISSUE #4112: ignore rm failures when machine has been deleted.
eg in spot instances, see also https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1947

parryjacob:
Fixes #4112 

This is especially useful when using Spot instances on EC2, as they may be removed at any time.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -952,6 +952,7 @@ func (d *Driver) terminate() error {
 	})
 
 	if strings.HasPrefix(err.Error(), "unknown instance") {
+		log.Warn("Remote instance does not exist, proceeding with removing local reference")
 		return nil
 	}
 
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -950,6 +950,11 @@ func (d *Driver) terminate() error {
 	_, err := d.getClient().TerminateInstances(&ec2.TerminateInstancesInput{
 		InstanceIds: []*string{&d.InstanceId},
 	})
+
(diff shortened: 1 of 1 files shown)


nathanleclaire:
Good idea, just one suggestion:

How about instead of returning `nil` you return something like `errors.New("Can't locate remote instance -- The instance may have been removed by another method")`? I'd still consider a `rm` that doesn't actually do anything an error (non-zero exit, just like how `rm i_dont_exist` would return 1).

parryjacob:
@nathanleclaire The general motivation behind this is the idea that when this function returns an error, it cascades up the stack and causes `docker-machine` to not remove the instance from its local state when it doesn't exist remotely either. Generally, the idea is that `rm` should remove the machine locally and remotely, and if it exists locally but not remotely it isn't really an error and it should still remove it locally instead of throwing an error.

This circumstance is most common when using Amazon EC2 Spot Instances, when the remote machine being gone isn't exceptional and is expected.

This also supposedly brings the `amazonec2` driver inline with the behaviour of other drivers like the `digitalocean` driver which apparently does not have this issue. (see related issue in the GitLab CI runner project: https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/1947#note_19955573)

nathanleclaire:
OK -- seems reasonable -- how about just a `log.Warn` (`Remote instance does not exist, proceeding with removing local reference`) before the `return nil` then.

parryjacob:
@nathanleclaire I've added that, thanks for the feedback!

nathanleclaire:
LGTM

PR #4118, opened 2017-06-02
Switch version in version.go to 'dev'

nathanleclaire:


nathanleclaire:
this should be sufficient since the release script will always replace this with the proper version

LGTM

Issue #4119, opened 2017-06-05
Shell prompt is not automatically changing after install the bash completion scripts.


lnshi:
As title described, after I set up the bash completion scripts, then did
```
eval "$(docker-machine env dev)"
```
the shell prompt is not automatically changing, still remains the same!

Only when I sourced my `~/.bash_profile` again, which the `PS1` is defined there, then the prompt got changed, it is boring.

Is that possible to make it automatically change after the `eval` command?

SakuraSound:
I know this thread is super old, but for anyone else who wants to have their bash completion happen all the time, make sure you prepend a backslash before the bash_completion function call. The backslash indicates to PS1 to evaluate after each command. i.e.

```shell
PS1="\$(__docker_machine_ps1)"
```

If you omit the backslash, it will compute the first time the terminal is opened, and then all subsequent calls will just use the result of the first call.

If this solves this issue, you can probably close it...

PR #4125, opened 2017-06-07
Fix issue #4119: automatically re-render 'PS1' in place upon you chan…

lnshi:
Scenario details:
1. Use Unix builtin function `bind` to bind one function `__update_shell_prompt` to `Enter key: \C-M`;

2. Upon any enter key hitting, the function `__update_shell_prompt` will be invoked, then based on whether you changed the 'docker-machine env' or not the function would try to re-render the `PS1`;

3. Since any executed command will be echoed back in the terminal, it is a bit annoying, so here I just use the Unix builtin function `tput` to omit the echoed `__update_shell_prompt` line, to make it looks like nothing ever happened;

4. Also since by default any executed command will be recored in the history(so you can use the `up` and `down` key to switch between them conveniently), so the above `2` will pollute your history, but as you can see the command ` __update_shell_prompt` which is bind to the `Enter key: \C-M` is prefixed with one space, so put `export HISTCONTROL=ignorespace` into your `~/.bash_profile` or `~/.bashrc`, then you can prevent this potential commands history pollution;

Then it works perfectly, enjoy 🙂

### DIFF
--- a/contrib/completion/bash/docker-machine-prompt.bash
+++ b/contrib/completion/bash/docker-machine-prompt.bash
@@ -66,4 +66,3 @@ __update_shell_prompt () {
 
 bind '"\C-M":"\n __update_shell_prompt\n"'
 
-
--- a/contrib/completion/bash/docker-machine-prompt.bash
+++ b/contrib/completion/bash/docker-machine-prompt.bash
@@ -64,5 +64,5 @@ __update_shell_prompt () {
     tput cup $((current_row-2)) 0 && tput el
 }
 
-bind '"\C-M":"\n __update_shell_prompt\n"'
+bind '"\e[C":"\n __update_shell_prompt\n"'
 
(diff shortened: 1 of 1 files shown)


lnshi:
Don't merge first, this will break the multiple lines commands copy paste into terminal behaviour, I am seeking the fix.

lnshi:
Give up to bind the function to `Enter Key`, coz if like that then will break the copy paste multiple lines commands to terminal behaviour, like if you try to copy and paste the follow commands:
```
docker-machine \
  create \
  --driver generic \
  --generic-ip-address 10.0.3.159 \
  --generic-ssh-user leonard \
  --generic-ssh-key ~/.ssh/id_rsa \
  --engine-opt dns=8.8.8.8 \
  --engine-insecure-registry 10.0.3.160:7777 \
  xxx.xxxx.com
```
then all the backslash `\` will trigger the function `__update_shell_prompt `, everything will be in messy...

**So I decided to bind it to the `Right Arrow Key`, so now after you run the `eval "$(docker-machine env dev)"`, then press the `Right Arrow Key` in your keyboard, then the shell prompt will be re-rendered.**

Guess this will never be merged, so if you need just copy from here...

PR #4128, opened 2017-06-09
Fix versioncmp for CE edition

StefanScherer:
I just tried to generate some TLS certs for a generic Ubuntu VM running Docker 17.06.0-ce-rc2, but after inserting TLS certs the Docker Engine couldn't be started any more.

The systemd service file is created with "docker daemon ..." instead of the "/usr/bin/dockerd ..." command line.

The version comparison was wrong if already running a 17.xx CE version.
I also added some tests.


StefanScherer:
This PR fixes the version check for https://github.com/docker/machine/blob/c604d46dc5894c42eb62abf1d7a1e6e7276137c2/libmachine/provision/systemd.go#L50 and https://github.com/docker/machine/blob/c604d46dc5894c42eb62abf1d7a1e6e7276137c2/libmachine/provision/coreos.go#L74

FrenchBen:
Also fixes https://github.com/docker/machine/issues/4156

FrenchBen:
Thanks for the patch @StefanScherer sorry it took so long to review and merge.

StefanScherer:
No problem. Glad it will make it in next version of docker-machine

brandontamm:
Is this the same issue I am having in #4189?

PR #4141, opened 2017-06-16
Don't use extglob in bash/docker-machine.bash

### LINKED ISSUE #4126: OSX - bash_completion.d/docker-machine.bash: line 86: syntax error in conditional expression: unexpected token `('
docker-machine 0.12.0, installed with homebrew on osx 10.12.5 (bash 3.2.57).

the bash_completion script is triggering a syntax error:

```
-bash: /usr/local/etc/bash_completion.d/docker-machine.bash: line 86: syntax error in conditional expression: unexpected token `('
-bash: /usr/local/etc/bash_completion.d/docker-machine.bash: line 86: syntax error near `@($'
-bash: /usr/local/etc/bash_completion.d/docker-machine.bash: line 86: `    [[ ${words[$glob_pos]} == @($glob) ]] && echo "$key"'
```

looks like it assumes `shopt extglob` is always enabled.
as `extglob` defaults to `off` this might be an incorrect assumption.

Empact:
extglob is not needed here - the only call is:
_docker_machine_map_key_of_current_option '--filter'

So $glob is always '--filter', so there isn't every a pattern-list to match.

According to my testing completion for docker-machine
ls --filter completion works for both driver and state

Fixes #4126
Introduced in fd9a0a683c69d94813eabecf8b5062ab38237e6c

### DIFF
--- a/contrib/completion/bash/docker-machine.bash
+++ b/contrib/completion/bash/docker-machine.bash
@@ -62,7 +62,7 @@ _docker_machine_value_of_option() {
 }
 
 # Returns `key` if we are currently completing the value of a map option
-# (`key=value`) which matches the extglob passed in as an argument.
+# (`key=value`) which matches the glob passed in as an argument.
 # This function is needed for key-specific argument completions.
 _docker_machine_map_key_of_current_option() {
     local glob="$1"
@@ -83,7 +83,7 @@ _docker_machine_map_key_of_current_option() {
 
     [ "${words[$glob_pos]}" = "=" ] && ((glob_pos--))  # --option=key=value syntax
 
-    [[ ${words[$glob_pos]} == @($glob) ]] && echo "$key"
+    [[ ${words[$glob_pos]} == $glob ]] && echo "$key"
 }
 
 # --- completion functions ---------------------------------------------------


Empact:
cc @albers

albers:
Thanks for the ping. I'll take a look at it.

albers:
Ah, Mac OS again with its ancient bash.
This function was taken from Docker's bash completion which takes care of temporarily setting the extglob option while loading the script.  I forgot to take this into account here.

@Empact You're right: no extglob is needed here, so it's probably best to keep the code simple. Thanks very much for fixing this.

To verify without a Mac:
Original code fails to load with bash 3.2.57:
```bash
$ docker run --rm --volume $(pwd)/contrib/completion/bash/docker-machine.bash:/docker-completion albers/bash-completion-mac bash -c ". /docker-completion" && echo OK
/docker-completion: line 86: syntax error in conditional expression: unexpected token `('
/docker-completion: line 86: syntax error near `@($'
/docker-completion: line 86: `    [[ ${words[$glob_pos]} == @($glob) ]] && echo "$key"'
```

With the fix it loads properly:
```bash
$ docker run --rm --volume $(pwd)/contrib/completion/bash/docker-machine.bash:/docker-completion albers/bash-completion-mac bash -c ". /docker-completion" && echo OK
OK
```
LGTM 

@shin- This is a bugfix and should go into 1.12.0.

jakirkham:
Would it be reasonable to test against `bash` 3.2.57 to catch errors like this during development?

albers:
@jakirkham I'm working on this. Similar to https://github.com/docker/cli/pull/266, a test like `docker run --rm --volume $(pwd)/contrib/completion/bash/docker-machine.bash:/docker-completion albers/bash-completion-mac bash -c ". /docker-completion"` should be added.
I'm not familiar with the build tooling, though, so this will take some time.

shin-:
Thank you both!

Issue #4146, opened 2017-06-22
docker-machine rm <host> for AWS


axelssonHakan:
docker-machine version 0.12.0, build 45c69ad

Then trying removing AWS hosts docker-machine fails
**Error removing host "<host>": unexpected EOF**

```
(<host>) DBG | ---[ RESPONSE ]--------------------------------------
(<host>) DBG | HTTP/1.1 200 OK
(<host>) DBG | Transfer-Encoding: chunked
(<host>) DBG | Content-Type: text/xml;charset=UTF-8
(<host>) DBG | Date: Thu, 22 Jun 2017 11:52:45 GMT
(<host>) DBG | Server: AmazonEC2
(<host>) DBG | Vary: Accept-Encoding
(<host>) DBG |
(<host>) DBG | 251
(<host>) DBG | <?xml version="1.0" encoding="UTF-8"?>
(<host>) DBG | <TerminateInstancesResponse xmlns="http://ec2.amazonaws.com/doc/2016-04-01/">
(<host>) DBG |     <requestId>...</requestId>
(<host>) DBG |     <instancesSet>
(<host>) DBG |         <item>
(<host>) DBG |             <instanceId>...</instanceId>
(<host>) DBG |             <currentState>
(<host>) DBG |                 <code>48</code>
(<host>) DBG |                 <name>terminated</name>
(<host>) DBG |             </currentState>
(<host>) DBG |             <previousState>
(<host>) DBG |                 <code>48</code>
(<host>) DBG |                 <name>terminated</name>
(<host>) DBG |             </previousState>
(<host>) DBG |         </item>
(<host>) DBG |     </instancesSet>
(<host>) DBG | </TerminateInstancesResponse>
(<host>) DBG | 0
(<host>) DBG |
(<host>) DBG |
(<host>) DBG | -----------------------------------------------------
(<host>) DBG | panic: runtime error: invalid memory address or nil pointer dereference
(<host>) DBG | [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x15535d8]
```

axelssonHakan:
Same issue in Docker Machine Version:  0.12.0-rc2, build 64fa459

axelssonHakan:
First time trigger for host - same issue different xml response
```
(<host>) DBG | <?xml version="1.0" encoding="UTF-8"?>
(<host>) DBG | <TerminateInstancesResponse xmlns="http://ec2.amazonaws.com/doc/2016-04-01/">
[...]

PR #4147, opened 2017-06-22
Reorder if's to ensure that err is not null before using it.

LordMike:
A bug seems to have been introduced that used `err` even though it might be null at the time. See issue #4146.

Now, I'm no go coder, have not compiled this or anything - but it seems like it should work from the code alone. Feel free to reject and re-do as necessary.


shin-:
Thank you! Merged in #4187

Issue #4156, opened 2017-06-29
Machine create fails with latest Docker


dminkovsky:
Hello

docker-machine version 0.12.0, build 45c69ad

`docker-machine create` fails now:

```
docker-machine -D create \
    --driver google \
    --google-project project \
    --google-zone us-east1-d \
    --google-machine-type n1-standard-1 \
    --google-disk-size 20 \
    --google-preemptible \
    build-vm2
```
The machine is created and Docker is installed, but it won't start. The problem appears to be a related to a new version of Docker getting installed by a new version of the install script over at https://get.docker.com. My installs went from 17.05.0-ce to 17.06.0-ce, and with that change, Docker installs but does not start.

    Jun 29 00:50:08 build-vm2 docker[5705]: `docker daemon` is not supported on Linux. Please run `dockerd` directly

or 

    Jun 29 00:56:12 build-vm2 dockerd[6407]: Error starting daemon: error initializing graphdriver: driver not supported

Unless I change:

    /usr/bin/docker daemon -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --storage-driver aufs --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=google

to 

    /usr/bin/dockerd -H tcp://0.0.0.0:2376 -H unix:///var/run/docker.sock --tlsverify --tlscacert /etc/docker/ca.pem --tlscert /etc/docker/server.pem --tlskey /etc/docker/server-key.pem --label provider=google

in `/etc/systemd/system/docker.service.d/10-machine.conf`.

thanhhx-zinza:
Same problem here
```
docker-machine create 
    --driver=digitalocean
    --digitalocean-access-token=XXX 
    --digitalocean-size=2gb
    machinename
```
Yesterday the same command worked fine with docker version 17.05.0-ce
Today my new machine's docker won't start (17.06.0-ce)
I've tried multiple time.

gnomus:
I can confirm this too:

```
dm create -d digitalocean \
--digitalocean-access-token XXX \
--digitalocean-size 4gb machine
```

therealppa:
Is there any idea for a workaround?

acouette:
I have the same issue.

docker version : Docker version 17.06.0-ce
docker-machine version : 0.12.0, build 45c69ad

[...]

PR #4157, opened 2017-06-29
Fix incorrect results on comparing `-ce` versions and previous versions

gussan:
related #4156

`dockerd` command should be used on docker version 1.12.0 and later, but actually `docker daemon` command has been used with docker "-ce" versions.
https://github.com/docker/machine/blob/master/libmachine/provision/systemd.go#L49-L52


gnomus:
This is a duplicate of
https://github.com/docker/machine/pull/4128

gussan:
Oops, I'm closing this. thx!

dminkovsky:
@gussan thanks for pointing out https://github.com/docker/machine/blob/master/libmachine/provision/systemd.go#L49-L52 in the code base. I couldn't find that last night in a brief search.

PR #4176, opened 2017-07-05
amazonec2: prevent nil pointer dereference on machine terminate

### LINKED ISSUE #4175: amazonec2: ssh key pairs are not being removed
Upon machine termination, when everything suceeds, the driver is panicking because the [error here is nil](https://github.com/docker/machine/blob/master/drivers/amazonec2/amazonec2.go#L954). This is causing the termination to fail and as a consequence the ssh key pairs are not being removed.

andrestc:
Fixes #4175.


shin-:
Thanks! Merged identical fix in #4187

PR #4204, opened 2017-07-25
driver: update exoscale driver

marcaurele:
Update exoscale driver to address an issue with the template selection


shin-:
Thanks!

PR #4232, opened 2017-08-23
libmachine/provision: retries pkg installation if fails to get lock

andrestc:
This commit is a follow-up on #3891 and makes sure we retry pkg installations if we fail to acquire a lock.

This is pretty useful if the user provides a userdata that also install pkgs and prevents errors such as:
Error creating machine: Error running provisioning: ssh command error:
command : DEBIAN_FRONTEND=noninteractive sudo -E apt-get install -y  curl
err     : exit status 100
output  : E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?

PTAL @shin- @SvenDowideit 

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -64,11 +64,7 @@ func (provisioner *DebianProvisioner) Package(name string, action pkgaction.Pack
 
 	log.Debugf("package: action=%s name=%s", action.String(), name)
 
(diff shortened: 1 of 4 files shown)


fabiorauber:
#4604

PR #4237, opened 2017-08-29
Modified error message output on host already or does not exist

londoncalling:
### What's changed

- Modified error message outputs for `docker-machine` actions that result in `host does not exist` and `host already exists` to be a little more user-friendly 

### Related

- Partially fixes an issue logged on the Docs repo as [#4379](https://github.com/docker/docker.github.io/issues/4379)

### Reviewers

@shin- please take a look, and if you don't like this you can just close the PR. We got some feedback on confusing error messages in [#4379](https://github.com/docker/docker.github.io/issues/4379). This slight change may help somewhat, and I can also see what else can do in the docs.

@johndmulhausen @mstanleyjones 

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>

### DIFF
--- a/libmachine/mcnerror/errors.go
+++ b/libmachine/mcnerror/errors.go
@@ -17,7 +17,7 @@ type ErrHostDoesNotExist struct {
 }
 
 func (e ErrHostDoesNotExist) Error() string {
-	return fmt.Sprintf("Docker machine \"%q\" does not exist. Use \"docker-machine ls\" to list machines. Use \"docker-machine create [OPTIONS]\" to add a new one.", e.Name)
+	return fmt.Sprintf("Docker machine \"%q\" does not exist. Use \"docker-machine ls\" to list machines. Use \"docker-machine create\" to add a new one.", e.Name)
 }
 
 type ErrHostAlreadyExists struct {
--- a/libmachine/mcnerror/errors.go
+++ b/libmachine/mcnerror/errors.go
@@ -17,7 +17,7 @@ type ErrHostDoesNotExist struct {
 }
 
 func (e ErrHostDoesNotExist) Error() string {
-	return fmt.Sprintf("Docker machine \"%q\" does not exist", e.Name)
+	return fmt.Sprintf("Docker machine \"%q\" does not exist. Use docker-machine create to add one.", e.Name)
 }
 
 type ErrHostAlreadyExists struct {
--- a/libmachine/mcnerror/errors.go
+++ b/libmachine/mcnerror/errors.go
@@ -17,7 +17,7 @@ type ErrHostDoesNotExist struct {
 }
 
 func (e ErrHostDoesNotExist) Error() string {
-	return fmt.Sprintf("Docker machine \"%q\" does not exist. Use docker-machine create to add one.", e.Name)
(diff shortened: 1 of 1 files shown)


shin-:
I'm inclined to say "host" is the correct terminology here, and it's a pretty common software eng term, so I'm not too sure what about it is confusing... :/

londoncalling:
Well there is confusion about what a Docker machine is.  I understand that
"host" is common but in this case the fact that it's generic is causing the
confusion... and since confusion over the concepts of a "Docker Machine"
have surfaced before, I thought something like this isn't a bad idea . I
was even thinking of adding a second sentence about using "docker-machine
create" to add a machine. Thus hint might have helped this user and
clarified the nature of the problem right at the CLI, giving breadcrumbs to
the "create" command:

londoncalling:
@shin- So, there was some confusion about what a Docker machine is.  

I understand that "host" is common but in this case the fact that it is a generic term, I think, caused the confusion. It's not a big deal, but there has been some confusion about the concept of a "Docker machine" have before. This was part of the impetus for me re-writing the [Docker Machine Overview](https://docs.docker.com/machine/overview/) page when I got here.

Given that, making the error message outputs a little more specific in this case might not be a bad idea, even if it's a little less standard (e.g., specifying "Docker machine" instead "host") . 

I was even thinking of adding a second sentence about using "docker-machine create" to add a machine. This hint might have helped this user and clarified the nature of the problem right at the CLI, giving breadcrumbs to the `docker-machine create` command.

That's my thinking on it, but if you disagree, of course will defer to you. You're the man! :-)

kolotyluk:
It might help if every docker web page has a link near the top, on the side, or some place easily spotted, to 'Glossary of Docker Terms' - in particular it should clarify the difference between generic terms like 'host' and specific terms like 'docker host'.

https://martinfowler.com/bliki/TwoHardThings.html

mstanleyjones:
There is a Glossary link at the top of every page. :)

kolotyluk:
Doh! Found it. I was looking below the header instead of at it.
Cheers, Eric

shin-:
@londoncalling Can you rebase this PR against the new master branch? This should clear the test failures. Changes LGTM.

libmachine/mcnerror/errors.go
```diff
@@ -17,15 +17,15 @@ type ErrHostDoesNotExist struct {
 }
 
 func (e ErrHostDoesNotExist) Error() string {
-	return fmt.Sprintf("Host does not exist: %q", e.Name)
+	return fmt.Sprintf("Docker machine \"%q\" does not exist. Use docker-machine ls to list machines. Use docker-machine create [OPTIONS] to add a new one.", e.Name)
```
shin-: I'd remove the `[OPTIONS]` part &ndash; it should be the general assumption that most commands have option flags and arguments, some mandatory. If the user attempts to run `docker-machine create`, a help message will be output listing the command's usage and options.

londoncalling: cool, makes sense

londoncalling:
@shin- let me take a look and see if I can do it

londoncalling:
@shin- okay I think you should take over updating the unit tests in `its/cli/create_rm_test.go`. There is some aspect of it I don't think I understand well enough.

PR #4241, opened 2017-09-01
Fixes incorrect parsing of ISO version number.

ggrussenmeyer:
It appears the full version contained within the ISO file may start with spaces.
As such, it must be trimmed before version number starting index is determined, else the version number won't be correctly extracted (and hence the ISO file will be downloaded each time as a version mismatch will always be detected).

### DIFF
--- a/libmachine/mcnutils/b2d.go
+++ b/libmachine/mcnutils/b2d.go
@@ -307,14 +307,16 @@ func (b *b2dISO) version() (string, error) {
 
 	fullVersion := string(isoMetadata)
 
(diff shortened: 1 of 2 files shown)


libmachine/mcnutils/b2d.go
```diff
@@ -307,14 +307,16 @@ func (b *b2dISO) version() (string, error) {
 
 	fullVersion := string(isoMetadata)
```
shin-: Do you reckon this could just be changed to `strings.TrimSpace(string(isoMetadata))`?

PR #4249, opened 2017-09-08
Azure driver: Fix issue #3976

### LINKED ISSUE #3976: [azure driver] docker-machine rm to remove the storage account as well
Hello,
Since you cannot choose a storage account when creating a docker-machine in Azure, the creation of each machine also creates its own, unique storage account with one blob inside (the virtual disk of the machine in question). This is fine, until you want to remove/recreate your machines - removing a machine deletes the virtual disk, but not the storage account itself. As a result I get an empty, unused and non-reusable storage account each time I `docker-machine rm`.
Issue [#3579](https://github.com/docker/machine/issues/3579) explains why we cannot specify a storage account upon machine creation and I can agree with the reasoning. However, I see no reason for the storage account to remain at all and if possible, `docker-machine rm` should remove the storage account as well.

kleinsasserm:
This PR will fix #3976

## Why
The current driver behavior does not delete the empty storage accounts. This will mess up your Microsoft Azure cloud resource group, like I wrote it as comment on the referenced issue. The current behavior will produce the following result if you, like we do, use for example Gitlab autoscaling runners.
screenshot

## What I've done
I added a functionality which checks, if there are storage blobs left over in the associated storage account of the virtual machine, after the virtual machine is deleted. If there are no container items left in the storage account, this PR will delete the corresponding storage account.

## How it is done
I modified the removeOSDiskBlob method and used the already existing client methods to retrieve a list of the containers associated with the storage account. If the retrieved list is empty, after the deletion of the virtual machine, the PR deletes the storage account too with the appropriate methods.

## Result
No more dangling storage accounts.

### DIFF
--- a/drivers/azure/azureutil/azureutil.go
+++ b/drivers/azure/azureutil/azureutil.go
@@ -460,6 +460,22 @@ func (a AzureClient) removeOSDiskBlob(resourceGroup, vmName, vhdURL string) erro
 	if err != nil {
 		log.Debugf("Container remove happened: %v", ok)
 	}
+
+	cts, err := bs.GetBlobService().ListContainers(blobstorage.ListContainersParameters{})
+	if err != nil {
+		return err
+	}
+
+	if len(cts.Containers) == 0 {
+		log.Debugf("No storage containers left. Deleting virtual machine storage account.")
+		resp, err := a.storageAccountsClient().Delete(resourceGroup, storageAccount)
+		if err != nil {
+			return err
+		}
+
+		log.Debugf("Storage account deletion happened: %v", resp.Response.Status)
+	}
+
 	return err
 }
 


kleinsasserm:
@shin @StefanScherer @nathanleclaire any comments on this from the committers? Is something missing or unclear? Can I do something to get this commited? Thank you!

nathanleclaire:
Thanks @kleinsasserm -- I am not maintainer anymore, but you want to actually ping @shin-

kleinsasserm:
Thanks to you @nathanleclaire for your response too!

drivers/azure/azureutil/azureutil.go
```diff
@@ -460,6 +460,21 @@ func (a AzureClient) removeOSDiskBlob(resourceGroup, vmName, vhdURL string) erro
 	if err != nil {
 		log.Debugf("Container remove happened: %v", ok)
 	}
+
+	cts, err := bs.GetBlobService().ListContainers(blobstorage.ListContainersParameters{})
+	if err != nil {
+		log.Debugf("Error listing containers.")
```
shin-: Probably need to `return err` here as well, as the value of `cts` will likely be undefined.

drivers/azure/azureutil/azureutil.go
```diff
@@ -460,6 +460,21 @@ func (a AzureClient) removeOSDiskBlob(resourceGroup, vmName, vhdURL string) erro
...
+	}
+
+	if len(cts.Containers) == 0 {
+		resp, err := a.storageAccountsClient().Delete(resourceGroup, storageAccount)
+		if err != nil {
+			return err
+		}
+
+		log.Debugf("No storage containers left. Virtual machine storage account deletion happened: %v", resp.Response.Status)
```
shin-: Move this before the delete call, and change the message to something like `No storage containers left. Deleting virtual machine storage account`

We can then add another debug message that reports the response status code.

kleinsasserm: Yes, that is correct, I will revise it! Thank you for your help!

kleinsasserm: @-shin What is your opinion?

kleinsasserm: @shin- what is your opinion?

drivers/azure/azureutil/azureutil.go
shin-: @kleinsasserm You're correct in that it will not cause a crash, but "falling through" here will cause the next condition `len(cts.Containers) == 0` to always be true, and could cause unwanted data loss if containers are in fact present. Also, returning `err` here will allow us to report the exact error that occurred downstream.

kleinsasserm: Omg. Now I see, this is now embarrassing. 😊 I will try to do it the correct way in the next days. Thx @shin- for showing me my fallacy!

kleinsasserm:
@shin- Hi Shin! I've updated the PR with your suggestions. Hopefully I've done the update of the PR with the correct git rebase and push commands.

shin-:
LGTM, thank you! :)

kleinsasserm:
Thank you too! :)

PR #4252, opened 2017-09-09
google driver: remove 404 machines with warning

### LINKED ISSUE #4251: Remove GCE machine if api returns 404
Removing a machine that already has been deleted should not result in an error. Instead it should just be deleted locally like implemented for ec2: #4112 

```
root@ci-broker:~# docker-machine ls
NAME         ACTIVE   DRIVER   STATE   URL   SWARM   DOCKER    ERRORS
runner-123   -        google                         Unknown   The resource 'projects/123/zones/europe-west1-d/instances/runner-123' was not found

root@ci-broker:~# docker-machine rm runner-123
About to remove runner-123
WARNING: This action will delete both local reference and remote instance.
Are you sure? (y/n): y
(runner-123) Deleting instance.
Error removing host "runner-123": googleapi: Error 404: The resource projects/123/zones/europe-west1-d/instances/runner-123' was not found, notFound

root@ci-broker:~# docker-machine ls
NAME         ACTIVE   DRIVER   STATE   URL   SWARM   DOCKER    ERRORS
runner-123   -        google                         Unknown   The resource 'projects/123/zones/europe-west1-d/instances/runner-123' was not found
```

fiws:
fixes #4251

Already deleted machines should be removed with a warning not throw an error. Use case: preemptive machines.

(Hopefully) same behaviour as the ec2 driver: https://github.com/docker/machine/pull/4116

### DIFF
--- a/drivers/google/google.go
+++ b/drivers/google/google.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"strings"
 
 	"github.com/docker/machine/libmachine/drivers"
@@ -11,6 +12,8 @@ import (
 	"github.com/docker/machine/libmachine/mcnflag"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/docker/machine/libmachine/state"
+
+	"google.golang.org/api/googleapi"
 )
 
 // Driver is a struct compatible with the docker.hosts.drivers.Driver interface.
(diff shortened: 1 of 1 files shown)


drivers/google/google.go
```diff
@@ -396,7 +396,11 @@ func (d *Driver) Remove() error {
 	}
 
 	if err := c.deleteInstance(); err != nil {
-		return err
+		if strings.Contains(err.Error(), "notFound") {
```
shin-: We should be able to cast this to the `googleapi.Error` type and check the `Code` field for a `http.StatusNotFound` code instead of relying on string matching.

fiws: Ok, i think i need some help with that. These were my first few lines in go.

I tried to change it, but I'm not sure at all, that it will work, is correct, or is clean code.

shin-: You aced it as far as I can tell :)

Thank you!

shin-:
(Merged manually)

PR #4254, opened 2017-09-12
support adding multiple block devices in softlayer driver

overmike:
This PR helps to reduce subsequence steps to add block devices to the docker-machine through softlayer portal/api. And this is backward compatible because we can specify one (old style) or more block devices (new style)


shin-:
Thank you for your contribution! Unfortunately, we're not looking to add any new features to Machine anymore at this point. Thank you for your understanding.

overmike:
Do you mean docker-machine stop development?

PR #4269, opened 2017-10-03
Modified error message output on host already or does not exist

shin-:
Closes #4237 - squashed and fixed unit test.

cc @londoncalling


londoncalling:
@shin- thanks!

PR #4274, opened 2017-10-09
Decapitalize error string;

243083df:
https://github.com/golang/go/wiki/Errors


shin-:
Thanks, but that's likely to create more problems than it solves.

PR #4280, opened 2017-10-14
Make vmwarefusion support VMware workstation

anfernee:
It makes the driver works for Mac, Linux and Windows.

I didn't change either the driver name or the parameters names inside the driver, because I am not sure how important it is to be compatible with old versions. I am free to change it as people started to review it.

I have tested the change in Linux (Ubuntu) and Windows 10 with VMware Workstation 12. 
I haven't tested on Mac yet. I will when I get a test machine and latest fusion installed.


anfernee:
I have tested fusion 10 on Mac as well. It also works.

anfernee:
anyone review this PR? It's required to solve https://github.com/kubernetes/minikube/issues/806

shin-:
Thank you for your contribution! Unfortunately, we're not looking to add any new features to Machine anymore at this point. Thank you for your understanding.

PR #4292, opened 2017-10-24
drivers/digitalocean: Increase ip assignment performance

243083df:


dgageot:
@243083df I'm not a big fan of the `break` with a label. Do you think the code can be made easier to read. Maybe even extract the code inside the loop?

PR #4296, opened 2017-10-27
 Replace sprintf to native concat and itoa.

243083df:


shin-:
Thanks, but I'd rather stick with the established coding style.

PR #4297, opened 2017-10-30
Check errors when creating virtual machine

tt:
Creating a virtual machine with the VMware Fusion driver lacks a lot of error checking which means it might continue trying to initialize a new machine even when an earlier step failed. This makes debugging a lot more difficult because the command actually failing might not be the one that caused it.


dgageot:
LGTM

shin-:
Are there any recoverable errors in this process? I'm assuming the provisioning was originally written that way for a reason.

dgageot:
@shin- If I remember correctly, the `vmrun` binary used to not always return a proper exit code in case of error. Wether this behaviour has changed or not, I think it's better the handle the errors on the go side.

PR #4300, opened 2017-11-01
Add bash completion for `docker-machine mount`

albers:
This adds bash completion for #4018.

Notes:
Bash completion supports completion for files and directory with its builtin `_filedir` function. Sadly, this only works for local files and directories. I don't know a simple way to do remote file system completions. So the first argument is just completed to the machine name and a trailing `:`. The user will have to add the remote path without support. The second argument is completed to local directories.

I added a helper function `_docker_machine_pos_first_nonflag` that allows to distinguish between the remote and local path arguments and to suppress completion for further arguments (which would be syntax errors).
This function was copied from [Docker's bash completion](https://github.com/docker/cli/blob/master/contrib/completion/bash/docker#L565-L594).

### DIFF
--- a/contrib/completion/bash/docker-machine.bash
+++ b/contrib/completion/bash/docker-machine.bash
@@ -86,6 +86,40 @@ _docker_machine_map_key_of_current_option() {
     [[ ${words[$glob_pos]} == $glob ]] && echo "$key"
 }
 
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

PR #4307, opened 2017-11-08
Add usage of ss to check Docker Engine listening port when netstat is not available

gfieni:
Hello,
This PR add the usage of the `ss` tool to check the Docker Engine listening port when `netstat` is unavailable. The package `net-tools` including `netstat` is deprecated and no longer installed by default in some distributions like [CentOS 7](https://bugzilla.redhat.com/show_bug.cgi?id=1119297) and [Debian 9](https://www.debian.org/releases/stable/amd64/release-notes/ch-information.en.html#iproute2).


dgageot:
LGTM

PR #4317, opened 2017-11-23
Ensure proper cleanup of dangling resources in case of failure

### LINKED ISSUE #3555: Provisioning spot instance can leave dangling spot request
Using docker-machine to provision a spot instance opens a spot request and waits for it to be fulfilled. If this fails, docker-machine exits but the spot request remains open and may be fulfilled later. This results in a dangling spot request with an active instance.

nolith:
In AWS EC2 driver we start creating resources that simply get lost in case
of failure during instance creation.

The worst case scenario is when a SpotInstance price is too low to be fulfilled
before the timeout.

In such case we already created a Key Pair, but we are not going to persist
`config.json` due to the failure, and we will never be able to delete such key.

Moreover the SpotInstanceRequest will remain in AWS and if it will be fulfilled
later a new instance will be created outside of docker-machine control.

This patch ensures that a proper cleanup is done in case of failure in
`amazonec2.Create()`

Furthermore `docker-machine rm` is now able to handle properly AWS machines
failed during creation thanks to this two new conditions:

1. Key Pair deletion will not fail in case of missing `KeyPairName`
2. Instance termination will not fail in case of missing `InstanceId`

Closes #3555

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -109,6 +109,8 @@ type Driver struct {
 	Endpoint                string
 	DisableSSL              bool
 	UserDataFile            string
+
+	spotInstanceRequestId string
 }
 
 type clientFactory interface {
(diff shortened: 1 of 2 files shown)


kostyrev:
When will this be merged?

dgageot:
ping @shin-

PR #4330, opened 2017-12-07
404 is a success when deleting GCE machine and disk

tmaczukin:
This MR refactors a little the `googleErr.Code == http.StatusNotFound` check and adds a `isNotFound` check for `c.deleteDisk()` operation.

As with `c.deleteInstance()`, disk deletion should not fail the API call when the disk is already removed (for any reason). In that case only a warning should be logged as it was already done for instance removing.

Signed-off-by: Tomasz Maczukin <tomasz@maczukin.pl>

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -3,6 +3,7 @@ package google
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


tmaczukin:
fyi: @ayufan

dgageot:
LGTM

PR #4331, opened 2017-12-07
Make DigitalOcean driver RateLimit aware

tmaczukin:
This MR makes the DigitalOcean driver aware of rate limiting that is present in DO's API.

If Docker Machine is used in a high concurrent environment with DO, it may start failing on rate limiting errors. This change should make this painless for the user.

Signed-off-by: Tomasz Maczukin <tomasz@maczukin.pl>

### DIFF
--- a/drivers/digitalocean/digitalocean.go
+++ b/drivers/digitalocean/digitalocean.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"net"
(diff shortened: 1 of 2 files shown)


tmaczukin:
fyi: @ayufan

ayufan:
@tmaczukin these changes should probably be DRY-up a little :)

tmaczukin:
I was squashing all commits related to this change - didn't want to break anything. I can clean it up a little now :)

Issue #4336, opened 2017-12-12
bash completion failed for --filter option


sailing0505:
-  How to reproduce this problem：
`#docker-machine ls --filter driver=<tab>`
`azure             driver=amazonec2  generic           hyperv            rackspace         virtualbox        vmwarevcloudair
digitalocean      exoscale          google            openstack         softlayer         vmwarefusion      vmwarevsphere
`
`#docker-machine ls --filter driver=vir<tab>`
`#docker-machine ls --filter virtualbox`

the above command is invalid, the valid command should be 
`#docker-machine ls --filter driver=virtualbox`

- env info
`$ docker-machine -v`
`docker-machine version 0.13.0, build 9ba6da9`

albers:
Thanks for the report. Unfortunately, I'm not able to reproduce this. Maybe it is specific to your environment?

Looks like on your machine, `$COMP_WORDBREAKS` does not contain `=`.

Please post the output of 
- `echo "$COMP_WORDBREAKS"`
- `bash --version`
- `uname -a`
Which version of bash completion do you have installed?
Are you using a mac?

sailing0505:
look like this problem only happens on particular bash version. I tried on RHEL7, this problem doesn't exist.

albers:
@sailing0505
Most likely, on your primary machine there is a package installed that removes `=` from 
`$COMP_WORDBREAKS` and that is not installed on the other machine (at least not in the same version).
`$COMP_WORDBREAKS` is a global variable that is shared across all completions. If one completion script modifies it, other completions might break.

Can you please check whether a completion script modifies `COMP_WORDBREAKS`?
```bash
grep COMP_WORDBREAKS= /usr/share/bash-completion/completions/* /etc/bash_completion.d/*
```
_Note: hits for git and gitk are OK, these completions conditionally restore the default `:` in case other scripts deleted it._

PR #4338, opened 2017-12-12
fix bash completion for "docker-machine node ls --filter driver=<tab>…

sailing0505:
fix bash completion for "docker-machine node ls --filter driver=<tab>" or "docker-machine node ls --filter state=<tab>" issue
Signed-off-by: Hanghai Wu <wuhanghai@gmail.com>

### DIFF
--- a/contrib/completion/bash/docker-machine.bash
+++ b/contrib/completion/bash/docker-machine.bash
@@ -217,11 +217,11 @@ _docker_machine_ls() {
     local key=$(_docker_machine_map_key_of_current_option '--filter')
     case "$key" in
         driver)
-            COMPREPLY=($(compgen -W "$(_docker_machine_drivers)" -- "${cur##*=}"))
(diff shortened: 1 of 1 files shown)


albers:
@sailing0505 Can you please elaborate what exactly the issue is? I can't reproduce problems with the completion here.

albers:
@sailing0505 I just found your [issue](https://github.com/docker/machine/issues/4336), so please ignore my previous comment.

albers:
This does not work for me:

```bash
$ docker-machine ls --filter driver=vir<tab>
```
expands to
```bash
$ docker-machine ls --filter driver=driver=virtualbox
```

sailing0505:
Thanks for your comments. I have tried different bash version. following is the result:
GNU bash, version 4.3.11 : not working
GNU bash, version 4.2.46:  working

anyway, I think we can close this request. as this problem is impacted by bash.

PR #4343, opened 2017-12-21
Rsync progress

### LINKED ISSUE #4318: Progress logging disparity between "scp" and "scp -d"
Hello,

There appears to be a disparity between the outputs of `docker-machine scp` and `docker-machine scp -d` 

The non-rsync version produces the following output when requesting the file `foo.txt` from a cloud machine:

```
$ docker-machine scp foo:/foo.txt /home/dsifford/repos/test/
foo.txt                                                     100%   25     0.7KB/s   00:00
```

However, the rsync version produces no output at all. 

Is this the expected behavior? Having a progress indicator is pretty handy, especially when transferring a large amount of data.

Thanks for the clarification.

afbjorklund:
Also adds a quiet option, to the scp command.

Closes #4318.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -359,6 +359,10 @@ var Commands = []cli.Command{
 				Name:  "delta, d",
 				Usage: "Reduce amount of data sent over network by sending only the differences (uses rsync)",
 			},
+			cli.BoolFlag{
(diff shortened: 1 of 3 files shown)


dgageot:
LGTM

PR #4351, opened 2018-01-01
Add forward command for tunneling ports to machine

afbjorklund:
It is not always possible to reach the machine IP address directly,
add a small "ssh" wrapper for setting up background port forwarding.

The _preferred_ long-term approach is to have the driver set up the
port forwarding, so that one doesn't have to run separate processes.

https://github.com/boot2docker/boot2docker/blob/master/doc/WORKAROUNDS.md#port-forwarding

But using ssh works for all drivers, and is also available remotely...

Eventually docker-machine could grow a -p parameter, just like docker.


afbjorklund:
Port forwarding was discussed in #41

dgageot:
LGTM. @shin- ?

commands/forward.go
```diff
@@ -0,0 +1,89 @@
...
+		sshArgs = append(sshArgs, "-o", "IdentitiesOnly=yes")
+	}
+
+	port, err := h.GetSSHPort()
+	if err == nil && port > 0 {
+		sshArgs = append(sshArgs, "-p", fmt.Sprintf("%d", port))
+	}
+
+	if h.GetSSHKeyPath() != "" {
+		sshArgs = append(sshArgs, "-o", fmt.Sprintf("IdentityFile=%q", h.GetSSHKeyPath()))
```
shin-: I'm concerned we'll see the same issue here as the one we had to fix https://github.com/docker/machine/pull/4388 on Windows.

afbjorklund: Are you saying that `%q` is broken on Windows ? That's unfortunate, but maybe there are workarounds...

afbjorklund: But you are probably right, since most was just copied from `scp`. Guess I will do the same then.

PR #4355, opened 2018-01-04
Google Driver: Enable AutoDelete of Disks for GCE

tmaier:
This is how currently my disk listing looks like (this are gitlab-runners, started/managed with docker-machine and using Preemptible VMs)
screen shot 2018-01-04 at 15 03 54
Note that most disks are there, but not attached to any VM.
The related VMs are gone.

This means disks are not reliably deleted. For example when using together with Preemptible VMs. It can happen that the Disk stays forever.

This Pull request enables the auto-delete feature for disks.
https://cloud.google.com/sdk/gcloud/reference/compute/instances/set-disk-auto-delete
The documentation states that this is a feature which is enabled by default usually.

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -242,7 +242,7 @@ func (c *ComputeUtil) createInstance(d *Driver) error {
 		Disks: []*raw.AttachedDisk{
 			{
 				Boot:       true,
-				AutoDelete: false,
+				AutoDelete: true,
 				Type:       "PERSISTENT",
 				Mode:       "READ_WRITE",
 			},


dgageot:
@tmaier The idea behind that was to be able to delete a machine and keep the disk to save money while not losing the data. That's what is done when we run `docker-machine stop VM`. Now, I admit that the code was written at the time where VMs couldn't be stopped on GCE.

What I'd rather see is a more important code change:
 + Disks are auto-deleted
 + `docker-machine stop VM` now actually stops a VM instead of deleting it.
 + `docker-machine rm VM` should delete the VM. The disk will then be deleted automatically
 + `docker-machine start VM` should start a stopped VM and complain if it's gone

Do you think it would address your issue?

dgageot:
@tmaier I changed my mind. I think you are right. I based my thinking on the old behavior of the driver. LGTM then!

tmaier:
Thank you!

PR #4356, opened 2018-01-04
[IMP] Add instance tenancy

lasley:
* Add option to control instance placement tenancy (default, dedicated, host)

I admittedly haven't compiled or tested this yet, but the theory seems sound. Is this something you would accept once tested?


dgageot:
@lasley You need to amend the commit so that the code is formatted according to the project rules. See https://github.com/docker/machine/blob/master/CONTRIBUTING.md#tests-and-validation

PR #4358, opened 2018-01-10
Fixes #4239 - openstack: only delete autogenerated keypairs on rm

### LINKED ISSUE #4239: open stack: do not delete key-pair on rm
When creating a machine it is possible to either provide no ssh-key in which case a key is automatically generated for each machine and registered with open stack or you can provide `--openstack-keypair-name` in which case this keypair is used.

When removing a machine the corresponding key-pair is deleted as well.  This is useful for the first case but when using an existing key-pair it should not automatically delete the key since docker-machine did not create it and doesn't conceptually own it.

davidarcher:
This fixes https://github.com/docker/machine/issues/4239 and https://github.com/docker/machine/issues/3261 by tracking when docker-machine creates the keypair and then only deleting the keypair if it was created by docker-machine.

This is the same logic and variable names used by the EC2 driver.

If this PR goes in, this should supercede https://github.com/docker/machine/pull/3969 .

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -50,6 +50,8 @@ type Driver struct {
 	IpVersion        int
 	ConfigDrive      bool
 	client           Client
+	// ExistingKey keeps track of whether the key was created by us or we used an existing one. If an existing one was used, we shouldn't delete it when the machine is deleted.
+	ExistingKey bool
 }
 
 const (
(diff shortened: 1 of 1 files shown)


davidarcher:
@nathanleclaire you looked at a similar proposed fix last year. Can you take a look when you have a moment? Thanks!

dgageot:
@davidarcher please rebase

davidarcher:
@dgageot Done

dgageot:
LGTM

PR #4364, opened 2018-01-12
Fixes #4362 - openstack: ignore 404 when removing machines

### LINKED ISSUE #4362: openstack: Remove machine locally if api returns 404
When using docker-machine with the GitLab CI runner, it is unable to remove the local references to machines that have been removed manually on the cloud instance.

```
time="2018-01-11T20:52:23-05:00" level=error msg="Error removing host \"runner-c75c2f88-1515719299-dea1b3b5\": Expected HTTP response code [202 204] when accessing [DELETE https://openstack.example.org:8774/v2.1/servers/219e7951-a5fa-4a5d-ad5c-3f754f5ca3f1], but got 404 instead" name=runner-c75c2f88-1515719299-dea1b3b5 operation=remove 

time="2018-01-11T20:52:23-05:00" level=error msg="{\"itemNotFound\": {\"message\": \"Instance 219e7951-a5fa-4a5d-ad5c-3f754f5ca3f1 could not be found.\", \"code\": 404}}" name=runner-c75c2f88-1515719299-dea1b3b5 operation=remove 
```

We fixed similar for [EC2](https://github.com/docker/machine/issues/4112) and [GCE](https://github.com/docker/machine/issues/4251) already by just warning and proceeding when the cloud provider API returns a 404. We should do the same for openstack.

davidarcher:
If a docker-machine instance is deleted in openstack, the current driver will fail when trying to rm the machine. We've fixed EC2 and GCE drivers to ignore the error in this case and to simply log a warning message. This implements the same logic for the openstack driver and is essentially the same code we used for GCE in https://github.com/docker/machine/pull/4252/commits/9c34400fee351f326b29d144759f962598449d2d .

Fixes #4362

### DIFF
--- a/drivers/openstack/openstack.go
+++ b/drivers/openstack/openstack.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"strings"
 	"time"
 
@@ -13,6 +14,8 @@ import (
 	"github.com/docker/machine/libmachine/mcnutils"
 	"github.com/docker/machine/libmachine/ssh"
 	"github.com/docker/machine/libmachine/state"
+
+	"github.com/rackspace/gophercloud"
 )
 
 type Driver struct {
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

PR #4334, opened 2017-12-11
Exoscale/upgrade

greut:
The library got upgraded to a tagged release and there are better default values.

PR #4374, opened 2018-01-29
Egoscale update

greut:
A regression was introduced in the lastest PR (thanks for the merge btw) #4334, between Ubuntu LTS and TLS... This fixes it and upgrade the Security groups to be even nicer towards the Docker Swarm.

screenshot from 2018-01-29 09-15-50

It previously had the 3376 port.

Thanks!


dgageot:
LGTM

PR #4390, opened 2018-02-10
Make VirtualBox driver build on OpenBSD

nogoegst:


dgageot:
@nogoegst You need to amend the commit so that the code is formatted according to the project rules. See https://github.com/docker/machine/blob/master/CONTRIBUTING.md#tests-and-validation

dgageot:
LGTM

PR #4403, opened 2018-02-23
exoscale/template size

greut:
Hey,

A little bump to make docker-machine even better with the Exoscale.com provider.

- **update** of the egoscale library to ~~v0.9.8~~ v0.9.9
- **feature** just like the generic driver, you may reuse an existing SSH keypair
- **feature** you may now use an arbitrary disk size rather than relying only on the provided templates
- better management of the SSH keys

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -129,8 +129,8 @@
 [[projects]]
   name = "github.com/exoscale/egoscale"
   packages = ["."]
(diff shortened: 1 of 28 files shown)


dgageot:
@greut I'm sorry, I just merged the PR that gets rid of Godep... You will have to use `go dep` instead

greut:
@dgageot very good news!! Congrats, I've rebased it.

dgageot:
LGTM

Issue #4342, opened 2017-12-19
Error with pre-create check: "Hyper-V PowerShell Module is not available"


rogermartensson:
Version:      17.09.

Command:
docker-machine create -d hyperv --hyperv-virtual-switch "myswitch" name1
Running pre-create checks...
Error with pre-create check: "Hyper-V PowerShell Module is not available"

I found problem in: https://github.com/docker/machine/blob/master/drivers/hyperv/powershell.go
The reason for the problem is the Check is looking for Get-VM. If you have VMware PowerCLI installed it can happen that the test will find the VMware version of the Get-VM command. And this isn't correct and will fail.

The solution is to add the modulename to the Get-VM name.

Wicaeed:
Currently affected by this issue. Is there a workaround?

cljk:
Hmmm.. I´m getting the same error... but I have the PowerShell Module installed and have NO Vmware CLI installed.... 

Any ideas!?

cljk:
Okay... I think I´m affected from issue #4342 - differenct cause (docker for windows installed)

capnchamomile:
Same issue here. I can run `Get-VM` and it will list VMs in Hyper-V but when I attempt to run `docker-machine create -d hyperv --hyperv-virtual-switch "myswitch" myvm1` I get the "Hyper-V Powershell Module is not available" error

SuperFlea2828:
Same issue. This started happening after the upgrade from 17.x to 18.03.0-ce-win59 (16762). Get-VM is also working. Running Win 10 Pro version 1803 OS build 17128.1

docker-machine create -d hyperv --hyperv-virtual-switch "myswitch" myvm1
Running pre-create checks...
Error with pre-create check: "Hyper-V PowerShell Module is not available"

Get-VM
Name        State   CPUUsage(%) MemoryAssigned(M) Uptime           Status             Version
----        -----   ----------- ----------------- ------           ------             -------
MobyLinuxVM Running 0           2048              02:10:08.2950000 Operating normally 8.3

cljk:
I now switched back to Virtualbox and removed Hyper-V completely. The whole network management of Hyper-V was a pain independent of the problems with docker-machine.

DrDoa:
I would use a hyper-v linux instead using docker directly

sincerekamal:
I'm using `docker-machine version 0.14.0, build 89b8332` in my macbook, but still I get this error, any update on fixing this issue ?

MikyChow:
[...]

PR #4404, opened 2018-02-23
Fix issue #4342 - PowerShell cmdlet name collision with VMware PowerCLI

davemay99:
When VMware PowerCLI is installed alongside Hyper-V, there are multiple command collisions which prevent the hyperv machine driver from working as expected.  This issue is reported in #4342, and is fixed by prepending the hyper-v module name to ensure the correct module is used.


dgageot:
LGTM

tannerchung:
I've installed version 0.14.0-rc1, build e918c74, but I'm getting the error reported: "Error with pre-create check: "Hyper-V PowerShell Module is not available". Does anyone have advice one what I should be looking out for?

PR #4401, opened 2018-02-16
Add --client-certs flag to regenerate-certs command

shin-:
Fixes #4397 
Fixes #4400 

cc @programmerq

PR #4412, opened 2018-03-02
Check NotAfter date on certs and recreate if necessary

shin-:
Follow-up to #4401 that no longer requires users to manually remove existing certificates.

### DIFF
--- a/libmachine/cert/bootstrap.go
+++ b/libmachine/cert/bootstrap.go
@@ -10,12 +10,73 @@ import (
 	"github.com/docker/machine/libmachine/mcnutils"
 )
 
-func BootstrapCertificates(authOptions *auth.Options) error {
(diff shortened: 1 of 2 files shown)


libmachine/cert/bootstrap.go
```diff
@@ -10,13 +10,74 @@ import (
...
+
+	log.Infof("Creating CA: %s", caCertPath)
+
+	// check if the key path exists; if so, error
+	if _, err := os.Stat(caPrivateKeyPath); err == nil {
+		return errors.New("certificate authority key already exists")
+	}
+
+	if err := GenerateCACertificate(caCertPath, caPrivateKeyPath, caOrg, bits); err != nil {
+		return fmt.Errorf("Generating CA certificate failed: %s", err)
```
dgageot: nitpick: Errors should start with a lowercase

PR #4414, opened 2018-03-04
Move cmd files under cmd/docker-machine to make it go-install-able

nogoegst:
Sometimes (most of the time for me) it's much easier to just `go get` docker-machine rather than cloning it somewhere and run `make build` (and have make!).
This PR moves some files around while standard `make` workflow remains the same. It enables to do this:

     $ go get github.com/docker/machine/cmd/...


dgageot:
@shin- WDYT? I like that but 2 years ago, when I tried something similar it was a bit of a no no for reasons I don't recall.

dgageot:
Let's merge this and see if someone cries

nogoegst:
@dgageot wow this is intense

PR #4415, opened 2018-03-06
EC2: Update AMIs to mitigate Meltdown and Spectre

meeee:
Also add the cn-northwest-1 region.

Verified Meltdown/Spectre (v1 and v2) mitigations using [speed47/spectre-meltdown-checker](https://github.com/speed47/spectre-meltdown-checker) in `eu-west-1` for the `20180228.1` AMIs.

Note: The latest AMIs for China are from January, so they most certainly
don't have all Meltdown/Spectre mitigations. I don't have access to a
China AWS account, so I can't check.

### DIFF
--- a/drivers/amazonec2/region.go
+++ b/drivers/amazonec2/region.go
@@ -8,26 +8,27 @@ type region struct {
 	AmiId string
 }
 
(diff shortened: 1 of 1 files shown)


dgageot:
LGTM

PR #4422, opened 2018-03-15
Switch Dockerfile to golang:alpine as the base image

nogoegst:
This reduces disk footprint from ~730MB to ~270MB for base image, speeds up package installation and leaves no package manager cache to be removed.
Also there is a minor syntax fix.


nogoegst:
Oh no, I completely forgot that race detector requires glibc and doesn't work on Alpine Linux (https://github.com/golang/go/issues/14481). I'll try to fix this somehow.

tlightsky:
please fix it, really painful can't use race in alpine

nogoegst:
sorry @tlightsky, what is painful?

tlightsky:
the fact that docker:alpine image can't use race :(

Issue #4424, opened 2018-03-16
Pre-create check: "Hyper-V PowerShell Module is not available"


pldmgg:
Please note: this is NOT a duplicate of #4342.

Running docker-for-windows 18.03.0.16511-edge.

Command:

    PS C:\Users\pdadmin> docker-machine --debug create -d hyperv --hyperv-virtual-switch "ToInternal" DockerHostA
    Docker Machine Version:  0.14.0, build 89b8332
    Found binary path at C:\Program Files\Docker\Docker\Resources\bin\docker-machine.exe
    Launching plugin server for driver hyperv
    Plugin server listening at address 127.0.0.1:49951
    () Calling .GetVersion
    Using API Version  1
    () Calling .SetConfigRaw
    () Calling .GetMachineName
    (flag-lookup) Calling .GetMachineName
    (flag-lookup) Calling .DriverName
    (flag-lookup) Calling .GetCreateFlags
    Found binary path at C:\Program Files\Docker\Docker\Resources\bin\docker-machine.exe
    Launching plugin server for driver hyperv
    Plugin server listening at address 127.0.0.1:49955
    () Calling .GetVersion
    Using API Version  1
    () Calling .SetConfigRaw
    () Calling .GetMachineName
    (DockerHostA) Calling .GetMachineName
    (DockerHostA) Calling .DriverName
    (DockerHostA) Calling .GetCreateFlags
    (DockerHostA) Calling .SetConfigFromFlags
    Reading certificate data from C:\Users\pdadmin\.docker\machine\certs\ca.pem
    Decoding PEM data...
    Parsing certificate...
    Reading certificate data from C:\Users\pdadmin\.docker\machine\certs\cert.pem
    Decoding PEM data...
    Parsing certificate...
    Running pre-create checks...
    (DockerHostA) Calling .PreCreateCheck
    (DockerHostA) DBG | [executing ==>] : C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive @(Get-Command hyper-v\Get-VM).ModuleName
    (DockerHostA) DBG | [stdout =====>] : hyper-v
    (DockerHostA) DBG |
    (DockerHostA) DBG | [stderr =====>] :
    Error with pre-create check: "Hyper-V PowerShell Module is not available"
    notifying bugsnag: [Error with pre-create check: "Hyper-V PowerShell Module is not available"]

[...]

PR #4426, opened 2018-03-19
Fix issue #4424 - Pre-create check: "Hyper-V PowerShell Module is not available"

beppler:
Pull request #4404 adds the ability to run Hyper-V driver when VMware PowerCLI is installed, but it make the pre-create check for Hyper-V module fail.

This corrects the test as indicated as best solution on #4424.

### DIFF
--- a/drivers/hyperv/hyperv.go
+++ b/drivers/hyperv/hyperv.go
@@ -128,7 +128,7 @@ func (d *Driver) GetURL() (string, error) {
 }
 
 func (d *Driver) GetState() (state.State, error) {
-	stdout, err := cmdOut("(", "hyper-v\\Get-VM", d.MachineName, ").state")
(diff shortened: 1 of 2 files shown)


dgageot:
LGTM

gbraad:
Can this be tagged as a new release?

krzysztofantczak:
Quick hack for people in hurry. Editing docker-machine binary file and replacing (ie. with vim or other editor - kind of - safe for binary files) :%s/Hyper-V/hyper-v/g works.

gbraad:
I'd rather not choose to go for local patches, as this will result in issues later on.
At the moment we pull a revision ... waiting for a tag

mtudor:
Although hacky, @krzysztofantczak suggestion did work for me (on Windows, C:\Program Files\Docker\Docker\resources\bin\docker-machine.exe - make a backup of the file first) if anyone is in a bind and needs a quick fix; at your own risk and bearing in mind the warning from @gbraad.

PR #4435, opened 2018-03-22
exoscale: library update

greut:
These changes do the following:
- upgrade egoscale to v0.9.14
- stop using the `GetTopology` which does many calls to the API that weren't required and where failing in some edge cases (e.g. too many templates, or too many security groups)
- add `context.Context` everywhere, like digital ocean's driver using `TODO()`

BTW, thanks for the long awaited 0.14 release!

Cheers


greut:
Thanks!

Issue #4038, opened 2017-03-28
Openstack floating IP race


bh9:
```
docker-machine ls 
NAME                                  ACTIVE   DRIVER      STATE     URL                        SWARM   DOCKER    ERRORS
runner-587241eb-1490702504-2a286bb1   -        openstack   Running   tcp://172.27.93.78:2376            v1.12.6   
runner-587241eb-1490702504-6c323aaf   -        openstack   Running   tcp://172.27.93.78:2376            v1.12.6   
runner-587241eb-1490702504-5024f19b   -        openstack   Running   tcp://172.27.93.78:2376            v1.12.6
```
the OpenStack driver doesn't seem to check the return code from attaching a floating IP. As a result, running multiple `docker-machine create` in parallel often causes instances to report having the same floating IP. The machines which don't have the floating IP are then removed if `docker-machine ip $NAME` is run. 
Expected behaviour: Docker-machine notices the instance failed to attach the floating IP and returns an error or tries again.
Actual behaviour: Docker-machine doesn't notice and carries on
Version: `docker-machine version 0.10.0, build 76ed2a6`

bh9:
```
docker-machine create --driver openstack --openstack-availability-zone nova --openstack-flavor-id 2002 --openstack-image-name xenial-gitlab-slave --openstack-net-name gitlab-autoscale --openstack-floatingip-pool nova --openstack-sec-groups ssh,default,icmp,autoscale --openstack-ssh-user "ubuntu" test1 > test1.txt | docker-machine create --driver openstack --openstack-availability-zone nova --openstack-flavor-id 2002 --openstack-image-name xenial-gitlab-slave --openstack-net-name gitlab-autoscale --openstack-floatingip-pool nova --openstack-sec-groups ssh,default,icmp,autoscale --openstack-ssh-user "ubuntu" test2 > test2.txt | docker-machine create --driver openstack --openstack-availability-zone nova --openstack-flavor-id 2002 --openstack-image-name xenial-gitlab-slave --openstack-net-name gitlab-autoscale  --openstack-floatingip-pool nova --openstack-sec-groups ssh,default,icmp,autoscale --openstack-ssh-user "ubuntu" test3 > test3.txt
```
returned 
```
Error creating machine: Error detecting OS: Too many retries waiting for SSH to be available. Last error: Maximum number of retries (60) exceeded
Error creating machine: Error detecting OS: Too many retries waiting for SSH to be available. Last error: Maximum number of retries (60) exceeded
```
[...]

PR #4438, opened 2018-03-26
Workaround openstack floating ip race condition

stefanandres:
docker-machine create the openstack driver
will look for available floating IPs, memorize them and try to allocate
(using floatingips.Update()) later to the VM.
When using multiple parallel create commands this will create a
race condition (https://github.com/docker/machine/issues/4038) because Update() does not properly check for success in
the upstream library.

As a workaround we can always allocate new floating IPs and delete them
properly when removing the VM.


stefanandres:
And even if the upstream library (https://github.com/hashicorp/terraform/blob/a1481805050729da9a7bd0a4d9207af007cc3158/vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips/requests.go#L127) would have better error handling, docker-machine would need to change the logic, but I don't have those means myself currently.

stefreak:
@stefanandres i think the travis job is failing because you forgot to run `go fmt` on the changed files

stefanandres:
Thanks, fixed that.

stefreak:
@dgageot @shin- hm, we are running the patched fork of docker machine in production since march now and are interested in getting this merged. What is missing? Are you interested in the patch?

drivers/openstack/openstack.go
```diff
@@ -753,7 +778,14 @@ func (d *Driver) assignFloatingIP() error {
...
 
-	log.Debugf("Looking for an available floating IP", map[string]string{
+	// This is currently broken
+	// When using parallel docker-machine create calls, it might get the same IP
+	// and fail to update the port, but somehow Update() does not fail
+	// and it will think the same IP has been given to different VMs
+	// https://github.com/docker/machine/issues/4038
+
+	_ = ips
+	/*log.Debugf("Looking for an available floating IP", map[string]string{
```
shin-: We shouldn't have dead code ; either 
1. This is never useful and it should be removed
2. It does something useful but breaks in some cases, in which case it needs to be rewritten and/or fixed.

drivers/openstack/client.go
```diff
@@ -388,6 +389,22 @@ func (c *GenericClient) assignNeutronFloatingIP(d *Driver, floatingIP *FloatingI
 	return nil
 }
 
+func (c *GenericClient) DeleteFloatingIP(d *Driver, floatingIP *FloatingIP) error {
+	if d.ComputeNetwork {
+		// Not implemented
```
shin-: Is that a FIXME, or is there something that prevents this from being implemented?

drivers/openstack/openstack.go
```diff
@@ -454,8 +454,33 @@ func (d *Driver) Kill() error {
...
 	log.Debug("deleting instance...", map[string]string{"MachineId": d.MachineId})
 	log.Info("Deleting OpenStack instance...")
+
+	if err := d.resolveIds(); err != nil {
+		return err
+	}
+	if ips, err := d.client.GetFloatingIPs(d); err != nil {
+		return err
+	} else {
+		for _, xip := range ips {
```
shin-: As far as I can tell, we should be able to add filters to `GetFloatingIPs` to only retrieve the relevant data and avoid this loop entirely. https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/floatingips/requests.go#L8-L13

drivers/openstack/openstack.go
stefanandres: I'd really like to have this fixed, too. 

In OpenStack the concept of Floating IPs works like this:
A VM has a network port then can be associated with a floating IP.
To associate a floating IP you have to allocate a floating IP to your OpenStack project.
This means you can have allocated floating IPs that are not associated (e.g. for later use). Those IP are shown as used in the quota of your project. Allocated IPs can also be deleted, this will free them from the quota.

I've tried digging around to see what the actual problem is in this code.
I've got so far:
assignFloatingIP() from openstack.go calls d.client.GetFloatingIPs().
This actually gets associated FloatingIPs if there are some.
This result is given to to client.go via d.client.AssignFloatingIP()
AssignFloatingIP() will then calls assignNeutronFloatingIP().
assignNeutronFloatingIP then either:
 - allocates  new IP if none were given as parameter by calling floatingips.Create()
   - this will actually fill up all floating IP quota because deleteFloatingIP() was not implemented yet
 - updates the floatingIP by calling floatingips.Update() and passing the portID of the VM that should get the allocated IP

So now floatingips.Update() will succeed if it is the only occurrence at this time, but when running this in parallel (this is what gitlab-runner does when spawning multiple VMs at the same time) floatingIPs.Update() will *always* succeed wether or not the floating IP could be updated with the port or not. This is actually triggered on every occurence because the current code will always try the first allocated IP (even if there are multiple).

This is where I have no idea how to fix this. https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/floatingips/requests.go#L132 actually looks valid enough that it should fail if it could not update the port.

Do you have any hints for this?

drivers/openstack/openstack.go
stefanandres: I believe I even tried this but failed to implement it. Do you have a hint on how it should look like in the code when used?

stefanandres:
Thanks for response! :-) I'll comment the in-line code comments

drivers/openstack/client.go
stefanandres: IMHO the entire nova-network code could be removed as it is marked as [deprecated](https://docs.openstack.org/nova/rocky/admin/networking-nova.html). Actually no sane OpenStack provider uses nova network anymore.

"New" features shouldn't care to implement this.

shin-: Fair enough! I'd still add a warning to the tune of "Floating IP will not be removed, please do so manually if necessary" or something in those lines, but I agree that it's very low-impact.

drivers/openstack/openstack.go
shin-: Something like this? https://gist.github.com/shin-/84599a9ad3dbc0f0b7872ad259c60c42
You'll want to test this against an actual openstack setup to make sure the filter works as expected, but that's the direction I was thinking of.

drivers/openstack/openstack.go
shin-: I see. Essentially we have N processes requesting a list of IPs from a remote resource and each claiming the first resource from the list as their own. One solution may be to create a lock file of some kind that lists which IPs have been claimed, but even that wouldn't be foolproof (if the concurrent requests happen in different VMs/containers for example).

The crux of the issue IMO is that `docker-machine` was never really designed to issue parallel `create` commands.

shin-: Actually the best solution might be to lock the whole `assignFloatingIP` sequence behind a lockfile using something like https://godoc.org/github.com/nightlyone/lockfile

That'll probably slow down the parallel execution scenario but slow's better than broken.

drivers/openstack/openstack.go
stefanandres: Thanks, I'll try that

stefanandres: Your patch seems to work so far for the filtering.
```
# docker-machine ls
NAME                                                       ACTIVE   DRIVER      STATE     URL                          SWARM   DOCKER        ERRORS
runner-d9685ed3-gitlabci-small-multi-1538471577-2212a5bc   -        openstack   Running   tcp://abc:2376           v18.06.1-ce
runner-d9685ed3-gitlabci-small-multi-1538471580-f9dce75d   -        openstack   Running   tcp://abc2:2376            v18.06.1-ce
runner-d9685ed3-gitlabci-small-multi-1538471583-35298647   -        openstack   Running   tcp://abc3:2376           v18.06.1-ce
```

drivers/openstack/openstack.go
stefreak: Couldn't it also be a solution to always request a new floating ip, and fully releasing the floating ip after use? That would be much simpler and allow parallel operation even with multiple VMs/ containers.

stefanandres: That is exactly what I have implemented in this pull request, but I left the old and currently broken (when used in parallel) logic commented out that you see there.

```
As a workaround we can always allocate new floating IPs and delete them
properly when removing the VM.
```
 :-)

If we can agree on this as only solution I could just remove the old logic.

stefreak: If the old logic is necessary for some users, you could also add some way to configure the floating ip strategy or something like that. But I can't think of a really good use case.

davidarcher: I'm running into this issue using gitlab-runner autoscaling docker-machine executor against multiple OpenStack private clouds. We pre-allocate enough floating IPs for the max number of concurrent builds we run so we have a known set of IPs for CI build network traffic. Maybe that's not a "really good" use case but it is what it is. :)

+1 for a lockfile

stefreak: that use case makes sense! 👍

baurmatt:
Can we please get this merged? :) As far as I can tell from the comments all reviews are resolved.

drivers/openstack/client.go
baurmatt: @stefanandres Can you please add a `log.Warn` here and and solve the discussion? :)

drivers/openstack/openstack.go
baurmatt: @stefanandres As the requested reviews seems to be included, can you please solve the discussion? :)

drivers/openstack/openstack.go
baurmatt: @stefanandres I think it's fine to just delete the old logic here. Commented out code isn't used anyway :)

baurmatt:
As discussed internally I've implemented the changes myself :)

PR #4455, opened 2018-04-19
EC2: add support for launching instance store based images

### LINKED ISSUE #1284: Proposal: add ec2 support for 'Root device type: instance-store'
Currently all ec2 instances created assume the AMI to be 'Root device type: ebs'.  If providing an AMI with 'Root device type: instance-store', docker-machine will error:

Error creating machine: Error launching instance: Problem with AWS API call: Non-200 API response: code=400 message=EBS block device mappings not supported for instance-store AMIs.

example AMI: ami-e9f1ced9

Jawshua:
This PR makes it possible to launch instance store backed instances. Currently when launching an instance store AMI you are greeted with an error that looks something like this:

> EBS block device mappings not supported for instance-store AMIs.

Resolves #1284


rsrchboy:
This would be useful; any status updates?

DJMatus23:
Just hopping in here too, this is something that would really help us too.  If @Jawshua is ok with it, I could try and make a new PR based off their work?

Jawshua:
> Just hopping in here too, this is something that would really help us too. If @Jawshua is ok with it, I could try and make a new PR based off their work?

By all means have at it. The only conflict appears to be updated AMI IDs, so it should be an easy branch to use as a basis.

PR #4457, opened 2018-04-20
Exoscale/fixes

greut:
Library upgrade and some fixes regarding the creation of the private key and automagically picking the username from the template themselves


dgageot:
LGTM

PR #4470, opened 2018-05-05
fix hyper-v pre check

jharrilim:
@(Get-Command hyper-v\Get-VM).ModuleName now returns "hyper-v".


jharrilim:
Closing because redundant; #4426 provides a better solution.

PR #4486, opened 2018-05-21
Feature/azure nsg Create VM with Named NSG

mitchellmaler:
Added argument for the Azure driver to created or use an already existing network security group. Added Network Security Group name for machine to saved state file. Will only remove NSG if there are no more NICs attached to it anymore which means the current VM being deleted was the last.


odhekar:
Is there a plan to approve this PR, IT restricts NSG creation and this PR will solve that problem.

HighwayofLife:
It's been a year and a half since this PR was opened, but it is needed... any status on accepting and merging?

Issue #4093, opened 2017-05-16
when creating machines on AWS all the security groups are modified


manuelsotoma:
Consider We create a machine using the amazonec2 driver, and we want to assign several security groups to that machine. Of course one of those security groups includes the rules to allow the docker-machine communication (2376 , 3376 and 50).

Once the machine is created rules to enable the ports to docker-machine and ssh are automatically opened on every security group assigned to the machine.
This is from my point of view an undesired behaviour.

Thanks

cadavre:
+1, we should check entered "amazonec2-security-group"s if required port are already opened and if not – then append. I cannot see any other solution how to prevent this.

Issue #3810, opened 2016-10-10
amazonec2 always opens port 22 on 0.0.0.0/0 for the security group which would be insecure.


yeonghoey:
## Problem
[amazonec2.go] opens `22`, `2376`, `3376` on `0.0.0.0/0` for the specified security group.
It would be insecure when docker-machine needs both public access and vpc access.

### An option for turning off the implicit call of [cofigureSecurityGroups].
```go
func (d *Driver) Create() error {
    ....
	if err := d.configureSecurityGroups(d.securityGroupNames()); err != nil {
		return err
	}
```
This [code](https://github.com/docker/machine/blob/master/drivers/amazonec2/amazonec2.go#L557) always open `22`, `2376`, `3376` on the specified security groups.
By making this routine optional, the user would control the access on their own.

### An option which specifies `ipRange` for the security group.
It seems that `ipRange` in [amazonec2.go] is constant.
Make this optional for more detailed configuration.

[amazonec2.go]: https://github.com/docker/machine/blob/master/drivers/amazonec2/amazonec2.go#L31
[cofigureSecurityGroups]: https://github.com/docker/machine/blob/master/drivers/amazonec2/amazonec2.go#L976

jurajseffer:
What is worse, specifying a perfectly valid security group for AWS provisioning is ignored and DM instead creates its own with 22 and 2376 open for the entire world. The documentation says that an existing SG passed via configuration will be modified but no such thing happens.

cyphrsonic:
This has been a big security hurdle for us - we are having to design custom scripts to audit security groups under the control of docker machine and re-update the ipRange on those groups to match our security policy.

PR #4490, opened 2018-05-23
drivers/amazonec2: adds flag to prevent mutating security groups

andrestc:
This adds a workaround for issues like #4093, #3810. The user can
set a boolean flag to prevent the driver from editing their security groups.
The default behavior is kept the same for compatibility reasons.


dgageot:
LGTM

PR #4491, opened 2018-05-24
The parentheses are needed for the shell commands to be executed properly.

mshindo:
If the shell command looks like

  "condition1 && command1 || condition2 && command2 || command3"

then condition2 will never be checked if condition1 is true and command1
exists normally. In order for this code to run expectedly, it should look
like

  "(condition1 && command1) || (condition2 && command2) || command3"

instead.

Specifically, with the code like

  "[ ! -d /Users ]&& sudo mkdir /Users; sudo mount --bind /mnt/hgfs/Users /Users || [ -f /usr/local/bin/vmhgfs-fuse ]&& sudo /usr/local/bin/vmhgfs-fuse -o allow_other .host:/Users /Users || sudo mount -t vmhgfs -o uid=$(id -u),gid=$(id -g) .host:/Users /Users"

if both "[ ! -d /Users ]" and "sudo mkdir /Users; sudo mount --bind
/mnt/hgfs/Users /Users" are true, then the existence of a file
/usr/local/bin/vmhgfs-fuse will never be examined and always be executed
even if the file doesn't exist. Consequently, it always fails.

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -403,7 +403,7 @@ func (d *Driver) Create() error {
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


mshindo:
Gordon,

Thanks for the instruction and sorry for not following the convention. I
hope this time is OK. If there's still any problem, please let me know.
Thanks.

Regards,

Motonori Shindo

2018年5月25日(金) 0:30 Gordon <notifications@github.com>:

dgageot:
LGTM

PR #4511, opened 2018-06-16
Support of new openSUSE releases

hombit:
- Only name for all SUSE family is used
- Prevent to rewrite `/usr/sbin/docker-runc` provided by modern `docker-runc` package

Signed-off-by: Konstantin Malanchev <hombit@gmail.com>

### DIFF
--- a/libmachine/provision/suse.go
+++ b/libmachine/provision/suse.go
@@ -15,32 +15,14 @@ import (
 )
 
 func init() {
-	Register("openSUSE", &RegisteredProvisioner{
-		New: NewOpenSUSEProvisioner,
-	})
-	Register("SUSE Linux Enterprise Desktop", &RegisteredProvisioner{
-		New: NewSLEDProvisioner,
-	})
-	Register("SUSE Linux Enterprise Server", &RegisteredProvisioner{
-		New: NewSLESProvisioner,
+	Register("SUSE", &RegisteredProvisioner{
+		New: NewSUSEProvisioner,
 	})
 }
 
-func NewSLEDProvisioner(d drivers.Driver) Provisioner {
+func NewSUSEProvisioner(d drivers.Driver) Provisioner {
 	return &SUSEProvisioner{
-		NewSystemdProvisioner("sled", d),
-	}
-}
-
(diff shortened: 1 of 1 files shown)


matwey:
Just for the reference: https://en.opensuse.org/SDB:Find_openSUSE_version#Understanding_the_os-release_fields

matwey:
@mssola

matwey:
Hi,

Anyone there? Could you please tell what's wrong with this PR?

libmachine/provision/suse.go
```diff
@@ -135,7 +123,7 @@ func (provisioner *SUSEProvisioner) Provision(swarmOptions swarm.Options, authOp
 		return err
 	}
 
-	if strings.ToLower(provisioner.OsReleaseInfo.ID) != "opensuse" {
+	if !strings.HasPrefix(strings.ToLower(provisioner.OsReleaseInfo.ID), "opensuse") {
 		// This is a SLE machine, enable the containers module to have access
 		// to the docker packages
 		if _, err := provisioner.SSHCommand("sudo -E SUSEConnect -p sle-module-containers/12/$(uname -m) -r ''"); err != nil {
```
flavio: Unrelated with the main topic of this PR, but here we hard code the `12` path which is no longer valid with SLE15

flavio:
> Anyone there? Could you please tell what's wrong with this PR?

I don't think there's anything wrong with this PR. I like how the code is simplified by using the `ID_LIKE` field instead of the `ID` one.

libmachine/provision/suse.go
hombit: Unfortunately, I haven't SLES and cannot check it, so I've just kept old code here. We should ask SLES expert, do we need this entire command and can we make it more flexible to support various SLES versions.

hombit: @flavio Is `VERSION` field of `/etc/os-release` equals `15` for SLE 15? Can we just use `provisioner.OsReleaseInfo.Version` instead of hard-coded `12`?

flavio: Leave it aside from this PR. I'll take care of that with another PR

PR #4515, opened 2018-06-21
Use latest docker/docker revision

otiai10:
# Question

why not use the latest revision of [`docker/docker`](https://github.com/moby/moby)?


shin-:
Thanks!

PR #4534, opened 2018-07-11
minor corrections to amazonec2 driver

feuGeneA:
*  The existing code is written as if `Ec2Client.DescribeAccountAttributes` does not even list a `default-vpc` attribute if no default VPC has been set.  However, in my experience (I have no default VPC set), there is indeed a `default-vpc` attribute, with a value of `none`.  This change accommodates the `none` value, and results in a meaningful error message rather than an irrelevant one.  More details are in the commit message.
*  When failing to find a subnet, the existing code complains only of not finding one in the region/availability zone.  However, the `ec2.Filter` used to search for a subnet has two critera, one for the region and one for the VPC ID.  This change ammends the error message to also show the VPC ID used in the search.

### DIFF
--- a/drivers/amazonec2/amazonec2.go
+++ b/drivers/amazonec2/amazonec2.go
@@ -503,7 +503,7 @@ func (d *Driver) checkPrereqs() error {
 		}
 
 		if len(subnets.Subnets) == 0 {
-			return fmt.Errorf("unable to find a subnet in the zone: %s", regionZone)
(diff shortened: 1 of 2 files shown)


feuGeneA:
Paging EC2 driver contributors, new and old: @ehazlett @nathanleclaire @dgageot @shin- @andrestc 

Super simple change here (6 lines, excluding tests).  Anyone care to review?

Thanks!

shin-:
Thanks, LGTM!

PR #4535, opened 2018-07-11
exoscale: bump egoscale to 0.10

greut:
the latest release of egoscale should be a bit more robust that it's 0.9.x predecessors.


greut:
outdated.

Issue #4542, opened 2018-07-27
Integration test of core commands failed


liu-sheng:
The core commands integration test will fail because of wrong tests result assertion. see:

```
=> test/integration/core/core-commands.bats
1..23
not ok 1 openstack: machine should not exist
# (in test file test/integration/core/core-commands.bats, line 11)
#   `[[ ${lines[0]} =~ "Host does not exist: \"UNKNOWN\"" ]]' failed
# Docker machine "UNKNOWN" does not exist. Use "docker-machine ls" to list machines. Use "docker-machine create" to add a new one.
ok 2 openstack: appears with ls
ok 3 openstack: has status 'started' appearing in ls
not ok 4 openstack: create with same name again fails
# (in test file test/integration/core/core-commands.bats, line 32)
#   `[[ ${lines[0]} == "Host already exists: \"$NAME\"" ]]' failed
# Docker machine "bats-openstack-test-shared-1532589132" already exists
ok 5 openstack: run busybox container
ok 6 openstack: url
ok 7 openstack: ip
ok 8 openstack: ssh
ok 9 openstack: version
ok 10 openstack: docker commands with the socket should work
ok 11 openstack: stop
ok 12 openstack: machine should show stopped after stop
ok 13 openstack: url should show an error when machine is stopped
ok 14 openstack: env should show an error when machine is stopped
ok 15 openstack: version should show an error when machine is stopped
ok 16 openstack: machine should upgrade even when machine is stopped
ok 17 openstack: start should show an error after upgrade
ok 18 openstack: machine should show running after start
ok 19 openstack: kill
ok 20 openstack: machine should show stopped after kill
ok 21 openstack: restart
ok 22 openstack: machine should show running after restart
ok 23 openstack: status
```

PR #4543, opened 2018-07-27
Fix integrate tests of core commands cases

liu-sheng:
## Description

There are two test cases in `test/integration/core/core-commands.bats` will fail because wrong test result assertions.

## Related issue(s)

#4542


shin-:
LGTM, Thank you!

Issue #4544, opened 2018-07-27
Integration test of engine options fail


liu-sheng:
Now the integration test of "test docker process envs" in test/integration/core/engine-options.bats will fail, see:

```
=> test/integration/core/engine-options.bats
1..5
ok 1 openstack: create with supported engine options
ok 2 openstack: check for engine label
ok 3 openstack: check for engine storage driver
not ok 4 openstack: test docker process envs
# (in test file test/integration/core/engine-options.bats, line 33)
#   `[ $status -eq 0 ]' failed
# cat: /proc//environ: No such file or directory exit status 1
ok 5 openstack: check created engine option (log driver)
```

Which is because the docker daemon process has been changed to `dockerd`
FYI: https://docs.docker.com/release-notes/docker-engine/

PR #4545, opened 2018-07-27
Fix wrong docker daemon process name of engine options tests

liu-sheng:
Signed-off-by: liusheng <liusheng@huawei.com>

## Description

In the integrate test "test docker process envs" in `test/integration/core/engine-options.bats`, there is an error of docker daemon process name, since now it has been changed to `dockerd` than the old `docker daemon`.

## Related issue(s)

#4544


shin-:
LGTM, thanks!

Issue #4549, opened 2018-08-03
Doesn't propagate error when we use "docker-machine --native-ssh ssh"


ukinau:
After created VM by docker-machine, I tried to run the command over ssh by `docker-machine --native-ssh <command>` and the command was failed in the target node but error information was not propagated to docker-machine command.  See actual logs

```
root@cattle-7f98d44b97-bf6gp:~# docker-machine ls
NAME        ACTIVE   DRIVER      STATE     URL                         SWARM   DOCKER        ERRORS
contrmanu   -        openstack   Running   tcp://10.127.155.130:2376           v17.03.2-ce
root@cattle-7f98d44b97-bf6gp:~# docker-machine --native-ssh ssh contrmanu ls /tjejkaj
ls: cannot access /tjejkaj: No such file or directory
root@cattle-7f98d44b97-bf6gp:~# echo $?
0
```

This is because of https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L326-L330. 
According to current implementation, docker-machine always grasp error information and return nil in the case of NativeClient . 
This make it difficult to know the command is succeeded or not if `docker-machine --native-ssh ssh` command is executed by something other system.

PR #4550, opened 2018-08-03
Return error of command exec in SSH NativeClient

ukinau:
## Description
Previously SSH NativeClient grasp error about session.Wait(),  session.Run() and always return nil indicating succeeded even if command execution on the destination was failed.
But the error of  session.Wait, Run is that definitely we need to feedback to user. so this PR changed the code to return error if session.Wait, Run failed in Shell func which is used by `docker-machine ssh`.

## Related issue(s)
https://github.com/docker/machine/issues/4549


shin-:
LGTM, thank you!

PR #4557, opened 2018-08-10
Fix Hyper-V Administrator group member detection

EagleIJoe:
## Description

The current implementation failed to recognize if the user invoking the script is a member of the Hyper-V Administrators group. The IsInRole method expects a SID object, which needs to be generated first.


EagleIJoe:
Well thats odd, I just ammended my signature. Before, travis was building fine (as you can see in the history) ... now it fails. How do I handle this?

gbraad:
reminder to myself: @gbraad, you might want to apply this patch to machine-drivers/machine and code-ready/machine

PR #4558, opened 2018-08-13
Update "aufs" to "overlay2" in several places

tianon:
## Description

Update the default storage driver from `aufs` to `overlay2`.

## Related issue(s)

- https://github.com/moby/moby/pull/34430

  - `overlay2` now preferred over `aufs` (since 17.09.0; https://github.com/docker/docker-ce/commit/479e9126c543937f0d16482d043b8003d8654a8f)

- https://docs.docker.com/docker-for-windows/release-notes/#docker-community-edition-18060-ce-win70-2018-07-25
  > AUFS storage driver is deprecated in Docker Desktop and AUFS support will be removed in the next major release. You can continue with AUFS in Docker Desktop 18.06.x, but you will need to reset disk image (in Settings > Reset menu) before updating to the next major update.

- https://docs.docker.com/docker-for-mac/release-notes/#docker-community-edition-18060-ce-mac70-2018-07-25
  > AUFS storage driver is deprecated in Docker Desktop and AUFS support will be removed in the next major release. You can continue with AUFS in Docker Desktop 18.06.x, but you will need to reset disk image (in Preferences > Reset menu) before updating to the next major update.

- https://github.com/boot2docker/boot2docker/issues/1326

Given that `overlay2` has been preferred over `aufs` since 17.09.0 _and_ that the next major releases of Docker Desktop will no longer include AUFS support _and_ that AUFS has consistently been a source of pain around updating the kernel in boot2docker, this makes a lot of sense.

To be completely honest, I'm not sure I understand why Docker Machine is so explicit about choosing a graph driver instead of simply letting Docker autodetect/choose, but I'm sure there's context I'm missing (so this PR simply replaces `aufs` with `overlay2` in several places). :+1: :heart:

### DIFF
--- a/libmachine/libmachine.go
+++ b/libmachine/libmachine.go
@@ -76,7 +76,7 @@ func (api *Client) NewHost(driverName string, rawDriver []byte) (*host.Host, err
 			},
 			EngineOptions: &engine.Options{
 				InstallURL:    drivers.DefaultEngineInstallURL,
-				StorageDriver: "aufs",
+				StorageDriver: "overlay2",
 				TLSVerify:     true,
 			},
 			SwarmOptions: &swarm.Options{
--- a/libmachine/provision/boot2docker.go
+++ b/libmachine/provision/boot2docker.go
@@ -232,7 +232,7 @@ func (provisioner *Boot2DockerProvisioner) Provision(swarmOptions swarm.Options,
 	swarmOptions.Env = engineOptions.Env
 
 	if provisioner.EngineOptions.StorageDriver == "" {
-		provisioner.EngineOptions.StorageDriver = "aufs"
+		provisioner.EngineOptions.StorageDriver = "overlay2"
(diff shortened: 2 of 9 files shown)


afbjorklund:
Shouldn't this be made conditional on the ISO version, or something ?

Not sure if there will any more releases, but the change makes sense...

tianon:
IMO it makes sense to use `overlay2` by default on any recent Docker version and kernel, but I can try to implement some kind of version-based logic if that's desired.

Alternatively we could try to remove a lot of this explicit graph driver choosing and let Docker itself DTRT, but with machine into maintenance-only mode, it seems like a bit much to make very drastic changes like that.

afbjorklund:
I suppose you could always use older versions of Machine for older versions of Docker.

shin-:
LGTM - thanks guys!

PR #4575, opened 2018-09-20
openstack: don't fail when trying to delete non-existing keypair

FlorianLudwig:
## Description

When deleting a machine and it's associated key-pair in the openstack driver the process will fail if the keypair does not exist.  If the vm instance is (already) gone only a warning is produced instead of an error. This PR applies the same logic to key-pairs.


shin-:
LGTM, thanks!

PR #4585, opened 2018-10-03
use CPUID instead of shelling out for VT-d detection

jda:
Signed-off-by: Jade Auer <jda@tapodi.net>

## Description

On FreeBSD, VT-d detection was broken because it relied on `/proc/cpuinfo` which does not exist on that platform. dmesg output is sometimes present in `/var/run/dmesg.boot`, but access to that can be restricted by common sysctl. 

This adds support for checking CPUID directly since that should be least brittle option.

Output of `make validate`: https://gist.github.com/f8881ec207276a8d6409e5486a22015c


jda:
Looks like I should use build tags to limit the VT-d checks to just x64 platforms since this makes builds fail on arm.

PR #4601, opened 2018-10-30
Update path of golint

klyr:
The new path is golang.org/x/lint/golint.
See https://github.com/golang/lint/commit/ead987a65e5c7e053cf9633f9eac1f734f6b4fe3.

This commit prevents the build from failing

Signed-off-by: Julien Barbot <jubarbot@cisco.com>


shin-:
Thanks!

PR #4668, opened 2019-02-15
Introduce flag for verbatim user-data

srueg:
This enables the configuration of user-data directly from a string.

## Description

Introduce a flag for the Exoscale driver to configure user-data in a string instead in a file.

## Related issue(s)

This would make some integrations a lot easier. E.g.: https://github.com/rancher/rancher/issues/17760

### DIFF
--- a/drivers/exoscale/exoscale.go
+++ b/drivers/exoscale/exoscale.go
@@ -40,6 +40,7 @@ type Driver struct {
 	PublicKey        string
 	UserDataFile     string
 	UserData         []byte
+	UserDataString   string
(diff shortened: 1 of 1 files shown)


srueg:
@greut your thoughts on this?

greut:
@srueg I'll try this tomorrow and come back at you. :clap:

greut:
@srueg docker-machine is a complicated story. I've integrated your suggestion into the long awaited... #4562 as well as into the rancher/machine fork.

srueg:
Very nice, thank you!

PR #4683, opened 2019-03-06
Add support for us-gov-east-1

rclai:
Adds support for the `us-gov-east-1` AMI.


digitaldavenyc:
@GordonTheTurtle It looks like @rclai made the required edits... can this be merged?

PR #4724, opened 2019-06-17
fix openFirewallPorts to build the network name with the same format …

perryao:
…used in createInstance, allowing the user to pass a fully-qualified url for network and subnetwork resources

## Description

https://github.com/docker/machine/pull/4335 added support for fully qualified network and subnetwork names in the google driver. This pr builds on that to fix the formatting of the network name in the openFirewallPorts method. 

Currently, the driver will append the fully qualified network name to the globalUrl instead of taking it as-is.


jvitor03:
This PR will be merged?

mateuszm123111:
Can we get this? Is anyone yet support docker machine project? I do not think so :)

Issue #4537, opened 2018-07-13
Docker Machine is now in maintenance mode


shin-:
As has been obvious for some time now, we've slowly stopped implementing or accepting new features for the project. Its desktop usage has mostly been supplanted by our [Docker](https://github.com/docker/for-mac) [Desktop](https://github.com/docker/for-win) product. Provisioning on a variety of Cloud providers is overall better achieved using [infrakit](https://github.com/docker/infrakit). Overall, pursuing active development on the project doesn't make sense anymore at this point, which is why **we're officially closing the faucet for non-bugfix changes, starting today.**

I'm sure many will want to chime in on this, please keep the discussion civil and keep it inside this thread so we can keep things manageable.

JJayet:
Oh :(

Well this was nice while it lasted :)
Thx everyone for the good work !

publicarray:
I would ask that you also update the readme.md

frebib:
If official support for machine is closing, what is the likelihood for continued community-driven support in this repo?

SvenDowideit:
for those of you needing machine, there is some activity in the https://github.com/machine-drivers organisation, and it may make sense for you to work on, and release from https://github.com/machine-drivers/machine...

gbraad:
We have already looked into dding patches to this organisation, as they seem to be held from being merged here: #4509. This are blocking for localized versions of Windows. Best is to move ahead with some form of releases, however for us: minikube and minishift we only need to link against a library.

justinclift:
@shin- As a thought, the new user "Getting Started" docs still use `docker-machine` as a central part of the intro.

For people interested in updating the docs, which should they walk people through instead?

* Windows: Seems like [Docker Desktop for Windows](https://github.com/docker/for-win)
* Mac: Seems like [Docker Desktop for macOS](https://github.com/docker/for-mac)
* Linux: Keep using the Machine docs?

efrecon:
I think that this really is a shame. The real power of machine was somewhere between the simplicity of getting things work on one machine using [Desktop](https://github.com/docker/for-win) and the complexity behind [infrakit](https://github.com/docker/infrakit). machine was perfect to fire up a few machines for testing distributed workloads at smaller scale.

gilbsgilbs:
Hi!

Does somebody have an alternative software for Linux? I don't want to run docker as root on my host machine so docker-machine gave at least some easy isolation in this respect. Is there any plan for Docker for Linux?

Thanks!

[...]

PR #4732, opened 2019-07-17
Hyper-V: Use Default Switch vSwitch by default

laozc:
## Description

Default Switch with NAT was skipped as docker-machine only looks for external vSwitches.

## Related issue(s)

* Currently it requires the user to create an external vSwitch to launch the VM. In some case the provision might fail when the external DHCP does not work.


blueelvis:
@afbjorklund - Can we merge this once the comments are addressed? I will then change our package source to this for Hyper-V. Sounds good?

tstromberg:
@afbjorklund - any chance this could be merged?

afbjorklund:
@tstromberg : merged downstream, not sure if this project is doing any more updates (due to #4537)

PR #4734, opened 2019-07-23
Fix Windows minikube ISO download

douglaswth:
- On Windows, `minikube` passes in a file URL like `"file://C:/Users/douglas/.minikube/cache/iso/minikube-v1.2.0.iso"`,
  however, due to the way that Go parses URLs, the `"C:"` part is actually
  parsed as the hostname portion instead of staying at the beginning of
  the path (golang/go#6027). If `minikube` is run with a working directory on a different drive
  from where its cache directory resides, it will fail since
  `libmachine` is only looking at the path portion (in this case:
  `"/Users/douglas/.minikube/cache/iso/minikube-v1.2.0.iso"`) so it
  fails to find the file (kubernetes/minikube#1574).


medyagh:
@douglaswth  @GordonTheTurtle is there any update on this ?

PR #4753, opened 2019-09-20
Add support for amazon linux 2

dodwmd:
Signed-off-by: Michael Dodwell <mdodwell@console.com.au>

## Description

Add support to the provisionor for Amazon Linux 2


justinclift:
Pretty sure this won't be accepted (here), as Docker Machine isn't actively developed any more: https://github.com/docker/machine/issues/4537

PR #4755, opened 2019-09-24
OpenStack: update gophercloud package and add new auth options

### LINKED ISSUE #3774: possible bug in docker-machine openstack: calls v2 of API for tenants when given v3 auth?
When calling `docker-machine` using the `openstack` interface with V3 of the API, I get an unexpected error which seems to suggest that `docker-machine` is calling `v2` of the API for the tenants.  Perhaps I'm doing something wrong in my call here, but given that the `v2` appears to come out of nowhere I wonder if this is a bug?  

I'm calling against the [openstack setup](https://iujetstream.atlassian.net/wiki/display/JWT/OpenStack+command+line)  of the [NSF Jetstream system](https://jetstream-cloud.org); trying to debug with the devs there but think it might be a docker-machine issue?

After sourcing in my [openrc.sh](https://iujetstream.atlassian.net/wiki/display/JWT/Setting+up+openrc.sh) env vars appropriately, here's the `docker-machine` command I am using (with secure/personal variables replaced):

```bash
docker-machine --debug create --driver openstack --openstack-flavor-id m1.tiny --openstack-image-id IMAGEIDHASH --openstack-keypair-name KEYPAIR --openstack-sec-groups default --openstack-tenant-name TENANTNAME --openstack-domain-name DOMAIN  --openstack-private-key-file ~/.ssh/id_rsa --openstack-auth-url  https://jblb.jetstream-cloud.org:XXXX/v3  testbed
```

Which gives the following:

```

Docker Machine Version:  0.8.1, build 41b3b25
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver openstack
Plugin server listening at address 127.0.0.1:39073
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(flag-lookup) Calling .GetMachineName
(flag-lookup) Calling .DriverName
(flag-lookup) Calling .GetCreateFlags
Found binary path at /usr/local/bin/docker-machine
Launching plugin server for driver openstack
Plugin server listening at address 127.0.0.1:44650
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(testbed) Calling .GetMachineName
(testbed) Calling .DriverName
(testbed) Calling .GetCreateFlags
(testbed) Calling .SetConfigFromFlags
Running pre-create checks...
(testbed) Calling .PreCreateCheck
(testbed) Calling .GetConfigRaw
Creating machine...
(testbed) Calling .Create
(testbed) DBG | Authenticating... map[DomainName:tacc Username:cboettig TenantName:TG-DEB160003 TenantID: AuthUrl:https://jblb.jetstream-cloud.org:XXXXX/v3 Insecure:false DomainID:]

Error creating machine: Error in driver during machine creation: Expected HTTP response code [200 204] when accessing [GET https://jblb.jetstream-cloud.org:XXXXX/v2.0/tenants], but got 403 instead
{"error": {"message": "You are not authorized to perform the requested action: admin_required", "code": 403, "title": "Forbidden"}}
```

I find it strange that the error message occurs on a call to `GET https://jblb.jetstream-cloud.org:XXXXX/v2.0/tenants`, surely this should be v3.0 instead? (Also confusing that it says earlier in the debug trace `Using API Version  1`, though perhaps that refers to something else?)

kayrus:
## Description

This PR adds new auth options into the openstack client and removes deprecated keystone V2 support.

Additionally it adds an ability to remove the VM, which was not created successfully. If it is not removed, new VMs cannot be created, because the resources quota can be reached.

## Related issue(s)

Resolves #3774 and https://github.com/rancher/rancher/issues/15796


DJAyth:
Could really use this as well, since it's causing me major pains. Can someone look at it?

dbagg:
Would be really cool if somebody could look at this PR.  This issue is preventing my team from giving rancher a shot.

kayrus:
Closing in favor of https://github.com/rancher/machine/pull/54
/cc DJAyth

Issue #2989, opened 2016-02-04
can't create machine as a user that has "Hyper-V Administration" rights


jtnord:
using `docker-machine create --driver hyperv` should not require administrator rights.

Users who are a member of "Hyper-V Administrators" groups can create/destroy Virtual machines.

```
C:\Users\luser>docker-machine create --driver hyperv --hyperv-virtual-switch "vEthernet (NAT)" boot2docker
Running pre-create checks...
Error creating machine: Error with pre-create check: Hyper-v commands have to be run as an Administrator
```

This is the [offending code](https://github.com/docker/machine/blob/b6be1b79bcb2bf7ef6f74031ed5da3e72b882b4b/drivers/hyperv/hyperv.go#L152-L159) - the check should check the user is an administrator or a member of "Hyper-V administrators"

see [this artical](https://blogs.msdn.microsoft.com/virtual_pc_guy/2014/06/11/allowing-non-administrators-to-control-hyper-vupdated/) for the old gory details for old OSes that are no longer needed with Windows 8+ Server 2012 R2+...

peschkaj:
I've managed to get this working locally (see https://github.com/peschkaj/machine/commit/3480af26881eef0483dee3ca7ce5c21a273c3ea3) but VM creation fails after the security checks pass:

```
λ .\docker-machine-Windows-x86_64.exe create -d hyperv test
Running pre-create checks...
Creating machine...
(test) Copying G:\docker\cache\boot2docker.iso to G:\docker\machines\test\boot2docker.iso...
(test) Creating SSH key...
(test) Creating VM...
(test) Using switch "docker"
(test) Creating VHD
Error creating machine: Error in driver during machine creation: exit status 1
```

Base VHDs are successfully created, but it looks like the VM isn't correctly being registered with Hyper-V. I can't see it using `Get-VM` or the Hyper-V Manager.

peschkaj:
Spoke too soon - after some good old fashioned `print` debugging, `Resize-VHD` is failing due to a permissions issue. The error appears to be `The system failed to resize 'g:\docker\machines\mytest\fixed.vhd': General access denied error (0x80070005).`

jtnord:
possibly due to the permissions on the folder where you are creating your VMs?

peschkaj:
I added myself to the Hyper-V Administrators local group, which should be enough permissions and that didn't work. I then gave Hyper-V Administrators full control of the base folder (G:\docker) and gave it cascading permissions. That also didn't work. 

[...]

PR #4770, opened 2019-11-19
Fix for #2989

olivbour:
## Description
A user member of "Hyper-V Administrators" group should be able to create a VM without also being member of "Administrators" group. 

For using an SID as parameter - which is better because of potential i18n issue with role name - "isInRole()" expect a "System.Security.Principal.SecurityIdentifier" object instead of a "String". 

That's because of this error that isHypervAdministrator() always return false and and consequently  "Administrator" role is requested.

## Related issue(s)

#2989

### DIFF
--- a/drivers/hyperv/powershell.go
+++ b/drivers/hyperv/powershell.go
@@ -85,7 +85,7 @@ func isAdministrator() (bool, error) {
 }
 
 func isHypervAdministrator() bool {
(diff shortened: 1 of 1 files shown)


stylobille:
Hello,

We are really waiting for a fix on this issue (giving everybody in the company administrator rights is not possible). Since the fix has already been coded, thanks to @olivbour, how much time will it take to see this in docker desktop ?

PR #4774, opened 2019-12-09
Fix support for GCP Shared VPC

JGiola:
This Merge Request aim to add support for having the docker-machine operate on a project that is using a shared VPC.

Part of the support is already baked in, like accepting full resource urls for the network and subnetwork parameters of the docker-machine; but is lacking support when checking the presence of the firewall rule to connect to the docker port.  
Until now the firewall rule will always searched, created and updated using the project set for the VMs, but this will break if the project is using a shared VPC hosted on another project and so can’t contain firewall rules.

The smaller fix that this merge contains is to extract the project value from the network project link if contains the `projects` part and fallback to the project variable if not found, and then using this new parameter for all the firewall requests (and to construct the subnetwork link if is not a self link for the same reasons).

### DIFF
--- a/drivers/google/compute_util.go
+++ b/drivers/google/compute_util.go
@@ -28,6 +28,7 @@ type ComputeUtil struct {
 	project           string
 	diskTypeURL       string
 	address           string
+	networkProject    string
 	network           string
 	subnetwork        string
 	preemptible       bool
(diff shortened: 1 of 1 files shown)


DanishKhakwani:
Can we get this merged?

mateuszm123111:
It'd be also very usefull for me, will it be merged?

PR #4788, opened 2020-03-17
Fix shell detection for Powershell core

e2e8:
## Description

This PR fixes shell detection for newer versions of powershell. The executable for Powershell Core is pwsh.exe as opposed to powershell.exe for previous versions.

## Related issue(s)

#4789

### DIFF
--- a/libmachine/shell/shell_windows.go
+++ b/libmachine/shell/shell_windows.go
@@ -56,7 +56,7 @@ func Detect() (string, error) {
 		if err != nil {
 			return "cmd", err // defaulting to cmd
 		}
-		if strings.Contains(strings.ToLower(shell), "powershell") {
+		if strings.Contains(strings.ToLower(shell), "powershell") || strings.Contains(strings.ToLower(shell), "pwsh") {
 			return "powershell", nil
 		} else if strings.Contains(strings.ToLower(shell), "cmd") {
 			return "cmd", nil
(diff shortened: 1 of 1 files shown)


SnirBroshi:
@guillaumerose Can you please merge this?

guillaumerose:
Sorry I can't merge anymore. Perhaps @jeanlaurent or @gtardif can help.

If you want a fix for minikube, I think you should open the PR here: https://github.com/machine-drivers/machine

SnirBroshi:
(to clarify, I don't know what minikube is and I'm only interested in docker-machine :)

afbjorklund:
> If you want a fix for minikube, I think you should open the PR here: https://github.com/machine-drivers/machine

The machine-drivers is _only_ for external drivers and for fixes not (yet) accepted into machine upstream.

```
 drivers/hyperv/hyperv.go      | 71 ++++++++++++++++++++++++++++++++++++++++++++---------------------------
 drivers/hyperv/powershell.go  |  4 ++--
 libmachine/mcnutils/b2d.go    |  7 ++++++-
 libmachine/provision/utils.go |  2 +-
 libmachine/ssh/keys.go        | 14 ++++++++++++++
 5 files changed, 67 insertions(+), 31 deletions(-)
```

It's **not** a place to do development, that is supposed to happen on https://github.com/docker/machine
But since this repository is not maintained anymore, we had to add some patches needed for hyperv...

The next step is to make a more proper fork, like crc has done. After that it will no longer be compatible.

@NotWearingPants : docker-machine is in "maintenance mode" and is not accepting feature patches :-(
I think support for newer Windows and newer Docker qualifies as feature, so it is basically dead/deprecated.

> Boot2Docker was only really still maintained for Windows 7 and Windows 10 Home users. The former should be off Windows 7 now that it's EOL, and the latter are now supported by Docker Desktop thanks to WSL2 (as noted in the changes in this PR). Docker Toolbox, which is the main way Windows 7 and Windows 10 Home users would've been consuming this hasn't seen a release since 19.03.1 (Jul 31, 2019). Docker Machine is in a similar boat with the last release being Sep 2, 2019.

PR #4790, opened 2020-03-21
Ensure that for OSes that specify ID_LIKE=debian, the DebianProvisioner is indicated as compatible (for ex: Raspbian)

aniongithub:
## Description

For OSes that indicate ID_LIKE=debian (for example Raspbian), the DebianProvisioner would return false. This simple one line fix fixes that. This is taken directly from the arch provisioner that works correctly for all Arch-like distros.

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -33,6 +33,10 @@ func (provisioner *DebianProvisioner) String() string {
 	return "debian"
 }
 
+func (provisioner *DebianProvisioner) CompatibleWithHost() bool {
+	return provisioner.OsReleaseInfo.ID == provisioner.OsReleaseID || provisioner.OsReleaseInfo.IDLike == provisioner.OsReleaseID
+}
+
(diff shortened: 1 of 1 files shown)


aniongithub:
_bump_

This is a trivial change and would help fix issues using docker-machine on debian-like platforms. Halp?

@ehazlett or any of the maintainers? Is there something I could do to help expedite this PR for merge?

dabeck:
This should also help for people moving to [flatcar container linux](https://github.com/flatcar-linux/Flatcar) which is a fork of Core OS container linux that reaches EOL by the end of may.

```
NAME="Flatcar Container Linux by Kinvolk"                                                                                                                                                                                                                                       
ID=flatcar                                                                                                                                                                                                                                                                      
ID_LIKE=coreos
```

aniongithub:
*bump* Any replies on this? This actually prevents usage on a popular operating system and is a valid use-case, HALP!

aniongithub:
Moving this PR to #4884 because I might as well continue my work on my own fork instead of waiting for this to be merged. If anyone sees this thread years later, come find my fork, you'll probably have better luck.

PR #4811, opened 2020-06-03
PR to change default storage-driver to overlay2

bentastic27:
To address https://github.com/rancher/rancher/issues/27414


bentastic27:
closed per not knowing git that well

PR #4847, opened 2020-11-13
Use vmrun to find guest IP

### LINKED ISSUE #4846: VMware Fusion can't find IP address on Big Sur
I'm not sure why, but instead of using `vmrun`, the `vmwarefusion` driver attempts to find the guest IP through [roundabout ways](https://github.com/docker/machine/blob/b170508bf44c3405e079e26d5fdffe35a64c6972/drivers/vmwarefusion/fusion_darwin.go#L182-L209) by looking into DHCP lease files.

With the advent of Big Sur, this doesn't work anymore as `vmnet1`/`vmnet8` have disappeared in favour of a native `bridge100` interface.

Comparatively, `vmrun getGuestIPAddress` just works:

```
~> vmrun list
Total running VMs: 1
/Users/lloeki/.docker/machine/machines/fusion/fusion.vmx
~> vmrun checkToolsState /Users/lloeki/.docker/machine/machines/fusion/fusion.vmx
running
~> vmrun getGuestIPAddress /Users/lloeki/.docker/machine/machines/fusion/fusion.vmx
172.16.159.2
```

I suppose it might just be a case of writing a `getIPfromVmrun()` or something.

lloeki:
Since there is no more vmnet devices on Big Sur, there's no DHCP lease update either. Find ip using vmrun's own feature first, falling back to the legacy ones next.

Fixes docker/machine#4846

### DIFF
--- a/drivers/vmwarefusion/fusion_darwin.go
+++ b/drivers/vmwarefusion/fusion_darwin.go
@@ -194,6 +194,11 @@ func (d *Driver) GetIP() (string, error) {
 		return "", err
 	}
 
+	// attempt to find the address from vmrun
+	if ip, err := d.getIPfromVmrun(); err == nil {
+		return ip, err
+	}
+
 	// attempt to find the address in the vmnet configuration
 	if ip, err := d.getIPfromVmnetConfiguration(macaddr); err == nil {
 		return ip, err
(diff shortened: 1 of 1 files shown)


drivers/vmwarefusion/fusion_darwin.go
```diff
@@ -535,6 +540,18 @@ func (d *Driver) getIPfromVmnetConfiguration(macaddr string) (string, error) {
 	return "", fmt.Errorf("IP not found for MAC %s in vmnet configuration files", macaddr)
 }
 
+func (d *Driver) getIPfromVmrun() (string, error) {
+	vmx := d.vmxPath()
+
+	ip := regexp.MustCompile(`(\d+\.\d+\.\d+\.\d+)`)
+	stdout, _, _ := vmrun("getGuestIPAddress", vmx)
```
zarenner, line 547: I don't think this works in VMWare headless mode - there appears to be a [bug](https://communities.vmware.com/t5/VMware-Fusion-Discussions/Using-vmrun-to-obtain-a-guests-IP-address-when-running-the-VM-in/td-p/2236978) that requires passing the `-wait` option to `getGuestIPAddress` otherwise it returns `Error: The VMware Tools are not running in the virtual machine`. 

Also see https://github.com/machine-drivers/docker-machine-driver-vmware/pull/31#discussion_r588061432

PR #4884, opened 2021-05-03
Ensure that for OSes that specify ID_LIKE=debian, the DebianProvisioner is indicated as compatible (for ex: Raspbian)

aniongithub:
For OSes that indicate ID_LIKE=debian (for example Raspbian), the DebianProvisioner would return false. This simple one line fix fixes that. This is taken directly from the arch provisioner that works correctly for all Arch-like distros.

### DIFF
--- a/libmachine/provision/debian.go
+++ b/libmachine/provision/debian.go
@@ -33,6 +33,10 @@ func (provisioner *DebianProvisioner) String() string {
 	return "debian"
 }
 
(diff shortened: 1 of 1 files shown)


aniongithub:
Same as #4790, just from a different branch. Looks like the maintainers of this are asleep so I might as well continue work on main. If anyone sees this thread years later, come find my fork, you'll probably have better luck.

